PR #21, opened 2017-12-04
Longhorn: commit the doc and Depolyment file for longhorn

jimmy-peng:


yasker:
Close since it has been partially merged.

PR #163, opened 2018-08-03
Update README.md for RancherOS support

niusmallnan:
Document how to enable Longhorn RancherOS support.

It should be noted that currently only Flexvolume is supported in RancherOS.

In my tests, CSI did have some problems. The `NodePublishVolume` is not executed, I can see some error in kubelet.
I recommend supporting it in the next release because I have no time to track it before the release of v0.3.

https://github.com/rancher/longhorn/issues/154

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,6 +302,47 @@ curl -s https://raw.githubusercontent.com/rancher/longhorn/v0.3-rc/deploy/longho
 kubectl create -f longhorn.yaml
 ```
 
+### RancherOS
+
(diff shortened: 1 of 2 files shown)


yasker:
OK, let's delay it to the next release then.

yasker:
BTW, RKE should already changed the bind mount point to `/var/lib/kubelet/volumeplugins` to be consistent with other OS. See https://github.com/rancher/rke/commit/82af829b8cfec15ff786e8a72356e076318fc5a2

yasker:
Close the PR. Let's do it in the next release.

PR #221, opened 2018-08-15
Update default reserved value for root disk

JacieChao:
Fixed issue: https://github.com/rancher/longhorn/issues/219

### DIFF
--- a/docs/multidisk.md
+++ b/docs/multidisk.md
@@ -11,7 +11,9 @@ To add any additional disks, user needs to:
 
 Longhorn will detect the storage information (e.g. maximum space, available space) about the disk automatically, and start scheduling to it if it's possible to accomodate the volume in there. A path mounted by the existing disk won't be allowed.
 
-User can reserve a certain amount of space of the disk to stop Longhorn from using it. It can be set in the `Space Reserved` field for the disk. It's useful for the non-dedicated storage disk on the node.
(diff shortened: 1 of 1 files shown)


yasker:
The MinimalPercentage should remain the same (10%) for all the disks. But we should set the default value of ReservedStorage in the root disk to workaround the disk pressure issue for that certain disk.

docs/multidisk.md
```diff
@@ -11,7 +11,7 @@ To add any additional disks, user needs to:
 
 Longhorn will detect the storage information (e.g. maximum space, available space) about the disk automatically, and start scheduling to it if it's possible to accomodate the volume in there. A path mounted by the existing disk won't be allowed.
 
-User can reserve a certain amount of space of the disk to stop Longhorn from using it. It can be set in the `Space Reserved` field for the disk. It's useful for the non-dedicated storage disk on the node.
+User can reserve a certain amount of space of the disk to stop Longhorn from using it. It can be set in the `Space Reserved` field for the disk. It's useful for the non-dedicated storage disk on the node. By default, longhorn reserved 30% of root disk space to avoid disk pressure.
```
yasker: "root disk space (`/var/lib/rancher/longhorn`)"

yasker: Also, add some descriptions for the Kubernetes disk pressure feature for the reason we're doing this.

PR #229, opened 2018-08-18
Add uninstall yaml for v0.1

LLParse:
https://github.com/rancher/longhorn/issues/208

@yasker We should create a `v0.1` branch and rebase this PR.

### DIFF
--- a/deploy/uninstall-for-upgrade.yaml
+++ b/deploy/uninstall-for-upgrade.yaml
@@ -1,0 +1,29 @@
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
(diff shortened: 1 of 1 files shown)


yasker:
https://github.com/rancher/longhorn/blob/v0.2/deploy/uninstall-for-upgrade.yaml should work as well?

LLParse:
Different default namespace of `longhorn` is the main difference. I also though it may be confusing to refer to a v0.1 upgrade to v0.3, but reference materials in v0.2 branch.

yasker:
Sounds good. I've created the `v0.1` branch. Please rebase.

PR #230, opened 2018-08-18
Document v0.1 to v0.3 upgrade path

LLParse:
https://github.com/rancher/longhorn/issues/208

### DIFF
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -1,100 +1,196 @@
 # Upgrade
 
-Here we would cover how to upgrade from Longhorn v0.2 to Longhorn v0.3 release.
+Here we cover how to upgrade to Longhorn v0.3 from all previous releases.
 
-## Backup your existing data
-1. It's recommended to create a latest backup for every volume to the backupstore before upgrade.
-2. Make sure no volume is in degraded or faulted state.
-3. Shutdown related Kubernetes pods. Detach all the volumes. Make sure all the volumes are detached before proceeding.
-4. Backup CRD yaml to local directory:
+## Backup Existing Volumes
+
+It's recommended to create a recent backup of every volume to the backupstore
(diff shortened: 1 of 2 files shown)


docs/upgrade.md
```diff
@@ -15,6 +15,59 @@ kubectl -n longhorn-system get settings.longhorn.rancher.io -o yaml > longhorn-v
...
+On Settings page, set Backup Target to `nfs://longhorn-test-nfs-svc.default:/opt/backupstore` and click `Save`.
+
+For all volumes you wish to retain post-upgrade, create a new snapshot and backup the snapshot.
+
+### Delete Volumes
+
+Delete all Longhorn volumes and wait for all engine and replica pods to complete.
+
+```
+kubectl -n longhorn get pod -w
```
yasker: We should be able to record all the yaml files, and create them in another namespace if needed. The v0.1 Chart is essentially v0.2 Longhorn, we shouldn't need to delete all the volumes.

If we backup all the CRDs, delete the volume manager first, then remove the finalizer manually and clean up the old namespace, create the new namespace then restore the CRDs, should it work?

LLParse: @yasker The primary problem is the namespace change. We went from `longhorn` to `longhorn-system`. I can't figure out how to migrate running replica/engine pods to the new namespace. All I could come up with is backup/restore. Any ideas?

If the namespaces were the same, I think the procedure you outlined will mostly work.

yasker: The volumes need to be detached (and the user have to do it manually during the upgrade since we cannot live-upgrade the engine image for v0.1/v0.2).

So there wouldn't be running replica/engine pods.

LLParse: Oh, duh, you're totally right.

PR #241, opened 2018-08-21
Remove cluster role before upgrade

LLParse:
https://github.com/rancher/longhorn/issues/202

### DIFF
--- a/deploy/uninstall-for-upgrade.yaml
+++ b/deploy/uninstall-for-upgrade.yaml
@@ -167,3 +167,7 @@ spec:
             #value: "/home/kubernetes/flexvolume/"
       serviceAccountName: longhorn-service-account
 ---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
+metadata:
+  name: longhorn-role


yasker:
Why we need this? `kubectl apply` should covered the upgrade for the roles.

LLParse:
We could just do `kubectl apply` for all of the resources and get rid of this altogether, right? Currently our deploy doc uses `create` so I'll adjust that as well: https://github.com/rancher/longhorn/tree/v0.3-rc#deployment

yasker:
The behavior for rolling upgrade manager etc needs to be double confirmed (and probably not working), that's why I still think it would be safer to remove the old components first. But at least it's safe to use `kubectl apply` for the cluster role.

Even the rolling upgrade using `kubectl apply` works, the old flexvolume driver deployer needs to be removed in some way.

Yeah, the README.md should be updated. And the link to the upgrade instruction should be put before the deployment, in case the user didn't check the sections below.

LLParse:
Hmm, the rolling upgrade didn't seem to happen automatically; the manager daemonset updated, but the pods didn't upgrade automatically. After deleting the pods, the upgrade worked. There's some rolling upgrade config we can play with later. Let's stick with removing old components first for now.

PR #271, opened 2018-08-31
Document Longhorn chart

LLParse:
https://github.com/rancher/longhorn/issues/261

### DIFF
--- a/docs/chart.md
+++ b/docs/chart.md
@@ -1,0 +1,64 @@
+# Rancher Longhorn Chart
+
+The following document pertains to running Longhorn from the Rancher 2.0 chart.
+
+## Source Code
+
+Longhorn is 100% open source software. Project source code is spread across a number of repos:
+
+1. Longhorn Engine -- Core controller/replica logic https://github.com/rancher/longhorn-engine
+2. Longhorn Manager -- Longhorn orchestration, includes Flexvolume driver for Kubernetes https://github.com/rancher/longhorn-manager
+3. Longhorn UI -- Dashboard https://github.com/rancher/longhorn-ui
+
+## Prerequisites
+
+1. Docker v1.13+
+2. Kubernetes v1.8+ cluster with 1 or more nodes and Mount Propagation feature enabled. If your Kubernetes cluster was provisioned by Rancher v2.0.7+ or later, MountPropagation feature is enabled by default. [Check your Kubernetes environment now](https://github.com/rancher/longhorn#environment-check-script). If MountPropagation is disabled, the Kubernetes Flexvolume driver will be deployed instead of the default CSI driver and the user would need to set the FLEXVOLUME DIR parameter correctly in the chart, based on the result of the environment check. Base Image feature will also be disabled if MountPropagation is disabled.
+3. Make sure `curl`, `findmnt`, `grep`, `awk` and `blkid` has been installed in all nodes of the Kubernetes cluster.
+4. Make sure `open-iscsi` has been installed in all nodes of the Kubernetes cluster. For GKE, recommended Ubuntu as guest OS image since it contains `open-iscsi` already.
+
+## Uninstallation
+
+**Do not attempt to immediately delete Longhorn App from Rancher UI. Longhorn propagates device mounts which can damage the host if not unmounted properly.** Follow this procedure instead.
+
+1. To prevent damage to the Kubernetes cluster, we recommend deleting all Kubernetes workloads using Longhorn volumes (PersistentVolume, PersistentVolumeClaim, StorageClass, Deployment, StatefulSet, DaemonSet, etc).
+
+2. From the `Cluster` tab, click `Launch kubectl` and run this command:
+
+```
+curl -sSfL https://raw.githubusercontent.com/rancher/longhorn-manager/master/deploy/scripts/cleanup.sh | bash
+```
+
+3. Delete Longhorn App from Rancher UI. This deletes the remaining chart components.
+
+## Troubleshooting
+
+### I deleted the Longhorn App from Rancher UI instead of following the uninstallation procedure
+
+Redeploy the (same version) Longhorn App. Follow the uninstallation procedure above.
+
+### Problems with CRDs
+
+If your CRD instances or the CRDs themselves can't be deleted for whatever reason, run the commands below to clean up. Caution: this will wipe all Longhorn state!
+
+```
+# Delete CRD finalizers, instances and definitions
+for crd in $(kubectl get crd -o jsonpath={.items[*].metadata.name} | tr ' ' '\n' | grep longhorn.rancher.io); do
+  kubectl -n ${NAMESPACE} get $crd -o yaml | sed "s/\- longhorn.rancher.io//g" | kubectl apply -f -
+  kubectl -n ${NAMESPACE} delete $crd --all
+  kubectl delete crd/$crd
+done
+```
+
+### Volume can be attached/detached from UI, but Kubernetes Pod/StatefulSet etc cannot use it
+
+Check if volume plugin directory has been set correctly.
+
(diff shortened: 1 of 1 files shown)


docs/chart.md
```diff
@@ -0,0 +1,75 @@
...
+Longhorn is 100% open source software. Project source code is spread across a number of repos:
+
+1. Longhorn Engine -- Core controller/replica logic https://github.com/rancher/longhorn-engine
+2. Longhorn Manager -- Longhorn orchestration, includes Flexvolume driver for Kubernetes https://github.com/rancher/longhorn-manager
+3. Longhorn UI -- Dashboard https://github.com/rancher/longhorn-ui
+
+## Prerequisites
+
+1. Docker v1.13+
+2. Kubernetes v1.8+ cluster with 1 or more nodes and Mount Propagation feature enabled. If your Kubernetes cluster was provisioned by Rancher v2.0.7+ or later, MountPropagation feature is enabled by default. [Check your Kubernetes environment now](https://github.com/rancher/longhorn#environment-check-script). If MountPropagation is disabled, legacy K8s FlexVolume driver will be deployed instead of the newer K8s CSI driver. Volumes backed by a Base Image will also be disabled.
```
yasker: `If MountPropagation is disabled, the Kubernetes Flexvolume driver will be deployed instead of the default CSI driver. And the user would need to set the FLEXVOLUME DIR parameter correctly in the chart, based on the result of the environment check. Base Image feature will also be disabled if MountPropagation is disabled.`

docs/chart.md
```diff
@@ -0,0 +1,75 @@
...
+3. Make sure `curl`, `findmnt`, `grep`, `awk` and `blkid` has been installed in all nodes of the Kubernetes cluster.
+4. Make sure `open-iscsi` has been installed in all nodes of the Kubernetes cluster. For GKE, recommended Ubuntu as guest OS image since it contains `open-iscsi` already.
+
+## Uninstallation
+
+**Do not attempt to immediately delete Longhorn App from Rancher UI. Longhorn propagates device mounts which can damage the host if not unmounted properly.** Follow this procedure instead.
+
+1. To prevent damage to the Kubernetes cluster, we recommend deleting all Kubernetes workloads using Longhorn volumes (Deployments, StatefulSets, DaemonSets, etc).
```
yasker: And PVs and PVCs?

docs/chart.md
```diff
@@ -0,0 +1,75 @@
...
+kubectl -n ${NAMESPACE} delete deployment.apps/longhorn-driver-deployer
+# Delete all other workloads
+kubectl -n ${NAMESPACE} get daemonset.apps -o yaml | kubectl delete -f -
+kubectl -n ${NAMESPACE} get deployment.apps -o yaml | kubectl delete -f -
+kubectl -n ${NAMESPACE} get replicaset.apps -o yaml | kubectl delete -f -
+kubectl -n ${NAMESPACE} get statefulset.apps -o yaml | kubectl delete -f -
+kubectl -n ${NAMESPACE} get pods -o yaml | kubectl delete -f -
+kubectl -n ${NAMESPACE} get service -o yaml | kubectl delete -f -
+# Delete CRD finalizers, instances and definition
+for crd in $(kubectl get crd -o jsonpath={.items[*].metadata.name} | tr ' ' '\n' | grep longhorn.rancher.io); do
```
yasker: Hmm, I don't think we should include removing finalizer in the normal uninstallation. Otherwise, the cleanup may not be completed. It's better for the troubleshooting steps only.

docs/chart.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Volume can be attached/detached from UI, but Kubernetes Pod/StatefulSet etc cannot use it
+
+Check if volume plugin directory has been set correctly.
+
+By default, Kubernetes use `/usr/libexec/kubernetes/kubelet-plugins/volume/exec/` as the directory for volume plugin drivers, as stated in the [official document](https://github.com/kubernetes/community/blob/master/contributors/devel/flexvolume.md#prerequisites).
+
+But some vendors may choose to change the directory due to various reasons. For example, GKE uses `/home/kubernetes/flexvolume` instead.
+
+User can find the correct directory by running `ps aux|grep kubelet` on the host and check the `--volume-plugin-dir` parameter. If there is none, the default `/usr/libexec/kubernetes/kubelet-plugins/volume/exec/` will be used.
```
yasker: Update to use the environment check script.

docs/chart.md
LLParse: Yeah, I hesitated to add that in case the user chose to  delete Longhorn App from UI, not redeploy Longhorn, and then attempt to initiate the uninstallation procedure. I'll move it to troubleshooting.

docs/chart.md
```diff
@@ -0,0 +1,108 @@
...
+
+1. Longhorn Engine -- Core controller/replica logic https://github.com/rancher/longhorn-engine
+2. Longhorn Manager -- Longhorn orchestration, includes Flexvolume driver for Kubernetes https://github.com/rancher/longhorn-manager
+3. Longhorn UI -- Dashboard https://github.com/rancher/longhorn-ui
+
+## Prerequisites
+
+1. Docker v1.13+
+2. Kubernetes v1.8+ cluster with 1 or more nodes and Mount Propagation feature enabled. If your Kubernetes cluster was provisioned by Rancher v2.0.7+ or later, MountPropagation feature is enabled by default. [Check your Kubernetes environment now](https://github.com/rancher/longhorn#environment-check-script). If MountPropagation is disabled, the Kubernetes Flexvolume driver will be deployed instead of the default CSI driver and the user would need to set the FLEXVOLUME DIR parameter correctly in the chart, based on the result of the environment check. Base Image feature will also be disabled if MountPropagation is disabled.
+If MountPropagation is disabled, legacy K8s FlexVolume driver will be deployed instead of the newer K8s CSI driver. Volumes backed by a Base Image will also be disabled.
```
yasker: Duplicate sentence.

docs/chart.md
```diff
@@ -0,0 +1,108 @@
...
+3. Make sure `curl`, `findmnt`, `grep`, `awk` and `blkid` has been installed in all nodes of the Kubernetes cluster.
+4. Make sure `open-iscsi` has been installed in all nodes of the Kubernetes cluster. For GKE, recommended Ubuntu as guest OS image since it contains `open-iscsi` already.
+
+## Uninstallation
+
+**Do not attempt to immediately delete Longhorn App from Rancher UI. Longhorn propagates device mounts which can damage the host if not unmounted properly.** Follow this procedure instead.
+
+1. To prevent damage to the Kubernetes cluster, we recommend deleting all Kubernetes workloads using Longhorn volumes (PersistentVolumes, PersistentVolumeClaims, Deployments, StatefulSets, DaemonSets, etc).
```
yasker: And StorageClass.

docs/chart.md
```diff
@@ -0,0 +1,108 @@
...
+    out=`kubectl -n ${NAMESPACE} get $crd -o yaml|grep 'items: \[\]'`
+    if [ $? -eq 0 ]; then
+      break
+    fi
+    echo $out
+    sleep 1
+  done
+}
+remove_and_wait volumes.longhorn.rancher.io
+remove_and_wait engines.longhorn.rancher.io
```
yasker: This is not right. Deletion of volume will remove `engine` and `replicas`. Those two shouldn't be deleted separately unless something goes wrong.

LLParse: I've observed replicas living beyond the volume CRD getting deleted gracefully (finalizer removed by manager), so there may be a bug.

yasker: It shouldn't. Replica deletion can be slower because we need to start another job to clean up the content on the disk. Try to wait for a little longer. The replica can still in the cleanup process while volume is gone, but not for long. For now, we just initiate the deletion process from volume, not waiting for the process to be completed. Maybe we should change that.

https://github.com/rancher/longhorn-manager/blob/master/controller/volume_controller.go#L281

Here we basically only make sure deletionTimestamp are set for the engines/replicas.

yasker: OK, filed https://github.com/rancher/longhorn/issues/273

PR #282, opened 2018-09-12
Update docs relevant to FlexVolume dirpath setting

LLParse:
https://github.com/rancher/longhorn/issues/266

@yasker I feel as though we may want to put this in a `v0.3` branch. Updating the master branch docs before releasing `0.3.1` is going to cause confusion for existing `0.3.0` users. Worse, it causes problems if user tries to deploy `longhorn.yaml` in master branch before we update the images.

Our docs all point to resources in the master branch as opposed to a specific release, so we have to be careful...

### DIFF
--- a/deploy/longhorn.yaml
+++ b/deploy/longhorn.yaml
@@ -336,9 +336,9 @@ spec:
               fieldRef:
                 fieldPath: spec.serviceAccountName
           # For auto detection, leave this parameter unset
-          #- name: FLEXVOLUME_DIR
+          - name: FLEXVOLUME_DIR
             # FOR RKE
-            #value: "/var/lib/kubelet/volumeplugins"
+            value: "/var/lib/kubelet/volumeplugins"
             # FOR GKE
             #value: "/home/kubernetes/flexvolume/"
       serviceAccountName: longhorn-service-account
--- a/README.md
+++ b/README.md
@@ -332,48 +332,21 @@ User can find the setting for the recurring snapshot and backup in the `Volume D
 
 ## Uninstall Longhorn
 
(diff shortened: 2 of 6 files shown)


yasker:
Yeah, I am in the process of updating the manager yaml. It should be fine if we keep the doc and yaml synchronized. I would update the yaml first, then merge the PR.

yasker:
Can you rebase this? Also, don't remove the FLEXVOLUME_DIR from the deployment yaml. It should still work if the user wants to override the auto-detection.

deploy/longhorn.yaml
```diff
@@ -336,9 +336,9 @@ spec:
               fieldRef:
                 fieldPath: spec.serviceAccountName
           # For auto detection, leave this parameter unset
-          #- name: FLEXVOLUME_DIR
+          - name: FLEXVOLUME_DIR
             # FOR RKE
-            #value: "/var/lib/kubelet/volumeplugins"
+            value: "/var/lib/kubelet/volumeplugins"
```
LLParse: I set default for RKE

LLParse:
@yasker rebased

deploy/longhorn.yaml
yasker: oh no, it should be empty by default. I just merged it.

OK, I will fix it.

yasker: Also, don't change the deployment yaml in this repo. It's generated from manager's `deploy` directory.

yasker:
..... I just noticed that you've changed the environment check script here as well. It should be done in the manager repo and sync here.

PR #448, opened 2019-03-26
add CoreOS support and rename rancheros.md to csi-config.md

shuo-wu:


README.md
```diff
@@ -264,7 +264,7 @@ Longhorn will always try to maintain at least given number of healthy replicas f
 ### [Google Kubernetes Engine](./docs/gke.md)
 ### [Upgrade](./docs/upgrade.md)
 ### [Deal with Kubernetes node failure](./docs/node-failure.md)
-### [Use CSI driver on RancherOS + RKE](./docs/rancheros.md)
+### [Use CSI driver on RancherOS/CoreOS + RKE](./docs/rke.md)
```
yasker: I don't think we should rename the file to `rke.md`. If RKE wasn't run with RancherOS or CoreOS, it doesn't need those configurations. 

How about `rke-rancheros.md`?

docs/rke.md
```diff
@@ -19,45 +19,64 @@
...
-  #### 3. Modify configuration for iscsi. 
+  ##### 3. Modify configuration for iscsi. 
     
   1. Open config file `/etc/iscsi/iscsid.conf`
   2. Comment `iscsid.startup = /bin/systemctl start iscsid.socket`
   3. Uncomment `iscsid.startup = /sbin/iscsid`
-    
 
+#### CoreOS:    
+  1. Start the iSCSI daemon:
```
yasker: The format is not consistent with RancherOS entry.

docs/rke.md
```diff
@@ -19,45 +19,64 @@
...
+  systemctl start iscsid
+  ```
+
+  2. If you want to connect to iSCSI targets automatically at boot, you need to enable the systemd service:
+
+  ```
+  sudo su
+  systemctl enable iscsid
+  ```
+  Then `reboot` to start the iSCSI daemon.
```
yasker: The instruction should able to enable iscsid automatically at boot and start iscsid daemon for the current session. Not sure who would want to start iscsid manually every time.

docs/rke.md
```diff
@@ -19,45 +19,64 @@
...
 
 ## Background 
-CSI doesn't work with RancherOS + RKE before Longhorn v0.4.1. The reason is:
+CSI doesn't work with RancherOS/CoreOS + RKE before Longhorn v0.4.1. The reason is:
 
-1. RancherOS sets argument `root-dir=/opt/rke/var/lib/kubelet` for kubelet, , which is different from the default value `/var/lib/kubelet`.
+1. iSCSI is not installed or configured.
```
yasker: Not really need this reason since it's about why CSI doesn't work.

README.md
shuo-wu: But this doc also involves in `CoreOS`...

yasker: When user googles it, CoreOS would be there since it's in the content. `rke-rancheros-coreos` is probably too long. And in fact, it may just cover k3s as well...

How about `csi-config.md`?

shuo-wu: yeah. `csi-config.md` is much better

README.md
```diff
@@ -264,7 +264,7 @@ Longhorn will always try to maintain at least given number of healthy replicas f
 ### [Google Kubernetes Engine](./docs/gke.md)
 ### [Upgrade](./docs/upgrade.md)
 ### [Deal with Kubernetes node failure](./docs/node-failure.md)
-### [Use CSI driver on RancherOS + RKE](./docs/rancheros.md)
+### [Use CSI driver on RancherOS/CoreOS + RKE](.docs/csi-config.md)
```
yasker: `./docs`

Also we can add k3s here.

shuo-wu: For k3s, we don't need any extra configurations. But maybe we can add a simple instruction for manually setting the root-dir.

yasker: Given the recent bug, we need to let users know they can manually set the root-dir (and what dir that would be).

docs/csi-config.md
```diff
@@ -0,0 +1,120 @@
...
+**For RancherOS/CoreOS**
+ 
+Run `ps aux | grep kubelet` and get argument `--root-dir` on host node. 
+
+e.g.
+```
+$ ps aux | grep kubelet
+root      3755  4.4  2.9 744404 120020 ?       Ssl  00:45   0:02 kubelet --root-dir=/opt/rke/var/lib/kubelet --volume-plugin-dir=/var/lib/kubelet/volumeplugins
+```
+You will find `root-dir` in the cmdline of proc `kubelet`. By default it is `/opt/rke/var/lil/kubelet` if you are using RKE.
```
yasker: `/opt/rke/var/lib/kubelet`. But in the case above, it's `/opt/rke/var/lib/kubelet`.

docs/csi-config.md
```diff
@@ -0,0 +1,120 @@
...
+Run `ps aux | grep k3s` and get argument `--data-dir` or `-d` on k3s node.
+
+e.g.
+```
+$ ps uax | grep k3s
+root      4160  0.0  0.0  51420  3948 pts/0    S+   00:55   0:00 sudo /usr/local/bin/k3s server --data-dir /opt/test/kubelet
+root      4161 49.0  4.0 259204 164292 pts/0   Sl+  00:55   0:04 /usr/local/bin/k3s server --data-dir /opt/test/kubelet
+``` 
+You will find `data-dir` in the cmdline of proc `k3s`. By default it is not set and `/var/lib/rancher/k3s` will be used. 
+Then joining `data-dir` with `/agent/kubelet` you will get the `root-dir`.
```
yasker: Don't add a new line here.

And add:

So the default `root-dir` is `/var/lib/rancher/k3s/agent/kubelet`.

docs/csi-config.md
```diff
@@ -0,0 +1,120 @@
...
+
+
+## Background 
+CSI doesn't work with RancherOS/CoreOS + RKE before Longhorn v0.4.1. The reason is:
+
+1. RKE sets argument `root-dir=/opt/rke/var/lib/kubelet` for kubelet, , which is different from the default value `/var/lib/kubelet`.
```
yasker: double comma here. Also noted: `in the case of RancherOS and CoreOS`.

docs/csi-config.md
yasker: missed this one. I will update it later.

PR #609, opened 2019-07-01
remove backupstore from S3 path

oskapt:
The docs say to use `s3://backupbucket@us-east-1/backupstore` but the app already appears to create `backupstore` in the bucket. This means we end up with `backupbucket/backupstore/backupstore/volumes/...`

This edit removes `backupstore` from the S3 path.

### DIFF
--- a/docs/snapshot-backup.md
+++ b/docs/snapshot-backup.md
@@ -92,7 +92,7 @@ kubectl create -f https://raw.githubusercontent.com/rancher/longhorn/master/depl
 
 Now set `Settings/General/BackupTarget` to
 ```
-s3://backupbucket@us-east-1/backupstore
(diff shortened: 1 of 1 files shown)


yasker:
We probably want to change the NFS option as well.

oskapt:
The NFS option is different. That would require a change to the code to not create a subdirectory of `backupstore` within the NFS mount.

The docs now use the example of mounting `server:/opt/backupstore` as the destination. Longhorn should then create a `volumes` directory. If it wants to create `backupstore/volumes`, then we would have to export and mount `server:/opt`...and that's never going to happen.

yasker:
Yeah, that makes sense. LGTM.

PR #614, opened 2019-07-04
del needless symbol

wusphinx:
There is a needless symbol ``` ` ``` in README.md which should be delete.


yasker:
Nice catch. Merged, thanks!

Issue #623, opened 2019-07-09
manager: Support custom setting during Longhorn deployment


yasker:
The general idea is to add `01-prerequisite/04-custom-setting.yaml`, which contains a config map and a `longhorn-manager` command to apply the config map included setting to the system. The config map is just an empty `map[string]string` json file by default. But users can add custom settings to it, e.g. specify `Create Default Disk With Node Labeled` and `Default Data Path` before Longhorn system was deployed.

Longhorn manager needs to add a command line parameter to create the settings for the setting json file. It would have one parameter for the json file. Though if the setting does already exist in the system, the value won't be updated.

The deployment method via Helm Chart should also be updated for this.

yasker:
Note: in order to test the chart, the modification to the images used are needed since we haven't cut rc1 yet.

meldafrawi:
Validation: **WIP**

Steps to test:
   - Set the following values to `longhorn-default-setting` ConfigMap, and deploy longhorn
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: longhorn-default-setting
  namespace: longhorn-system
data:
  default-setting.yaml: |-
    backup-target: "s3://backupbucket@us-east-1/"
    backup-target-credential-secret: "minio-secret"
    create-default-disk-labeled-nodes: True
    default-data-path: "/var/lib/longhorn-data"
    replica-soft-anti-affinity: False
    storage-over-provisioning-percentage: 700
    storage-minimal-available-percentage: 50
    upgrade-checker: True
    default-replica-count: 5
    guaranteed-engine-cpu: 30
    default-longhorn-static-storage-class: "longhorn-new-static"
    backupstore-poll-interval: 600
```

>> Expected Result: Values should be reflected to longhorn settings in UI

PR #681, opened 2019-08-12
chart: Update chart for customized default settings

shuo-wu:
https://github.com/longhorn/longhorn/issues/623

### DIFF
--- a/chart/templates/default-setting.yaml
+++ b/chart/templates/default-setting.yaml
@@ -1,0 +1,19 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: longhorn-default-setting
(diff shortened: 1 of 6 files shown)


chart/templates/default-setting.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: longhorn-default-setting
+  namespace: {{ .Release.Namespace }}
+data:
+  default-setting.yaml: |-
+    backup-target:
```
yasker: ?

shuo-wu:
Need to update other yaml files for instance manager

Issue #806, opened 2019-10-09
Longhorn UI Taint-toleration settings


HHT02:
Related to https://github.com/longhorn/longhorn/issues/805.

After editing the default setting YAML and applying it to our deployment, we were seeing a different result within the Longhorn UI than was in our deployment. We previously added an incorrect key/value pair to the deployment and when trying to overwrite it we were still seeing the old key/value pair instead of the newer one.

Our default-setting.yaml looked like the following: 

```
backup-target:
backup-target-credential-secret:
create-default-disk-labeled-nodes: 
default-data-path: /var/lib/rancher/longhorn/
replica-soft-anti-affinity: true
storage-over-provisioning-percentage: 600
storage-minimal-available-percentage: 15
upgrade-checker: false
default-replica-count: 2
guaranteed-engine-cpu:
default-longhorn-static-storage-class: longhorn
backupstore-poll-interval:
taint-toleration: dedicated=storage:NoSchedule
```

shuo-wu:
I think you just want to modify the toleration setting for the existing Longhorn system. If Yes, you can check this doc for more info and instructions: [taint toleration setting](https://github.com/longhorn/longhorn/blob/master/docs/taint-toleration.md#after-longhorn-has-been-installed)

The default setting YAML file is for the Longhorn system that hasn't been deployed only, rather than the existing Longhorn system. The docs for these new features may confuse you since we forget to include these details in them. We will update the doc for these cases soon. 

Thank you for your reporting! And feel free to post issues if there are more questions confusing you.

ChrisScottThomas:
> The default setting YAML file is for the Longhorn system that hasn't been deployed only, rather than the existing Longhorn system. The docs for these new features may confuse you since we forget to include these details in them. We will update the doc for these cases soon.

The confusion comes because this isn't really Kubernetes intuitive.  We want, and expect, to be able to declare the state of workloads via the YAMLs, not via a UI.  Really this configMap, if used, should be the source of truth for the Longhorn workloads, not the UI.

yasker:
@ChrisScottThomas There is no way for us to validate the config map. So it cannot be the source of truth.

We're using CRDs in Kubernetes to store the metadata for Longhorn managers. Currently, we don't recommend to modify it directly, since there was no way of validation. We're working on #791 to adding validation hooks and transiting the API to more Kubernetes native way.

[...]

Issue #805, opened 2019-10-09
Taint-toleration issues


HHT02:
Hi, 

We've been having some issues when trying to add the taint-toleration via deployment. After changing the recommended YAML file (https://github.com/longhorn/longhorn/blob/master/docs/customized-default-setting.md) and applying it to our deployment we can see the appropriate tolerations in our manager pods, however the volumes still seem to attach to all nodes, not just the storage nodes. Checking these storage nodes shows the correct taints on them, with no taints being seen on the other compute/infrastructure nodes.

Perhaps this is us misunderstanding where to apply the taints/tolerations, but it feels like we have done this properly, with the volumes still attaching to the incorrect nodes. Any help would be appreciated.

Toleration seen on pod:

```
  tolerations:
  - effect: NoSchedule
    key: dedicated
    operator: Equal
    value: storage
```

Taints on storage node:

```
spec:
  taints:
  - effect: NoSchedule
    key: dedicated
    value: storage
```

No Taints on computing node

```
spec:
  podCIDR: 172.30.2.0/24

```

shuo-wu:
I think you may misunderstand the usage of taint and tolerations. The tolerations of one pod mean this pod can be scheduled to the node containing the corresponding taints. It doesn't mean that pod cannot be scheduled to the node without those taints. 

In other words, the storage nodes(the nodes contain dedicated taints) should block all other workloads and accept Longhorn workloads only as long as you set the tolerations for Longhorn correctly. And the Longhorn workloads can still be applied to the nodes without any taints. 

If you do want to exclude/evict the Longhorn system from some specific nodes, you can add some new taints for those nodes and make sure the related tolerations don't exist in the Longhorn toleration setting. In this case, you also need to make sure the workloads that will be scheduled to those nodes contain the corresponding tolerations.

ChrisScottThomas:
The description of this usage is:

> If users want to create nodes with large storage spaces and/or CPU resources for Longhorn only (to store replica data) and reject other general workloads, they can taint those nodes and add tolerations for Longhorn components. Then Longhorn can be deployed on those nodes.

This is confusing, as we _only_ want storage and replicas on our _storage_ nodes, however longhorn workloads appear on all the nodes _and_ the nodes without taints are still marked as schedulable.

Ideally we want to guarantee a longhorn state via declarative config, and without having to meddle in the UI.  This is the nature of Kubernetes after all.

We can achieve this in some fashion by adding:

```
spec:
  template:
    spec:
[...]

PR #807, opened 2019-10-09
Add more descriptions for the default setting doc and the taint toler…

shuo-wu:
Some details are not included in the docs and they may confuse users:
#805 
#806

### DIFF
--- a/docs/customized-default-setting.md
+++ b/docs/customized-default-setting.md
@@ -4,6 +4,10 @@
 During Longhorn system deployment, users can customize the default settings for Longhorn. e.g. specify `Create Default Disk With Node Labeled` and `Default Data Path` before starting the Longhorn system.
 
 ## Usage
+### Note:
+1. This default setting is only for Longhorn system that hasn't been deployed. And it has no impact on the existing Longhorn system.
+2. The users should modify the settings for an existing Longhorn system via UI. 
+
 ### Via Rancher UI
 [Cluster] -> System -> Apps -> Launch -> longhorn -> LONGHORN DEFAULT SETTINGS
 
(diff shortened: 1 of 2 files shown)


docs/customized-default-setting.md
```diff
@@ -4,6 +4,10 @@
 During Longhorn system deployment, users can customize the default settings for Longhorn. e.g. specify `Create Default Disk With Node Labeled` and `Default Data Path` before starting the Longhorn system.
 
 ## Usage
+### Warning:
```
yasker: `### Note`

It won't cause severe harm if you get it wrong, so it shouldn't be a `warning`.

docs/customized-default-setting.md
```diff
@@ -4,6 +4,10 @@
...
+1. This default setting is for Longhorn system that hasn't been deployed only.
```
yasker: `is only for` , and emphasize it with `it has no impact on the existing Longhorn system`.

docs/customized-default-setting.md
```diff
@@ -4,6 +4,10 @@
...
+2. If you want to update the settings for the existing Longhorn system, you can modify them via Longhorn UI.
```
yasker: `The users should modify the settings for an existing Longhorn system via UI`

docs/taint-toleration.md
```diff
@@ -3,6 +3,7 @@
 ## Overview
 If users want to create nodes with large storage spaces and/or CPU resources for Longhorn only (to store replica data) and reject other general workloads, they can taint those nodes and add tolerations for Longhorn components. Then Longhorn can be deployed on those nodes.
 
+Notice that Setting tolerations for one workload doesn't mean this workload should be scheduled to the nodes containing corresponding taints only. The nodes don't contain taints can accept all kinds of workloads regardless of the tolerations.
```
yasker: >Notice that the taint tolerations setting for one workload will not prevent it from being scheduled to the nodes that don't contain the corresponding taints. See [Kubernetes's taint and toleration](url) for details

PR #821, opened 2019-10-16
Clarify the expectations for failed node recovery 

shuo-wu:


docs/node-failure-reboot.md
```diff
@@ -13,3 +13,9 @@ running on the lost node. See [here](https://kubernetes.io/docs/tasks/run-applic
     on the lost Node.
 4. If you decide to delete the Pod manually (and forcefully), Kubernetes will take about another **six minutes** to delete the VolumeAttachment
 object associated with the Pod, thus finally detach the Longhorn volume from the lost Node and allow it to be used by the new Pod.
+
+## What to expect when rebooting a failed Kubernetes Node
+1. If the node is **rebooted within 6 minutes** of the failure, Kubernetes will restart pods, unmount then re-mount volumes without volume re-attaching and VolumeAttachment cleanup.
+   In this case, Longhorn will automatically detach and re-attach the volumes so that the pods can re-mount/reuse the volumes safely.
```
yasker: The responses from Longhorn are the same from two scenarios. Without more details on what's driven Longhorn's behavior, it's hard to differentiate and explain why we're documenting this.

README.md
```diff
@@ -212,7 +212,7 @@ More examples are available at `./examples/`
 
 ### [Restoring Stateful Set volumes](./docs/restore_statefulset.md)
 ### [Google Kubernetes Engine](./docs/gke.md)
-### [Deal with Kubernetes node failure](./docs/node-failure.md)
+### [Deal with Kubernetes node failure/reboot](./docs/node-failure-reboot.md)
```
yasker: No need to change the filename. It's not an exact match anyway.

docs/node-failure-reboot.md
shuo-wu: It's different. For scenario 1, Longhorn needs to detach and re-attach the volume by itself. But for scenario 2, Longhorn doesn't need to do anything.

yasker: Yes. So I meant we need to add more details on why it's different from Longhorn's perspective.

docs/node-failure.md
```diff
@@ -13,3 +13,9 @@ running on the lost node. See [here](https://kubernetes.io/docs/tasks/run-applic
     on the lost Node.
 4. If you decide to delete the Pod manually (and forcefully), Kubernetes will take about another **six minutes** to delete the VolumeAttachment
 object associated with the Pod, thus finally detach the Longhorn volume from the lost Node and allow it to be used by the new Pod.
+
+## What to expect when rebooting a failed Kubernetes Node
+1. If the node is **rebooted within 6 minutes** of the failure, Kubernetes will restart pods, unmount then re-mount volumes without volume re-attaching and VolumeAttachment cleanup.
```
yasker: 1. Need a source for `6 minutes` here. If someone asked how we know it, we should able to answer.
2. What does `If the node is **rebooted within 6 minutes** of the failure` mean? Did you mean kubelet becomes unavailable for 6 minutes or less? IIRC, normally the node reboot caused the failure. The sentence sounds like the failure would happen first, then start counting down from 6 minutes.

shuo-wu: Yes. I don't know the current statement may cause misunderstanding. How about `If users can reboot the node within 6 minutes since the failure`?

PR #841, opened 2019-10-29
Update Helm Chart

shuo-wu:


chart/templates/storageclass.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
   annotations:
     storageclass.beta.kubernetes.io/is-default-class: "false"
   {{- end }}
-provisioner: rancher.io/longhorn
+provisioner: driver.longhorn.io
```
yasker: This upgrade might fail.

yasker:
The first and third commits have been covered.

Issue #851, opened 2019-11-05
Automatically attach volume if was detached unexpected


yasker:
As long as there are still healthy replicas available, we can attach the volume.

It should cover with node reboot #375 , docker reboot #762 , Kubernetes upgrade #703 , and recovery from volume remounted as read-only #381 

https://github.com/longhorn/longhorn-manager/pull/453 should help.

chrisbulgaria:
Yes !!! 
Thats what I have been waiting for .

meldafrawi:
Validation: **Failed**
Longhorn version: `0.7.0-rc1`

Steps to test:
  - Create Longhorn StorageClass using 
` kubectl apply -f https://raw.githubusercontent.com/longhorn/longhorn-manager/master/examples/storageclass.yaml`
  - Create a pod using Longhorn volume as persistent storage
` kubectl apply -f https://raw.githubusercontent.com/longhorn/longhorn-manager/master/examples/pvc.yaml`
- Inside the pod, create a file in `/data` directory
- Restart docker service on the node which the pod is scheduled to using `systemctl restart docker`
>Expected result:  pod should be restarted , and Longhorn volume should be detached and reattached again. (PASSED)
- Check the content of the file
> **FAILED**:  listing the content of `/data` shows no files
> I had to manually delete and re-create the pods to be able to access the created file.

yasker:
@shuo-wu can you check? Seems remount doesn't work.

shuo-wu:
Need to restart the pod container to bring the volume back.

yasker:
We need document for this.

meldafrawi:
Validation: **FAILED**

Case: Kubernetes upgrade

Steps to reproduce: 
  - Using:
    - rancher v2.3.2
    - k8s v1.14.8
    - Longhorn `v0.7.0-rc1`
- Create a pod using the example in `https://github.com/shuo-wu/longhorn/blob/14b524130eccb0c32eb3d1fecfaed51d7612b1d0/docs/restore-volume.md`
- Create a new file and write some data to it.
- Using rancher, upgrade k8s version from `v1.14.8` to `v1.15.5`
- Expected result: volume get detached and re-attached automatically, and volume become accessible after pod is restarted.

> **Error: pod get stuck in CrashLoopBackOff loop**

```
  Type     Reason                  Age                     From                     Message
  ----     ------                  ----                    ----                     -------
[...]

PR #861, opened 2019-11-09
doc: Restore volume after unexpected detachment

shuo-wu:
#851


yasker:
Licenses probe is preferred so it should be put in front of the doc. Manually restart with docker should be the last resort

PR #864, opened 2019-11-11
chart: Update post upgrade job

shuo-wu:


yasker:
Why do we need to change this?

shuo-wu:
1. I don't want to fail the whole upgrade when the upgrader errors out for one time.

2. We can avoid changing the `backoffLimit` if we set a longer wait period in the upgrader. But we still need to set `toVersion` for the upgrader:
https://github.com/longhorn/longhorn-manager/blob/7cefc46ec36e193bf78f15d3c306dee4b011e751/app/post_upgrade.go#L88

yasker:
After removing `deleteOrphanedReplicaJobs`, are we using fromVersion or toVersion for anything? If not, we can remove it for now.

shuo-wu:
No. I will remove it.

PR #911, opened 2019-11-23
add units for the staleReplicaTimeout

tedder:
I can't find a single "spec" for the parameters, so I figure it's best to explain it in the examples and documentation. It took me a while to find the units in the code.

I'm still curious the mechanism of it- is this for a single replica before it gets replaced? What are the hazards of setting it too low or high?


tedder:
force-pushed for the signoff.

Issue #17, opened 2017-11-04
Volume expansion (resize)


mhemrg:
Is it possible to update a running volume's size?

yasker:
Not for now. And it's not on the plan yet.

mhemrg:
So can I create a new volume (larger than previous) based on a snapshot?

yasker:
No, currently you cannot do that as well.

errorsandwarnings:
How is it planned to increase a volume size then ? That seems a big limitation

yasker:
Increase volume size is not hard to implement with Longhorn. But it will need cooperation from filesystem right now.

It's not high in priority at the moment.

dnauck:
Article about this topic with CSI: https://kubernetes.io/blog/2018/08/02/dynamically-expand-volume-with-csi-and-kubernetes/

mariusstaicu:
this is a big blocker for using longhorn atm, as well as for using rancher2. is it gonna be postponed indefinitely like other rancher features ?

yasker:
@mariusstaicu We're not pushing back the feature indefinitely. As you see there is a milestone assigned to the feature, that means we're tracking and planning to work on it. 

More and more users are asking for this feature, so we've prioritized it from "no milestone" to v0.6.0 this year. But later we've decided to introduce an architecture refactor in v0.6.0, so we have to push some features to v0.7.0 instead.

Hope you can understand. If you have questions regarding other features/issues planning, please let me know as well in the issue.

shuo-wu:
The requirement for expanding PVC of CSI volume:
Kubernetes version v1.16: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#csi-volume-expansion
CSI Version v1.1.0: https://kubernetes-csi.github.io/docs/volume-expansion.html
New sidecar container named `CSI external-resizer`: https://kubernetes-csi.github.io/docs/external-resizer.html

For now, we can provide an option on the Longhorn UI page to directly resize Longhorn volumes.

gittygoo:
+1 that this is needed, cant understand how this is not a high priority feature

yasker:
Due to the short cycle of v0.7.0 release (only have about a month after v0.6.2), we have no choice but to move this feature along with some other features originally planned for v0.7.0 like #46 to v0.8.0 release instead.

At this moment, the volume expansion feature is at the top of our list, and we will get to it as soon as we can.

gittygoo:
[...]

PR #998, opened 2020-01-02
doc: Update expansion.doc for revert case

shuo-wu:
https://github.com/longhorn/longhorn/issues/17

### DIFF
--- a/docs/expansion.md
+++ b/docs/expansion.md
@@ -70,11 +70,35 @@ Modify `spec.resources.requests.storage` of this PVC.
  
 
 ## Filesystem expansion
-Longhorn will try to expand the file system only if:
+#### Longhorn will try to expand the file system only if:
 1. The expanded size should be greater than the current size.
 2. There is a Linux filesystem in the Longhorn volume. 
 3. The filesystem used in the Longhorn volume is one of the followings:
-  3.1 ext4
-  3.2 XFS
+    1. ext4
+    2. XFS
 4. The Longhorn volume is not in maintanence mode
 5. The Longhorn volume is using block device frontend. 
+
+#### Handling volume revert:
+If users revert a volume to a snapshot with smaller size, the frontend of the volume is still holding the expanded size. But the filesystem size will be the same as that of the reverted snapshot. In this case, users need to handle the filesystem manually:
+  1. Attach the volume to a random nodes.
+  2. Log into the corresponding node, expand the filesystem:
+  - If the filesystem is `ext4`, the volume might need to be mounted and umounted once before resizing the filesystem manually. Otherwise, executing `resize2fs` might result in an error:
+    ```
+    resize2fs: Superblock checksum does not match superblock while trying to open ......
+    Couldn't find valid filesystem superblock.
+    ```
+    Follow the steps below to resize the filesystem:  
+    ```
+    mount /dev/longhorn/<volume name> <arbitrary mount directory>
+    umount /dev/longhorn/<volume name>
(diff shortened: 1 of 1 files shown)


docs/expansion.md
```diff
@@ -70,11 +70,34 @@ Modify `spec.resources.requests.storage` of this PVC.
...
 
 ## Filesystem expansion
-Longhorn will try to expand the file system only if:
+#### Longhorn will try to expand the file system only if:
 1. The expanded size should be greater than the current size.
 2. There is a Linux filesystem in the Longhorn volume. 
 3. The filesystem used in the Longhorn volume is one of the followings:
-  3.1 ext4
-  3.2 XFS
+
```
yasker: Why do we need the extra blank lines?

docs/expansion.md
```diff
@@ -70,11 +70,34 @@ Modify `spec.resources.requests.storage` of this PVC.
...
+    3.1. ext4
+
+    3.2. XFS
+
 4. The Longhorn volume is not in maintanence mode
 5. The Longhorn volume is using block device frontend. 
+
+#### Warning:
```
yasker: Don't need `warning` line.

docs/expansion.md
```diff
@@ -70,11 +70,34 @@ Modify `spec.resources.requests.storage` of this PVC.
...
+##### Volume revert
+If users revert a volume to a snapshot with smaller size, the frontend of the volume is still holding the expanded size. But the filesystem size will be the same as that of the reverted snapshot. In this case, users need to handle the filesystem manually:
+  1. Attach the volume to a random nodes.
+  2. Log into the corresponding node, expand the filesystem:
+  - If the filesystem is `ext2/3/4`:
+    ```
+    mount /dev/longhorn/<volume name> <arbitrary mount directory>
+    umount /dev/longhorn/<volume name>
+    mount /dev/longhorn/<volume name> <arbitrary mount directory>
```
yasker: Need to at least mentioned that without the first mount/umount the user might get superblock error.

docs/expansion.md
shuo-wu: They are in the same line if there is no extra blank line:
https://github.com/longhorn/longhorn/blob/v0.8.0/docs/expansion.md#filesystem-expansion

yasker: Use four space in the beginning of the line and `1./2.` instead

```
    1. ext4
    2. XFS
```

docs/expansion.md
```diff
@@ -70,11 +70,30 @@ Modify `spec.resources.requests.storage` of this PVC.
...
+    1. ext4
+    2. XFS
 4. The Longhorn volume is not in maintanence mode
 5. The Longhorn volume is using block device frontend. 
+
+#### Handling volume revert:
+If users revert a volume to a snapshot with smaller size, the frontend of the volume is still holding the expanded size. But the filesystem size will be the same as that of the reverted snapshot. In this case, users need to handle the filesystem manually:
+  1. Attach the volume to a random nodes.
+  2. Log into the corresponding node, expand the filesystem:
+  - If the filesystem is `ext2/3/4`, the volume should be mounted twice. Otherwise using `resize2fs` to expand the filesystem will get the superblock error `resize2fs: Superblock checksum does not match superblock while trying to open ......`.
```
yasker: If the filesystem is `ext4`, the volume might need to be mount/umount once before starting the manual filesystem resizing. Otherwise, executing `resize2fs` might result in an error:
```
resize2fs: Superblock checksum does not match superblock while trying to open ......
```
Follow the steps below to resize the filesystem:

Issue #1065, opened 2020-02-22
security improvement webui should be cluster ip by default


runningman84:
The webui should not be exposed as a loabalancer by default. This creates a security issue because most users would assume they have to setup an ingress with some kind of http auth...

Exposing it additionally using a Loadbalancer creates unnecessary costs and risks.

yasker:
Make sense. @meldafrawi Can you work on this? Be careful around the upgrade case.

yasker:
@runningman84 I think `NodePort` makes more sense. `ClusterIP` will require the user to manually configure the service. `NodePort` only expose the service on the internal node's port, which should be a good compromise.

runningman84:
@yasker no user would have to manually configure the service because it is common to use an ingress for just about anything. An ingress offers options like basic auth, ip auth and so on.

If your users deploy longhorn to a cheap vps provider with no firewall (which is normally the default) they expose the longhorn storage to the internet. In this case an attacker just enters a new backup target and can download all volume data from there.

This would also expose it to search engines like https://www.shodan.io/ which just scans for open ports and their content...

runningman84:
Why do you want to go the same route as mongodb and elasticsearch which had a lot of data leaked due to insecure default configs?

yasker:
I have to say the name `mongodb` and `elasticsearch` works, lol.

OK, let's do the secure way by default. 

@meldafrawi Let's use ClusterIP.
 
The user needs to be aware of this change. We need to update the document with some examples too.

runningman84:
The examples for the ingress should also include some http basic auth or ip filter... otherwise you just move the problem to a different part of the system...

yasker:
@runningman84 suggestions are welcome! PRs on documents are even better. :smirk:

PR #1073, opened 2020-02-26
Chart: change frontend service to type ClusterIP

meldafrawi:
longhorn/longhorn#1065

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,18 +113,21 @@ longhorn-ui-f849dcd85-cgkgg                 1/1       Running   0          5d
 
 ### Accessing the UI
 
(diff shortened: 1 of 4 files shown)


runningman84:
I would advice to use ClusterIP instead of NodePort. NodePort also opens another security hole which cannot be protected well.

Using ClusterIP with an ingress in front is the best option!

yasker:
For the document part, we need to add a link in the main README.md to the ingress setup, e.g. if user doesn't use Rancher to authenticate the UI.

meldafrawi:
Updated main README.md and added link to the ingress setup.

Issue #1131, opened 2020-03-20
[Doc] Migrate document from `longhorn/longhorn` to `longhorn/website`


yasker:
We need to migrate docs for v0.8.0 from this repo to `longhorn/website` repo.

For any future releases, we should update the document in `longhorn/website` instead. Docs section in https://github.com/longhorn/website should be the source of truth of documentation for Longhorn.

JasonvanBrackel:
@yasker I don't have access to assign myself.  Please assign me to that.  There are currently 23 docs in the website. Here's the process I'll use.

1. Find the analogous document in the website docs
2. Diff the docs
3. Add information to the website docs assuming the docs in the longhorn repository are currently more up to date. 

There will be two PRs with this.  One in this repository and one in the websites repository.

I'll start the migration today.  Should be done before end of next week, given my current other commitments.

yasker:
Sounds good, assigned, thanks @JasonvanBrackel

yasker:
Done.

PR #1187, opened 2020-04-14
Remove longhorn.io docs from longhorn/longhorn

catherineluse:
This PR should be merged after https://github.com/longhorn/website/pull/48 is merged because it links to the longhorn.io docs.


yasker:
https://github.com/longhorn/longhorn/issues/1131

PR #1215, opened 2020-04-17
add Longhorn and CSI use default image question

meldafrawi:
Signed-off-by: meldafrawi <mohamed.eldafrawi@rancher.com>

### DIFF
--- a/chart/questions.yml
+++ b/chart/questions.yml
@@ -4,78 +4,110 @@ labels:
   io.rancher.certified: experimental
 namespace: longhorn-system
 questions:
-- variable: image.longhorn.manager
-  default: longhornio/longhorn-manager
-  description: "Specify Longhorn Manager Image Name"
+- variable: image.defaultImage
+  default: "true"
+  description: "Use default Longhorn images"
+  label: Use Default Images
+  type: boolean
+  show_subquestion_if: false
+  group: "Longhorn Images"
+  subquestions:
+  - variable: image.longhorn.manager
+    default: longhornio/longhorn-manager
+    description: "Specify Longhorn Manager Image Name"
+    type: string
+    label: Longhorn Manager Image Name
+    group: "Longhorn Images Settings"
+  - variable: image.longhorn.managerTag
+    default: v0.8.1-rc1
+    description: "Specify Longhorn Manager Image Tag"
+    type: string
+    label: Longhorn Manager Image Tag
+    group: "Longhorn Images Settings"
+  - variable: image.longhorn.engine
+    default: longhornio/longhorn-engine
+    description: "Specify Longhorn Engine Image Name"
+    type: string
+    label: Longhorn Engine Image Name
+    group: "Longhorn Images Settings"
+  - variable: image.longhorn.engineTag
+    default: v0.8.1-rc1
+    description: "Specify Longhorn Engine Image Tag"
+    type: string
+    label: Longhorn Engine Image Tag
+    group: "Longhorn Images Settings"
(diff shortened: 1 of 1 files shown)


meldafrawi:
- Longhorn & CSI `Use Default Images` = True

image


- Longhorn  `Use Default Images` = False

image


-  CSI `Use Default Images` = False
image

yasker:
Good idea on attaching the picture.

We can merge the default CSI image section with the default image section, since it's uncommon to enable one but disable another. We can also move them to near the private registry setting, though I don't want to put those two together under the same switch since someone can use public registry with different images.

Btw, two issues with the view:
1. Image name and tag are not on the same line for default images.
2. I found a `generate` button next to the `Private registry password`. It doesn't make sense since this password should be predefined by the user. It's better if we can remove that button, but not a deal-breaker.

meldafrawi:
- `Use Default Images` = True

image


- `Use Default Images` = False
image

yasker:
Can we try if put on/off toggle in a different group helps with the layout?

meldafrawi:
@yasker Already tried that,  it will not affect the order of the field

yasker:
We will address the format issue later. @meldafrawi can you file an issue to track that?

meldafrawi:
#1226

yasker:
@meldafrawi Need to update your signed-off-by.

chart/questions.yml
```diff
@@ -145,19 +177,19 @@ questions:
   default: "true"
 - variable: defaultSettings.storageOverProvisioningPercentage
   label: Storage Over Provisioning Percentage
-  description: "The over-provisioning percentage defines how much storage can be allocated relative to the hard drive's capacity. By default 200."
+  description: "The over-provisioning percentage defines how much storage can be allocated relative to the hard drive's capacity. By default 500."
   group: "Longhorn Default Settings"
   type: int
   min: 0
-  default: 200
+  default: 500
```
yasker, line 184: Why did those changed?

chart/questions.yml
```diff
@@ -145,19 +177,19 @@ questions:
...
 - variable: defaultSettings.storageMinimalAvailablePercentage
   label: Storage Minimal Available Percentage
-  description: "If one disk's available capacity to it's maximum capacity in % is less than the minimal available percentage, the disk would become unschedulable until more space freed up. By default 25."
+  description: "If one disk's available capacity to it's maximum capacity in % is less than the minimal available percentage, the disk would become unschedulable until more space freed up. By default 10."
   group: "Longhorn Default Settings"
   type: int
   min: 0
   max: 100
-  default: 25
+  default: 10
```
yasker, line 192: And here.

chart/questions.yml
meldafrawi: fixed

chart/questions.yml
meldafrawi: fixed

chart/questions.yml
```diff
@@ -4,78 +4,110 @@ labels:
...
+  group: "Longhorn Images"
+  subquestions:
+  - variable: image.longhorn.manager
+    default: longhornio/longhorn-manager
+    description: "Specify Longhorn Manager Image Name"
+    type: string
+    label: Longhorn Manager Image Name
+    group: "Longhorn Images Settings"
+  - variable: image.longhorn.managerTag
+    default: v0.8.1
```
yasker, line 22: should be `v0.8.1-rc1`, which is the same as the previous default value.

meldafrawi: fixed

Issue #1163, opened 2020-04-05
[FEATURE] Update helm charts to support volume expansion


dbason:
**Is your feature request related to a problem? Please describe.**
I've installed Longhorn using the app catalog in Rancher, and there are no options for volume resizing in the answers, or in the chart.

**Describe the solution you'd like**
It would be great if options to enable this were added in the chart

**Describe alternatives you've considered**
I've tried simply modifying the storage class to enable the allow volume expansion flag, however the csi resizer ran in to permission issues:
```
I0405 23:44:44.866764       1 controller.go:224] Started PVC processing "elastic/elasticsearch-data-poc-es-default-0"
I0405 23:44:44.867331       1 controller.go:224] Started PVC processing "elastic/elasticsearch-data-poc-es-default-1"
I0405 23:44:44.867354       1 controller.go:243] No need to resize PVC "elastic/elasticsearch-data-poc-es-default-1"
I0405 23:44:44.867369       1 controller.go:224] Started PVC processing "elastic/elasticsearch-data-poc-es-default-2"
I0405 23:44:44.867385       1 controller.go:243] No need to resize PVC "elastic/elasticsearch-data-poc-es-default-2"
E0405 23:44:44.868652       1 controller.go:319] Mark pvc "elastic/elasticsearch-data-poc-es-default-0" as resizing failed: can't patch status of  PVC elastic/elasticsearch-data-poc-es-default-0 with persistentvolumeclaims "elasticsearch-data-poc-es-default-0" is forbidden: User "system:serviceaccount:longhorn-system:longhorn-service-account" cannot patch resource "persistentvolumeclaims/status" in API group "" in the namespace "elastic"
```

shuo-wu:
Updating the ClusterRole is needed. And we can provide an option to help users set `allowVolumeExpansion` in the Longhorn default StorageClass.

yasker:
Done

PR #1222, opened 2020-04-20
Update chart

shuo-wu:
https://github.com/longhorn/longhorn/issues/1163


yasker:
LGTM. Need to test upgrade though.

yasker:
Also need to add instruction on how to use the script. (just run it? but longhorn/longhorn is normally not in the GOPATH)

shuo-wu:
I have removed the script for now since it's not powerful and easy-to-use. I will polish/refactor it later.

chart/templates/storageclass.yaml
```diff
@@ -10,8 +10,8 @@ metadata:
     storageclass.beta.kubernetes.io/is-default-class: "false"
   {{- end }}
 provisioner: driver.longhorn.io
+allowVolumeExpansion: true
 parameters:
   numberOfReplicas: "{{ .Values.persistence.defaultClassReplicaCount }}"
-  staleReplicaTimeout: "30"
+  staleReplicaTimeout: "2880"
```
yasker, line 16: Don't change this for this release. We can add this to part of upgrade to 1.0.

chart/templates/deployment-ui.yaml
```diff
@@ -18,6 +18,7 @@ spec:
       containers:
       - name: longhorn-ui
         image: "{{ .Values.image.longhorn.ui }}:{{ .Values.image.longhorn.uiTag }}"
+        imagePullPolicy: Always
```
yasker, line 21: I don't think we have this in the manager?

chart/templates/deployment-ui.yaml
```diff
@@ -28,6 +29,7 @@ spec:
       imagePullSecrets:
       - name: {{ .Values.defaultSettings.registrySecret }}
       {{- end }}
+      serviceAccountName: longhorn-service-account
```
yasker, line 32: Seems unnecessary. Need to remove from manager yaml if it's there.

Issue #608, opened 2019-06-29
Pod Security Policy Should be used instead of privileged


maaztariq:
Our cluster does not allow privilege mode, we should be able to run using POD Security policy with right permission instead of privilege = true https://github.com/longhorn/longhorn/blob/master/chart/templates/daemonset-sa.yaml#L22

yasker:
Thanks for the reporting @maaztariq 

Internally Longhorn also deploys a few containers so we would need to look into PSP to find out what would be requirements and ways to address those issues.

janeczku:
The Longhorn Helm Chart should install the correct PSP granting required privileges.

janeczku:
Judging by the official Helm repo, it appears standard practice to include a custom PSP  when an application requires privileged pods: https://github.com/helm/charts/search?q=podsecuritypolicy&unscoped_q=podsecuritypolicy

yasker:
We've shipped PSP with deployment since 1.0.2.

yasker:
Close the issue since we've added PSP to Longhorn.

PR #1281, opened 2020-05-01
Set securityContext for deployer and ui (for use with PSP)

aspettl:
When pod security policies are used, the typical default of a very restricted policy does not allow root permissions. Even when a more permissive policy is assigned to the service account, we need to inform k8s that we need root permissions so that the correct policy can be selected. This PR does that.

Note: when PSPs are enforced, one still needs to explicitly set the correct PSP for both the default account in the longhorn-system namespace (allow root user) and the "longhorn-service-account" (allow everything).

### DIFF
--- a/chart/templates/deployment-driver.yaml
+++ b/chart/templates/deployment-driver.yaml
@@ -75,3 +75,5 @@ spec:
       - name: {{ .Values.defaultSettings.registrySecret }}
       {{- end }}
       serviceAccountName: longhorn-service-account
+      securityContext:
+        runAsUser: 0
--- a/chart/templates/deployment-ui.yaml
+++ b/chart/templates/deployment-ui.yaml
@@ -19,6 +19,8 @@ spec:
       - name: longhorn-ui
         image: "{{ .Values.image.longhorn.ui }}:{{ .Values.image.longhorn.uiTag }}"
         imagePullPolicy: Always
+        securityContext:
+          runAsUser: 0
         ports:
         - containerPort: 8000
           name: http
--- a/deploy/longhorn.yaml
+++ b/deploy/longhorn.yaml
@@ -348,6 +348,8 @@ spec:
       - name: longhorn-ui
         image: longhornio/longhorn-ui:v0.8.1
         imagePullPolicy: Always
+        securityContext:
+          runAsUser: 0
         ports:
         - containerPort: 8000
           name: http
@@ -439,6 +441,8 @@ spec:
       #imagePullSecrets:
       #- name:
       serviceAccountName: longhorn-service-account
+      securityContext:
+        runAsUser: 0
 ---
 kind: StorageClass
 apiVersion: storage.k8s.io/v1


aspettl:
One note: For the UI "runAsUser: 0" should be fine. For the deployer, I'm not sure. Maybe it should even have "privileged: true" like the manager (because both use the same service account anyway). Please let me know if I should change this.

yasker:
Is this related to https://github.com/longhorn/longhorn/issues/608 ?

I am not familiar with PSP configuration. Can you help to layout the reproduce steps for the issue? Thanks!

aspettl:
Yes, it is related. But, if I understand correctly, in the existing issue it's also about the need for a privileged container at all (however, considering what Longhorn needs to do, I guess at least one privileged container on each node is required).

I'm using Longhorn with PSP enabled and it works. With the change in this PR, I only need to assign the two service accounts the correct PSP. It sure is not as restricted as it could be (then one would probably need more service accounts), but I think it is good enough. Basically I am just disabling all restrictions for Longhorn.
* Cluster-wide default: I have a restricted PSP that does not allow root or privileged containers.
* I have a PSP that allows root, but not privileged containers. I assigned that to the "default" service account in longhorn-system. It's needed for the UI.
* I have a PSP for Longhorn that allows everything. I assigned that to the "longhorn-service-account".

When several PSPs are available for a service account, Kubernetes will still choose the cluster default when it thinks that those permissions are sufficient. So the way to really use another PSP is to tell in the pod definition that those permissions are required (by having a securityContext which should ideally exactly list what is needed, but in practice one requirement leading to the selection of the right PSP works).

**More details on my setup:**

For the "default" service account, the PSP including cluster role binding looks like this for me:
```
---
apiVersion: policy/v1beta1
kind: PodSecurityPolicy
metadata:
  name: longhorn-default-psp
  annotations:
    seccomp.security.alpha.kubernetes.io/allowedProfileNames: docker/default
    seccomp.security.alpha.kubernetes.io/defaultProfileName: docker/default
spec:
  privileged: false
  defaultAllowPrivilegeEscalation: false
  volumes:
    - 'configMap'
    - 'emptyDir'
    - 'projected'
    - 'secret'
    - 'downwardAPI'
  hostNetwork: false
  hostIPC: false
  hostPID: false
  runAsUser:
    rule: 'RunAsAny'
  seLinux:
    rule: 'RunAsAny'
  supplementalGroups:
    rule: 'RunAsAny'
  fsGroup:
    rule: 'RunAsAny'
  readOnlyRootFilesystem: false
---
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: longhorn-default-psp
rules:
- apiGroups:
  - policy
  resources:
  - podsecuritypolicies
  resourceNames:
  - longhorn-default-psp
  verbs:
  - use
---
kind: ClusterRoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: longhorn-default-psp-binding
roleRef:
  kind: ClusterRole
  name: longhorn-default-psp
  apiGroup: rbac.authorization.k8s.io
subjects:
- kind: ServiceAccount
  name: default
  namespace: longhorn-system
```

For "longhorn-service-account", the PSP including cluster role binding looks like this for me:
```
---
apiVersion: policy/v1beta1
kind: PodSecurityPolicy
metadata:
  name: longhorn-privileged-psp
  annotations:
    seccomp.security.alpha.kubernetes.io/allowedProfileNames: '*'
spec:
  privileged: true
  allowPrivilegeEscalation: true
  allowedCapabilities:
    - '*'
  volumes:
    - '*'
  hostNetwork: true
  hostPorts:
    - min: 0
      max: 65535
  hostIPC: true
  hostPID: true
  runAsUser:
    rule: 'RunAsAny'
  seLinux:
    rule: 'RunAsAny'
  supplementalGroups:
    rule: 'RunAsAny'
  fsGroup:
    rule: 'RunAsAny'
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: longhorn-privileged-psp
rules:
- apiGroups:
  - policy
  resources:
  - podsecuritypolicies
  resourceNames:
  - longhorn-privileged-psp
  verbs:
  - use
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: longhorn-privileged-psp-binding
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: longhorn-privileged-psp
subjects:
- kind: ServiceAccount
  name: longhorn-service-account
  namespace: longhorn-system
```

aspettl:
Hi, is there any blocker for this PR?

I at least see no way how this change can hurt. When someone does not use pod security policies, it's just not relevant.

yasker:
@aspettl Sorry for the late reply. We've been too busy working on the upcoming v1.0.0 release lately.

Regarding the PR, I can understand the driver might need more permissions. But why does UI need extra permission?

aspettl:
Happy to hear 1.0.0 is coming :-)

I do not assign any new permissions, I only state what the containers already require. Both the driver and the UI image use the "root" user. (Changing the UI image so that it runs without root user could be another PR.)
So this is only about informing Kubernetes what the containers need so that the correct PSP can be selected.

yasker:
@aspettl Sounds good. LGTM.

PR #1343, opened 2020-05-14
feat: add csi.resizerReplicaCount to chart values

f0rmiga:
I was deploying longhorn on a local single-node cluster for local development and I didn't want many replicas. Let me know if this is something not desirable.


yasker:
Thanks @f0rmiga .

Btw: It's using the environment variable from https://github.com/longhorn/longhorn-manager/blob/master/app/driver.go#L46 .

f0rmiga:
Thanks @yasker, that was quick. :)

PR #1344, opened 2020-05-14
fix: chart improvements

f0rmiga:


yasker:
Thanks for the PR @f0rmiga . Any reasons to prefer one over another? I don't know much about those nuances in the helm chart..

f0rmiga:
@yasker For the YAML `~`, it's just a matter of readability, especially because of indentation matters in YAML and depending on what you get on the next line may make your `foo:` become a map or an array. `foo: ~` makes it clear that it's actually `null`. For the `quote` instead of literal quotes, it is just another good general practice when creating Helm charts. It uses the `%q` for the values. https://play.golang.org/p/ePLIiIk-BCy

yasker:
Sounds good. LGTM.

PR #1351, opened 2020-05-15
Add option to modify the ServiceAccount name & Add it to the UI pod spec

tibuntu:
Hey there!

When deploying Longhorn using this chart, I noticed that the UI pod was not able to start in our k8s-environment.

We do have to add OSS software ServiceAccounts to a ClusterRoleBinding if root access is demanded by any container.

But the UI spec did not use the ServiceAccount at all.

Beside adding it to the template, I also updated all existing occurrences in order to make the name configureable.

My PR might be conflicting with https://github.com/longhorn/longhorn/pull/1281, so ping my if I shall rebase mine after it was merged.

Thanks and awesome work!

### DIFF
--- a/chart/questions.yml
+++ b/chart/questions.yml
@@ -262,7 +262,7 @@ questions:
   show_if: "ingress.enabled=false"
   group: "Services and Load Balancing"
   show_subquestion_if: "NodePort"
-  subquestions: 
+  subquestions:
   - variable: service.ui.nodePort
     default: ""
     description: "NodePort port number(to set explicitly, choose port between 30000-32767)"
@@ -271,3 +271,9 @@ questions:
     max: 32767
     show_if: "service.ui.type=NodePort||service.ui.type=LoadBalancer"
     label: UI Service NodePort number
+- variable: serviceAccount.name
+  default: "longhorn-service-account"
+  description: "The name of the Kubernetes ServiceAccount to create"
+  type: string
+  group: "Kubernetes RBAC"
+  label: Kubernetes ServiceAccount name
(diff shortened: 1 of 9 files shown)


aspettl:
I'm not sure I completely understand what your requirements are. The old behavior is to use the default service account of the namespace for the UI. I agree that having this configurable would be nice.
But: Basically you now use the existing service account (used for the manager etc.) also for the UI. I think you probably do not want that? The UI service account should not have the longhorn-role, right?

tibuntu:
Hey @aspettl,

maybe my wording was poor, so let me try to rephrase it.

Basically, what we want (And what might not make sense from your perspective), is to use one serviceaccount per application-stack/helm-chart to be added to use our existing RoleBinding (Which basically allows to use a PodSecurityPolicy that allows root permission for the containers).

Now, when deploying Longhorn via this chart, it turned out that the UI pod will attempt to start but it goes into an error state due to:

```
2020/05/18 05:30:34 [emerg] 1#1: chown("/var/cache/nginx/client_temp", 101) failed (1: Operation not permitted)
nginx: [emerg] chown("/var/cache/nginx/client_temp", 101) failed (1: Operation not permitted)
```

So even though the container does not need root permission from the actual deployment spec perspective, it can't start without actual root permissions.

For us, it would be the easiest solution to have it using the same SA like all other Longhorn containers but I guess we could also add the default SA of the namespace to our RoleBinding.

What would also work, would be using the non-privileged version of the nginx container. Then, the UI pod wouldn't need any root privileges at all.

Last but not least regarding your question about the longhorn-role that is deployed with this chart: I really don't know, as long as the UI container doesn't make use of the permissions granted, nothing speaks against it, or?

aspettl:
Thanks, now I understand better :-) I really want to keep a separate service account for the UI container. That's just a standard security argument: restrict permissions where possible. And the UI is one thing that may have attack surface.

So imho it would be nice to specify the service account name for both (longhorn, longhorn-ui) and when they are identical, then just use the union of permissions. Or move to a non-root version of the UI directly. Then both use cases are possible.

Issue #1349, opened 2020-05-15
Sync up manager deployment's latest setting with the chart default setting


yasker:

yasker:
Done.

Issue #1208, opened 2020-04-16
[BUG ]Error when mounting ext4 by longhorn on SLES 12 SP3


gytxmu:
- Describe the bug:

create volume by longhorn UI and create pv and pvc also by longhornUI.
create a deployment of busybox by RancherUI using the pvc created by the longhornUI.
but the pvc cannot mount on Pod.
create pv/pvc by example yaml got the same problem.

- Log:

- the log of Rancher UI or describe pod：

MountVolume.SetUp failed for volume "pvc-73a45b97-3b7e-4966-a1e2-23080eb3a9cd" : kubernetes.io/csi: mounter.SetupAt failed: rpc error: code = Internal desc = mount failed: exit status 32 Mounting command: mount Mounting arguments: -t ext4 -o defaults /dev/longhorn/pvc-73a45b97-3b7e-4966-a1e2-23080eb3a9cd /var/lib/kubelet/pods/9c75a3f9-1663-4384-a220-28207144b40b/volumes/kubernetes.iocsi/pvc-73a45b97-3b7e-4966-a1e2-23080eb3a9cd/mount: wrong fs type, bad option, bad superblock on /dev/longhorn/pvc-73a45b97-3b7e-4966-a1e2-23080eb3a9cd, missing codepage or helper program, or other error.

- the log of dmesg:

#dmesg|tail
[ 4133.324354] sd 3:0:0:1: [sde] Write cache: enabled, read cache: enabled, supports DPO and FUA
[ 4133.363342] sd 3:0:0:1: [sde] Attached SCSI disk
[ 4138.962559] EXT4-fs (sde): Couldn't mount RDWR because of SUSE-unsupported optional feature METADATA_CSUM. Load module with allow_unsupported=1.
[ 4139.704904] EXT4-fs (sde): Couldn't mount RDWR because of SUSE-unsupported optional feature METADATA_CSUM. Load module with allow_unsupported=1.

- Enviroment:

Rancher Version 2.3.5
K8S Version 1.17.2
Docker Version 18.9.1
OS:Suse12SP3
kernel:4.4.180-94.107-default
longhorn version:v0.8.0

- yaml for pv/pvc:

apiVersion: v1
kind: PersistentVolume
metadata:
name: longhorn-volext4-pv
spec:
capacity:
storage: 2Gi
volumeMode: Filesystem
accessModes:
- ReadWriteOnce
persistentVolumeReclaimPolicy: Delete
storageClassName: longhorn
csi:
[...]

PR #1372, opened 2020-05-20
chart: Update settings

shuo-wu:
longhorn/longhorn#1349


yasker:
We might have another setting option coming from https://github.com/longhorn/longhorn/issues/1208

yasker:
Fix for #1208 is merged. There is a new setting option added.

shuo-wu:
The new setting has been added.

PR #1418, opened 2020-05-28
Update README.md to specify NFS version is v4

dmayle:
Attempting to use longhorn with NFSv3 for backuip causes errors related to rpc.statd, which is not managed by longhorn processes.  Longhorn works properly with NFSv4 servers only.


yasker:
Thanks @dmayle ! Would you mind submitting a PR for https://github.com/longhorn/website as well?

Issue #1183, opened 2020-04-13
[Doc] Add NFS-on-Longhorn documentation


bgulla:
As a user of Longhorn, I would like to know how I can help support the applications I deploy that require RWX. It has been brought to my attention that running an NFS layer atop longhorn will facilitate that type of use case. Can we add supporting documentation to help walk through users in the setup of this approach?

yasker:
Depends on the result of https://github.com/longhorn/longhorn/issues/1105

yasker:
We need a test plan for NFS. Ideally, we should test it end to end for the workload cases.

joshimoo:
## Pre-merged Checklist
* [ ] Does the PR include the explanation for the fix or the feature?

* [ ] Is the backend code merged?
The PR is at

* [ ] Is the reproduce steps/test steps documented?

* [ ] **If the fix introduces the code for backward compatibility** Has an separate issue been filed with the label `release/obsolete-compatibility`?
The compatibility issue is filed at

* [ ] **If labeled: area/ui** Has the UI issue filed or ready to be merged?
The UI issue/PR is at

* [ ] **if labeled: require/doc** Has the necessary document PR submitted or merged?
The Doc issue/PR is at 

* [ ] **If labeled: require/automation-e2e** Has the end-to-end test plan been merged? Have QAs agreed on the automation test case?
The automation skeleton PR is at
The automation test case PR is at

* [ ] **if labeled: require/automation-engine** Has the engine integration test been merged?
The engine automation PR is at

* [ ] **if labeled: require/manual-test-plan** Has the manual test plan been documented?
The updated manual test plan is at

yasker:
One additional thing to consider is when after we implement #1470 , it should be easy for the user to upgrade. My biggest concern is if the user can use the same StorageClass name `longhorn-nfs` in the later version and seamlessly migrate to the new solution.

khushboo-rancher:
Validation - **Passed**

Followed the docs https://staging.longhorn.io/docs/1.0.0/advanced-resources/rwx-workloads/ example workloads are coming up healthy properly.
The links of yamls are broken currently but will be fixed with https://github.com/longhorn/website/pull/159/files

RWX functionality is working fine.
Test cases executed from https://rancher.atlassian.net/wiki/spaces/LON/pages/515376490/RWX+Volumes

jaredkipe:
[...]

PR #1425, opened 2020-05-28
Add example deployment of nfs on top of a longhorn for RWX volumes

joshimoo:
Below are the example deployment for nfs on top of a longhorn volume.
The nfs volume can then be used in RWX mode.

**For quicker pod eviction:**
We add tolerations so that nfs provisioner pod gets evicted from a failing
node after 60 second + 30 grace period (relevant for va recovery policy).
Add liveness + readyness probe, so that no traffic gets routed to a failed
nfs server. Disable device based fsids (major:minor) since our block device
mapping can change from node to node, which makes the id's unstable.

**To enable nfs reconnection to the new pod ip via the service:**
We add noresvport to the nfs mount options
This makes the nfs client use a new src port for each tcp reconnect.
This way after a crash the faulty connection isn't kept alive in the
connection cache (nat). This should allow to resolve the cluster ip to
the new destination pod ip.

**Hardcoded service cluster ip**
We hardcode nfs service cluster ip, so that on a delete & recreate of the service the previous pv's still point to this nfs-provisioner. We cannot use the hostname since the actual host doesn't know how to resolve service addresses inside of the cluster.

longhorn/longhorn#1183

### DIFF
--- a/examples/rwx/02-longhorn-nfs-provisioner.yaml
+++ b/examples/rwx/02-longhorn-nfs-provisioner.yaml
@@ -102,6 +102,7 @@ spec:
                 - SYS_RESOURCE
           args:
             - "-provisioner=nfs.longhorn.io"
+            - "-device-based-fsids=false"
           env:
             - name: POD_IP
               valueFrom:
@@ -114,6 +115,20 @@ spec:
                 fieldRef:
                   fieldPath: metadata.namespace
           imagePullPolicy: "IfNotPresent"
+          readinessProbe:
+            exec:
+              command:
+                - ls
+                - /export
+            initialDelaySeconds: 5
+            periodSeconds: 5
+          livenessProbe:
+            exec:
+              command:
+                - ls
+                - /export
+            initialDelaySeconds: 5
+            periodSeconds: 5
           volumeMounts:
             - name: export-volume
               mountPath: /export
@@ -121,6 +136,17 @@ spec:
         - name: export-volume
           persistentVolumeClaim:
             claimName: longhorn-nfs-provisioner
+      # we want really quick failover
+      terminationGracePeriodSeconds: 30
+      tolerations:
+        - effect: NoExecute
+          key: node.kubernetes.io/not-ready
+          operator: Exists
+          tolerationSeconds: 60
+        - effect: NoExecute
+          key: node.kubernetes.io/unreachable
+          operator: Exists
+          tolerationSeconds: 60
 ---
 apiVersion: v1
 kind: PersistentVolumeClaim
@@ -139,5 +165,5 @@ kind: StorageClass
 metadata:
   name: longhorn-nfs # workload storage class
 provisioner: nfs.longhorn.io
-parameters:
-  mountOptions: "vers=4.1"
+mountOptions:
+  - "vers=4.1"
--- a/examples/rwx/02-longhorn-nfs-provisioner.yaml
+++ b/examples/rwx/02-longhorn-nfs-provisioner.yaml
@@ -1,0 +1,143 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: longhorn-nfs-provisioner
+---
+kind: Service
+apiVersion: v1
+metadata:
+  name: longhorn-nfs-provisioner
+  labels:
+    app: longhorn-nfs-provisioner
+spec:
+  ports:
+    - name: nfs
+      port: 2049
+    - name: nfs-udp
+      port: 2049
+      protocol: UDP
+    - name: nlockmgr
+      port: 32803
+    - name: nlockmgr-udp
+      port: 32803
+      protocol: UDP
+    - name: mountd
+      port: 20048
+    - name: mountd-udp
+      port: 20048
+      protocol: UDP
+    - name: rquotad
+      port: 875
+    - name: rquotad-udp
+      port: 875
+      protocol: UDP
+    - name: rpcbind
+      port: 111
+    - name: rpcbind-udp
+      port: 111
+      protocol: UDP
+    - name: statd
+      port: 662
+    - name: statd-udp
+      port: 662
+      protocol: UDP
+  selector:
+    app: longhorn-nfs-provisioner
+---
+kind: Deployment
+apiVersion: apps/v1
+metadata:
+  name: longhorn-nfs-provisioner
+spec:
+  selector:
+    matchLabels:
+      app: longhorn-nfs-provisioner
+  replicas: 1
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      labels:
+        app: longhorn-nfs-provisioner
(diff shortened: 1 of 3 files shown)


joshimoo:
Need to add liveness check

examples/nfs/01-security.yaml
```diff
@@ -0,0 +1,87 @@
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: longhorn-nfs-provisioner
```
meldafrawi, line 4: `longhorn-nfs-provisioner` is used across the file with different resource types
can we change the resources names to reflect it's types ?

joshimoo: Do we generally do that?

examples/rwx/02-longhorn-nfs-provisioner.yaml
```diff
@@ -139,5 +165,5 @@ kind: StorageClass
 metadata:
   name: longhorn-nfs # workload storage class
 provisioner: longhorn.io/nfs
-parameters:
-  mountOptions: "vers=4.1"
+mountOptions:
```
shuo-wu, line 172: Directly modify this part in the 1st commit

examples/rwx/02-longhorn-nfs-provisioner.yaml
```diff
@@ -114,13 +115,38 @@ spec:
...
+              command:
+                - ls
+                - /export
+            initialDelaySeconds: 5
+            periodSeconds: 5
+          livenessProbe:
+            exec:
+              command:
+                - ls
+                - /export
```
shuo-wu, line 133: Does this work?
If the NFS provisioner pod is using ext4, should this be `/export/lost+found`?

examples/rwx/02-longhorn-nfs-provisioner.yaml
```diff
@@ -102,6 +102,7 @@ spec:
                 - SYS_RESOURCE
           args:
             - "-provisioner=longhorn.io/nfs"
+            - "-device-based-fsids=false"
```
shuo-wu, line 109: For this arg, I cannot follow the explanation in the comment.
Ideally, once this pod is launched, the Longhorn volume should keep attached hence the mapping won't change. (In other words, I don't think the major/minor number of the Longhorn volume can be changed while the volume keeps attached.) If the volume gets detached unexpectedly, can the NFS server and its connections still work fine?

joshimoo: We cannot use the fsid since when the node the server lives on fails and the nfs server pod get's evicted and rescheduled onto another node. There can be a different set of volumes attached, so the major:minor is not stable across pods.

examples/rwx/02-longhorn-nfs-provisioner.yaml
joshimoo: The mount point is /export, so a ls operation will fail if the volume is unavailable:
`Liveness probe failed: ls: reading directory '/export': Input/output error`

examples/rwx/02-longhorn-nfs-provisioner.yaml
shuo-wu: If the arg is enabled and NFS-provisioner pod suddenly crashes then gets restarted, what will happen for the volumes created by the provisioned pods? Will those volumes be no longer usable? 
IMO, once the pod somehow is restarted on another node, all the client connections are invalid and users need to find a way to recover them. Since this arg `might affect client connections on restarts of the nfs provisioner pod` and the connections are already invalid, it's fine to set as True.

joshimoo: The way  I see it we want stable fsids, we cannot use the major:minor since it depends on the set of volumes attached on a node. So it's not stable across different nodes, therefore we should disable this.

examples/rwx/02-longhorn-nfs-provisioner.yaml
```diff
@@ -10,6 +10,10 @@ metadata:
   labels:
     app: longhorn-nfs-provisioner
 spec:
+  # hardcode a cluster ip for the service
+  # so that on delete & recreate of the service the previous pv's still point
+  # to this nfs-provisioner, pick a new ip for each new nfs provisioner
```
yasker, line 15: `rather than picking a new ip`?

yasker: Another concern is what if this cluster IP is taken? We probably want some IP that's obvious and document if the service creation fail, what should the user do.

examples/rwx/02-longhorn-nfs-provisioner.yaml
yasker: Since the new commit message which explicitly explained why is this needed, this is mostly fine.

The alternative would be adding the commit message to the first commit as well, which may seem a bit weird to explain one of the parameters but ignoring others.

examples/rwx/02-longhorn-nfs-provisioner.yaml
joshimoo: The reason this is necessary is that the pvc get's the servers ip set as the target. Unfortunatly we cannot use the service name, since the nodes by default aren't running kube-dns and cannot resolve internal cluster services. So we need to stick with a fixed service ip, but the choice of the service ip can be arbitrary by the user. By default Kubernetes assigns one ip randomly when the service gets created. We could continue with that, approach but then on deletion + recreation of the service the old pvc's would no longer be accessible.

I will add information about this to the documentation. Is there a specific ip that we should take?

yasker: I don't have a preference for IP. I will leave that to you.

examples/rwx/02-longhorn-nfs-provisioner.yaml
```diff
@@ -0,0 +1,174 @@
...
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: "20G" # make this 10% bigger then the workload pvc
+---
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: longhorn-nfs # workload storage class
+provisioner: longhorn.io/nfs
```
yasker, line 171: Change the provisioner to `nfs.longhorn.io` to be consistent with `driver.longhorn.io` which we used for RWO volumes.

examples/rwx/02-longhorn-nfs-provisioner.yaml
joshimoo: I changed the ip to: `10.43.111.111` and will mention this in the documentation :)

yasker:
LGTM. @meldafrawi @shuo-wu ?

joshimoo:
@yasker can we merge?

Issue #520, opened 2019-05-10
Deleting a volume will not delete its PV/PVC. 


meldafrawi:
- Using UI:
  - Create a volume.
  - Create PV/PVC
  - Delete the volume.
- Using `kubectl get pv`
  - PV/PVC not deleted.

yasker:
We need to think more about deleting a volume.

Currently deleting a volume won't be blocked by any operations.
1. PV can be deleted while deleting the volume. But PVC is the declaration of a workload. User can manually create it for its workload. I don't think we can just delete it.
2. If there is workload current using the volume, should we allow deleting the volume? Or, should we even allow detach the volume?

yasker:
keep it open

ttpcodes:
@joshimoo and I had a discussion regarding handling of `Volume` deletion in various cases in which a `PersistentVolume` and `PersistentVolumeClaim` are involved. This is what we eventually decided on:
- It's possible that there are other inactive workloads such as a `Deployment` scaled to zero `Pods` that depend on the `Volume`. While we'd be able to successfully delete the `PersistentVolume` and `PersistentVolumeClaim`, it would cause problems once the `Deployment` scales back up and discovers that the `PersistentVolumeClaim` and other resources that it depends on no longer exists. We determined that it might be best to warn the user before deletion that there is a `PersistentVolume` and possibly `PersistentVolumeClaim` related to the `Volume` and  to be aware in case that it affects their workloads. Afterwards, they can delete the `Volume` with the option to have `Longhorn` handle the `PersistentVolume` and `PersistentVolumeClaim` deletion for them.
- Deleting a `Volume` with running workloads would cause obvious problems for the end user, and we should at the very least, block a user from deleting a `Volume` that has a workload attached.

I think it's safe to go ahead with working on the second case and preventing deletion of `Volumes` with associated workloads. As for the first case, @yasker do you have any additional thoughts on how this should be handled? Or does this look good and can we proceed with implementing it?

yasker:
@ttpcodes 

I've thought about this a bit more.

1. We should know if the volume is referred by a PV and PVC, since any reference to the PVC will be written in the `claimRef` in the PV. We also know if the workload using the volume exists or not. We can provide information to the user, and make sure user checks the confirmation checkbox before deleting the volume.
[...]

PR #1528, opened 2020-06-26
Add Longhorn Enhancement Proposal for handling Volume deletion

ttpcodes:
This PR adds a Longhorn Enhancement Proposal for how to handle `Volume` deletion in the cases that a `Volume` has associated resources in `Kubernetes` or is `Attached`. The changes outlined in this proposal will make deletion of `Volumes` in these situations safer by avoiding leaving leftover `Persistent Volumes` or `Persistent Volume Claims` that would be unusable and warning users about possible repercussions as a result of deleting an attached `Volume`.

This proposal provides a solution for the cases described in #520.

### DIFF
--- a/enhancements/20200625-volume-deletion-flows.md
+++ b/enhancements/20200625-volume-deletion-flows.md
@@ -1,0 +1,78 @@
+# Volume Deletion Flows
+
+## Summary
+This enhancement modifies the flow a user would follow for handling deletion of `Volumes` that are `Attached` or otherwise have resources such as a `Persistent Volume` associated with them. Specifically, this adds warnings in the `longhorn-ui` when deleting a `Volume` in these specific cases and provides a means for the `longhorn-manager` to clean up any leftover resources in `Kubernetes` associated with a deleted `Volume`.
+
+### Related Issues
+https://github.com/longhorn/longhorn/issues/520
+
+## Motivation
+
+### Goals
+The goal of this enhancement is to either address or warn users about situations in which deleting a `Volume` could cause potential problems. In handling the case of cleaning up an associated `Persistent Volume` (and possibly `Persistent Volume Claim`), we can prevent there being leftover unusable `Volume`-related resources in `Kubernetes`. In warning about deletion when the `Volume` is attached, we can inform the user about possible consequences the deletion would have on existing workloads so the user can handle this accordingly.
+
+### Non-Goals
+This enhancement is not intended on completely blocking a user from pursuing any dangerous operations. For example, if a user insists on deleting a currently attached `Volume`, they should not be forbidden from doing so in case the user is absolutely sure that they want to follow through.
+
+## Proposal
+When a user wishes to delete a `Volume` from the `Longhorn UI`, the system should check to see if the `Volume` has a resource tied to it or is currently `Attached`:
+- If the `Volume` is `Attached`, the user should be warned about the potential consequences of deleting the `Volume` (namely that any applications currently using the `Volume` will no longer have access to it and likely error out) before they can confirm the deletion or cancel it.
+- If the `Volume` is tied to a `Volume` that is tied to a `Persistent Volume` (and possibly a `Persistent Volume Claim`), the user should be informed of this information and the fact that we will clean up those resources if the `Volume` is deleted. If the `Volume` is tied to a `Persistent Volume Claim`, the user should also be warned that there may be `Deployments` or `Stateful Sets` that depend on this `Volume` that could no longer work should the user delete the `Volume` (we cannot explicitly see this without having to monitor all `Deployments` and `Stateful Sets` to check if they use a `Longhorn`-backed `Persistent Volume Claim`). Afterwards, the user can confirm the deletion if they wish, which will lead to cleanup of the associated resources and deletion of the `Volume`.
+
+### User Stories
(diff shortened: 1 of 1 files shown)


enhancements/20200625-volume-deletion-flows.md
```diff
@@ -0,0 +1,74 @@
+# Volume Deletion Flows
+
+## Summary
+This enhancement modifies the flow a user would follow for handling deletion of `Volumes` that are `Attached` or otherwise have resources such as a `Persistent Volume` associated with them. Specifically, this adds warnings in the `longhorn-ui` when deleting a `Volume` in these specific cases and provides a means for the `longhorn-manager` to clean up any leftover resources in `Kubernetes` associated with a deleted `Volume`.
+
+### Related Issues
+#520
```
yasker, line 7: Use a link here.

enhancements/20200625-volume-deletion-flows.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Non-Goals
+This enhancement is not intended on completely blocking a user from pursuing any dangerous operations. For example, if a user insists on deleting a currently attached `Volume`, they should not be forbidden from doing so in case the user is absolutely sure that they want to follow through.
+
+## Proposal
+When a user wishes to delete a `Volume` from the `Longhorn UI`, the system should check to see if the `Volume` has a resource tied to it or is currently `Attached`:
+- If the `Volume` is `Attached`, the user should be warned about the potential consequences of deleting the `Volume` (namely that any applications currently using the `Volume` will no longer have access to it and likely error out) before they can confirm the deletion or cancel it.
+- If the `Volume` is tied to a `Volume` that is tied to a `Persistent Volume` (and possibly a `Persistent Volume Claim`), the user should be informed of this information and the fact that we will clean up those resources if the `Volume` is deleted. Afterwards, the user can confirm the deletion if they wish, which will lead to cleanup of the associated resources and deletion of the `Volume`.
```
yasker, line 20: Should we give users an option to clean up PV/PVC as well or we should just do it? Trying to think of the cases that user want to delete the volume but not PV/PVC. Restore? Though the default will be clean up PV/PVC if we decide to give the user a choice.

yasker:
One thing I have a bit concern is how are we going to delete the volume provisioned by the storage class. We might want to clarify that, since deleting the PVC/PV will automatically trigger the volume deletion.

enhancements/20200625-volume-deletion-flows.md
ttpcodes: Joshua and I were thinking about the `Persistent Volume` and `Persistent Volume Claim` cleanup and also thought about the `Restoration` case. We weren't sure if users would more frequently create a new `Persistent Volume` and `Persistent Volume Claim` during the restoration process or leave the original resources alone and then restore on top of them.

If it's the latter, then I think it's reasonable to provide an option to clean up `Persistent Volume`. I would need to rethink the deletion operation then though, because the `DELETE` methods generated by `go-rancher` don't seem to take in additional input for the `DELETE` method and we'd need to figure out how to work around that.

yasker: We can add parameters to the DELETE API call. Empty would be the default. See https://github.com/longhorn/longhorn-manager/blob/master/api/model.go#L125 for an example.

ttpcodes: It appears that the default `Delete` method that's generated by `go-rancher` for us to use the `API` programmatically in `Go` can't take parameters, and requiring parameters in the `API` would break our generated client. For now, we can delete the `Persistent Volume` and `Persistent Volume Claim` automatically and then wait for changes in the generated client that would allow us to add parameters to the `DELETE` call.

ttpcodes:
@yasker There shouldn't be problems if the `Volume` was provisioned by the `Storage Class`. The `CSI` driver would attempt to delete the `Volume` and the call would return successfully, allowing the `Persistent Volume` and `Persistent Volume Claim` to be deleted and for the `Volume` itself to actually delete.

@joshimoo Since the `Volume` would be gone already and the `Retain` policy for a `Persistent Volume` is primarily relevant in determining what to do if the `Volume` still exists and the `Persistent Volume Claim` has been deleted, I think it makes more sense for any decision regarding retaining the `Persistent Volume` and `Persistent Volume Claim` to be made as part of the `Volume` deletion dialog in the `longhorn-ui`. That's currently blocked though since `go-rancher` doesn't take a request body input in it's `Resource` deletion calls.

@shuo-wu I didn't end up commenting on `Deployments` and `Stateful Sets` because in my original discussion with @joshimoo on this issue, we determined that we'd have to monitor all `Deployments`, `Stateful Sets`, and any other resources with possible `Persistent Volume Claim` references to find those that did have references to a `Longhorn`-backed `Volume` (which seemed less than ideal). Even if we did pick up on the existence of the `Deployments` and `Stateful Sets`, handling them as part of our cleanup of associated resources seems a step too far (though we could give the user a warning that these `Deployments` and `Stateful Sets` depend on our `Volume`.

shuo-wu:
OK. Then I think adding a warning message before the PV/PVC cleanup is enough. And it may be better to explicitly mention the reason for this warning in the doc.

yasker:
LGTM. @shuo-wu ?

PR #1550, opened 2020-07-06
example: Fix example rwx

shuo-wu:
1. Use different names for the 2 roles
(and the related role bindings)
2. Make sure the role binding namespace is the same as the
namespace where nfs provisioner is deployed

### DIFF
--- a/examples/rwx/01-security.yaml
+++ b/examples/rwx/01-security.yaml
@@ -63,7 +63,8 @@ roleRef:
 kind: Role
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: longhorn-nfs-provisioner
+  name: leader-locking-longhorn-nfs-provisioner
+  namespace: longhorn-system
 rules:
   - apiGroups: [""]
     resources: ["endpoints"]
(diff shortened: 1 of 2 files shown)


examples/rwx/01-security.yaml
```diff
@@ -54,7 +54,7 @@ metadata:
 subjects:
   - kind: ServiceAccount
     name: longhorn-nfs-provisioner
-    namespace: longhorn-system
+    namespace: default
```
joshimoo, line 57: QA asked me to move these into the longhorn-system namespace. If I forgot to do so for 02-provisioners please feel free to change it there.

examples/rwx/01-security.yaml
```diff
@@ -63,7 +63,7 @@ roleRef:
 kind: Role
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: longhorn-nfs-provisioner
+  name: leader-locking-longhorn-nfs-provisioner
```
joshimoo, line 66: Not sure why they need different names since they are different types?
But feel free to change the name :)

shuo-wu: Using the same name with the above ClusterRole confuses me a little bit at a first glance. I think it might be better to rename this Role and the related RoleBinding.

PR #1573, opened 2020-07-14
Include Priority Class in v1.0.1-rc1 chart

ttpcodes:
This PR updates the Longhorn v1.0.1-rc1 chart to include missing changes to the Longhorn deploy files such as the Priority Class access needed in the Cluster Role for the Longhorn Manager to work properly.


yasker:
We always use the master for the latest stable release. So we won't merge rc1 into master now.

PR #1582, opened 2020-07-15
chart: Sync up with longhorn-manager deployment files

shuo-wu:


yasker:
It should be against the v1.0.1 branch.

shuo-wu:
Michael had updated the related files. Sorry for this duplicate PR.

Issue #1590, opened 2020-07-16
[BUG] The longhorn-csi-plugin still try to pull images from public registry in air-gap setup


PhanLe1010:
**Describe the bug**
Install longhorn using Rancher UI with private registry. All components are pulling images from the private registry except for the `longhorn-csi-plugin` . The `longhorn-csi-plugin` still try to pull images from public registry.

**To Reproduce**
Install longhorn using Rancher UI with a private registry.

**Expected behavior**
All components are pulling images from the private registry.

**Environment:**
 - Longhorn version: v1.0.1-rc1
 - Kubernetes version: v1.17.6
 - Node OS type and version:  Ubuntu 18.04.4 LTS

**Additional context**
This problem happens when install Longhorn using Rancher UI. `kubectl apply -f` works fine. Not sure about `Helm chart` installation.

PhanLe1010:
## Pre-merged Checklist
* [x] Does the PR include the explanation for the fix or the feature?

* [x] Is the backend code merged (Manager, Engine, Instance Manager, BackupStore etc)?
The PR is at https://github.com/longhorn/longhorn/pull/1591

* [x] Is the reproduce steps/test steps documented?

* [ ] Which areas/issues this PR might have potential impacts on?
Area air-gap installation with private registry

* [x]~~ **If the fix introduces the code for backward compatibility** Has an separate issue been filed with the label `release/obsolete-compatibility`?
The compatibility issue is filed at~~

* [x] ~~**If labeled: area/ui** Has the UI issue filed or ready to be merged?
The UI issue/PR is at~~

* [x] ~~**if labeled: require/doc** Has the necessary document PR submitted or merged?
The Doc issue/PR is at~~

* [x] ~~**If labeled: require/automation-e2e** Has the end-to-end test plan been merged? Have QAs agreed on the automation test case?
The automation skeleton PR is at
The automation test case PR is at~~

* [x] ~~**if labeled: require/automation-engine** Has the engine integration test been merged?
The engine automation PR is at~~

* [x] ~~**if labeled: require/manual-test-plan** Has the manual test plan been documented?
The updated manual test plan is at~~

PhanLe1010:
**To Reproduce**

Install longhorn using Rancher UI and setup a private registry to pull images from. 

Bug: `longhorn-csi-plugin` DaemonSet still trying to pull image `csi-node-driver-registrar` from public registry.

meldafrawi:
Validation: **PASSED**

PR #1591, opened 2020-07-16
chart: change the CSI Driver Registrar image variable in question.yaml to have the same name as in values.yaml

PhanLe1010:
This makes sure that the Rancher chart correctly passes the user-provided CSI Driver Registrar image registry value to `values.yaml` and therefore to the yaml template `deployment-driver.yaml`

longhorn/longhorn#1590

### DIFF
--- a/chart/questions.yml
+++ b/chart/questions.yml
@@ -70,11 +70,11 @@ questions:
     type: string
     label: Longhorn CSI Provisioner Image
     group: "Longhorn CSI Driver Images"
-  - variable: csi.driverRegistrarImage
+  - variable: csi.nodeDriverRegistrarImage
     default:
-    description: "Specify CSI Driver Registrar image. Leave blank to autodetect."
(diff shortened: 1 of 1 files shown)


meldafrawi:
LGTM

chart/questions.yml
```diff
@@ -70,7 +70,7 @@ questions:
     type: string
     label: Longhorn CSI Provisioner Image
     group: "Longhorn CSI Driver Images"
-  - variable: csi.driverRegistrarImage
+  - variable: csi.nodeDriverRegistrarImage
     default:
     description: "Specify CSI Driver Registrar image. Leave blank to autodetect."
```
yasker, line 75: Need to change the name of the field to `Node Driver Registrar` (instead of `Driver Registrar`)

Same below.

`Driver Registrar` is another component of CSI.

See https://github.com/kubernetes-csi/node-driver-registrar vs https://github.com/kubernetes-csi/driver-registrar

PR #1634, opened 2020-07-24
Helm chart ships the pod security policy required to run LH in CIS hardened clusters

janeczku:
This PR allows Longhorn to successfully run in a CIS hardened Cluster using the Helm chart install. See https://github.com/longhorn/longhorn/issues/608

To test:
1. Create a cluster hardened according to CIS Kubernetes benchmark
2. Deploy with current release Helm chart -> Longhorn pods fail to start / crashloop
3. Deploy with Helm chart per this PR -> Longhorn successfully installed


kursattokpinar:
Is PSP used here designed with 'least privileges'?

janeczku:
@kursattokpinar PSP here uses recommended "restricted" CIS Kubernetes benchmark policy as starting point,  extended only with the least privileges required to run Longhorn pods.

yasker:
@janeczku In fact, Kubernetes community is planning to depreciate the PSP soon, see https://github.com/kubernetes/enhancements/issues/5#issuecomment-637066475

chart/values.yaml
```diff
@@ -114,3 +114,7 @@ ingress:
   # - name: longhorn.local-tls
   #   key:
   #   certificate:
+
+# Setup pod security policy in the Longhorn namespace to allow privileged pods
+podSecurityPolicy:
+  enabled: true
```
ttpcodes, line 120: Are there any other fields that we'd put under the `podSecurityPolicy` value? If not, I think it makes sense to just condense this key into something like `enablePSP` for the user.

janeczku: Modified per your suggestion.

PR #1644, opened 2020-07-29
fix: remove beta annotation from storage class

f0rmiga:
Since the storage.k8s.io/v1 API, the annotation for is-default-class is no longer in beta and was deprecated.
https://github.com/kubernetes/kubernetes/pull/40088


yasker:
Thanks @f0rmiga . I think you meant the `beta` version was deprecated, not the annotation.

Otherwise LGTM.

Issue #1455, opened 2020-06-04
[FEATURE] Backup/Restore Longhorn System


yasker:
Rancher has introduced ETCD snapshot/restore, but [we found that](#1969) it doesn't work well with Longhorn. More importantly, there is currently no recommended way to backup the Longhorn system itself. It can be useful for the following scenarios:
1. Rollback in the case of an upgrade failure.
2. Recreate/replicate the cluster

In general, we can dump the related CRDs and rollback to them if something went wrong.

It requires some designs to offer the best experience to the end user.

REF: https://github.com/longhorn/longhorn/pull/1648

innobead:
This is what we have now for backup/restore longhorn.
- https://longhorn.io/docs/1.3.0/advanced-resources/cluster-restore/restore-to-a-cluster-contains-data-using-rancher-snapshot/
- https://longhorn.io/docs/1.3.0/advanced-resources/cluster-restore/restore-to-a-new-cluster-using-velero/

c3y1huang:
Note: @FrankYang0529 is currently working on backing image backup and restore.
1. Add backup cmd to the backing image manager to upload to the backup targets.
2. Add fields `lastBackingImage`... to the related CR

For this issue, we can leverage the field `lastBackingImage` for restoring the rollout.

Relate issue https://github.com/harvester/harvester/issues/2237

c3y1huang:
Backing image backup/restore, ref https://github.com/longhorn/longhorn/pull/4385.

khushboo-rancher:
Testing in progress, will update the result here.

PR #1648, opened 2020-07-30
Add Longhorn Enhancement Proposal for upgrade rollback

ttpcodes:
This PR adds a Longhorn Enhancement Proposal outlining a mechanism for supporting upgrade rollback to a previously installed version, a feature requested in #1455. The basic mechanism outlined in this proposal suggests storing the state of Longhorn's CRD instances on an annotation for that instance, and then having the Upgrade Controller reapply that annotation's state in the case the user has rolled back and is starting a previous version of Longhorn.

### DIFF
--- a/enhancements/20200727-upgrade-rollback.md
+++ b/enhancements/20200727-upgrade-rollback.md
@@ -1,0 +1,58 @@
+# Upgrade Rollback
+
+## Summary
+This enhancement adds a mechanism for users to be able to rollback from later versions of Longhorn and restore previous versions of their CRDs.
+
+### Related Issues
+https://github.com/longhorn/longhorn/issues/1455
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


ttpcodes:
The proposal is currently a draft, but I'd like some feedback on this before I keep tuning up this proposal.

shuo-wu:
I have some questions after checking this LEP:
1. For CRD instance managers, the state of the CRD objects is determined by the actual status of the related engine/replica processes. Can we directly overwrite the current status with the value in the label during the rollback?
2. What if users decide to do rollback after upgrading the image for some volumes? And do we need to consider the compatibility issue if there is CLI API version modification during the upgrade & rollback? (I think we need more user stories to clarify these cases.)
3. Do we need to consider the CRD definition update?

Issue #298, opened 2018-09-21
Manager: Add replica eviction support for node and disk


yasker:
https://github.com/rancher/longhorn/issues/292

Overview:
Eviction mode would give users the ability to work on scheduled maintenance without impact the volumes.

Pre-requirement:
1. The eviction can only be called if the scheduling was disabled.
2. During the eviction process, the scheduling cannot be re-enabled.

Description:
The eviction would result in all the replicas on the node or disk being moved to other nodes/disks.

Design:
1. If the volume is not in the healthy state, the detection of evicting would be skipped.
2. The volume of the replica which being moved will temporarily increase replica count by one. The rebuilding process should be triggered then, and the new replica will be automatically scheduled to another suitable place.
2.1. There can be multiple volumes in the replica moving process at the same time, but only one replica for each volume can be moved at any given time, even there are multiple disk/nodes are being evicted.
3. If there is no space on other nodes/disks to fit the new replica, the volume will report `unable to schedule` error, but keep functioning. The user can choose to stop evicting for the nodes/disks at this point or clean up more spaces for the replicas.
4. If the user chooses to stop evicting when one volume is `unable to schedule`, the evicting for that volume would be canceled, and the replica would remain at the disk, and the replica count for the volume would be automatically restored back to the original setting.
5. If the new replica has been successfully rebuilt, the volume would have `replica count + 1` healthy replicas. After the new replica has been confirmed by Longhorn engine, the controller would restore the replica count to the original setting, then destroy the replica on the node/disk that being evicted. The controller will make sure that at any given time, the healthy replica counts won't be below the original replica count.
7. After one replica has been successfully moved, the controller would repeat this process for the other replicas of the volume. Until there is no replica needs to be moved.

Additional requirement:
Make sure to cover the case of disk replacement/node replacement.

boknowswiki:
I think we need @smallteeths to help adding buttons for `Eviction` and `Cancel Eviction` for disks and nodes from UI.

yasker:
This issue has been in Review for 13 days. What's blocking this issue moving to Merged?

boknowswiki:
> This issue has been in Review for 13 days. What's blocking this issue moving to Merged?

I think it's because of [this part](https://github.com/longhorn/longhorn-manager/pull/651#discussion_r474952588). I will sync with @shuo-wu about it later today.

boknowswiki:
[...]

PR #1668, opened 2020-08-05
Add replica eviction support for disks and nodes enhancement document

boknowswiki:
https://github.com/longhorn/longhorn/issues/298

Signed-off-by: Bo Tao <bo.tao@rancher.com>

### DIFF
--- a/enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
+++ b/enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
@@ -1,0 +1,86 @@
+# Replica Eviction Support for Disks and Nodes
+
+## Summary
+This enhancement is to simplify and automatically evict the replicas on the selected disabled disks or nodes to other suitable disks and nodes per user's request. Meanwhile keep the same level of fault tolerance during this eviction period of time.
+
+### Related Issues
+https://github.com/longhorn/longhorn/issues/292
+https://github.com/longhorn/longhorn/issues/298
+
+## Motivation
+
+### Goals
+1. Allow user easily evict the replicas on the selected disks or nodes to other disks or nodes without impact the user defined `Volume.Spec.numberOfReplicas` and keep the same level of fault tolerance. This means we don't change the user defined replica number.
+2. Report any error to user during the eviction time.
+3. Allow user to cancel the eviction at any time.
+
+### Non-goals
+
+For now, we only focus on `Healthy` volumes, which means `Attached` and `Healthy` or the volume can be in `Degraded` state with rebuilding a new replica. If the volume is `Detached`, the eviction won't work for this volume.
+
+## Proposal
+1. Add `Eviction Requested` with `true` and `false` selection buttons for disks and nodes. This is for user to evict or cancel the eviction of the disks or the nodes.
+2. Add new `evictionRequested` field to `Node.Spec`, `Node.Spec.disks` Spec and `Replica.Status`. These will help tracking the request from user and trigger replica controller to update `Replica.Status` and volume controler to do the eviction. And this will reconcile with `scheduledReplica` of selected disks on the nodes.
+3. Display `fail to evict` error message to `Dashboard` and any other eviction errors to the `Event log`.
+
+### User Stories
+### Disks and Nodes Eviction
+For disk replacement or node replacement, the eviction needs to be done successfully in order to guarantee Longhorn volume function properly.
+
+Before, when user wants to evict a disk or a node they need to do the following steps:
+
+1. User needs to disable the disk or the node.
+2. User needs to scale up the replica count for the volume which has replica on disabled disks or nodes, and wait for the rebuild complete, scale down the replica count, then delete the replicas on this disk or node.
+
+After this enhancement, user can click `true` to the `Eviction Requested` on scheduling disabled disks or nodes. Or select `Disable` for scheduling and `true` to the `Eviction Requested` at the same time then save this change. The backend will take care of the eviction for the disks or nodes and cleanup for all the replicas on disks or nodes.
+
+### User Experience In Detail
+#### Disks and Nodes Eviction
+1. User can select `true` to the `Eviction Requested` from `Longhorn UI` for disks or nodes. And user has to make sure the selected disks or nodes have been disabled, or select the `Disable` Scheduling at the same time of `true` to the `Eviction Requested`.
+2. Once `Eviction Requested` has been set to `true` on the disks or nodes, they can not be enabled for `Scheduling`.
+3. If the disks or the nodes haven't been disabled for `Scheduling`, there will be error message showed in `Dashboard` immediatly to indicate that user need to disable the disk or node for eviction.
+4. And user will wait for the replica number for the disks or nodes to be 0.
+5. If there is any error e.g. no space or couldn't find other schedulable disk, the error message will be logged in the `Event log`. And the eviction will be suspended until either user sets the `Eviction Requested` to `false` or cleanup more disk spaces for the new replicas.
+6. If user cancel the eviction by setting the `Eviction Requested` to `false`, the remaining replicas on the selected disks or nodes will remain on the disks or nodes.
+
+### API changes
+From an API perspective, the call to set `Eviction Requested` to `true` or `false` on the `Node` or `Disk` eviction should look the same. The logic for handling the new field `Eviction Requested` `true` or `false` should to be in the `Node Controller` and `Volume Controller`.
+
+## Design
+
+### Implementation Overview
+
+1. On `Longhorn UI` `Node` page, for nodes eviction, adding `Eviction Requested` `true` and `false` options in the `Edit Node` sub-selection, next to `Node Scheduling`. For disks eviction, adding `Eviction Requested` `true` and `false` options in `Edit node and disks` sub-selection under `Operation` column next to each disk `Scheduling` options. This is for user to evict or cancel the eviction of the disks or the nodes.
+2. Add new `evictionRequested` field to `Node.Spec`, `Node.Spec.disks` Spec and `Replica.Status`. These will help tracking the request from user and trigger replica controller to update `Replica.Status` and volume controler to do the eviction. And this will reconcile with `scheduledReplica` of selected disks on the nodes.
+3. Add a informer in `Replica Controller` to get these information and update `evictionRequested` field in `Replica.Status`.
+4. Once `Eviction Requested` has been set to `true` for disks or nodes, the `evictionRequested` fields for the disks and nodes will be set to `true` (default is `false`).
+5. `Replica Controller` will update `evictionRequested` field in `Replica.Status` and `Volume Controller` to get these information from it's replicas.
+6. During reconcile the engine replica, based on `Replica.Status.EvictionRequested` of the volume replicas to trigger rebuild for different volumes' replicas. And remove one replica with `evictionRequested` `true`.
+7. Logged the errors to `Event log` during the reconcile process.
+8. By the end from `Longhorn UI`, the replica number on the eviction disks or nodes should be 0, this mean eviction is success.
+
+### Test plan
+
+#### Manual Test Plan For Disks and Nodes Eviction
+Positive Case:
+
+For both `Replica Node Level Soft Anti-Affinity` has been enabled and disabled.
+1. User can select one or more disks or nodes for eviction. Select `Eviction Requested` to `true` on the disabled disks or nodes, Longhorn should start rebuild replicas for the volumes which have replicas on the eviction disks or nodes, and after rebuild success, the replica number on the evicted disks or nodes should be 0. E.g. When there are 3 nodes in the cluster, and with `Replica Node Level Soft Anti-Affinity` is set to `false`, disable one node, and create a volume with replica count 2. And then evict one of them, the eviction should get stuck, then set `Replica Node Level Soft Anti-Affinity` to `true`, the eviction should go through.
+
+Negative Cases:
+1. If user selects the disks or nodes have not been disabled scheduling, Longhorn should display the error message on `Dashboard` immediatly. Or during the eviction, the disabled disk or node can not be re-enabled again.
+2. If there is no enough disk spaces or nodes for disks or nodes eviction, Longhorn should log the error message in the `Event Log`. And once the disk spaces or nodes resources are good enough, the eviction should continue. Or if the user selects `Eviction Requested` to `false`, Longhorn should stop eviction and clear the `evictionRequested` fields for nodes, disks and volumes crd objects. E.g. When there are 3 nodes in the cluster, and the volume replica count is 3, the eviction should get stuck when the `Replica Node Level Soft Anti-Affinity` is `false`.
+
+#### Integration Test Plan
+For `Replica Node Level Soft Anti-Affinity` is enabled, create 2 replicas on the same disk or node, and then evict this disk or node, the 2 replicas should goto another disk of node.
+
+For `Replica Node Level Soft Anti-Affinity` is disabled, create 1 replica on a disk, and evict this disk or node, the replica should goto the other disk of node.
+
+For node eviction, Longhorn will process the evition based on the disks for the node, this is like disk eviction. After eviction success, the replica number on the evicted node should be 0.
+
+#### Error Indication
+During the eviction, user can click the `Replicas Number` on the `Node` page, and set which replicas are left from eviction, and click the `Replica Name` will redirect user to the `Volume` page to set if there is any error for this volume. If there is any error during the rebuild, Longhorn should display the error message from UI. The error could be `failed to schedule a replica` due to disk space or based on schedule policy, can not find a valid disk to put the relica.
+
+### Upgrade strategy
+No special upgrade strategy is necessary. Once the user upgrades to the new version of `Longhorn`, these new capabilities will be accessible from the `longhorn-ui` without any special work.
+


enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,96 @@
...
+- For disk replacement or node replacement, the eviction needs to be done successfully in order to guarantee longhorn volume function properly.
+
+On `Longhorn UI` `Node` page, for nodes eviction, adding eviction options in the `Edit Node` sub-selection, next to `Node Scheduling`. For disks eviction, adding eviction options in `Edit node and disks` sub-selection under `Operation` column next to each disk `Scheduling` options.
+
+For eviction `Enable`, this will starts to evict the replicas on the selected disks or nodes. Any error happened during the eviction will be showed from `Longhorn UI`. Once the eviction is done for disks or nodes, the replica count displayed by `Replicas` column will be 0.
+For eviction `Disable`, this will be the default setting, and no operation will be done.
+
+### User Stories
+### Disk Eviction
+Before, when user wants to evict a disk, user needs to disable the disk first, then delete all the replicas on this disks after the deleted replica has been rebuilt successfully.
```
shuo-wu, line 28: Add some detailed descriptions. e.g., `disable scheduling of the disk first, then find out and delete all the replicas on this disks for each volume manually`

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,96 @@
...
+
+After this enhancement, user can do only one click to enable the eviction on scheduling disabled disk or select `Disable` for scheduling and `Enable` for eviction at the same time then save this change. The backend will take care of the replica reuild and cleanup for all the replicas on this disk.
+
+#### Node Eviction
+Before, when user wants to evict a node, user needs to disable the node first, then delete all the replicas on the disks of this node one by one after the deleted replica has been rebuilt successfully.
+
+After this enhancement, user can do only one click to enable the eviction on scheduling disabled node or select `Disable` for scheduling and `Enable` for eviction at the same time then save this change. The backend will take care of the replica reuild and cleanup for all the replicas on the disks of this node.
+
+### User Experience In Detail
+#### Disk Eviction
```
shuo-wu, line 38: Step 2~4 are like the implementation details rather than the user experience.

Similar below.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,96 @@
...
+1. User can selects a disk or disks to evict the replicas from `Longhorn UI`.
+2. From the disks, longhorn can get all the replicas currently saving on the disks.
+3. From the replicas, longhorn can get all the volumes of the replicas.
+4. Longhorn will temporarily increase the replica conut by one to the volumes, and this will trigger the rebuid process for the volumes, and new replica will be scheduled to another schedulable disks. Only one replica for each volume can be triggered for rebuild at any given time.
+5. If the disk or disks eviction finished successfully, the replica count on the `Replica` column should be 0. And all the replicas on the selected disk or disks will be removed automatically.
+6. If there is any error e.g. no space or couldn't find other schedulable disk, the error message will be display on the `Longhorn UI`. And the eviction for the failed volume would be changed to `Disable` automatically.
```
shuo-wu, line 44: How do users find the related errors? In `Dashboard/Event Log`?

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,96 @@
...
+Add a new condition to node and disk named `Eviction`, and status could be `True` or `False`.
+
+For the `replica eviction state`, there could be 3 states:
+1. `Disabled` means not eviction.
+2. `Enabled` means eviction.
+3. `Evited` means eviction is done successfully on this replica.
+
+And also add a `errMessage` field for each replica, and if there is any error, we update the error here, and display them in the UI.
+
+Also disk and node controller can take this `errMessage` field to indicate the eviction for the disk or the node has failed, and updatet the disk and node eviction condition to `Disable`.
```
shuo-wu, line 73: 1. If the eviction fails, can we put the error in the condition message then somehow show it in UI?
2. Typo `updatet`

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,96 @@
...
+
+Initially, for every replica created, the default stats is `Disabled`. And if user select a node or disk to evcit, and the replica on that disk will be set to `Enabled`, and this should trigger volume replica count increase by 1, and check the volume or engine rebuild status. And if the reuild success and the volume ends at healthy state, the replica state will be set to `Evicted`, and this replica can be removed safely. Otherwise report the error and set it's state to `Disabled`, and decrease the volume replica count by 1.
```
shuo-wu, line 75: 1. Typo `stats`
2. Put this part together with the `replica eviction state` introduction.
3. It seems we need a new field in `replica.Spec` besides `replica eviction state`.
4. Why using a new field `replica eviction state` rather than adding a condition `Eviction` in `replica.Status`?

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,96 @@
...
+
+Replica controller will check each replicas' state and then process for the next state.
+Each replica state transition should be independent to each other, can be process parallel. And for the `Enabled` state, can check if the rebuild for the volume is done and the volume is in healthy state.
+
+### Test plan
+
+#### For Disk Eviction
```
shuo-wu, line 82: It this manual test or integration test? What's the test skeleton/step? Putting a draft version here is fine.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,96 @@
...
+#### For Node Eviction
+For one node and multiple node case, user can select one or more node for eviction. Select eviction on a disabled node, and Longhorn will process the evition based on the disk for each node, this is like disk eviction. After eviction success, the replica number on the evicted node should be 0.
+
+#### Error Indication
+If there is any error during the rebuild, Longhorn should display the error message from UI. The error could be `failed to schedule a replica` due to disk space or based on schedule policy, can not find a valid disk to put the relica.
+
+### Upgrade strategy
+No special upgrade strategy is necessary. Once the user upgrades to the new version of `Longhorn`, these new capabilities will be accessible from the `longhorn-ui` without any special work.
+
+## Note [optional]
```
shuo-wu, line 94: Remove it if it's unnecessary.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,82 @@
...
+
+### API changes
+From an API perspective, the call made to `Evict` and `Unevict` the `Node` or `Disk` eviction should look the same. The logic for handling the new field `Evict` and `Unevict` should to be in the `Node Controller` and `Volume Controller`.
+
+## Design
+
+### Implementation Overview
+
+1. On `Longhorn UI` `Node` page, for nodes eviction, adding `Evict` and `Unevict` options in the `Edit Node` sub-selection, next to `Node Scheduling`. For disks eviction, adding `Evict` and `Unevict` options in `Edit node and disks` sub-selection under `Operation` column next to each disk `Scheduling` options. This is for user to evict or cancel the eviction of the disks or the nodes.
+2. Add new `requestEviction`  `type.Source` field to `Node.Spec` and `Node.Spec.disks`. This is to help tracking the request from user and trigger volume controler.
```
shuo-wu, line 52: What is `type.Source` for?

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,82 @@
...
+This enhancement is to simpler and automatically to evict the replicas on the selected disabled disks or nodes to other suitable disks and nodes. Meanwhile keep the same level of fault tolerance during this eviction period of time.
+
+### Related Issues
+https://github.com/longhorn/longhorn/issues/292
+https://github.com/longhorn/longhorn/issues/298
+
+## Motivation
+
+### Goals
+1. Allow user easily evict the replicas on the selected disks or nodes to other disks or nodes without impact the user defined `Volume.Spec` and keep the same level of fault tolerance.
```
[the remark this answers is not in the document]
boknowswiki: I think I should say `Volume.Spec.numberOfReplicas` to be more clear. This means we don't change the user defined replica number.

boknowswiki:
@shuo-wu , after synced with @yasker , we will focus on the `Attached` volume for now, and will revisit the `Detached` volume case after refactoring. So I will start to implement the code as the LEP described.

boknowswiki:
@shuo-wu I just update this as a RP, please take a look. Thanks.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,87 @@
...
+### API changes
+From an API perspective, the call made to `Evict` and `Unevict` the `Node` or `Disk` eviction should look the same. The logic for handling the new field `Evict` and `Unevict` should to be in the `Node Controller` and `Volume Controller`.
+
+## Design
+
+### Implementation Overview
+
+1. On `Longhorn UI` `Node` page, for nodes eviction, adding `Evict` and `Unevict` options in the `Edit Node` sub-selection, next to `Node Scheduling`. For disks eviction, adding `Evict` and `Unevict` options in `Edit node and disks` sub-selection under `Operation` column next to each disk `Scheduling` options. This is for user to evict or cancel the eviction of the disks or the nodes.
+2. Add new `requestEviction` field to `Node.Spec` and `Node.Spec.disks` Spec. This is to help tracking the request from user and trigger volume controler. And this will reconsile with `scheduledReplica` of selected disks on the nodes.
+3. Add new `ReplicaEvictionRequested` slice of string to collecting eviction replicas. Use this for the eviction creating extra replica or deleting eviction requested replica.
```
shuo-wu, line 56: Add the new field to `volume.status`?

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,87 @@
...
+Success Case:
+
+For both `Replica Node Level Soft Anti-Affinity` has been enabled and disabled, and for the node that volume has been mount or not mount.
+1. User can select one or more disks or nodes for eviction. Select `Evict` on the disabled disks or nodes, Longhorn should start rebuild replicas for the volumes which have replicas on the eviction disks or nodes, and after rebuild success, the replica number on the evicted disks or nodes should be 0.
+
+Failure Cases:
+1. If user selects the disks or nodes have not been disabled scheduling, Longhorn should display the error message on `Dashboard` immediatly.
+2. If there is no enough disk spaces or nodes for disks or nodes eviction, Longhorn should log the error message in the `Event Log`. And once the disk spaces or nodes resources are good enough, the eviction should continue. Or if the user clicks `Unevict`, Longhorn should stop eviction and clear the `requestEviction` fields for nodes, disks and volumes crd objects.
+
+#### Integration Test Plan
```
shuo-wu, line 74: The description of this part is too vague. And we can add more integration tests. e.g.:
1. Test eviction canceling.
2. Test if the suspending eviction will move on once there is one disk/node suddenly becoming available.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,87 @@
...
+2. Add new `requestEviction` field to `Node.Spec` and `Node.Spec.disks` Spec. This is to help tracking the request from user and trigger volume controler. And this will reconsile with `scheduledReplica` of selected disks on the nodes.
+3. Add new `ReplicaEvictionRequested` slice of string to collecting eviction replicas. Use this for the eviction creating extra replica or deleting eviction requested replica.
+4. Once `Evict` has been clicked for disks or nodes, the `requestEviction` fields for the disks and nodes will be set to `True` (default is `False`).
+5. Add a informer in `Volume Controller` to get these information from disks and nodes.
+6. During `ReconcileEngineReplicaState`, based on this map to trigger rebuild for different volumes' replicas. And remove the replicas have been evicted from the map.
```
shuo-wu, line 59: Do not need to mention the concrete function name you prepare to modify. And it may be changed during the implementation.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,87 @@
...
+Failure Cases:
+1. If user selects the disks or nodes have not been disabled scheduling, Longhorn should display the error message on `Dashboard` immediatly.
+2. If there is no enough disk spaces or nodes for disks or nodes eviction, Longhorn should log the error message in the `Event Log`. And once the disk spaces or nodes resources are good enough, the eviction should continue. Or if the user clicks `Unevict`, Longhorn should stop eviction and clear the `requestEviction` fields for nodes, disks and volumes crd objects.
+
+#### Integration Test Plan
+For `Replica Node Level Soft Anti-Affinity` is enabled, create 2 replicas on the same disk or node, and then evict this disk or node, the 2 replicas should goto another disk of node.
+
+For `Replica Node Level Soft Anti-Affinity` is disabled, create 1 replica on a disk, and evict this disk or node, the replica should goto the other disk of node.
+
+#### For Node Eviction
```
shuo-wu, line 79: Does this still belong to `Test Plan`?  
Same below.

shuo-wu:
And it's better to mention how you know the eviction is done in the implementation part. This would help us understand why you introduce those new fields.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
boknowswiki: Sorry, this should be removed. I will remove it.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
boknowswiki: I think for the two test cases you described should be in manual test, not integration test. Since in the integration test, it's hard to predict when to cancel the eviction and what the status may be looked like.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
boknowswiki: Ok, let me update this part.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
boknowswiki: Will update it to make it more clear.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
boknowswiki: Yes, let me remove the header, it's confusing.

boknowswiki:
> And it's better to mention how you know the eviction is done in the implementation part. This would help us understand why you introduce those new fields.

Updated, please take a look. Thanks.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,88 @@
+# Replica Eviction Support for Disks and Nodes
+
+## Summary
+This enhancement is to simpler and automatically to evict the replicas on the selected disabled disks or nodes to other suitable disks and nodes. Meanwhile keep the same level of fault tolerance during this eviction period of time.
```
yasker, line 4: `simpler`? You meant `simplify`?

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,88 @@
...
+2. Report any error to user during the eviction time.
+3. Allow user to cancel the eviction at any time.
+
+### Non-goals
+
+For now, we only focus on `Healthy` volumes, which means `Attached` and `Healthy`. If the volume is `Detached`, the eviction won't work for this volume.
+
+
+## Proposal
+1. Add `Evict` and `Unevict` buttons for disks and nodes. This is for user to evict or cancel the eviction of the disks or the nodes.
```
yasker, line 23: `Evict` and `cancel eviction`. `Unevict` is not commonly used.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,88 @@
...
+2. Add new `requestEviction` field to `Node.Spec` and `Node.Spec.disks` Spec. This is to help tracking the request from user and trigger volume controler. And this will reconsile with `scheduledReplica` of selected disks on the nodes.
```
yasker, line 24: Should be `evictionRequested`, following the naming convention of other fields.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,88 @@
...
+3. Display can not evict error to `Dashboard` and any other eviction errors to the `Event log`.
```
yasker, line 25: Display `fail to evict` error

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,88 @@
...
+
+### User Stories
+### Disks and Nodes Eviction
+For disk replacement or node replacement, the eviction needs to be done successfully in order to guarantee longhorn volume function properly.
```
yasker, line 29: `Longhorn`

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,88 @@
...
+
+Before, when user wants to evict a disk or a node they need to do the following steps:
+
+1. User needs to disable the disk or the node.
+2. User needs to scale up the replica count for the volume which has replica on disabled disks or nodes, and wait for the rebuild complete, then delete the replicas on this disk or node.
```
yasker, line 34: In fact the user needs to scale up the replica count, wait for the rebuilding complete, **scale down the replica count**, then delete the replica.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,88 @@
...
+
+After this enhancement, user can click the `Evict` on scheduling disabled disks or nodes. Or select `Disable` for scheduling and `Evict` at the same time then save this change. The backend will take care of the eviction for the disks or nodes and cleanup for all the replicas on disks or nodes.
+
+### User Experience In Detail
+#### Disks and Nodes Eviction
+1. User can selects `Evict` from `Longhorn UI` for disks or nodes. And user has to make sure the selected disks or nodes have been disabled, or select the `Disable` Scheduling at the same time of `Evict`.
```
yasker, line 40: Thinking about that page, it makes sense to present `Evict` as `Request Eviction` status, since it will be submitted as a form. `Evict` is a verb so we need to use the noun to describe the status.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,88 @@
...
+2. Once `Evict` has been selected on the disks or nodes, they can not be enabled for `Scheduling`.
+3. If the disks or the nodes haven't been disabled for `Scheduling`, there will be error message showed in `Dashboard` immediatly to indicate that user need to disable the disk or node for eviction.
+4. And user will wait for the replica number for the disks or nodes to be 0.
+5. If there is any error e.g. no space or couldn't find other schedulable disk, the error message will be logged in the `Event log`. And the eviction will be suspended until either user cancel the eviction or cleanup more disk spaces for the new replicas.
+6. If user cancels the eviction, the remaining replicas on the selected disks or nodes will remain on the disks or nodes.
+
+### API changes
+From an API perspective, the call made to `Evict` and `Unevict` the `Node` or `Disk` eviction should look the same. The logic for handling the new field `Evict` and `Unevict` should to be in the `Node Controller` and `Volume Controller`.
```
yasker, line 48: Probably do `Eviction requested` and `true` or `false`.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,88 @@
...
+
+### User Experience In Detail
+#### Disks and Nodes Eviction
+1. User can selects `Evict` from `Longhorn UI` for disks or nodes. And user has to make sure the selected disks or nodes have been disabled, or select the `Disable` Scheduling at the same time of `Evict`.
+2. Once `Evict` has been selected on the disks or nodes, they can not be enabled for `Scheduling`.
+3. If the disks or the nodes haven't been disabled for `Scheduling`, there will be error message showed in `Dashboard` immediatly to indicate that user need to disable the disk or node for eviction.
+4. And user will wait for the replica number for the disks or nodes to be 0.
+5. If there is any error e.g. no space or couldn't find other schedulable disk, the error message will be logged in the `Event log`. And the eviction will be suspended until either user cancel the eviction or cleanup more disk spaces for the new replicas.
```
yasker, line 44: We can add a condition in the node/disk as well, to indicate the reason of the eviction failure. That would be easier for the user, instead of looking into the event log (which can be written to automatically when updating the condition), or the manager log.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
boknowswiki: Yes, will fix.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
boknowswiki: Will fix.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
boknowswiki: This is in the same struct with `AllowScheduling`, so I think `requestEviction` matches the naming. Please let me know if you still think `evictionRequested` is better.

yasker: I see. `AllowScheduling` is named a long time ago, so it doesn't follow the pattern as well. It's still better we can follow the pattern as `noun-verb` from now on.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
boknowswiki: Will fix.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
boknowswiki: Sure, will change the UI button name from `Eviction` to `Request Eviction`. And `Unevict` to `Cancel Eviction`.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
boknowswiki: Yes, this sound better. At beginning, I was thinking of `Eviction`/ `enable` and `disable`. But I think the `Eviction requested`/`true` and `false` sounds better.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
boknowswiki: Got it, I will rename it to `evictionRequested`.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
boknowswiki: Yes, it will be logged in manager log.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
shuo-wu: For the 1st case, when the volume replica count is 3, the eviction should get stuck when the soft anti-affinity setting is disabled.
For the 2nd case, you can disable one node scheduling, unset soft anti-affinity, and create a volume with 2 replicas. Then the eviction of the volume will get stuck until you enable the node scheduling.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
boknowswiki: @yasker  and @shuo-wu , I thought about the design, and I am thinking it's better to let replica controller to listen on the nodes changes, and update the new field `evictionRequested` in `Replica.Spec`. And once volume controller gets the replicas will have this field set, and then trigger replenish replica.

Also about adding a condition in volume. I think we don't need it once we have [this UI redirect replica name to volume page feature](https://github.com/longhorn/longhorn/issues/1697), once in the volume page, if there is any like `Schedule failure` or `Replica rebuild failure`, they will be displayed there.

Please let me know what you think. Thanks.

yasker: @boknowswiki Good idea. We should the field in `replica.Status` instead of `spec` though. No controller can change it's own spec.

I am OK with no condition change now too. We can add a condition later if it's needed.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Design
+
+### Implementation Overview
+
+1. On `Longhorn UI` `Node` page, for nodes eviction, adding `Eviction Requested` `true` and `false` options in the `Edit Node` sub-selection, next to `Node Scheduling`. For disks eviction, adding `Eviction Requested` `true` and `false` options in `Edit node and disks` sub-selection under `Operation` column next to each disk `Scheduling` options. This is for user to evict or cancel the eviction of the disks or the nodes.
+2. Add new `evictionRequested` field to `Node.Spec`, `Node.Spec.disks` Spec and `Replica.Status`. These will help tracking the request from user and trigger replica controller to update `Replica.Status` and volume controler to do the eviction. And this will reconsile with `scheduledReplica` of selected disks on the nodes.
+3. Add a informer in `Replica Controller` to get these information and update `evictionRequested` field in `Replica.Status`.
+4. Add new `ReplicaEvictionRequested` slice of string in volume controller as local variable to collecting eviction replicas. Use this for the eviction creating extra replica or deleting eviction requested replica.
```
yasker, line 57: Do we still need this after introducing the new field in the replica?

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,89 @@
...
+8. Logged the errors to `Event log` during the reconcile process.
+9. Once the `ReplicaEvictionRequested` slice of string is empty, this means the eviction for this volume is done.
+10. By the end from `Longhorn UI`, the replica number on the eviction disks or nodes should be 0, this mean eviction is success.
+
+### Test plan
+
+#### Manual Test Plan For Disks and Nodes Eviction
+Success Case:
+
+For both `Replica Node Level Soft Anti-Affinity` has been enabled and disabled, and for the node that volume has been mount or not mount.
```
yasker, line 70: Anti-affinity setting and mount or not shouldn't impact this feature, so we don't need to add specific test for them (unless we want to use anti-affinity to simulate scheduling failure).

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,89 @@
...
+1. User can select one or more disks or nodes for eviction. Select `Eviction Requested` to `true` on the disabled disks or nodes, Longhorn should start rebuild replicas for the volumes which have replicas on the eviction disks or nodes, and after rebuild success, the replica number on the evicted disks or nodes should be 0. E.g. When there are 3 nodes in the cluster, and with `Replica Node Level Soft Anti-Affinity` is set to `false`, disable one node, and create a volume with replica count 2. And then evict one of them, the eviction should get stuck, then set `Replica Node Level Soft Anti-Affinity` to `true`, the eviction should go through.
+
+Failure Cases:
```
yasker, line 73: `Negative case`

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,89 @@
...
+6. `Replica Controller` will update `evictionRequested` field in `Replica.Status` and `Volume Controller` to get these information from it's replicas.
+7. During reconcile the engine replica, based on this map to trigger rebuild for different volumes' replicas. And remove the replicas have been evicted from the map.
+8. Logged the errors to `Event log` during the reconcile process.
+9. Once the `ReplicaEvictionRequested` slice of string is empty, this means the eviction for this volume is done.
+10. By the end from `Longhorn UI`, the replica number on the eviction disks or nodes should be 0, this mean eviction is success.
+
+### Test plan
+
+#### Manual Test Plan For Disks and Nodes Eviction
+Success Case:
```
yasker, line 68: Positive Case

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
boknowswiki: Yes, there could be multiple replicas need to be evicted.
With the new field in the replica, this could simplify the logic in volume controller to tell if the replica has been requested eviction.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
boknowswiki: Ok, I see. I was thinking they could be different scenarios for tests. And yes, I was using the anti-affinity to simulate the scheduling failure case. So I mentioned all the cases here for reference.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
yasker: Volume controller has the full list of the replica in the memory, it can just query that.

boknowswiki: Yes, you are right, we can drop this variable as well. Let me update it.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,87 @@
...
+2. Add new `evictionRequested` field to `Node.Spec`, `Node.Spec.disks` Spec and `Replica.Status`. These will help tracking the request from user and trigger replica controller to update `Replica.Status` and volume controler to do the eviction. And this will reconsile with `scheduledReplica` of selected disks on the nodes.
+3. Add a informer in `Replica Controller` to get these information and update `evictionRequested` field in `Replica.Status`.
+4. Once `Eviction Requested` has been set to `true` for disks or nodes, the `evictionRequested` fields for the disks and nodes will be set to `true` (default is `false`).
+5. `Replica Controller` will update `evictionRequested` field in `Replica.Status` and `Volume Controller` to get these information from it's replicas.
+6. During reconcile the engine replica, based on `Replica.Status.EvictionRequested` of the volume replicas to trigger rebuild for different volumes' replicas. And remove the replicas have been evicted from the map.
```
yasker, line 59: There is no map now.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,87 @@
+# Replica Eviction Support for Disks and Nodes
+
+## Summary
+This enhancement is to simplify and automatically to evict the replicas on the selected disabled disks or nodes to other suitable disks and nodes. Meanwhile keep the same level of fault tolerance during this eviction period of time.
```
yasker, line 4: `to simplify and automatically evict .... per user's request.`

shuo-wu (changes requested):
In general LGTM except for the test part.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Goals
+1. Allow user easily evict the replicas on the selected disks or nodes to other disks or nodes without impact the user defined `Volume.Spec.numberOfReplicas` and keep the same level of fault tolerance. This means we don't change the user defined replica number.
+2. Report any error to user during the eviction time.
+3. Allow user to cancel the eviction at any time.
+
+### Non-goals
+
+For now, we only focus on `Healthy` volumes, which means `Attached` and `Healthy`. If the volume is `Detached`, the eviction won't work for this volume.
+
```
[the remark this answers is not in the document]
boknowswiki: Fixed.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Proposal
+1. Add `Eviction Requested` with `true` and `false` selection buttons for disks and nodes. This is for user to evict or cancel the eviction of the disks or the nodes.
+2. Add new `evictionRequested` field to `Node.Spec`, `Node.Spec.disks` Spec and `Replica.Status`. These will help tracking the request from user and trigger replica controller to update `Replica.Status` and volume controler to do the eviction. And this will reconsile with `scheduledReplica` of selected disks on the nodes.
```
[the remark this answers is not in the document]
boknowswiki: Fixed.

enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md
shuo-wu: Here is the integration test example for case 1. Can you check if it's possible?
1. Disable the setting soft anti-affinity
2. Create and attach a 3-replica volume.
3. Do eviction for a random node.
4. Verify there is a scheduling failed replica and the volume is still healthy
5. Verify the evictionRequested field in node/replica.
6. Cancel the eviction.
7. Re-verify the evictionRequested field in node/replica.

shuo-wu (approved):

boknowswiki:
> What if the volume get detached during the eviction process?

I think for this case, we can cover it with [detached volume case](https://github.com/longhorn/longhorn/issues/1704), since we need to auto-attach it and do the eviction.

So far I am thinking of the eviction is depends on the volume status, if it's attached, then we can trigger rebuild and once it success, we can remove the replica. If the volume become detached during this eviction time or the volume is detached at the beginning, we need to auto-attach it and do the rebuild and once it success we remove the replica and then auto-detach by the end(need to take a deep look at the code about the auto-detach logic since there could be the case a pod which the volume attached is crashed, and the volume become detached. And after the eviction, we need to keep it attached).

Please let me know if you have other concern @yasker  and @shuo-wu . Thanks.

yasker:
We haven't implemented the eviction for the detached volume yet so we can leave this part to that.

Otherwise LGTM. We can update the proposal once we start working on the evict detached volume as well.

Issue #1269, opened 2020-04-28
[FEATURE] Preserve the data from existing disks on the new nodes


all4innov:
**Description**
As shown in the figure, Longhorn is completely off when all EC2 k8s workers was recreated via an ASG (the red down nodes are the old and the green schedulable are the new) knowing that the new instances are attaching the same EBS volumes with the same datapath (so in theory preserving all replicas).

![image](https://user-images.githubusercontent.com/520080/80543253-d3dcd000-89ae-11ea-8242-0b7f70f4f197.png)

**NB**
- RKE cluster deployed as custom via Rancher 2.3
- Kubernetes 1.15

yasker:
This might impact the design on https://github.com/longhorn/longhorn/issues/1151

shuo-wu:
Test step 1:
1. Deploy Longhorn v1.0.2.
2. Add an extra disk to a node and disable the default disk for the node. (So that all replicas scheduled on this node will use the extra disk.)
3. Create and attach volumes then write data to the volumes.
4. Upgrade Longhorn to the latest version.
5. Verify all running volumes **keep** `Healthy` during/after the upgrade.
6. Verify the existing volume rebuilding works fine.
7. Detach some running volumes.
8. Directly shut down the node containing the extra disk. Then verify all running replicas on the node will become failed.
9. Verify if all disk of the node becomes `Disconnected` in the UI Node page.
10. In the cloud vendor platform, Attach the extra physical disk to another running node (not Longhorn disk and Longhorn node). Then mount the disk to a mount point in the node host.
11. On the Longhorn Node page, Click `connect`, use the mount point as a Longhorn disk for the node. 
12. Verify 
    1. the disconnected Longhorn disk object (which represents the extra physical disk) will be reused by the node.
    2. the node ID of all replicas (failed replicas are included) that use the extra disk will be updated.
13. Attach the `detached` volumes. Verify r/w works fine.

Test step 2:
    1. Launch the latest Longhorn.
    2. Create and attach some volumes.
    3. Disable the setting `retain-disks-during-node-deletion`.
    4. Delete a node.
    5. Verify all disks on the node are gone with the node.
    6 Verify all volumes/replicas on the removed node/disks can be deleted.

yasker:
Need doc and test skeleton. Also, add one more test case (also doable by e2e test)
1. Add a disk to node A in path A (e.g. `/opt/disk1`).
2. Create a volume, write data to it, detach the volume.
[...]

Issue #1304, opened 2020-05-07
[FEATURE] Rebuild using existing replica data


yasker:
Using a new replica every time for rebuild will waste lots of network bandwidth and disk space. We should try to rebuild using existing replica's data.

Require #1269

yasker:
The PR is missing an explanation for how the feature works.

Also, more detailed test steps are needed, especially around the scheduling part.

shuo-wu:
More explanations are added as the implementation comments.
And one more test skeleton is introduced for the replica reuse with scheduling case.

khushboo-rancher:
Verified with longhorn-master - 10/27/2020

Validation - **Fail**

Scenarios 3 and 4 failed. When a node is rebooted, the replica on it got deleted and recreated. The ```Replica Replenishment Wait Interval``` setting was set to 600 sec.

Validated with a cluster up of 1 etcd/control plane, 3 worker nodes.
Scenarios tested:
1. Failed one replica by crashing the replica instance manager. - Replica went into rebuilding state for a while and becomes healthy, data validated.
2. Deleted volume-head data in a replica - The same replica rebuilt and become healthy.
3. Had volume with 2 replicas and rebooted one of the replica node - A new replica rebuilt on an available node.
4. Had volume with 3 replicas and rebooted one of the replica node - A new replica rebuilt on an available node.

[longhorn-support-bundle_7a2a3332-a6b3-4eae-84f4-c267b2844b59_2020-10-28T00-49-06Z.zip](https://github.com/longhorn/longhorn/files/5448859/longhorn-support-bundle_7a2a3332-a6b3-4eae-84f4-c267b2844b59_2020-10-28T00-49-06Z.zip)
Volume name - pvc-68411985-ea3b-4b4f-8d0b-55f8f5e436b2
Node- khushboo-test-lh-wk2 rebooted
Replica before reboot - pvc-68411985-ea3b-4b4f-8d0b-55f8f5e436b2-6841b31a
Replica after reboot - pvc-68411985-ea3b-4b4f-8d0b-55f8f5e436b2-f19caabb
Time  - ~00:43:40

shuo-wu:
The reason of the above test failure:
When the node is rebooted, the node ready condition hasn't been updated. Then Longhorn immediately and continuously retries reusing the failed replica on the node for 3 times, which reaches the max retry count. Then Longhorn considers the failed replica as no longer available. 
Solution:
Add backoff interval for the retry.

[...]

Issue #1232, opened 2020-04-21
[BUG] Nodes deleted from cluster and re created with the same node name are seen as Unschedulable


sowmyav27:
**Describe the bug**
Nodes deleted from cluster and re created with the same node name are seen as Unschedulable

**To Reproduce**

1. Create an RKE DO cluster from Rancher - 1 etcd, 1 control plane and 4 worker nodes
2. Enable/Deploy Longhorn in the cluster
3. Delete a worker node from Rancher.
4. The worker node is recreated (node pool) with the same node name
5. Navigate to Node page in Longhorn.
6. The node is seen in `Unschedulable` state 

**Expected behavior**
The node should NOT be in "Unschedulable" but be in `Schedulable` state. Longhorn should delete data about nodes which have been deleted from the cluster

**Note:**

- Currently user has to disable scheduling on the disk and delete the disk.
- Add a new disk on the node and enable scheduling for the node to come in `Schedulable` state.

**Environment:**
 - Longhorn version: master - 04/21/2020
 - Kubernetes version: 1.17.4
 - Node OS type and version: rke DO linux

khushboo-rancher:
Verified with longhorn-master - 10/21/2020

Validation - **Pass**

The corresponding Kubernetes node object gets deleted on the deletion of the nodes and adding another node with the same name works fine. Validated the case of schedulable/down/disabled and multiple nodes getting deleted.

Issue #1151, opened 2020-03-31
[FEATURE] Scale down Longhorn nodes with Kubernetes


dkeightley:
**Is your feature request related to a problem? Please describe.** Deleting a node from Longhorn outside of the UI involves a number of unique steps

**Describe the solution you'd like** Simplified process for deleting node during a rolling update of nodes in the cluster, when using kubectl to self-drain the node and delete itself from the cluster, having the ability to delete the node via it's CRD object would provide a universal approach

**Describe alternatives you've considered** API calls to the Longhorn API

yasker:
I am thinking we can add a setting to allow automatically delete the node if it's deleted from Kubernetes.

yasker:
Need to consider the design with #1269

yasker:
Move out to v1.1.0 due to we need to redesign the feature due to the user may want to retain the disk on the new nodes, e.g. extra EBS volume. See #1269

yasker:
Retain metadata of disk or not can be implemented as a flag for each disk, with a global default setting.

yasker:
In order to implement this feature, we will need to promote disk to the first-class object.

yasker:
The automatic removal of the node feature has been implemented in Longhorn v1.1.0.

PR #1699, opened 2020-08-18
enhancements: Add LEP 'Identify Disk Reconnection' 

shuo-wu:
longhorn/longhorn#1269

### DIFF
--- a/enhancements/20200817-identify-disk-reconnection.md
+++ b/enhancements/20200817-identify-disk-reconnection.md
@@ -1,0 +1,136 @@
+# Identify Disk Reconnection
+
+## Summary
+When disks are reconnected to other Longhorn nodes, Longhorn should be able to figure out the disk reconnection and update the node ID, disk ID as well as the data path for the related replicas (including failed replicas). 
+
+### Related Issues
+https://github.com/longhorn/longhorn/issues/1269
+
+## Motivation
+### Goals
+The ultimate goal of this feature and [#1304](https://github.com/longhorn/longhorn/issues/1304) is to reuse the existing data to speed up replica rebuild as well as save bandwidth. 
+As the first step, Longhorn should identify the reconnected disks which contain the existing data left by the failed replicas. 
+Then Longhorn needs to make sure the failed replicas are still able to find out the data after the disk migration.
+Finally, there is an opportunity to reuse the existing data to rebuild new replicas with the existing data.
+
+### Non-Goals
+As for how to reuse the existing data and handle rebuild related feature, it will be left to #1304, which is not the intention of this enhancement.
+
+## Proposal
+1. Use separate CRD resources to manage disk lifecycle, so that the disks and nodes are decoupled. Then there is a chance to reuse replica data on the disks:
+    1. The new disk objects will be created automatically (as the default/customized disks) when a Longhorn node is introduced.
+    2. The disk creation/deletion/update calls will be separated from the node editing. 
+    3. Launch Disk Controller to monitor and update the disk.
+    4. When a physical disk previously used by Longhorn is reconnected to a node, Longhorn should be able to figure it out and update all replicas using the disk.
+2. Modify the node deletion logic.
+    1. Right now the ownership of replicas is taken by disk rather than node. We can remove the replica check in the node deletion function.
+    2. We can add an setting `Retain disks during node deletion` to determine if Longhorn needs to clean up disks.
+3. As long as the failed replicas is not cleaned up (it depends on the replica stale timeout), Longhorn may ba able to reuse the data of the failed replicas later.
+
+### User Stories
+#### Migrate the existing disks to new nodes
+Before the enhancement, users need to:
+1. disable the scheduling for the disk
+2. attach all volumes using the disk
+3. disable scheduling then evict the disk (this guarantees the HA)
+4. reconnect the disk to another node
+
+After the enhancement, this can be done by:
+1. detach the volumes using the disk
+2. Reconnect the disk to the another node (both nodes keep running)
+3. reattach the related volumes
+
+#### Scale down the node but reuse the disks on the node
+Before the enhancement, there is no chance to reuse the failed replicas on the node.
+
+After the enhancement, the node ID of all failed replicas on the node will be updated once the related disks are reconnected to other running nodes. Then the data of those failed replicas can be reused after implementing #1304.
+
+### User Experience In Detail
+#### Migrate the existing disks to new nodes
+1. Detach all related volumes using the disk before the disk migration.
+2. Directly move the disk to the new node (physically or in cloud vendor) and mount the disk.
+3. Click `Connect Disk` with new the mount point and the new node ID in Longhorn Node page.
+4. Attach the volumes for the workloads.
+
+#### Scale down the node but reuse the disks on the node
+1. Directly shut down the node when there are replicas on the node. Then the replicas on the node will fail.
+2. Move the disks on the down node to other running nodes (physically or in cloud vendor).
+3. Click `Connect Disk` with new the mount point and the new node ID in Longhorn Node page.
+4. Wait then verify the node ID of the failed replicas will be updated.
+5. Longhorn will reuse the failed replicas to speed up the rebuilding once the feature [#1304](https://github.com/longhorn/longhorn/issues/1304) is done.
+
+### API Changes
+Disk related HTTP APIs are required.
+
+## Design
+### Implementation Overview
+#### longhorn-manager:
+1. Create a CRD for `disk`.
+    1. The field `Disks` and `DiskStatus` will be deprecated from `NodeSpec` and `NodeStatus`, respectively. 
+    2. A new field `NodeID` will be added to `DiskSpec`. It will be set during the disk creation.
+    3. A new field `State` will be added to `DiskStatus`. The state of a disk is always `connected` or `disconnected`.
+2. Add a new setting `Retain disks during node deletion`.
+3. Create disk HTTP API calls. And separate the disk related logic from the node API if necessary.
+    1. The disk creation call will blindly create the disk objects for active nodes.
+    2. The disk deletion function will be responsible for setting the deletion timestamp the disk objects.
+    3. The node creation call will invoke the disk creation function for default/customized disks.
+    4. The node deletion call may invoke the disk deletion function based on the setting `Retain disks during node deletion`.
+4. Launch Disk Controller to handle the disk status monitoring and update.
+    1. Move the disk update logic from Node Controller to Disk Controller.
+    2. `DiskStatus.State` will be `DiskStateDisconnected` in one of the following cases:
+        1. The disk is not handled by the conroller  of the preferred node (disk.Spec.NodeID). in Disk Controller: If `DiskSpec.NodeID` is empty or the related node is down, the state . Otherwise, the state is `DiskStateConnected`.
+        2. The preferred node is down or deleted.
+        3. The disk info cannot be fetched.
+        4. There is another disk has the same FSID or DiskUUID.
+    3. When there is an existing DiskUUID (meta file) in the new disk and this UUID is still used in a disconnected disk, it means the physical disk is migrated. In this case, we need to transfer the ownership of the related replicas and update the replica fields so that the new disk can take care of those replicas. 
+       Then the field `Status.DiskUUID` will be cleaned up for the existing disconnected disk.
+    4. The running replicas will be marked as failure if the related disk beomces `disconnected`.
+5. Need to update nodes and replicas as well as create disks for Longhorn v1.1.0 upgrade. 
+
+#### longhorn-ui:
+1. Group all disconnected disks into a special list. Then present it in Node Page, too.
+2. Add button `Operation` for each disk, which contains the following calls:
+    1. `Delete`: Add warning "Directly deleting the disk will fail replicas using the disk".
+    2. `Update`: This call receives 3 inputs `AllowScheduling`, `EvictionRequested`, `StorageReserved`, and `Path`.
+3. Rename button `Edit Node and Disks` to `Edit Node`. And remove disk related operations from the node editing page.
+4. Add button `Connect Disk` at the top right of Node Page.
+    1. Input `Name` is optional.
+    2. Input `Path` and `NodeID` are required.
+    3. Input `AllowScheduling` and `EvictionRequested` are True and False by default, respectively. 
+
+### Test Plan
+##### Disk migration
+1. Add an extra disk for a Kubernetes worker node.
+2. Connect the disk to the Longhorn node via UI or [the node annotations](https://longhorn.io/docs/1.0.2/advanced-resources/default-disk-and-node-config/#launch-longhorn-with-multiple-disks).
+3. Disable the default disk for the nodes containing extra disks.
+4. Create and attach a 3-replica volume. Make sure there is one replica using the extra disk.
+5. Write data to the volume and get the checksum.
+6. Detach the volume.
+7. Migrate the extra disk used by the volume replica to another worker node (in the cloud vendor management platform).
+8. Connect the disk with the new mount path and new node ID in Node Page.
+9. Verify the old Longhorn disk object using the extra path becomes `disconnected`. And the new Longhorn disk object using the extra path is `connected`. 
+10. Attach the volume. Verify the volume works fine and there is no failed replica.
+11. Verify the data. 
+
+##### Scale down the node but reuse the disks on the node
+1. Launch Longhorn in a 3-node cluster.
+2. Connect an extra disk with scheduling enabled for one worker node Node1, and disable the default disk for Node1.
+3. Create and attach a 3-replica volume. Write data to the volume. (Make sure the volume is not attached to Node1.)
+4. Shut down the worker node.
+5. Move the extra disk used by the volume replica to another worker node Node2 (in the cloud vendor management platform). 
+6. Disconnect the disk then reconnect it to the corresponding Longhorn node Node2 in Node Page.
+7. Verify there is a failed replica. And the node ID of the failed replica is updated to Node2.
+
+##### Longhorn upgrade
+1. Deploy Longhorn v1.0.2.
+2. Launch some volumes and write data.
+3. Upgrade Longhorn to the latest version.
+4. Make sure the disk count is unchanged and there is no disconnected disk.
+5. Verify all volumes work fine, and the related data is correct.
+6. Verify volume CRUD work fine.
+
+### Upgrade strategy
+No special upgrade strategy is required.
+
+### Notes
+The hard anti-affinity restriction will be broken for the existing volumes after the disk migration.


enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,75 @@
...
+#### longhorn-manager:
+1. Create a CRD for `disk`.
+2. Add a new field `NodeID` in `disk.Spec`.
+3. Create separate HTTP REST APIs for `disk`.
+4. Separate the disk create/update/delete calls from node calls.
+5. Modify the condition check for node/disk removal.
+6. Create `Disk Controller` to monitor and update the disk status.
+7. Move the disk update logic from Node Controller to Disk Controller.
+8. Rely on disk objects to update the related nodes in Node Controller.
+9. Update Node ID for the related replicas in Disk Controller when an existing disk is re-attached to a node.
```
yasker, line 58: Not sure if the disk controller should update the node ID of replica, since replica spec is not created by the disk controller. Also, update a node ID for a "running" replica means it needs to be stopped (if the user didn't detach the volume before changing the disk). I feel like the volume controller is the better place to do it.

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,101 @@
...
+The goal of this enhancement is to eliminate unnecessary replica crashing when the existing disks are migrated.
+
+### Non-Goals
+This enhancement is not intended to retain the running replicas when the related disks is forcely migrated to other nodes. 
+
+## Proposal
+1. Use separate CRD resources to manage disk lifecycle, so that the disks and nodes are decoupled. Then there is a chance to reuse the disks:
+  1. The new disk object will be created when a disk is added/attached to a Longhorn node.
+  2. When a node is deleted, the related disks won't be cleaned up immediately since Longhorn doesn't know if those disks are gone from the Kubernetes cluster. 
+     Instead, there is a corruption timeout for the disk. Once the disk is resued/added back before the timeout, the disk won't be deleted.
```
yasker, line 21: `Disconnection timeout`. It's not an indication of corruption.

`reused`

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,101 @@
...
+  3. Launch Disk Controller to handle the disk status monitoring and update rather than directly using Node Controller.
+2. Remove the field `NodeID` in `replica.Spec`. Instead, the disk UUID will be stored in `replica.Spec`, and the node ID will be detected only when the replica needs to be started.
```
yasker, line 23: `move the nodeid into replica.status.`

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,101 @@
...
+  1. For stopped replicas, nothing happens even if the corresponding disks are removed or migrated.
+  2. For running replicas, they will be marked as `ERROR` by the replica manager once the related disks don't exist, or the related disks exist but become state unknown. Since its directory is suddenly unavailable.
+
+### User Stories
+#### Longhorn worker nodes in AWS Auto Scaling group
+Before the enhancement, all volumes will become `Faulted` and the auto salvage doesn't work when the instance replacement happen. As a result, it's hard to recover the data.
+
+After the enhancement, all volumes will become `Faulted` first after the replacement. Then the auto salvage will reuse the data in the old disks to recover the replicas. As a result, all volumes will be recovered automatically and the data won't be lost.
```
yasker, line 31: Not all the volumes become faulted.

Also, there is a possibility that managed EKS will drain the node first before replacing it.

There is a difference between the volume attached to the node and the volume has the replica on the node but not attached to it.

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,101 @@
...
+
+#### Migrate the existing disks to new nodes without crashing replicas
+Before the enhancement, users need to:
+1. disable the scheduling for the disk
+2. attach all volumes using the disk
+3. scale up, wait for the rebuild, then scale down the volumes (this guarantees the HA)
```
yasker, line 37: New replica eviction feature can be used.

User can do:
1. Evict Longhorn replicas.
2. Drain the node

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,101 @@
...
+### API Changes
+A bunch of disk related HTTP APIs will be separate from node related APIs.
+
+## Design
+### Implementation Overview
+#### longhorn-manager:
+1. Create a CRD for `disk`.
+2. Separate disk related HTTP APIs from `node` APIs.
+  1. Add a new option `Clean up the host path during the initialization`. Sometimes users don't need the existing data when they want to reuse the disks.
+  2. The disk configuration (e.g., reserved space) is always determined by the users. Longhorn won't apply the previous configuration when reusing the old disks.
```
yasker, line 65: why? we have the same object.

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,101 @@
...
+
+### API Changes
+A bunch of disk related HTTP APIs will be separate from node related APIs.
+
+## Design
+### Implementation Overview
+#### longhorn-manager:
+1. Create a CRD for `disk`.
+2. Separate disk related HTTP APIs from `node` APIs.
+  1. Add a new option `Clean up the host path during the initialization`. Sometimes users don't need the existing data when they want to reuse the disks.
```
yasker, line 64: I prefer to let the user do that. If they don't want to reuse the old replicas, they should wipe the disk.

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,101 @@
...
+  2. The disk configuration (e.g., reserved space) is always determined by the users. Longhorn won't apply the previous configuration when reusing the old disks. 
+3. Modify the condition check for node/disk removal.
+4. Launch Disk Controller to handle the disk status monitoring and update:
+  1. Add a new field `NodeID` in `disk.Spec` and set it when adding the disk to a node.
+  2. Add a new field `DetachedAt` in `disk.Status`. This field will be set when the related node is down. And Disk controller will clean up the disk if this field is set and the timeout is passed.
```
yasker, line 69: Disk controller shouldn't delete a disk object.

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,101 @@
...
+## Design
+### Implementation Overview
+#### longhorn-manager:
+1. Create a CRD for `disk`.
+2. Separate disk related HTTP APIs from `node` APIs.
+  1. Add a new option `Clean up the host path during the initialization`. Sometimes users don't need the existing data when they want to reuse the disks.
+  2. The disk configuration (e.g., reserved space) is always determined by the users. Longhorn won't apply the previous configuration when reusing the old disks. 
+3. Modify the condition check for node/disk removal.
+4. Launch Disk Controller to handle the disk status monitoring and update:
+  1. Add a new field `NodeID` in `disk.Spec` and set it when adding the disk to a node.
```
yasker, line 68: Needs to be careful here on who is responsible for creating the disk.

If it's the user, the user will need to manually move the disk.

It might be easier if it's the node controller to create/migrate/cleanup the disk. Cleanup needs to be done by someone else if it's disconnected disk though.

Thinking about it, I am not sure if a timeout for the disk removal makes sense. If users don't want the disk, they will set timeout to 0; if they want to keep the disk, that means they are planning to reattach the disk, so removing the disk automatically anytime in the future seems weird. I think it makes better sense if the option is just true/false to keep the disk after the node removal. By default, it will be true since losing data is more worrisome than keeping something that you need to clean up.

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,101 @@
...
+  2. Add a new field `DetachedAt` in `disk.Status`. This field will be set when the related node is down. And Disk controller will clean up the disk if this field is set and the timeout is passed.
+  3. During the disk initialization/re-initialization, the UUID will be loaded if the meta file is found, field `DetachedAt` will be unset, and other status fields like `ScheduledReplica` will be detected based on the current status.
```
yasker, line 70: Use `DisconnectedAt` to differentiate from the volume.

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,108 @@
...
+  2. Update the node ID for replicas when the related disks are migrated.
+
+### Non-Goals
+This enhancement is not intended to hanlde replica rebuild with the existing data, which is tracked in [#1304](https://github.com/longhorn/longhorn/issues/1304)  
+
+## Proposal
+1. Use separate CRD resources to manage disk lifecycle, so that the disks and nodes are decoupled. Then there is a chance to reuse replica data the disks:
+    1. The new disk object will be created when a disk is added/attached to a Longhorn node.
+    2. When a node is deleted, Longhorn doesn't know if those disks are gone with node from the Kubernetes cluster. Hence we can provide two kinds of disk deletion policies:
+        1. `Immediate`: Always delete disk immediately when the node is removed. This means the disk migration is not taken into consideration. This will be the default policy.
```
yasker, line 24: Why `Immediate` is the default option? This changes the default behavior from now and might result in data loss.

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,108 @@
...
+        2. `Delayed`: Set a long deletion timestamp when the node is gone. When the disks are back later, the deletion timestamp will be unset and the disks will be re-initialized.
```
yasker, line 25: Will deletion timestamp be used later? Or it's just a marker for when the disk was disconnected. Sounds like `DisconnectedAt` timestamp is more proper is that's the case.

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,108 @@
...
+    1. The new disk object will be created when a disk is added/attached to a Longhorn node.
+    2. When a node is deleted, Longhorn doesn't know if those disks are gone with node from the Kubernetes cluster. Hence we can provide two kinds of disk deletion policies:
+        1. `Immediate`: Always delete disk immediately when the node is removed. This means the disk migration is not taken into consideration. This will be the default policy.
+        2. `Delayed`: Set a long deletion timestamp when the node is gone. When the disks are back later, the deletion timestamp will be unset and the disks will be re-initialized. 
+           But the ultimate goal of this feature and #1304 is to reuse the existing data. If the rebuilding is started before the disks being back, the existing data in the disk is useless. 
+           Hence, the **failed replicas should be retained** and **the rebuild will be delayed** before the related disks are completely removed. Then Longhorn can rebuild new replicas based on (the data) the failed replicas after the disk migration success.
```
yasker, line 27: Good catch. This is a problem not only happens with disk migration. If the replica failed due to a temporary network interruption, rebuild replica immediately is not a good idea too.

We can introduce a setting named e.g. `ReplicaRebuildWaitTime` (in seconds) to delay the rebuild for existing replicas, assuming the network and disk can be mounted back automatically after a few minutes.

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,108 @@
...
+    2. When a node is deleted, Longhorn doesn't know if those disks are gone with node from the Kubernetes cluster. Hence we can provide two kinds of disk deletion policies:
+        1. `Immediate`: Always delete disk immediately when the node is removed. This means the disk migration is not taken into consideration. This will be the default policy.
+        2. `Delayed`: Set a long deletion timestamp when the node is gone. When the disks are back later, the deletion timestamp will be unset and the disks will be re-initialized. 
+           But the ultimate goal of this feature and #1304 is to reuse the existing data. If the rebuilding is started before the disks being back, the existing data in the disk is useless. 
+           Hence, the **failed replicas should be retained** and **the rebuild will be delayed** before the related disks are completely removed. Then Longhorn can rebuild new replicas based on (the data) the failed replicas after the disk migration success.  
+  3. Launch Disk Controller to handle the disk status monitoring and update rather than directly using Node Controller.
```
yasker, line 28: `update disk`?

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,108 @@
...
+#### Migrate the existing disks to new nodes
+Before the enhancement, users need to:
+1. disable the scheduling for the disk
+2. attach all volumes using the disk
+3. disable scheduling then evict the disk (this guarantees the HA)
+4. migrate the disk to another nodes
+
+After the enhancement, this can be done by:
+1. detach the volumes using the disk
+2. directly attach the disk to the another node
```
yasker, line 52: `reattach` or `reconnect`

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,108 @@
...
+3. reattach the related volumes
+
+### User Experience In Detail
+#### Longhorn worker nodes in AWS Auto Scaling group
+Users don't need to do anything except for setting the disk deletion policy to `Delayed` in advance.
+
+#### Migrate the existing disks to new nodes
+1. Detach all related volumes using the disk before the disk migration.
+2. Directly Detach the disk from the old node, and reattach it to the new node (physically or in cloud vendor).
```
yasker, line 61: We can use `connect` instead of `attach` for the disk case, otherwise, it gets easily confused with the volume.

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,108 @@
...
+1. disable the scheduling for the disk
+2. attach all volumes using the disk
+3. disable scheduling then evict the disk (this guarantees the HA)
+4. migrate the disk to another nodes
+
+After the enhancement, this can be done by:
+1. detach the volumes using the disk
+2. directly attach the disk to the another node
+3. reattach the related volumes
+
```
yasker, line 54: Add a case of scaling down the Kubernetes cluster automatically and a case of decommissioning a node.

In theory, as long as the user only scales down less than replica count number of nodes (e.g. 1 or 2 nodes) every time, and wait for the rebuild to complete, the cluster should be fine.

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,108 @@
...
+### API Changes
+No API change is needed.
+
+## Design
+### Implementation Overview
+#### longhorn-manager:
+1. Add a setting named `DiskDeletionPolicy`
+    1. `Immediate`: Do not add grace period for disk deletion.
+    2. `Delayed`: Add a grace period for disk deletion. But this will block/delay the replica rebuild.
```
yasker, line 73: As I mentioned above, I feel it's better to leave this timeout to the rebuilding of the existing replica feature.

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,108 @@
...
+
+##### Disk migration
+1. Launch Longhorn in a 3-node cluster.
+2. Add an extra disk for one worker node, and disable the default disk for the node.
+3. Create and attach a 3-replica volume. Write data to the volume.
+4. Detach the volume.
+5. Attach the extra disk to another worker node (in the cloud vendor management platform). And add the disk for the corresponding Longhorn node in Longhorn Node Page.
+6. Attach the volume. Verify the volume works fine and there is no failed replica.
+7. Verify the data. 
+
```
yasker, line 103: Add other two cases here as well.

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
```diff
@@ -0,0 +1,108 @@
...
+### Upgrade strategy
+No special upgrade strategy is required.
+
+### Notes
+The soft anti-affinity restriction will be broken after the disk migration.
```
yasker, line 108: hmm? why only the soft anti-affinity? how about hard anti-affinity?

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
shuo-wu: This means the disk will be removed only when the node is deleted rather than when the node is down. In order to delete the node, all replicas should be removed by users first. Hence I don't think this policy will cause data loss.

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
shuo-wu: The deletion timestamp will be set in the node & disk deletion API call. If a field `DisconnectedAt` is used for the deletion delay, the node controller should take care of the actual disk deletion later, which increases the unnecessary complexity.

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
shuo-wu: Yeah. Add this setting is definitely a good idea.

enhancements/20200817-preserve-replicas-after-the-disks-migration.md
shuo-wu: Sorry for the typo. I mean `hard anti-affinity` here.

enhancements/20200817-identify-disk-reconnection.md
```diff
@@ -0,0 +1,109 @@
...
+Finally, there is an opportunity to reuse the existing data to rebuild new replicas with the existing data.
+
+### Non-Goals
+As for how to reuse the existing data and handle rebuild related feature, it will be left to #1304, which is not the intention of this enhancement.
+
+## Proposal
+1. Use separate CRD resources to manage disk lifecycle, so that the disks and nodes are decoupled. Then there is a chance to reuse replica data the disks:
+    1. The new disk object will be created when a disk is added/attached to a Longhorn node.
+    2. When a disk is removed via the UI, there is no doubt that the disk object should be removed.
+    3. When a node is/can be manually removed via the UI, there is no replica using the related disks. It means there is no existing data in the related disks.
```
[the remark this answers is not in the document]
shuo-wu: We can add a notification for the node deletion or rename the button name to `Remove node and related disks`

enhancements/20200817-identify-disk-reconnection.md
```diff
@@ -0,0 +1,109 @@
...
+4. Attach the volumes for the workloads.
+
+#### Scale down the node but reuse the disks on the node
+1. Directly shut down the node when there are replicas on the node. Then the replicas on the node will fail.
+2. Reconnect the disks on the down node to other running nodes.
+3. Config the reconnected disks for the running Longhorn nodes.
+4. Wait then verify the node ID of the failed replicas will be updated.
+
+### API Changes
+Creating disk related HTTP APIs is required.
```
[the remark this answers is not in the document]
shuo-wu: It's just HTTP APIs. Since we may require users to delete the disk manually.

enhancements/20200817-identify-disk-reconnection.md
```diff
@@ -0,0 +1,109 @@
...
+2. Directly reconnect the disk to the new node (physically or in cloud vendor).
+3. Config the disk for the new Longhorn node in Longhorn Node page.
+4. Attach the volumes for the workloads.
+
+#### Scale down the node but reuse the disks on the node
+1. Directly shut down the node when there are replicas on the node. Then the replicas on the node will fail.
+2. Reconnect the disks on the down node to other running nodes.
+3. Config the reconnected disks for the running Longhorn nodes.
+4. Wait then verify the node ID of the failed replicas will be updated.
+
```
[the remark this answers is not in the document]
shuo-wu: I am not sure why we need to remove the disk with the node here?  I am trying to reuse the disk on the down node, rather than cleaning up the down node.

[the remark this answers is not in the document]
yasker: I meant if the user decides the disk will be gone if the node is gone (so no reattachment), we should remove the disk automatically, so the user won't need to do it manually. See https://github.com/longhorn/longhorn/issues/1151

enhancements/20200817-identify-disk-reconnection.md
```diff
@@ -0,0 +1,127 @@
...
+    2. The disk deletion function will be responsible for setting the deletion timestamp the disk objects. And the function will error out if there is a replica using the disk.
+    3. The disk connection call will be introduced, which requires 2 inputs `NodeID` and `Path`. As a counterpart, there is a disk disconnection call.
+    4. The disk update call is used for updating Scheduling status, tags, and the reserved space.
+    5. The node creation call will invoke the disk creation function for default/customized disks.
+    6. An new option `Delete all related disks` will be added in the node deletion call. If this option is checked, all disks on the node will be listed then deleted. Notice that the disk deletion failure will lead to the node deletion error out.
+3. Launch Disk Controller to handle the disk status monitoring and update.
+    1. Move the disk update logic from Node Controller to Disk Controller.
+    2. Update `DiskStatus.State` in Disk Controller: If `DiskSpec.NodeID` is empty or the related node is down, the state will be `DiskStateConnected`. Otherwise, the state is `DiskStateConnected`.
```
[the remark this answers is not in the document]
shuo-wu: Yes. Thank you for pointing out the typo!

enhancements/20200817-identify-disk-reconnection.md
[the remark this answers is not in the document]
shuo-wu: 1. What if there are (unknown) engines on the down node? Do we need to break the node deletion restriction thoroughly? 
2. We can provide an option/setting to do automatically cleanup (volumes, engines, replicas, disks) for down nodes, but I don't think this is related to this feature. It can be handled in #1151 separately. IMO, the main purpose of this feature is to reuse the existing data on the disk. Adding more complexity is not a good choice.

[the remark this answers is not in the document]
shuo-wu: 1. What if there are (unknown) engines on the down node? Do we need to break the node deletion restriction thoroughly? 
2. We can provide an option/setting to do automatically cleanup (volumes, engines, replicas, disks) for down nodes, but I don't think this is related to this feature. It can be handled in #1151 separately. IMO, the main purpose of this feature is to reuse the existing data on the disk. Adding more complexity is not a good choice.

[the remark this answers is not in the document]
yasker: 1. The node has been removed from the cluster, there is no way there still be an engine attached to it. I feel detach the volume automatically and don't attach it until the Spec.NodeID changes to any existing nodes.
2. There are two ways of scaling down the cluster, and both are included in the epic issue https://github.com/longhorn/longhorn/issues/1151 (and  https://github.com/longhorn/longhorn/issues/1232 ). The goal of v1.1.0 is to support both cases.

enhancements/20200817-identify-disk-reconnection.md
```diff
@@ -0,0 +1,136 @@
...
+The ultimate goal of this feature and [#1304](https://github.com/longhorn/longhorn/issues/1304) is to reuse the existing data to speed up replica rebuild as well as save bandwidth. 
+As the first step, Longhorn should identify the reconnected disks which contain the existing data left by the failed replicas. 
+Then Longhorn needs to make sure the failed replicas are still able to find out the data after the disk migration.
+Finally, there is an opportunity to reuse the existing data to rebuild new replicas with the existing data.
+
+### Non-Goals
+As for how to reuse the existing data and handle rebuild related feature, it will be left to #1304, which is not the intention of this enhancement.
+
+## Proposal
+1. Use separate CRD resources to manage disk lifecycle, so that the disks and nodes are decoupled. Then there is a chance to reuse replica data the disks:
```
boknowswiki, line 20:
```suggestion
1. Use separate CRD resources to manage disk lifecycle, so that the disks and nodes are decoupled. Then there is a chance to reuse replica data  on the disks:
```

enhancements/20200817-identify-disk-reconnection.md
```diff
@@ -0,0 +1,136 @@
...
+    1. The new disk objects will be created automatically (as the default/customized disks) when a Longhorn node is introduced.
+    2. The disk creation/deletion/update calls will be separated from the node editing. 
+    3. Launch Disk Controller to monitor and update the disk.
+    4. When a physical disk previously used by Longhorn is reconnected to a node, Longhorn should be able to figure it out and update all replicas using the disk.
+2. Modify the node deletion logic.
+    1. Right now the ownership of replicas is taken by disk rather than node. We can remove the replica check in the node deletion function.
+    2. We can add an setting `Retain disks during node deletion` to determine if Longhorn needs to clean up disks.
```
boknowswiki, line 27: Does this mean user can delete a node from Longhorn, but leave the disks in Longhorn? Could you help me to understand this part? Thanks.

enhancements/20200817-identify-disk-reconnection.md
```diff
@@ -0,0 +1,136 @@
...
+11. Verify the data. 
+
+##### Scale down the node but reuse the disks on the node
+1. Launch Longhorn in a 3-node cluster.
+2. Connect an extra disk with scheduling enabled for one worker node Node1, and disable the default disk for Node1.
+3. Create and attach a 3-replica volume. Write data to the volume. (Make sure the volume is not attached to Node1.)
+4. Shut down the worker node.
+5. Move the extra disk used by the volume replica to another worker node Node2 (in the cloud vendor management platform). 
+6. Disconnect the disk then reconnect it to the corresponding Longhorn node Node2 in Node Page.
+7. Verify there is a failed replica. And the node ID of the failed replica is updated to Node2.
```
boknowswiki, line 122: Why 'Verify these is a failed replica'? Should there be 3 replicas, and 2 replicas are on Node2 one on the default disk, one on the extra disk which reconnected from Node1 to Node2?

boknowswiki (changes requested):

enhancements/20200817-identify-disk-reconnection.md
shuo-wu: The main purpose of this test is to verify if Longhorn will update DiskID/NodeID/DataPath for the failed replica after the migration. Since the feature #1304 is not implemented, Longhorn won't reuse the failed replica and the existing data to do the rebuild.

enhancements/20200817-identify-disk-reconnection.md
boknowswiki: What about this part? Is there a missing "on"? The sentence doesn't seem right to me.

shuo-wu: Oh, sorry. I forgot to push the update commit for the PR.

yasker:
Let's update this one to the new design.

shuo-wu:
Close this PR since the new design is totally different from this version.

Issue #1045, opened 2020-02-06
Support `Data Locality` option to always keep a replica local to the engine


yasker:
Try to keep a replica local.

1. Need to delay the scheduling to after the volume being used for the first time.
2. Need to migrate replica automatically if there is no local replica on the node.

yasker:
Thought more about this:
1. The setting `Ensure One Replica co-located With Engine` should be exposed as a global option.
    1. We don't need to allow per volume configuration for now because the option will be enabled only when the network condition is bad.
2. There is still a question if we can delay the scheduling of replica, or does it make sense to do that.
3. The basic idea is, if a volume is attached to a node that has no replica, then we will start rebuilding the replica on the node after the volume is attached. Once the rebuilding process is done, we will randomly remove one of the other replicas to keep the replica count as specified.
    1. If the volume was detached before the rebuilding was done, the half-finished replica will be removed automatically even in the current mechanism.

PhanLe1010:
@yasker 
When users choose the option to keep a replica on the same node as the engine. If for some reasons (e.g node tag, disk tag conflict, or not enough space), the replica cannot be migrated to the node. Do we error out and stop the volume or just skip the migration?

yasker:
It's going to be the best effort. We cannot give user zero availability if there is no local disk satisfied the requirement. We should notify the user about that in some way.

PhanLe1010:
**reproduce/test steps:**
1. Create a cluster of 6 nodes and install Longhorn. Having more nodes helps us to be more confident in verifying the PR because the chance of randomly scheduling a replica onto same node as the engine is small. 
1. Create volume `testvol` with `Number of Replicas` = 2
####Case 1: `keep-local-replica-on-same-node-as-engine` set to `false`:
1. Attach `testvol` to a node which does not contain any replica
1. Expected: No new replica is created
####Case 2: `keep-local-replica-on-same-node-as-engine` set to `true`:
1. Attach `testvol` to a node which does not contain any replica
1. Expected: Longhorn starts rebuilding a local replica on the engine's node. After finishing rebuilding the local replica, Longhorn remove one of the other replicas to keep the total number of replicas == 2
1. Assume that the engine is currently on `node-1`. Delete a replica that is not in `node-1`. Verify that Longhorn creates and rebuild a new replica.
[...]

PR #1705, opened 2020-08-20
enhancement: Add LEP 'support Data Locality option to keep a replica local to the engine'

PhanLe1010:
Issue: longhorn/longhorn#1045

### DIFF
--- a/enhancements/20200819-keep-a-local-replica-to-engine.md
+++ b/enhancements/20200819-keep-a-local-replica-to-engine.md
@@ -1,0 +1,244 @@
+# Data Locality - Option To Keep A Local Replica On The Same Node As The Engine
+
+## Summary
+
+A Longhorn volume can be backed by replicas on some nodes in the cluster and accessed by a pod running on any node in the cluster. 
+In the current implementation of Longhorn, the pod which uses Longhorn volume could be on a node that doesn't contain any replica of the volume. 
+In some cases, it is desired to have a local replica on the same node as the consuming pod. 
+In this document, we refer to the property of having a local replica as having `data locality`
+
+This enhancement gives the users option to have a local replica on the same node as the engine which means on the same node as the consuming pod. 
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/1045
+
+## Motivation
+
+### Goals
+
+Provide users an option to try to migrate a replica to the same node as the consuming pod.
+
+### Non-goals
+
+Another approach to achieve data locality is trying to influence Kubernetes scheduling decision so that pods get scheduled onto the nodes which contain volume's replicas.
+However, this is not a goal in this LEP. See https://github.com/longhorn/longhorn/issues/1045 for more discussion about this approach. 
+
+## Proposal
+
+We give user 2 options for data locality setting: `disabled` and `best-effort`. 
+In `disabled` mode, there may be a local replica of the volume on the same node as the consuming pod or there may not be. 
+Longhorn doesn't do anything.
+In `best-effort` mode, if a volume is attached to a node that has no replica, the Volume Controller will start rebuilding the replica on the node after the volume is attached. 
+Once the rebuilding process is done, it will remove one of the other replicas to keep the replica count as specified.
+
+### User Stories
+
+Sometimes, having `data locality` is critical. 
+For example, when the network is bad or the node is temporarily disconnected, having local replica will keep the consuming pod running. 
+Another case is that sometimes the application workload can do replication itself (e.g. database) and it wants to have a volume of 1 replica for each pod. 
+Without the `data locality` feature, multiple replicas may end up on the same node which destroys the replication intention of the workload. See more in [Story 1](#story-2)
+ 
+In the current implementation of Longhorn, the users cannot ensure that pod will have a local replica. 
+After the enhancement implemented, users can have options to choose among `disabled` (default setting) or `best-effort`
+
+#### Story 1
+
+A user has three hyper-converged nodes and default settings with: `default-replica-count: 2`. 
+He wants to ensure a pod always runs with at least one local replica would reduce the amount of network traffic needed to keep the data in sync. 
+There does not appear to be an obvious way for him to schedule the pod using affinities.
+
+#### Story 2
+
+A user runs a database application that can do replication itself. 
+The database app creates multiple pods and each pod uses a Longhorn volume with `replica-count = 1`. 
+The database application knows how to schedule pods into different nodes so that they achieve HA. 
+The problem is that replicas of multiple volumes could land on the same node which destroys the HA capability. 
+With the `data locality` feature we can ensure that replicas are on the same nodes with the consuming pods and therefore they are on different nodes.
+
+### User Experience In Detail
+
+* Users create a new volume using Longhorn UI with `dataLocality` set to `best-effort`. 
+* If users attach the volume a node which doesn't contain any replica, they will see that Longhorn migrate a local replica to the node.
+* Users create a storageclass with dataLocality: best-effort set
+* Users launch a statefulset with the storageclass.
+* Users will find that there is always a replica on the node where the pod resides on
+* Users update  `dataLocality` to `disable`, detach the volume, and attach it to a node which doesn't have any replica
+* Users will see that Longhorn does not create a local replica on the new node.
+
+### API changes
+
+There are 2 API changes:
+1. When creating a new volume, the body of the request sent to `/v1/volumes` has a new field `dataLocality` set to either `disabled` or `best-effort`.
+1. Implement a new API for users to update `dataLocality` setting for individual volume.
+The new API could be `/v1/volumes/<VOLUME_NAME>?action=updateDataLocality`. This API expects the request's body to have the form `{dataLocality:<DATA_LOCALITY_MODE>}`.
+
+## Design
+
+### Implementation Overview
+There are 2 modes for `dataLocality`:
+1. `disabled` is the default mode. 
+   There may be a local replica of the volume on the same node as the consuming pod or there may not be. 
+   Longhorn doesn't do anything.
+1. `best-effort` mode instructs Longhorn to try to keep a local replica on the same node as the consuming pod. 
+   If Longhorn cannot keep the local replica (due to not having enough disk space, incompatible disk tags, etc...), Longhorn does not stop the volume.
+
+There are 3 settings the user can change for `data locality`:
+
+1. Global default setting inside Longhorn UI settings.
+   The global setting should only function as a default value, like replica count. 
+   It doesn't change any existing volume's setting
+1. specify `dataLocality` mode for individual volume upon creation using UI
+1. specify `dataLocality` mode as a parameter on Storage Class.
+
+Implementation steps:
+
+1. Add a global setting `DefaultDataLocality`
+1. Add the new field `DataLocality` to `VolumeSpec`
+1. Modify the volume creation API so that it extracts, verifies, and sets the `dataLocality` mode for the new volume.
+If the volume creation request doesn't have `dataLocality` field inside its body, we use the `DefaultDataLocality` for the new volume.
+1. Modify the `CreateVolume` function inside the CSI package so that it extracts, verifies, and sets the `dataLocality` mode for the new volume.
+This makes sure that Kubernetes can use CSI to create Longhorn volume with a specified `datLocality` mode.
+1. Inside `volume controller`'s sync logic, we add a new function `ReconcileLocalReplica`.
+1. When a volume enters the `volume controller`'s sync logic, function `ReconcileLocalReplica` checks the `dataLocality` mode of the volume.
+   If the `dataLocality` is `disabled`, it will do nothing and return.
+1. If the `dataLocality` is `best-effort`, `ReconcileLocalReplica` checks whether there is a local replica on the same node as the volume.
+   1. If there is no local replica, we create an in-memory replica struct.
+      We don't create a replica in DS using createReplica() directly because we may need to delete the new replica if it fails to ScheduleReplicaToNode.
+   	  This prevents UI from repeatedly show creating/deleting the new replica.
+      Then we try to schedule the replica struct onto the consuming pod's node.
+      If the scheduling fails, we don't do anything. The replica struct will be collected by Go's garbage collector.
+      If the scheduling success, we save the replica struct to the data store. This will trigger replica rebuilding on the consuming pod's node.
+   1. If there already exists a local replica on the consuming pod's node, we check to see if there are more healthy replica than specified on the volume's spec.
+      If there are more healthy replicas than specified on the volume's spec, we remove a replica on the other nodes. 
+      We prefer to delete replicas on the same disk, then replicas on the same node, then replicas on the same zone.
+
+UI modification:
+1. On volume creation, add an input field for `dataLocality`
+1. On volume detail page:
+   * On the right volume info panel, add a <div> to display `selectedVolume.dataLocality`
+   * On the right volume panel, in the Health row, add an icon for data locality status. 
+     Specifically, if  `dataLocality=best-effort` but there is not a local replica then display a warning icon.
+     Similar to the replica node redundancy wanring [here](https://github.com/longhorn/longhorn-ui/blob/0a52c1f0bef172d8ececdf4e1e953bfe78c86f29/src/routes/volume/detail/VolumeInfo.js#L47)
+   * In the volume's actions dropdown, add a new action to update `dataLocality`
+1. In Rancher UI, add a parameter `dataLocality` when create storage class using Longhorn provisioner. 
+   
+### Test plan
+
+#### Manually Test Plan
+1. Create a cluster of 9 worker nodes and install Longhorn. 
+Having more nodes helps us to be more confident because the chance of randomly scheduling a replica onto the same node as the engine is small.
+
+##### Test volume creation with `dataLocality` is `best-effort`:
+
+1. Create volume `testvol` with `Number of Replicas = 2` and `dataLocality` is `best-effort`
+1. Attach `testvol` to a node that doesn't contain any replica.
+1. Verify that Longhorn schedules a local replica to the same node as the consuming pod. 
+   After finishing rebuilding the local replica. Longhorn removes a replica on other nodes to keep the number of replicas is 2.
+
+##### Test volume creation with `dataLocality` is `disabled`:
+
+1. Create another volume, `testvol2`  with `Number of Replicas = 2` and `dataLocality` is `disabled`
+1. Attach `testvol2` to a node that doesn't contain any replica.
+1. Verify that Longhorn doesn't move replica
+
+##### Test volume creation with `dataLocality` is unspecified and `DefaultDataLocality` setting as `disabled`:
+1. Leave the `DefaultDataLocality` setting as `disabled` in Longhorn UI.
+1. Create another volume, `testvol3`  with `Number of Replicas = 2` and `dataLocality` is empty
+1. Attach `testvol3` to a node that doesn't contain any replica.
+1. Verify that the `dataLocality` of `testvol3` is `disabled` and that Longhorn doesn't move replica.
+
+##### Test volume creation with `dataLocality` is unspecified and `DefaultDataLocality` setting as `best-effort`:
+1. Set the `DefaultDataLocality` setting to `best-effort` in Longhorn UI.
+1. Create another volume, `testvol4`  with `Number of Replicas = 2` and `dataLocality` is empty
+1. Attach `testvol4` to a node that doesn't contain any replica.
+1. Verify that the `dataLocality` of `testvol4` is `best-effort`.
+1. Verify that Longhorn schedules a local replica to the same node as the consuming pod. 
+   After finishing rebuilding the local replica. 
+   Longhorn removes a replica on other nodes to keep the number of replicas is 2.
+
+##### Test `updateDataLocality` from `disabled` to `best-effort`:
+1. Change `dataLocality` to `best-effort` for `testvol2`
+1. Verify that Longhorn schedules a local replica to the same node as the consuming pod. 
+   After finishing rebuilding the local replica. 
+   Longhorn removes a replica on other nodes to keep the number of replicas which is 2.
+   
+##### Test `updateDataLocality` from `best-effort` to `disabled` :
+1. Change `dataLocality` to `disabled` for `testvol2`
+1. Go to Longhorn UI, increase the `number of replicas` to 3. Wait until the new replica finishes rebuilding.
+1. Delete the local replica on the same node as the consuming pod.
+1. Verify that Longhorn doesn't move replica 
+
+##### Test volume creation by using storage class with `dataLocality` parameter is `disabled`:
+1. Create `disabled-longhorn` storage class with from this yaml file:
+    ```yaml
+    kind: StorageClass
+    apiVersion: storage.k8s.io/v1
+    metadata:
+      name: disabled-longhorn
+    provisioner: driver.longhorn.io
+    allowVolumeExpansion: true
+    parameters:
+      numberOfReplicas: "1"
+      dataLocality: "disabled"
+      staleReplicaTimeout: "2880" # 48 hours in minutes
+      fromBackup: ""
+    ```
+1. create a deployment of 1 pod using PVC dynamically created by `disabled-longhorn` storage class.
+1. The consuming pod is likely scheduled onto a different node than the replica. 
+   If this happens, verify that Longhorn doesn't move replica
+   
+##### Test volume creation by using storage class with `dataLocality` parameter is `best-effort`:
+1. Create `best-effort-longhorn` storage class with from this yaml file:
+    ```yaml
+    kind: StorageClass
+    apiVersion: storage.k8s.io/v1
+    metadata:
+      name: best-effort-longhorn
+    provisioner: driver.longhorn.io
+    allowVolumeExpansion: true
+    parameters:
+      numberOfReplicas: "1"
+      dataLocality: "best-effort"
+      staleReplicaTimeout: "2880" # 48 hours in minutes
+      fromBackup: ""
+    ```
+1. create a shell deployment of 1 pod using the PVC dynamically created by `best-effort-longhorn` storage class.
+1. The consuming pod is likely scheduled onto a different node than the replica. 
+1. If this happens, verify that Longhorn schedules a local replica to the same node as the consuming pod. 
+   After finishing rebuilding the local replica, Longhorn removes a replica on other nodes to keep the number of replicas which is 1.
+1. verify that the volume CRD has `dataLocality` is `best-effort`
+
+##### Test volume creation by using storage class with `dataLocality` parameter is unspecified`:
+1. Create `unspecified-longhorn` storage class with from this yaml file:
+    ```yaml
+    kind: StorageClass
+    apiVersion: storage.k8s.io/v1
+    metadata:
+      name: unspecified-longhorn
+    provisioner: driver.longhorn.io
+    allowVolumeExpansion: true
+    parameters:
+      numberOfReplicas: "1"
+      staleReplicaTimeout: "2880" # 48 hours in minutes
+      fromBackup: ""
+    ```
+1. create a shell deployment of 1 pod using PVC dynamically created by `unspecified-longhorn` storage class.
+1. The consuming pod is likely scheduled onto a different node than the replica. 
+1. If this happens, depend on `DefaultDataLocality` setting in Longhorn UI, verify that Longhorn does/doesn't migrate a local replica to the same node as the consuming pod.
+
+#####  Tests for the volumes created in old versions:
+
+1. The volumes created in old Longhorn versions don't have the field `dataLocality`. 
+1. We treat those volumes the same as having `dataLocality` set to `disabled`
+1. Verify that Longhorn doesn't migrate replicas for those volumes. 
+
+### Upgrade strategy
+
+No special upgrade strategy is required.
+
+We are adding the new field, `dataLocality`, to volume CRD's spec.
+Then we use this field to check whether we need to migrate a replica to the same node as the consuming pod.
+When users upgrade Longhorn to this new version, it is possible that some volumes don't have this field.
+This is not a problem because we only migrate replica when `dataLocality` is `best-effort`.
+So, the empty `dataLocality` field is fine.


PhanLe1010:
I will fix spelling and grammar soon. Please help me to verify the logic first 😁

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,141 @@
...
+     * Communication cost: The data are transferred in Http(s) between the default scheduler and scheduler extenders. So be aware of the associated costs of performing serialization and deserialization.
+     * Limited extension points: As you might have noticed, extenders can only be involved at the end of certain phrases, such as “Filter” and “Prioritize.” They can’t be called in the beginning or middle of any phase.
+     * Subtraction over addition: Compared to the node candidate list that was passed by the default scheduler, you might have a good reason to add a new one. But it’s risky doing so because it’s not guaranteed that the new node can pass other requirements. So the scheduler extender is preferable to perform “subtraction” (further filtering) instead of “addition” (adding nodes).
+     * Cache sharing: In practice, you need to give scheduling decisions by looking at the status of the whole cluster. The default scheduler makes scheduling decisions pretty well, but its caching can’t be shared which means you have to build and maintain your own ones.
+
+1. **Kubernetes Scheduler Framework and Plugins**
```
shuo-wu, line 53: It's seems that the main disadvantage of this approach is that users need to specify `spec.schedulerName` in pods' spec?

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Non-goals
+
+1. This enhancement doesn't make any guarantee if the users misconfigure the `data locality` setting. When the users misconfigure the `data locality` setting, the `data-locality` behavior will be undefined. 
+
+## Proposal
+
+When designing this enhancement, I notice that there are many different moving parts/components from which we can use to build up a solution. I will first list all the available components and then propose a complete solution.
+
+1. **Longhorn's ability to move replica to a new node**
```
shuo-wu, line 30: Maybe this part can be put in the `Design` or `Implementation Overview`. And summarising the essence of the design here is enough.

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,141 @@
...
+
+A user has three hyper-converged nodes and default settings with: `default-replica-count: 2`. He wants to ensure a pod always runs with at least one local replica would reduce the amount of network traffic needed to keep the data in sync. There does not appear to be an obvious way for him to schedule the pod using affinities.
+
+#### Story 2
+
+A user runs a database application that is able to do replication itself. The database app creates multiple pods and each pod uses a Longhorn volume with `replica-count = 1`. The database application knows how to schedule pods into different nodes so that they achieve HA. The problem is that replicas of multiple volumes could land on the same node which destroys the HA capability. With the `data locality` feature we can ensure that replicas are on the same nodes with the consuming pods and therefore they are on different nodes.
+
+### User Experience In Detail
+
+There are 3 modes for `dataLocality`:
```
shuo-wu, line 97: The mode naming `dynamic` and `static` is a little bit confusing.

shuo-wu (changes requested):
Great investigation and analysis! In general LGTM.

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,141 @@
...
+
+     1. First, in the `filter` phase, the `default scheduler` filters out all the un-schedulable nodes base on the requirements (like CPU, memory, selector, affinity, ..). At the end of the `filter` phase, the result consists of a list of nodes that are compatible with the given Pod. This list of compatible nodes is sent to the first webhook. Longhorn can build an Http server that listens on the webhook. When the Http server receives the list of compatible nodes, it can decide whether it needs to filter out nodes that don't have a replica of volume depend on the `data locality` setting. After that, a new list of nodes is passed to the second phase.
+     1. In the `prioritize` phase, the `default scheduler` will score the nodes based on attributes such as the state. Longhorn can register the Http server as a webhook at the end of this phase. The Http server can try to influence Kubernetes scheduling decisions by giving more points for the node which has the volume's replica.
```
joshimoo, line 45: Since points are additive, we can score non replica nodes with negative scores in our scoring function :)
This way we are more or less in control of the scheduling decisions.

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,141 @@
...
+
+     The scheduling framework contains 2 cyles: `Scheduling Cycle` and `Binding Cycle`. The `Scheduling Cycle` exposes 7 extension points. Longhorn can implement 2 plugins: one for the `Filter` extension point (to filter out the nodes which don't have volume's replica) and one for the `Scoring` extension point (To give higher priority for nodes which have volume's replica). We can then add the 2 plugins into the `default scheduler`. Re-compile the `default scheduler` and create a scheduler profile name: `longhorn-scheduler`. The end result is that there only 1 single scheduler in the cluster which has 2 profiles: `default-scheduler` and `longhorn-scheduler`. Longhorn can then uses `longhorn-scheduler` for pods that use Longhorn volumes and leaves other pods to use the `default-scheduler`.
```
joshimoo, line 57: `filter` implementation is only required for `enforced` data locality
`prioritize` for `best effort` data locality

enhancements/20200819-keep-a-local-replica-to-engine.md
joshimoo: Agree how about `best-effort` and `required/enforced/guaranteed` data locality?

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,141 @@
...
+
+### User Experience In Detail
+
+There are 3 modes for `dataLocality`:
+1. `do-nothing`: this is the default setting. Longhorn will not attempt to keep a local replica on the same node as the consuming pod
+2. `dynamic`: Longhorn will try the best to keep a local replica on the same node as the consuming pod. Longhorn can achieve it by schedule pod on to the node which already contains a replica. If the pod cannot be scheduled into the nodes which contain the existing replica, Longhorn will try to move replica to the node where the pod is running. If Longhorn is unable to move replica to the node where the pod is running, it will not stop the pod. This probably is the most popular option. 
+3. `static`: pod has to be scheduled onto the node which contains an existing replica. Otherwise, the pod will be stuck in a pending unscheduled state. The idea here is that for some workloads such as distributed databases or message buses, it might be preferable to avoid scheduling a pod that would run more slowly than its peers, ensuring all running instances respond with similar latency. 
+
+There are 3 things the user can change for `data locality` setting:
```
joshimoo, line 102: Consider exposing the data locality setting  as a parameter on the Storage Class.
This way the user can create different Storage Classes for different setups, we can also then ship common Storage Classes for different usage scenarios.

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,141 @@
...
+
+1. need to specify `spec.schedulerName: longhorn-scheduler` in the spec of all pods which use Longhorn volumes. 
+1. specify `dataLocality` mode for each volume
+1. specify `dataLocality` mode for all volume in the global setting
```
joshimoo, line 106: Not sure if we need a global setting, switching from one mode of data locality to another seems complicated.

In my mind data locality mode is a setting that should be specified when a `volume is created`, same as the access mode. One wouldn't expect to be able to change a RWO to an RWX volume after creation :)

If we do want to support changing this, then the per volume setting is fine otherwise let's consider the storage class approach?
@yasker @shuo-wu @PhanLe1010

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,141 @@
...
+
+1. The pod is created with the `spec.schedulerName: longhorn-scheduler` and `dataLocality: dynamic/static` set for Volume
+1. The pod goes into the scheduler queue to be scheduled on to a node. The scheduler will use the `longhorn-scheduler` profile to schedule to the pod. 
+1. After going through all default plugins in the `Filter` extension point, A list of compatible nodes is passed into `longhorn-filer-plugin`.
+1. Inside `longhorn-filer-plugin`, we inspect pod's spec. Find the corresponding PV. Find the list of nodes that contain existing healthy replicas of the volume. If `dataLocality: static` then we filer out the node which doesn't contain any existing replica and return a new list of nodes.  If `dataLocality: dynamic`, then we don't do anything and return the input list of nodes.
+1. The list of nodes then passed into the `Scoring` extension point. After going through all default plugins in the `Scoring` extension point, the list is passed into `longhorn-soring-plugin`.
+1. In `longhorn-soring-plugin`, If `dataLocality: dynamic`, then we give more points for the nodes in the input list which contain existing volume's replica.
```
joshimoo, line 127: Also subtract points for nodes, that don't have a local replica.
Also need to look at the Storage Capacity of the available nodes.

joshimoo (commented):
@PhanLe1010 good job on the analysis and the write up  👍 

Left some comments.

One Sentence per line please, this helps with the diffs and reviews.
Markdown will format as required :)

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,141 @@
...
+### Goals
+
+This enhancement has 3 goals:
+1. Provide user options to achieve `data locality`
+1. The enhancement is designed so that it is easy for the users to use without having to do many manual setups.
+1. Don't have a negative performance/functionality impact on the other workloads which don't use Longhorn volume.
+
+### Non-goals
+
+1. This enhancement doesn't make any guarantee if the users misconfigure the `data locality` setting. When the users misconfigure the `data locality` setting, the `data-locality` behavior will be undefined.
```
yasker, line 24: What does this mean? How can the user misconfigure it?

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,141 @@
...
+     * User will have to do more setup (e.g. maintain a cert for the external webhook)
+     * It adds one more single point of failure: every pod creation has to go through the `mutating admission controller` webhook. This might also be a problem when the cluster getting bigger and bigger.
+
+1. **Dynamic Resource Binding**
+     
+     Users can create a StorageClass with `volumeBindingMode` set to `WaitForFirstConsumer`. This new setting instructs the volume provisioner to not create a volume immediately, and instead, wait for a pod using an associated PVC to run through scheduling. Longhorn can use this setting as a signal to wait for the pod to have `NodeID` before scheduling replicas. As we can see, this setting is just informational to instruct CSI driver. Also, the logic in the `Volume Controller` will get more complicated. On the other hand, if we can implement this feature, it only helps when the volume is used the first time, it still doesn't help in case pod move to a different node.
+
+**My propose solution** is the combination of the component **Longhorn's ability to move replica to a new node** and **Kubernetes Scheduler Framework and Plugins**
```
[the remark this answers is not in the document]
yasker: I also don't understand the reason behind picking this approach. Recompile the scheduler is big. Ask the user to change the Pod spec is a big ask.

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,141 @@
...
+A user has three hyper-converged nodes and default settings with: `default-replica-count: 2`. He wants to ensure a pod always runs with at least one local replica would reduce the amount of network traffic needed to keep the data in sync. There does not appear to be an obvious way for him to schedule the pod using affinities.
+
+#### Story 2
+
+A user runs a database application that is able to do replication itself. The database app creates multiple pods and each pod uses a Longhorn volume with `replica-count = 1`. The database application knows how to schedule pods into different nodes so that they achieve HA. The problem is that replicas of multiple volumes could land on the same node which destroys the HA capability. With the `data locality` feature we can ensure that replicas are on the same nodes with the consuming pods and therefore they are on different nodes.
+
+### User Experience In Detail
+
+There are 3 modes for `dataLocality`:
+1. `do-nothing`: this is the default setting. Longhorn will not attempt to keep a local replica on the same node as the consuming pod
```
yasker, line 98: Just simply `Disabled`.

enhancements/20200819-keep-a-local-replica-to-engine.md
yasker: `best-effort` and `enforced` sounds good.

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,141 @@
...
+2. `dynamic`: Longhorn will try the best to keep a local replica on the same node as the consuming pod. Longhorn can achieve it by schedule pod on to the node which already contains a replica. If the pod cannot be scheduled into the nodes which contain the existing replica, Longhorn will try to move replica to the node where the pod is running. If Longhorn is unable to move replica to the node where the pod is running, it will not stop the pod. This probably is the most popular option. 
+3. `static`: pod has to be scheduled onto the node which contains an existing replica. Otherwise, the pod will be stuck in a pending unscheduled state. The idea here is that for some workloads such as distributed databases or message buses, it might be preferable to avoid scheduling a pod that would run more slowly than its peers, ensuring all running instances respond with similar latency. 
+
+There are 3 things the user can change for `data locality` setting:
+
+1. need to specify `spec.schedulerName: longhorn-scheduler` in the spec of all pods which use Longhorn volumes.
```
yasker, line 104: Ask I mentioned above, it's a big ask.

enhancements/20200819-keep-a-local-replica-to-engine.md
yasker: The global setting should only function as a default value, like replica count. It shouldn't change any existing volume's setting.

On change volume's setting side, it seems the setting can be changed without much consequence once the volume is detached. Changing the attached volume sounds like another story.

And yes, definitely should be an option per storageclass.

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,141 @@
...
+
+1. need to specify `spec.schedulerName: longhorn-scheduler` in the spec of all pods which use Longhorn volumes. 
+1. specify `dataLocality` mode for each volume
+1. specify `dataLocality` mode for all volume in the global setting
+
+**Note:** If the user forgets to add `spec.schedulerName` but the user turns on `data-locality`, the `data-locality` behavior will be `undefined`
+
+### API changes
+
+No API changes
```
yasker, line 112: There is setting added per global and in volume level, probably both creation and update. So there are API changes and we need to define the API names etc.

enhancements/20200819-keep-a-local-replica-to-engine.md
PhanLe1010: In my proposal solution, user need to specify 2 things for `dataLocality` settings:
1. need to specify `spec.schedulerName: longhorn-scheduler` in the spec of all pods which use Longhorn volumes.
1. need to specify `dataLocality` mode for each volume

By saying the user misconfigures, I mean that user specify specify `dataLocality` mode for each volume (e.g. the setting in storageclass) but forget to  specify `spec.schedulerName: longhorn-scheduler`. In this case, the behavior of Longhorn will be undefined. 

For example, user  set `dataLocality=enforced` mode in storageclass but  forget to  specify `spec.schedulerName: longhorn-scheduler`. In this case Kubernetes does not use `longhorn-scheduler` profile, therefore Longhorn cannot influence the scheduling descision.

enhancements/20200819-keep-a-local-replica-to-engine.md
PhanLe1010: Thank you! I will fix this

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,141 @@
...
+
+     1. First, in the `filter` phase, the `default scheduler` filters out all the un-schedulable nodes base on the requirements (like CPU, memory, selector, affinity, ..). At the end of the `filter` phase, the result consists of a list of nodes that are compatible with the given Pod. This list of compatible nodes is sent to the first webhook. Longhorn can build an Http server that listens on the webhook. When the Http server receives the list of compatible nodes, it can decide whether it needs to filter out nodes that don't have a replica of volume depend on the `data locality` setting. After that, a new list of nodes is passed to the second phase.
```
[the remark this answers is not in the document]
PhanLe1010: > We will need filter in the case where we enforce data locality.
For best effort data locality we shouldn't filter out any nodes and instead only use a custom scoring function in the prioritize call, otherwise we potentially end up with an empty list after the filter call.

Yes, the propose solution is doing this. I will fix the doc to reflect this. 

> Also it's not a new list it's the resulting list (output of filter) that gets passed to the prioritize call.

Will fix this

enhancements/20200819-keep-a-local-replica-to-engine.md
PhanLe1010: I will try this

enhancements/20200819-keep-a-local-replica-to-engine.md
PhanLe1010: Yes, you are correct! If we want Longhorn to automatically set ` spec.schedulerName:longhorn-scheaduler `, we can implement a `Mutating Admission Controller` as mentioned in this document. StorageOS is doing this. However, there are disadvantages when using a `Mutating Admission Controller` (also mentioned on this doc) which discourage me from using a `Mutating Admission Controller`

enhancements/20200819-keep-a-local-replica-to-engine.md
PhanLe1010: Thank you! I will update the document to reflect this.

enhancements/20200819-keep-a-local-replica-to-engine.md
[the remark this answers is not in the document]
PhanLe1010: > Recompile the scheduler is big.

So sorry, I was wrong here. I mentioned before that by using `Kubernetes Scheduler Framework`, we will modify the user's default scheduler and there will be only 1 scheduler in the cluster. I was wrong. The proposing solution is actually building a separate custom scheduler using `Kubernetes Scheduler Framework`. We will build 2 custom plugins for `filter` and `scoring` extension points and use `Kubernetes Scheduler Framework` to compile the 2 plugins with all the default plugins. The end result is a custom, performant scheduler without having to fork the core scheduler's code.  The building and compiling happen on our end.  We then can push and image of this scheduler onto DockerHub. Then can deploy the custom scheduler to the user's cluster using a deployment of 1 pod. 

> Ask the user to change the Pod spec is a big ask.

In order to use a custom scheduler, the pod's spec must have `schedulerName` set to a profile name in the custom scheduler. If we don't want users to manually set this field, we can implement a `Mutating Admission Controller` as mentioned in this document.  The `Mutating Admission Controller` will automatically `schedulerName`. StorageOS is doing this. However, there are disadvantages when using a Mutating Admission Controller (also mentioned on this doc) which discourage me from using it.

enhancements/20200819-keep-a-local-replica-to-engine.md
PhanLe1010: thank you! I will fix this

enhancements/20200819-keep-a-local-replica-to-engine.md
PhanLe1010: Same as https://github.com/longhorn/longhorn/pull/1705/files#r474361317

enhancements/20200819-keep-a-local-replica-to-engine.md
PhanLe1010: Thank you!

enhancements/20200819-keep-a-local-replica-to-engine.md
PhanLe1010: Thank you. I will try to subtract.

Note sure why we need to `look at the Storage Capacity of the available nodes.` We only give more point for the nodes which contains existing replicas. It means that the node already has enough  Storage Capacity. Is it correct?

yasker:
We need to clean up this LEP to make it qualify as a Longhorn LEP. This document in the current format is more like notes of discussion and brainstorming, rather than a formal design document for a feature.

Here are a few things we can do about this doc:
1. For the LEP itself, limit the topic to migrating the replica to the pod/node. State the scheduling change as a `non-goal` for this part of enhancement, and refer to the issue for details.
2. Summary the discussion and thought process and post them to https://github.com/longhorn/longhorn/issues/1045 instead of a part of LEP.
    1. Remove creating our own scheduler. It's out of the question.
    2. State the pros and cons with the other potential scheduler changes (either webhook or the new scheduling framework).
    3. Potential enhancement of Kubernetes.
    4. Anything else you think useful for understanding the choice.

PhanLe1010:
Thanks Sheng! working on this now.

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,224 @@
...
+### Test plan
+
+#### Manually Test Plan
+1. Create a cluster of 9 worker nodes and install Longhorn. 
+Having more nodes helps us to be more confident because the chance of randomly scheduling a replica onto the same node as the engine is small.
+1. Using Rancher to install an L7 ingress pointing to `longhorn-backend` service
+
+##### Test volume creation with `dataLocality` is `best-effort`:
+
+1. Open the ingress and go to volume API at `v1/volumes`
```
shuo-wu, line 113: Ideally, QA will test the case after the UI modification. I don't need to mention this. Same below

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,224 @@
...
+   After finishing rebuilding the local replica. Longhorn removes a random replica on other nodes to keep the number of replicas is 2.
+
+##### Test volume creation with `dataLocality` is `disable`:
+
+1. Open the ingress and go to volume API at `v1/volumes`
+1. Create another volume, `testvol2`  with `Number of Replicas = 2` and `dataLocality` is `disable`
+1. Attach `testvol2` to a node that doesn't contain any replica.
+1. Verify that Longhorn doesn't move replica
+
+##### Test volume creation with `dataLocality` is unspecified and `DefaultDataLocality` setting as `diable`:
```
shuo-wu, line 126: Typo `diable`

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,224 @@
...
+1. Leave the `DefaultDataLocality` setting as `diable` in Longhorn UI.
```
shuo-wu, line 127: Typo `diable`

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,224 @@
...
+      name: disable-longhorn
+    provisioner: driver.longhorn.io
+    allowVolumeExpansion: true
+    parameters:
+      numberOfReplicas: "1"
+      dataLocality: "disable"
+      staleReplicaTimeout: "2880" # 48 hours in minutes
+      fromBackup: ""
+    ```
+1. create a shell deployment of 1 pod using `ubuntu:xenial` image, and the PVC dynamically created by `disable-longhorn` storage class.
```
shuo-wu, line 172: I don't think we need to mention this kind of detail in LEP. Same below

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,224 @@
...
+    allowVolumeExpansion: true
+    parameters:
+      numberOfReplicas: "1"
+      staleReplicaTimeout: "2880" # 48 hours in minutes
+      fromBackup: ""
+    ```
+1. create a shell deployment of 1 pod using `ubuntu:xenial` image, and the PVC dynamically created by `unspecified-longhorn` storage class.
+1. The consuming pod is likely scheduled onto a different node than the replica. 
+1. If this happens, depend on `DefaultDataLocality` setting in Longhorn UI, verify that Longhorn does/doesn't migrate a local replica to the same node as the consuming pod.
+
```
shuo-wu, line 214: Add tests for the volumes created in old versions.

shuo-wu (changes requested):

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,224 @@
...
+A user runs a database application that can do replication itself. 
+The database app creates multiple pods and each pod uses a Longhorn volume with `replica-count = 1`. 
+The database application knows how to schedule pods into different nodes so that they achieve HA. 
+The problem is that replicas of multiple volumes could land on the same node which destroys the HA capability. 
+With the `data locality` feature we can ensure that replicas are on the same nodes with the consuming pods and therefore they are on different nodes.
+
+### User Experience In Detail
+
+There are 2 modes for `dataLocality`:
```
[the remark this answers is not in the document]
PhanLe1010: Thank you! Fixed

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,224 @@
...
+      If the scheduling success, we save the replica struct to the data store. This will trigger replica rebuilding on the consuming pod's node.
+   1. If there already exists a local replica on the consuming pod's node, we check to see if there are more healthy replica than specified on the volume's spec.
+      If there are more healthy replicas than specified on the volume's spec, we randomly remove a replica on the other nodes.
+
+### Test plan
+
+#### Manually Test Plan
+1. Create a cluster of 9 worker nodes and install Longhorn. 
+Having more nodes helps us to be more confident because the chance of randomly scheduling a replica onto the same node as the engine is small.
+1. Using Rancher to install an L7 ingress pointing to `longhorn-backend` service
```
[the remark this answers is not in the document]
PhanLe1010: Thank you! Fixed

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,224 @@
...
+1. When a volume enters the `volume controller`'s sync logic, function `ReconcileLocalReplica` checks the `dataLocality` mode of the volume.
+   If the `dataLocality` is `disable`, it will do nothing and return.
+1. If the `dataLocality` is `best-effort`, `ReconcileLocalReplica` checks whether there is a local replica on the same node as the volume.
+   1. If there is no local replica, we create an in-memory replica struct.
+      Then we try to schedule the replica struct onto the consuming pod's node.
+      If the scheduling fails, we don't do anything. The replica struct will be collected by Go's garbage collector.
+      If the scheduling success, we save the replica struct to the data store. This will trigger replica rebuilding on the consuming pod's node.
+   1. If there already exists a local replica on the consuming pod's node, we check to see if there are more healthy replica than specified on the volume's spec.
+      If there are more healthy replicas than specified on the volume's spec, we randomly remove a replica on the other nodes.
+
```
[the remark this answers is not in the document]
PhanLe1010: Thank you! Added

enhancements/20200819-keep-a-local-replica-to-engine.md
PhanLe1010: Thank you! Fixed

enhancements/20200819-keep-a-local-replica-to-engine.md
PhanLe1010: Thank you! Fixed

enhancements/20200819-keep-a-local-replica-to-engine.md
PhanLe1010: Thank you! Fixed

enhancements/20200819-keep-a-local-replica-to-engine.md
PhanLe1010: Thank you! Fixed

enhancements/20200819-keep-a-local-replica-to-engine.md
PhanLe1010: Thank you! Added

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,229 @@
...
+A user runs a database application that can do replication itself. 
+The database app creates multiple pods and each pod uses a Longhorn volume with `replica-count = 1`. 
+The database application knows how to schedule pods into different nodes so that they achieve HA. 
+The problem is that replicas of multiple volumes could land on the same node which destroys the HA capability. 
+With the `data locality` feature we can ensure that replicas are on the same nodes with the consuming pods and therefore they are on different nodes.
+
+### User Experience In Detail
+
+User can use this enhancement feature to solve problems as described in [Story 1](#story-1) and [Story 2](#story-2)
```
shuo-wu, line 61: The experience is more like: Users do something first to enable the feature, then users will see something.
e.g.:
1. Users create a storageclass with `dataLocality: best-effort` set
2. Users launch a statefulset with the storageclass.
3. Users will find that there is always a replica on the node where the pod resides on

enhancements/20200819-keep-a-local-replica-to-engine.md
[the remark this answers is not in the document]
shuo-wu: Use a separate sub-section to clarify which kind of UI modification should be applied.

shuo-wu (changes requested):

enhancements/20200819-keep-a-local-replica-to-engine.md
PhanLe1010: Thank you. I fixed it

enhancements/20200819-keep-a-local-replica-to-engine.md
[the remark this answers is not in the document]
PhanLe1010: Thank you. Added

enhancements/20200819-keep-a-local-replica-to-engine.md
```diff
@@ -0,0 +1,244 @@
...
+      If there are more healthy replicas than specified on the volume's spec, we remove a replica on the other nodes. 
+      We prefer to delete replicas on the same disk, then replicas on the same node, then replicas on the same zone.
+
+UI modification:
+1. On volume creation, add an input field for `dataLocality`
+1. On volume detail page:
+   * On the right volume info panel, add a <div> to display `selectedVolume.dataLocality`
+   * On the right volume panel, in the Health row, add an icon for data locality status. 
+     Specifically, if  `dataLocality=best-effort` but there is not a local replica then display a warning icon.
+     Similar to the replica node redundancy wanring [here](https://github.com/longhorn/longhorn-ui/blob/0a52c1f0bef172d8ececdf4e1e953bfe78c86f29/src/routes/volume/detail/VolumeInfo.js#L47)
```
shuo-wu, line 122: typo `wanring `

shuo-wu (changes requested):
LGTM except for the typo

enhancements/20200819-keep-a-local-replica-to-engine.md
PhanLe1010: Thank you! Fixed. I am sorry for the late response.

shuo-wu: Wait... The typo is still there.

yasker (approved):

Issue #508, opened 2019-05-02
Provide a choice to replace revision counter


yasker:
The record of revision counter results in one additional direct_io writes to the disk, which caused the performance penalty, especially this feature is more for data validation purpose.

We will see if we can replace the revision counter using some existing mechanism.

yasker:
The basic design:
1. The engine will ask replica for the FIEMAP of the volume head and compare them. They should be the same.
2. If not all replicas report the same FIEMAP, check the timestamp for the last access time of volume head (precision is nanoseconds). Accept the latest one and the ones with the same FIEMAP as it.
    2.1. This will require servers to have NTP servers running.
    2.2. One enhancement is Longhorn manager should warn about time disparity on the different nodes.

On top of this, the user can still manually choose which replicas to be taken into consideration when `salvage` the volume from `faulted` state.

boknowswiki:
Created UI issue at https://github.com/longhorn/longhorn/issues/1839

boknowswiki:
@khushboo-rancher , can you use this UI image `wangsiye/longhorn-ui:71f804a` to test this feature? Since I just filed a UI issue for this feature, and it's a long holiday in China now, @smallteeths probably can't work on this UI changes very soon, and we can unblock our side with this UI image. Please let me know if you have any concern. Thanks.

boknowswiki:
Filed engine integration test at [#1840 ](https://github.com/longhorn/longhorn/issues/1840)

meldafrawi:
Validation: **WIP**

**Steps to test:**
-  `Diable Revision Counter` Setting should be disabled by default.  (**PASS**)
- `Disable Revision Counter` property can be set per volume at creation time.

General Steps:
1. Disable scheduling on Node2, and Node3
2. Set `Replica Node Level Soft Anti-Affinity` to True
3. Create a volume with 2 replicas, all replicas should scheduled to Node1
4. Attach a volume to a node, mount it to a directory
5. Identify and exec to `instance-manager-r-xxx` pod scheduled to Node1
6. Identify volume replicas names and PIDs.

| Replica Name  | Replica PID |
| ------------- | ------------- |
| Replica 1  | PID1  |
| Replica 2  | PID2  |

7. Using the following line, write data to the volume to `data.txt` file
`$ for x in {1..1000000}; do echo $x >> <PATH_TO_VOLUME_MNT_DIR>/data.txt ; sleep 0.01; sync; echo $x ;done`

[...]

PR #1710, opened 2020-08-22
Add replace revision counter in longhorn engine enhancement document

boknowswiki:
https://github.com/longhorn/longhorn/issues/508

Signed-off-by: Bo Tao <bo.tao@rancher.com>

### DIFF
--- a/enhancements/20200821-add-revision-counter-disable-support.md
+++ b/enhancements/20200821-add-revision-counter-disable-support.md
@@ -1,0 +1,124 @@
+# Add Revision Counter Disable Support
+
+## Summary
+
+This enhancement adds revision counter disable support globally and individually per each volume with a logic based on latest modified time and total block counts of volume-head image file to select the most updated replica rather than the biggest revision counter for salvage recovering.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/508
+
+## Motivation
+
+### Goals
+
+1. By default 'DisableRevisionCounter' is 'false', but Longhorn provides an optional for user to disable it.
+2. Once user set 'DisableRevisionCounter' to 'true' globally or individually, this will improve Longhorn data path performance.
+3. And for 'DisableRevisionCounter' is 'true', Longhorn will keep the ability to find the most suitable replica to recover the volume when the engine is faulted(all the replicas are in 'ERR' state).
+4. Also during Longhorn Engine starting, with head file information it's unlikly to find out out of synced replicas. So will skip the check.
+
+## Proposal
+
+Currently Longhorn Replica will update revision counter to the 'revision.counter' file on every write operation, this impacts the performance a lot. And the main purpose of this revision counter is to pick the most updated replica to recover the volume.
+
+Also every time when Longhorn Engine starts, it will check all the replicas' revision counter to make sure the consistency.
+
+Having an option to disable the revision counter with a new logic which selects the most updated replica based on the last modified time and the head file size of the replica, and removing the revision counter mechanism. These can improve the Longhorn performance and keep the salvage feature as the same.
+
+Longhorn Engine will not check the synchronization of replicas during the starting time.
+
+### User Stories
+
+Some user has concern about Longhorn's performance. And with this 'DisableRevisionCounter' option, Longhorn's performance will be improved a lot.
+
+#### TODO Will update the performance data here later.
+
+### User Experience In Detail
+
+In Longhorn UI 'Setting' 'General', by default 'DisableRevisionCounter' is 'false', user can set 'DisableRevisionCounter' to 'true'. This only impacts the UI setting.
+
+Or from StorageClass yaml file, user can set 'parameters' 'revisionCounterDisabled' to true. And all the volume created based on this storageclass will have the same 'revisionCounterDisabled' setting.
+
+User can also set 'DisableRevisionCounter' for each individual volumes created by Longhorn UI this individual setting will over write the global setting.
+
+Once the volume has 'DisableRevisionCounter' to 'true', there won't be revision counter file. And the 'Automatic salvage' is 'true', when the engine is fauled, the engine will pick the most suitable replica as 'Source of Truth' to recover the volume.
+
+### API changes
+
+1. Add 'DisableRevisionCounter' global setting in Longhorn UI 'Setting' 'General'.
+2. Add 'DisableRevisionCounter' individual setting in Longhorn UI 'Volume' 'Create Volume' for each volume.
+3. For CSI driver, need to update volume creation API.
+4. Add new parameter to 'ReplicaProcessCreate' to disable revision counter. And new parameter to 'EngineProcessCreate' to indicate the salvage requested mode for recovering.
+5. Update Longhorn Engine Replica proto 'message Replica' struct with two new fields 'last_modified_time' and 'head_file_size' of the head file. This is for Longhorn Engine Control to get these information from Longhorn Engine Replica.
+
+## Design
+
+### Implementation Overview
+
+This enhancement has two phases, the first phase is to enable the setting to disable revision counter. The second phase is the implement the new gRPC APIs with new logic for salvage.
+
+And for the API compatibility issues, always check the 'EngineImage.Statue.cliAPIVersion' before making the call.
+
+#### Disable Revision Counter
+
+1. Add 'Volume.Spec.RevisionCounterDisabled', 'Replica.Spec.RevisionCounterDisabled' and 'Engine.Spec.RevisionCounterDisabled' to volume, replica and engine objects.
+2. Once 'RevisionCounterDisabled' is 'true', volume controller will set 'Volume.Spec.RevisionCounterDisabled' to true, 'Replica.Spec.RevisionCounterDisabled' and 'Engine.Spec.RevisionCounterDisabled' will set to true. And during 'ReplicaProcessCreate' and 'EngineProcessCreate' , this will be passed to engine replica process and engine controller process to start a replica and controller without revision counter.
+3. During 'ReplicaProcessCreate' and 'EngineProcessCreate', if 'Replica.Spec.RevisionCounterDisabled' or 'Engine.Spec.RevisionCounterDisabled' is true, it will pass extra parameter to engine replica to start replica without revision counter or to engine controller to start controller without revision counter support, otherwise keep it the same as current and engine replica will use the default value 'false' for this extra paramter. This is the same as the engine controller to set the 'salvageRequested' flag.
+4. Add 'RevisionCounterDisabled' in 'ReplicaInfo', when engine controller start, it will get all replica information.
+4. For engine controlloer starting cases:
+- If revision counter is not disabled, stay with the current logic.
+- If revision counter is disabled, engine will not check the synchronization of the replicas.
+- If unexpected case (engine controller has revision counter diabled but any of the replica doesn't, or engine controller has revision counter enabled, but any of the replica doesn't), engine controller will log this as error and mark unmatched replicas to 'ERR'.
+
+#### Add New Logic for Salvage
+
+Once the revision counter has been disabled.
+
+1. Add 'SalvageRequested' in 'InstanceSpec' and 'SalvageExecuted' in 'InstanceStatus' to indicate salvage recovering status. If all replicas are failed, 'Volume Controller' will set 'Spec.SalvageRequested' to 'true'.
+2. In 'Engine Controller' will pass 'Spec.SalvageRequested' to 'EngineProcessCreate' to trigger engine controller to start with 'salvageRequested' is 'true' for the salvage logic.
+3. The salvage logic gets details of replicas to get the most suitable replica for salvage.
+- Based on 'volume-head-xxx.img' last modified time, to get the latest one and any one within 5 second can be put in the candidate replicas for now.
+- Compare the head file size for all the candidate replicas, pick the one with the most block numbers as the 'Source of Truth'.
+- Only mark one candidate replica to 'RW' mode, the rest of replicas would be marked as 'ERR' mode.
+4. Once this is done, set 'SalvageExecuted' to 'true' to indicate the salvage is done and change 'SalvageRequested' back to false.
+
+### Test plan
+
+#### Disable revision counter option test case
+
+Disable revision counter option should return error if only Longhorn Manager got upgraded, not Longhorn Engine.
+
+This is when user trying to disable revision counter with new Longhorn Manager but the Longhorn Engine is still the over version which doesn't have this feature support. In this case, UI will shows error message to user.
+
+#### Disable revision counter options
+
+Revision counter can be disabled globally via UI by set 'Setting' 'General' 'DisableRevisionCounter' to 'true'.
+
+It can be set locally per volume via UI by set 'Volume' 'Create Volume''DisableRevisionCounter' to true.
+
+It can be set via 'StorageClass', and every PV created by this 'StorageClass' will inherited the same setting:
+```yaml
+    kind: StorageClass
+    apiVersion: storage.k8s.io/v1
+    metadata:
+      name: best-effort-longhorn
+    provisioner: driver.longhorn.io
+    allowVolumeExpansion: true
+    parameters:
+      numberOfReplicas: "1"
+      disableRevisionCounter: "true"
+      staleReplicaTimeout: "2880" # 48 hours in minutes
+      fromBackup: ""
+```
+
+#### Integration Test Plan
+1. Disable the revision counter.
+2. Create a volume with 3 replicas.
+3. Attach the volume to a node, and start to write data to the volume.
+4. Kill the engine process during the data writing.
+5. Verify the volume still works fine.
+6. Repeat the above test multiple times.
+
+### Upgrade strategy
+
+Deploy Longhorn image with v1.0.2 and upgrade Longhorn Manager, salvage function should still work. And then update Longhorn Engine, the revision counter disabled feature should be availabe.


enhancements/20200821-replace-revision-counter-in-longhorn-engine.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Motivation
+
+### Goals
+
+To improve Longhorn data path performance for reading and writing. And when the engine is failed (all the replicas are in 'ERR' state), Longhorn can select one most updated replica to recover the volume.
+
+## Proposal
+
+Currently Longhorn Replica will update revision counter to the 'revision.counter' file on every write operation, this impacts the performance a lot. And the main purpose of this revision counter is to pick the most updated replica to recover the volume.
```
shuo-wu, line 19: We will check the revision counter to prevent inconsistency among replicas each time the engine is started, too:
https://github.com/longhorn/longhorn-engine/blob/17c3e3ea20ea660a025b0b8c14e2c3f6e3f186e8/pkg/controller/control.go#L552

Remember to update the summary and goal with this info, too.

enhancements/20200821-replace-revision-counter-in-longhorn-engine.md
```diff
@@ -0,0 +1,56 @@
...
+
+Having a new logic which selects the most updated replica based on the last modified time and the total block numbers of the replica, and removing the revision counter mechanism. These can improve the Longhorn performance and keep the salvage feature as the same.
+
+### User Stories
+
+This enhancement doesn't impact any user experience.
```
shuo-wu, line 25: Not sure if we need to mention the performance improvement here.

enhancements/20200821-replace-revision-counter-in-longhorn-engine.md
```diff
@@ -0,0 +1,56 @@
...
+
+### API changes
+
+No API change.
+
+## Design
+
+### Implementation Overview
+
+1. Add Longhorn Engine Replica new gRPC API like 'ReplicaVolumeHeadStatGet' to get 'volume-head-xxx.img' stat information which includes 'last modified time' and 'block numbers' allocated to this file. This is for Longhorn Engine Control to get these information from Longhorn Engine Replica.
```
shuo-wu, line 39: Need to add this kind of gRPC API changes in Section `API changes`

enhancements/20200821-replace-revision-counter-in-longhorn-engine.md
```diff
@@ -0,0 +1,56 @@
...
+2. Add Longhorn Engine Control new gRPC API like 'GetCandidateReplicas' for Longhorn Manager to get the candidate replicas for salvage from Longhorn Engine Control. 
+	* Based on 'volume-head-xxx.img' last modified time, to get the latest one and any one within 5 second can be put in the candidate replicas for now.
+	* Compare the block numbers for all the candidate replicas, pick the one with the most block numbers as the 'Source of Truth', if some other candidate replicas have the same block numbers, they can be keep in the candidate replicas, otherwise remove them from the list.
+	* Return this candidate replicas list to the caller.
+3. Update the logic in Volume Controller 'ReconcileVolumeState', when the volume is at 'allFaulted' state and 'Automatic salvage' has been enabled, get the candidate replicas for salvage from 'GetCandidateReplicas' function call. And mark the rest replicas as 'ERR' state.
```
shuo-wu, line 44: 1. Need to consider the engine status. IIRC, the engine will be detached and reattached once all replicas become faulted. And the gRPC call cannot be invoked once the engine process is gone.
2. Need to consider the compatibility issue.

shuo-wu (changes requested):

enhancements/20200821-replace-revision-counter-in-longhorn-engine.md
boknowswiki: I will update the 'Proposal and Goal'. Do you think we need to add this specific case to the 'Summary'? I think the 'Summary' is more like a high level description. Please let me know. Thanks.

enhancements/20200821-replace-revision-counter-in-longhorn-engine.md
boknowswiki: I am not sure either. Let me add it for now and see how @yasker says.

enhancements/20200821-replace-revision-counter-in-longhorn-engine.md
boknowswiki: Oh, I see. I was thinking the 'API changes' was for the UI. I will update this part.

enhancements/20200821-replace-revision-counter-in-longhorn-engine.md
boknowswiki: Thanks for pointing this, let me think about the engine status case and sync up with you later.

enhancements/20200821-replace-revision-counter-in-longhorn-engine.md
shuo-wu: No. `Summary` LGTM.

enhancements/20200821-replace-revision-counter-in-longhorn-engine.md
boknowswiki: You are right about the engine will be detached. So I am thinking to call the replica gRPC to get the stat directly, what do you think this idea?
And for the compatibility issue, I am thinking if we call the new gRPC failed, then we do the original way.
Please let me know what do you think. Thanks.

boknowswiki: After discuss with @yasker , I should stay with the original idea, and to auto-attach the engine during the 'Faulted' state, and then auto-detach it. I will update this part.

enhancements/20200821-add-revision-counter-disable-support.md
```diff
@@ -0,0 +1,89 @@
...
+https://github.com/longhorn/longhorn/issues/508
+
+## Motivation
+
+### Goals
+
+1. By default 'RevisionCounterDisabled' is 'false', but Longhorn provides an optional for user to disable it.
+2. Once user set 'RevisionCounterDisabled' to 'true' globally or individually, this will improve Longhorn data path performance.
+3. And for 'RevisionCounterDisabled' is 'true', Longhorn will keep the ability to find the most suitable replica to recover the volume when the engine is faulted(all the replicas are in 'ERR' state).
+4. Also during Longhorn Engine starting, be able to connect to all the synced replicas and mark the out of synced replicas as 'ERR' state.
```
yasker, line 18: We might want to reconsider if it's possible if we disabled the revision counter.

enhancements/20200821-add-revision-counter-disable-support.md
```diff
@@ -0,0 +1,89 @@
...
+
+Also every time when Longhorn Engine starts, it will check all the replicas' revision counter to make sure the consistency.
+
+Having an option to disable the revision counter with a new logic which selects the most updated replica based on the last modified time and the total block numbers of the replica, and removing the revision counter mechanism. These can improve the Longhorn performance and keep the salvage feature as the same.
+
+With the head file information, Longhorn Engine can still mark the out of synced replica as 'ERR' during the starting time.
+
+### User Stories
+
+Some user has concern about Longhorn's performance. And with this 'RevisionCounterDisabled' option, Longhorn's performance will be improved a lot. And give user better experience.
```
yasker, line 32: Remove `And give user better experience` since it's unclear what is `experience` referring to here.

Also, we can collect some preliminary performance data here.

enhancements/20200821-add-revision-counter-disable-support.md
```diff
@@ -0,0 +1,89 @@
...
+
+### User Experience In Detail
+
+In Longhorn UI 'Setting' 'General', by default 'RevisionCounterDisabled' is 'false', user can set 'RevisionCounterDisabled' to 'true', this will apply to all the new volumes both from Longhoron UI or StorageClass.
```
yasker, line 36: Should be applied only to the volumes created by the Longhorn UI. StorageClass has a separate setting for that.

enhancements/20200821-add-revision-counter-disable-support.md
```diff
@@ -0,0 +1,89 @@
...
+
+1. The logic details of 'ReplicaList' with 'salvage' to get the most suitable replica for salvage.
+- Based on 'volume-head-xxx.img' last modified time, to get the latest one and any one within 5 second can be put in the candidate replicas for now.
+- Compare the block numbers for all the candidate replicas, pick the one with the most block numbers as the 'Source of Truth', if some other candidate replicas have the same block numbers, they can be keep in the candidate replicas, otherwise remove them from the list.
+- Only return one candidate replica to the caller.
+2. Update the logic in Volume Controller 'ReconcileVolumeState', when the volume is at 'allFaulted' state, get the candidate replicas for salvage from 'GetCandidateReplicas' function call. And mark the rest replicas as 'ERR' state.
+
+### Test plan
+
+####TODO how to make engine to fauled state. One way is to failed to ping monitor, like connection or modifing the replica files.
```
c3y1huang, line 81:
```suggestion
####TODO how to make engine to a failed state. One way is to failed to ping monitor, like connection or modifying the replica files.
```

enhancements/20200821-add-revision-counter-disable-support.md
```diff
@@ -0,0 +1,89 @@
...
+And for the API compatibility issues, always check the 'EngineImage.Statue.cliAPIVersion' before making the call.
+
+#### Disable Revision Counter
+
+1. Add 'Volume.Spec.RevisionCounterDisabled' and 'Replica.Status.RevisionCounterDisabled' to volume and replica objects.
+2. Once 'RevisionCounterDisabled' is 'true' either from global setting or individual setting, volume controller will set 'Volume.Spec.RevisionCounterDisabled' to true, and 'Replica.Status.RevisionCounterDisabled' to true. And during 'ReplicaCreateProcess', this will be passed to engine replica process to start a replica without revision counter.
+3. For engine control, it can get 'Replica.Status.RevisionCounterDisabled' from each replicas.
+- If revision counter is not disabled, stay with the current logic.
+- If revision counter is disabled, comparing the head file size to mark the out of synced replica as 'ERR' state.
+- If Mixed case, use the head file size for the comparision. And print warning about it.
```
c3y1huang, line 67:
```suggestion
- If Mixed case, use the head file size for the comparison. And print warning about it.
```

enhancements/20200821-replace-revision-counter-in-longhorn-engine.md
shuo-wu: According to Sheng's comment, we need to collect the performance data here.

enhancements/20200821-add-revision-counter-disable-support.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Proposal
+
+Currently Longhorn Replica will update revision counter to the 'revision.counter' file on every write operation, this impacts the performance a lot. And the main purpose of this revision counter is to pick the most updated replica to recover the volume.
+
+Also every time when Longhorn Engine starts, it will check all the replicas' revision counter to make sure the consistency.
+
+Having an option to disable the revision counter with a new logic which selects the most updated replica based on the last modified time and the total block numbers of the replica, and removing the revision counter mechanism. These can improve the Longhorn performance and keep the salvage feature as the same.
+
+With the head file information, Longhorn Engine can still mark the out of synced replica as 'ERR' during the starting time.
```
shuo-wu, line 28: Clarify how Longhorn can verify the out-of-date replica via the head file information.

enhancements/20200821-add-revision-counter-disable-support.md
```diff
@@ -0,0 +1,89 @@
...
+User can also set 'RevisionCounterDisabled' for each individual volumes created by Longhorn UI or StorageClass, this individual setting will take higher priority than the global setting.
+
+Once the volume has 'RevisionCounterDisabled' to 'true', there won't be revision counter file. And the 'Automatic salvage' is 'true', when the engine is fauled, the engine will pick the most suitable replica as 'Source of Truth' to recover the volume.
+
+### API changes
+
+1. Add 'RevisionCounterDisabled' global setting in Longhorn UI 'Setting' 'General'.
+2. Add 'RevisionCounterDisabled' individual setting in Longhorn UI 'Volume' 'Create Volume' for each volume.
+3. For CSI driver, need to update volume creation API.
+4. Add new parameter to 'ReplicaCreateProcess' to disable revision counter. And a new gRPC API for engine control to get this setting from replica.
```
shuo-wu, line 47: It seems that both instance manager gRPC API and the replica gRPC API need to be updated. Then you may need to clarify how to handle the compatibility issue in this doc.

shuo-wu (changes requested):

enhancements/20200821-add-revision-counter-disable-support.md
boknowswiki: I will add a 'TODO' in this section, and update this part once we have the data.

enhancements/20200821-add-revision-counter-disable-support.md
boknowswiki: Sure, will update this part.

enhancements/20200821-add-revision-counter-disable-support.md
boknowswiki: Will fix, thanls.

enhancements/20200821-add-revision-counter-disable-support.md
boknowswiki: Will fix, thanks.

enhancements/20200821-add-revision-counter-disable-support.md
boknowswiki: Will add more details there. thanks.

enhancements/20200821-add-revision-counter-disable-support.md
boknowswiki: Let me think more detail about it and update it later. May need to sync with you. Thanks.

enhancements/20200821-add-revision-counter-disable-support.md
boknowswiki: @shuo-wu , per discussion with @yasker , so far we haven't think a good way to tell if any replica is out of sync if some of the latest writes are on the same block. We will skip the check for now. Please let me know if you have any good idea about this part. Thanks.

enhancements/20200821-add-revision-counter-disable-support.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/508
+
+## Motivation
+
+### Goals
+
+1. By default 'RevisionCounterDisabled' is 'false', but Longhorn provides an optional for user to disable it.
```
yasker, line 15: It's better to call the setting `DisableRevisionCounter` (like `Disable Scheduling on Cordoned Node`) with the options `true` or `false`, or `RevisionCounter` with the options `enabled`/`disabled`.

boknowswiki: Sure. I will update it to 'DisableRevisionCounter'. Thanks.

enhancements/20200821-add-revision-counter-disable-support.md
shuo-wu: I think adding a new field in `replica/ReplicaGetResponse` is enough. We don't need a new separate engine GRPC API for it.

enhancements/20200821-add-revision-counter-disable-support.md
```diff
@@ -0,0 +1,128 @@
...
+### Implementation Overview
+
+This enhancement has two phases, the first phase is to enable the setting to disable revision counter. The second phase is the implement the new gRPC APIs with new logic for salvage.
+
+And for the API compatibility issues, always check the 'EngineImage.Statue.cliAPIVersion' before making the call.
+
+#### Disable Revision Counter
+
+1. Add 'Volume.Spec.RevisionCounterDisabled' and 'Replica.Status.RevisionCounterDisabled' to volume and replica objects.
+2. Once 'RevisionCounterDisabled' is 'true' either from global setting or individual setting, volume controller will set 'Volume.Spec.RevisionCounterDisabled' to true, and 'Replica.Status.RevisionCounterDisabled' to true. And during 'ReplicaCreateProcess', this will be passed to engine replica process to start a replica without revision counter.
```
shuo-wu, line 67: Why not put `RevisionCounterDisabled` in `Replica.Spec`?

enhancements/20200821-add-revision-counter-disable-support.md
```diff
@@ -0,0 +1,128 @@
...
+3. For engine control, it can get 'Replica.Status.RevisionCounterDisabled' from each replicas.
+- If revision counter is not disabled, stay with the current logic.
+- If revision counter is disabled, engine will not check the synchronization of the replicas.
+- If Mixed case, engine will not check the synchronation of the replicas. And print warning about it.
```
shuo-wu, line 71: What's this?

enhancements/20200821-add-revision-counter-disable-support.md
```diff
@@ -0,0 +1,128 @@
...
+4. Add new parameter to 'ReplicaCreateProcess' to disable revision counter. This probably won't have compatibility issue since it passed by arguments.
+5. Add new gRPC API in engine control may have a compatibility issue when a new version Longhorn manager calling an old version Longhorn engine which doesn't have this new API. And this issue should be able to resolve by checking the 'EngineImage.Statue.cliAPIVersion' before the call.
+- If 'EngineImage.Statue.cliAPIVersion' is less than 4, using salvage logic.Which clears the 'FailedAt' field.
+- If 'EngineImage.Statue.cliAPIVersion' is not less than 4, Longhorn manager can call the new gRPC API to get the replica candidate for salvage.
+
+#### Add New Logic for Salvage
+
+Once the revision counter has been disabled.
+
+1. The logic details of 'ReplicaList' with 'salvage' to get the most suitable replica for salvage.
```
shuo-wu, line 81: Do you mean the command in `longhorn-manager/engineapi`?

enhancements/20200821-add-revision-counter-disable-support.md
```diff
@@ -0,0 +1,128 @@
...
+- Based on 'volume-head-xxx.img' last modified time, to get the latest one and any one within 5 second can be put in the candidate replicas for now.
```
shuo-wu, line 82: Do you mean the engine process is responsible for figuring out the candidates? When all replica processes become ERROR, how does the engine process get this kind of info?

enhancements/20200821-add-revision-counter-disable-support.md
```diff
@@ -0,0 +1,128 @@
...
+- Compare the block numbers for all the candidate replicas, pick the one with the most block numbers as the 'Source of Truth', if some other candidate replicas have the same block numbers, they can be keep in the candidate replicas, otherwise remove them from the list.
+- Only return one candidate replica to the caller.
+2. Update the logic in Volume Controller 'ReconcileVolumeState', when the volume is at 'allFaulted' state, get the candidate replicas for salvage from 'GetCandidateReplicas' function call. And mark the rest replicas as 'ERR' state.
+
+### Test plan
+
```
shuo-wu, line 88: After skipping the longhorn engine starting check, adding a regression test would be better.
e.g.,
1. Disable the revision counter.
2. Kill the engine process during the data writing.
3. Verify the volume still works fine.
4. Repeat the above test multiple times.

enhancements/20200821-add-revision-counter-disable-support.md
```diff
@@ -0,0 +1,128 @@
...
+#### Revision counter disable option test case
+
+Revision counter disbled option should return error if only Longhorn Manager got upgraded, not Longhorn Engine.
```
shuo-wu, line 91: You mean disabling revision counter for old volumes will error out?

shuo-wu (changes requested):

enhancements/20200821-add-revision-counter-disable-support.md
boknowswiki: Sure, will fix it. Thanks.

enhancements/20200821-add-revision-counter-disable-support.md
boknowswiki: will clarify this part. Thanks.

enhancements/20200821-add-revision-counter-disable-support.md
boknowswiki: After discussion, let me take more look at this part.

enhancements/20200821-add-revision-counter-disable-support.md
boknowswiki: After discussion, will think about this part. Thanks.

enhancements/20200821-add-revision-counter-disable-support.md
boknowswiki: Will update this part for clarification.

boknowswiki:
@shuo-wu , I have updated this LEP based on our discussion, please help to take a look. Thanks.

enhancements/20200821-add-revision-counter-disable-support.md
```diff
@@ -0,0 +1,121 @@
...
+#### Add New Logic for Salvage
+
+Once the revision counter has been disabled.
+
+1. Add 'SalvageRequested' in 'InstanceSpec' and 'SalvageExecuted' in 'InstanceStatus' to indicate salvage recovering status.
+2. The logic details of 'ReplicaList' with 'salvage' to get the most suitable replica for salvage.
+- Based on 'volume-head-xxx.img' last modified time, to get the latest one and any one within 5 second can be put in the candidate replicas for now.
+- Compare the head file size for all the candidate replicas, pick the one with the most block numbers as the 'Source of Truth'.
+- Only mark one candidate replica to 'RW' mode, the rest of replicas would be marked as 'ERR' mode.
+3. Update the logic in Volume Controller 'ReconcileVolumeState', when the volume is at 'allFaulted' state, set 'SalvageRequested' in 'InstanceSpec' to 'true' and start engine controller with 'salvageRequested' is 'true', and this will trigger the new logic during engine controller starting to get only one replica as 'RW' and mark the rest to 'ERR'. One this is one, set 'SalvageExecuted' to 'true' to indicate the salvage is done and change 'SalvageRequested' back to false.
```
shuo-wu, line 80: During the auto salvage, longhorn-manager will still launch replicas first then launch the engine. And as I mentioned in the discussion, we need to consider what if some replicas cannot be launched.

enhancements/20200821-add-revision-counter-disable-support.md
```diff
@@ -0,0 +1,121 @@
...
+#### Disable Revision Counter
+
+1. Add 'Volume.Spec.RevisionCounterDisabled' and 'Replica.Spec.RevisionCounterDisabled' to volume and replica objects.
+2. Once 'RevisionCounterDisabled' is 'true', volume controller will set 'Volume.Spec.RevisionCounterDisabled' to true, and 'Replica.Spec.RevisionCounterDisabled' to true. And during 'ReplicaProcessCreate', this will be passed to engine replica process to start a replica without revision counter.
+3. During 'ReplicaProcessCreate', if 'Replica.Spec.RevisionCounterDisabled' is true, it will pass extra parameter to engine replica to start replica without revision counter, otherwise keep it the same as current and engine replica will use the default value 'false' for this extra paramter. This is the same as the engine controller to set the 'salvageRequested' flag.
+4. For engine controlloer starting cases:
+- If revision counter is not disabled, stay with the current logic.
+- If revision counter is disabled, engine will not check the synchronization of the replicas.
+
+#### Add New Logic for Salvage
```
shuo-wu, line 71: In this part, it's better to follow the salvage workflow to clarify what will happen after the refactoring. And explicitly clarifying which part (`longhorn-engine/controller`, `longhorn-manager/engineapi`, `longhorn-manager/controller`) the modification is applied to will be more easy-to-understand.

enhancements/20200821-add-revision-counter-disable-support.md
shuo-wu: Remember to update this statement. And there is a typo `disbled `

enhancements/20200821-add-revision-counter-disable-support.md
```diff
@@ -0,0 +1,121 @@
...
+    provisioner: driver.longhorn.io
+    allowVolumeExpansion: true
+    parameters:
+      numberOfReplicas: "1"
+      disableRevisionCounter: "true"
+      staleReplicaTimeout: "2880" # 48 hours in minutes
+      fromBackup: ""
+```
+
+#### Salvage replica when revision counter disabled
```
shuo-wu, line 111: Mention this test as a regression test.

shuo-wu (changes requested):

enhancements/20200821-add-revision-counter-disable-support.md
boknowswiki: Yes, we have discussed this, and I think @yasker mentioned replicas couldn't be launched was a separate case which our current design also had this issue. If the replica can not be launched, then engine can not connect to them, they won't be selected as candidate for recovering, right?

enhancements/20200821-add-revision-counter-disable-support.md
boknowswiki: Sure, will update.

boknowswiki:
@shuo-wu , any comments on the last PR? Thanks.

enhancements/20200821-add-revision-counter-disable-support.md
boknowswiki: Hi Shuo, I have tried the steps per our discussion (kill the replica pods, and then delete one replica directory), I don't hit the issue that engine won't be up until all the replicas are up. So I think it's fine. Also I think this feature doesn't change previous behavior. Please let me know what you think. Thanks.

shuo-wu (approved):

boknowswiki:
Hi @shuo-wu and @yasker , I just updated this LEP mainly the implementation overview part to add 'Engine.Spec.RevisionCounterDisabled' to engine object, which pass extra flag to engine controller to be aware of revision counter disabled or not. Please take a look and let me know. Thanks.

shuo-wu (approved):

yasker (approved):

PR #1773, opened 2020-09-09
Helm default schedule

sauber:
Allow to specify a schedule for snapshots and backups in default storage class created by helm.

values.yaml example:

```
persistence:
  defaultRecurringJobs: '[{\"name\":\"snap\", \"task\":\"snapshot\", \"cron\":\"*/1 * * * *\", \"retain\":1}, {\"name\":\"backup\", \"task\":\"backup\", \"cron\":\"*/2 * * * *\", \"retain\":1}]'
```

### DIFF
--- a/chart/questions.yml
+++ b/chart/questions.yml
@@ -175,6 +175,11 @@ questions:
     min: 1
     max: 10
     label: Default Storage Class Replica Count
+  - variable: persistence.defaultRecurringJobs
+    description: "Set snapshot and backup jobs for default StorageClass"
+    group: "Longhorn CSI Driver Settings"
+    type: string
+    default: '[]'
   - variable: defaultSettings.backupTarget
     label: Backup Target
     description: "The endpoint used to access the backupstore. NFS and S3 are supported."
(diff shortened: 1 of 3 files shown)


pertoft:
+1

yasker:
We have been pretty cautious about updating the default storage class since Kubernetes doesn't allow any modification to the parameter section of the storage class. Any modification to the default storage class can result in upgrade failure since the new storage class configuration cannot be applied directly. That's why we recommend creating separate storage classes instead of customizing the default storage class for the workload.

So I am leaning towards not adding recurring schedule in the storage class, since it gives an impression that this parameter can be tweaked. Also, this is a parameter that can easily get wrong. We also trying to see if if there is a better way to solve this problem. If storage class parameter can be changed somehow, then we would be glad to accept this PR.

yasker:
Since the last time we talked, we've implemented a way to upgrade the storage class in Longhorn. So recurring job field should be already covered by @PhanLe1010 on https://github.com/longhorn/longhorn/commit/9c3ae72954c02c8e9dc1396e1efb04d42ac1e346 , see https://github.com/longhorn/longhorn/blob/v1.1.0/chart/templates/storageclass.yaml.

sauber:
Thanks for the good news. The implementation looks promising. I look forward to test when v1.1.0 is released.

yasker:
Thanks @sauber . Close this PR, in favor of https://github.com/longhorn/longhorn/commit/9c3ae72954c02c8e9dc1396e1efb04d42ac1e346

PR #1834, opened 2020-09-29
chart: Remove some redundancy for the chart

shuo-wu:


chart/templates/_helpers.tpl
```diff
@@ -36,3 +36,12 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 app.kubernetes.io/version: {{ .Chart.AppVersion }}
 {{- end -}}
+
+
+{{- define "system_default_registry" -}}
```
yasker, line 41: What if the user is not using Rancher, how will the user install the chart in the air gap environment?

yasker: What's the answer?

shuo-wu: They can still specify the private registry by using flag `--set global.cattle.systemDefaultRegistry=private-registry` or `--set privateRegistry.registryUrl=private-registry` in the `helm install` command.

https://github.com/longhorn/longhorn/pull/1834/commits/7ff9bab624cf52797abf7b91d48be3ef0aaa1563#diff-a695820e1f93e7decd902e54573160d357bc57f3faeec7e501cc7d2c607baaaaR49

yasker: Do we have this documented anywhere? And should we change our questions.yaml to reflect that too? I am not clear what the air gap installation will look like without the option.

shuo-wu: We don't need to add anything in question.yaml. And the air gap installation will remain the same without this option.

This option is required by Rancher: https://rancher.atlassian.net/wiki/spaces/EN/pages/563609933/Helm+Charts+Requirements
Here is the example the Rancher team provide:
https://github.com/rancher/charts/blob/b55e6ec01907964723942fdbd9f63b9afef7a1ed/packages/rancher-cis-benchmark/charts/templates/_helpers.tpl#L6-L12

PR #1836, opened 2020-09-30
Add recurringJobs to StorageClass

### LINKED ISSUE #1129: [FEATURE] Set up recurring jobs from the helm chart
**Is your feature request related to a problem? Please describe.**

I would like to install the helm chart enabling recurring jobs 

**Describe the solution you'd like**

Based on how to [Set up recurring jobs using StorageClass](https://github.com/longhorn/longhorn/blob/c91680aa56beedb60c751163194ec5a26177615e/docs/snapshot-backup.md#set-up-recurring-jobs-using-storageclass), be able to use values for `reccurringJobs` parameter from the [chart/templates/storageclass.yaml](https://raw.githubusercontent.com/longhorn/longhorn/master/chart/templates/storageclass.yaml)

**Describe alternatives you've considered**
```yml
parameters:
  numberOfReplicas: "{{ .Values.persistence.defaultClassReplicaCount }}"
  staleReplicaTimeout: "30"
  fromBackup: ""
  baseImage: ""
  {{- if .Values.recurringJobs.enable }}
  recurringJobs: "{{ .Values.recurringJobs.jobsList }}"
  {{- end }}
```

**Additional context**
Same goes for `fromBackup` parameter BTW

jackivanov:
Fixes #1129

### DIFF
--- a/chart/templates/storageclass.yaml
+++ b/chart/templates/storageclass.yaml
@@ -12,3 +12,6 @@ parameters:
   staleReplicaTimeout: "30"
   fromBackup: ""
   baseImage: ""
+  {{- if .Values.recurringJobs.enable }}
+  recurringJobs: '{{ .Values.recurringJobs.jobsList | toPrettyJson | indent 2 | trim }}'
+  {{- end }}
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -25,6 +25,14 @@ persistence:
   defaultClass: true
   defaultClassReplicaCount: 3
 
+recurringJobs:
+  enable: false
+  jobsList: []
+    # - name: snap
+    #   task: snapshot
+    #   cron: '*/1 * * * *'
+    #   retain: 1
+
 csi:
   attacherImage: longhornio/csi-attacher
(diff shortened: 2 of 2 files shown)


yasker:
We can add an option to the storage class for recurring jobs, but we cannot add a default value to it, since the user should choose how they would want to setup the recurring job. Also, if there is no backup target, we cannot have a recurring backup job as well.

jackivanov:
@yasker I corrected the templates a bit, take a look please

chart/templates/storageclass.yaml
```diff
@@ -12,6 +12,6 @@ parameters:
   staleReplicaTimeout: "30"
   fromBackup: ""
   baseImage: ""
-  {{- if .Values.recurringJobs.enable }}
+  {{- if and .Values.recurringJobs.enable .Values.defaultSettings.backupTarget }}
```
yasker, line 15: In fact we don't need to check the backupTarget since recurring snapshot is a valid choice.

yasker:
Also, please squash the commits into one.

Issue #1527, opened 2020-06-26
[BUG] Prevent upgrade failure due to changed StorageClass


yasker:
When using Helm, the upgrade may fail due to Kubernetes cannot change StorageClass's certain parameters. It is inconvenient to the users. We need to figure out a way to deal with it automatically. Some ideas:
1. Should we make Helm chart replace the `longhorn` storageclass without complaining?
2. Should we compare the current `longhorn` storageclass with our record of latest `longhorn` storageclass and stop the upgrade process if there is a problem?
3. Should we move the StorageClass creation to the code instead of YAML?
4. What's the workaround for users with StorageClass upgrade failure?

Currently, we don't recommend modifying the StorageClass (https://longhorn.io/docs/1.0.0/best-practices/#storageclass ). But a predictable upgrade failure always sounds bad.

yasker:
@khushboo-rancher Can you help to try the following scenario and see if the steps we provide at https://longhorn.io/docs/0.8.1/deploy/upgrade/longhorn-manager/#error-longhorn-is-invalid-provisioner-forbidden-updates-to-provisioner-are-forbidden still works for v0.8.1 to v1.0.0?

1. Install Longhorn v0.8.1 using Rancher.
2. Modify the StorageClass `longhorn`, e.g. change replica count from 3 to 2.
3. Upgrade to v1.0.0.
    1. Observe the upgrade error
4. Try if https://longhorn.io/docs/0.8.1/deploy/upgrade/longhorn-manager/#error-longhorn-is-invalid-provisioner-forbidden-updates-to-provisioner-are-forbidden works to fix the upgrade.

PhanLe1010:
## Upgrading using Rancher from v0.8.1 to v1.0.0
1. Install Longhorn v.0.8.1 using Rancher
1. Delete `longhorn` storageclass. Create `longhorn` storageclass with new parameter `numberOfReplicas: 2`
1. Upgrade to Longhorn v.1.0.0 using Rancher
Upgrading succeed. 

## Upgrading using `kubectl apply -f` from v1.0.0 to v1.0.2
1. Install Longhorn v1.0.0 using:
   ```
   kubectl apply -f https://raw.githubusercontent.com/longhorn/longhorn/v1.0.0/deploy/longhorn.yaml
   ```
1. Delete `longhorn` storageclass. Create `longhorn` storageclass with new parameter `numberOfReplicas: 2`
1. Install Longhorn v1.0.2 using:
   ```
[...]

PR #1902, opened 2020-10-22
Prevent Longhorn upgrade failure due to changed StorageClass when using Helm installation/upgrade

PhanLe1010:
Add the logic to `storageclass.yaml` template to detect the existing `longhorn` storageclass. There are 3 cases:

1. If there is no `longhorn` storageclass, `storageclass.yaml` template output a yaml file for `longhorn` storageclass. Helm install/upgrade will succeed.
2. If there is a longhorn storageclass with provisoner `driver.longhorn.io`, storageclass.yaml template outputs an identical storageclass. Helm install/upgrade will succeed.
3. If there is a `longhorn` storageclass but its provisoner is not `driver.longhorn.io`, storageclass.yaml template outputs an error. Helm install will fail. In this case, users have to follow the [document](https://longhorn.io/docs/0.8.1/deploy/upgrade/longhorn-manager/#error-longhorn-is-invalid-provisioner-forbidden-updates-to-provisioner-are-forbidden) to resolve the problem. (In short, users have to delete the existing `longhorn` storageclass and retry)

longhorn/longhorn#1527

### DIFF
--- a/chart/templates/storageclass.yaml
+++ b/chart/templates/storageclass.yaml
@@ -1,3 +1,5 @@
+{{- $longhornSC := (lookup "storage.k8s.io/v1" "StorageClass" "" "longhorn") -}}
+{{- if not $longhornSC }}
 kind: StorageClass
 apiVersion: storage.k8s.io/v1
 metadata:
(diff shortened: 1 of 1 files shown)


PhanLe1010:
I need to do more testing on this one.

PR #1953, opened 2020-11-04
Reference correct PodSecurityPolicy name

zbialik:
ClusterRoleBinding must declare use of the PodSecurityPolicy properly. The current yaml template references a PSP named "nfs-provisioner", when the name of the PSP defined in this yaml is "longhorn-nfs-provisioner"

### DIFF
--- a/examples/rwx/01-security.yaml
+++ b/examples/rwx/01-security.yaml
@@ -44,7 +44,7 @@ rules:
     verbs: ["get"]
   - apiGroups: ["extensions"]
     resources: ["podsecuritypolicies"]
-    resourceNames: ["nfs-provisioner"]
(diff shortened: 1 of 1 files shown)


yasker:
Hmm, how did this happen? @khushboo-rancher any ideas?

@zbialik this file is from https://github.com/longhorn/longhorn-manager/blob/master/examples/rwx/01-security.yaml , can you update that one instead? We will sync it before the release.

zbialik:
@yasker Thanks for the help, I am going to try to create a new pull request. I work at an air-gapped facility so I can't use git command line nor and IDE to submit these commits, etc. So I am doing so in the Github Web UI directly.

PR #1954, opened 2020-11-04
Updating ClusterRole to reference correct PSP

zbialik:
Updating the ClusterRole such that it references the correct name of the PodSecurityPolicy. Current yaml references a PSP named "nfs-provisioner", while the yaml created a PSP named "longhorn-nfs-provisioner"

Signed-off-by: Zach Bialik <zachbialik1207@gmail.com>


yasker:
Hi @zbialik , please make the change to https://github.com/longhorn/longhorn-manager as I mentioned in https://github.com/longhorn/longhorn/pull/1953 . Thanks.

Issue #1818, opened 2020-09-24
task: Clean up Longhorn chart


yasker:
1. Remove questions.yaml
2. Move CRDs into `/crd` directory.

yasker:
Regarding this issue, @boknowswiki please work with @shuo-wu to understand the context and what should we do.

yasker:
We won't able to remove `questions.yaml` for now since we don't have the new UI ready yet.

shuo-wu:
It seems that we cannot move `crd.yaml` in this version. Since Helm will try to delete the CRDs defined in the crd yaml file when the crd yaml file is moved out from the template directory. 
In order to let Helm ignore the `templates/crd.yaml` removal, we can add an annotation `"helm.sh/hook": "pre-install"` for all CRDs in advance. But I haven't found a way to automatically add this annotation at the 1st step of the upgrade. Instead, I think we can modify the package `longhorn-manager/upgrade` to append the annotation. Then we can move the crd file after v1.1.0.

yasker:
Do we have any solution?

shuo-wu:
We can add [a special annotation `helm.sh/resource-policy: keep`](https://helm.sh/docs/topics/charts_hooks/#hook-resources-are-not-managed-with-corresponding-releases) for the CRDs in v1.1.0 release, then separate the CRD file in the following release. Since some users may not follow the upgrade guide, requiring users to add the annotation manually and separating the CRD file in v1.1.0 is not a good choice.

yasker:
@shuo-wu sounds good to me. Can you create a new issue to remind us to move it in v1.2.0?

shuo-wu:
OK. The CRD separation is tracked in https://github.com/longhorn/longhorn/issues/1879

yasker:
@shuo-wu this issue has been moved directly to `Merged` without having test steps ready.

shuo-wu:
Use the branch `master` of `https://github.com/rancher/charts` to test the updated chart.
The version `v1.1.0` in this branch is actually using the `master` images. And this chart is not for Rancher dev-v2.5. Instead, it's for the old-style Rancher APP.

Test 1:
Test if the private registry setting still works fine

Test 2:
1. Launch Longhorn v1.0.2
2. Deploy some workloads using Longhorn volumes.
3. Upgrade Longhorn to v1.1.0 via Rancher APP
4. Check if Longhorn system still works fine.

yasker:
[...]

Issue #1813, opened 2020-09-23
[Question] support for CustomResourceDefinition in kuberntes 1.19.2


arturmon:
Good day. updated kubernetes to 1.19.2 and got an error:  
```
Error from server (Invalid): error when creating "longhorn.yaml": CustomResourceDefinition.apiextensions.k8s.io "engines.longhorn.io" is invalid: spec.versions[0].schema.openAPIV3Schema: Required value: schemas are required
Error from server (Invalid): error when creating "longhorn.yaml": CustomResourceDefinition.apiextensions.k8s.io "replicas.longhorn.io" is invalid: spec.versions[0].schema.openAPIV3Schema: Required value: schemas are required
Error from server (Invalid): error when creating "longhorn.yaml": CustomResourceDefinition.apiextensions.k8s.io "settings.longhorn.io" is invalid: spec.versions[0].schema.openAPIV3Schema: Required value: schemas are required
Error from server (Invalid): error when creating "longhorn.yaml": CustomResourceDefinition.apiextensions.k8s.io "volumes.longhorn.io" is invalid: spec.versions[0].schema.openAPIV3Schema: Required value: schemas are required
Error from server (Invalid): error when creating "longhorn.yaml": CustomResourceDefinition.apiextensions.k8s.io "engineimages.longhorn.io" is invalid: spec.versions[0].schema.openAPIV3Schema: Required value: schemas are required
Error from server (Invalid): error when creating "longhorn.yaml": CustomResourceDefinition.apiextensions.k8s.io "nodes.longhorn.io" is invalid: spec.versions[0].schema.openAPIV3Schema: Required value: schemas are required
Error from server (Invalid): error when creating "longhorn.yaml": CustomResourceDefinition.apiextensions.k8s.io "instancemanagers.longhorn.io" is invalid: spec.versions[0].schema.openAPIV3Schema: Required value: schemas are required
```
after tweaking CustomResourceDefinition.
```
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  labels:
    longhorn-manager: Engine
  name: engines.longhorn.io
spec:
  group: longhorn.io
  names:
    kind: Engine
    listKind: EngineList
    plural: engines
    shortNames:
    - lhe
    singular: engine
  scope: Namespaced
  versions:
    - name: v1beta1
      served: true
      storage: true
---
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  labels:
[...]

PR #1971, opened 2020-11-11
chart: Sync chart with the YAML files in longhorn-manager

shuo-wu:
#1818, #1813


yasker:
Missing question for the recurring job.

shuo-wu:
There is no way to add variable type `array` for the StorageClass recurring job in `questions.yml`: https://github.com/rancher/rancher/issues/24918

Rancher team recommends to directly EDIT YAML file for this case.

yasker:
Is it possible to add the recurring job as a string though?

shuo-wu:
> Is it possible to add the recurring job as a string though? e.g. #1773

Yes. Using string will avoid the issue. But users need to take care of converting an array/map to a JSON string.

shuo-wu:
This PR is blocked by #1980

shuo-wu:
This PR is ready for review & merge

chart/questions.yml
```diff
@@ -322,15 +334,115 @@ WARNING: DO NOT CHANGE THIS SETTING WITH ATTACHED VOLUMES.'
...
+- variable: persistence.defaultClass
+  default: "true"
+  description: "Set as default StorageClass for Longhorn"
+  group: "Longhorn Storage Class Settings"
+  type: boolean
+  required: true
+  label: Default Storage Class
+- variable: persistence.reclaimPolicy
+  default: "Delete"
+  description: "Define reclaim policy (Retain or Delete)"
```
meldafrawi, line 419: I think it should be better for StorageClass reclaim policy to be an option with Retain or Delete instead of a string.

yasker:
Why this is PR against master? We have v1.1.0 branch.

shuo-wu:
It seems that branch v1.1.0 doesn't hold the latest commit then there are some conflicts.

meldafrawi:
@shuo-wu should we change Longhorn image tags to v1.1.0 ?  [Here](https://github.com/shuo-wu/longhorn/blob/8527d310eda2ef9bc6d26350db1606701c0c6c72/chart/values.yaml#L12)
Also modify Chart version [Here](https://github.com/shuo-wu/longhorn/blob/chart/chart/Chart.yaml#L3)

Since PR is created against v1.1.0 not the master.

yasker:
@shuo-wu v1.1.0 is newer for chart since it contains https://github.com/longhorn/longhorn/commit/9c3ae72954c02c8e9dc1396e1efb04d42ac1e346

chart/templates/storageclass.yaml
```diff
@@ -12,7 +12,7 @@ data:
       name: longhorn
       annotations:
         storageclass.kubernetes.io/is-default-class: {{ .Values.persistence.defaultClass | quote }}
-      labels: {{- include "longhorn.labels" . | nindent 4 }}
+      labels: {{- include "longhorn.labels" . | nindent 8 }}
```
PhanLe1010, line 15: Nice catch

PhanLe1010: Actually, I think that we shouldn't put `longhorn.labels` into the actual StorageClass since `longhorn.labels` generates the standard Helm labels. However, we don't want Helm to manage the actual StorageClass resource.

PhanLe1010 (commented):
LGTM except for a comment

chart/templates/storageclass.yaml
shuo-wu: Agree.

PhanLe1010 (approved):

Issue #1986, opened 2020-11-16
[BUG] Post-upgrade job uses "always" image pull policy


gitlawr:
**Describe the bug**
I'm running longhorn in an offline environment without a private registry. All images are imported to the nodes. When I try to upgrade longhorn via helm, the post-upgrade pod stucks in `ContainerCreating`, given that the imagePullPolicy is `Always`.

**To Reproduce**
1. Import offline images to the nodes.
2. Install longhorn
3. helm upgrade the longhorn release
4. wait and check pods' status.

**Expected behavior**
Be able to upgrade successfully.
I would expect that the `image.pullPolicy` value applies to the job as well.

**Log**
N/A

**Environment:**
 - Longhorn version: v1.0.2
 - Kubernetes version: v1.18.9+k3s1
 - Node OS type and version: k3OS v0.11.1

PhanLe1010:
https://github.com/longhorn/longhorn/pull/2007 should affect/fix this issue.

PhanLe1010:
**Reproduce steps:**
1. Using Rancher, create a cluster of 1 node (with controller, etcd, worker roles) on Amazon EC2 provider. It is easier to control the network for a one-node cluster. 
1. Wait until the cluster is created, SSH into the EC2. 
1. Follow the document at https://longhorn.io/docs/1.0.2/advanced-resources/deploy/airgap/ to download all images of Longhorn v1.0.1, Longhorn v1.0.2, Longhorn master to the EC2
1. Log in to AWS console, find the EC2, open the security group setting for the EC2. We disconnect the EC2 from the internet by setting the  the security group as below:
   <img width="1622" alt="Screen Shot 2020-11-24 at 4 49 21 PM" src="https://user-images.githubusercontent.com/22139961/100167927-10a17480-2e75-11eb-98c1-ba496923bedb.png">
   Make sure to open the appropriate ports so that Rancher and your computer can communicate with the EC2
1. From Rancher, Install Longhorn v1.0.1
1. Create a volume, write some data, attached it to the node, verify everything is ok.
1. From Rancher, Upgrade Longhorn to v1.0.2. Verify that the `Post-upgrade` job is stuck since its `imagePullPolicy=Always` and it cannot pull the image.

PhanLe1010:
@gitlawr Thank you for reporting.

We will fix this bug in the coming v1.1.0 release. When v1.1.0 is out, you can download v1.0.1 images to the cluster and update Longhorn.

[...]

Issue #1670, opened 2020-08-07
[BUG]The image cannot be pulled from the private image repository due to lack of secret if not customize setting


xiaoluhong:
**Describe the bug**
Even if the privateRegistry parameter is configured correctly, the image cannot be pulled from the private image repository when using Default Settings

**To Reproduce**
1. Set Longhorn's image repo to private mode(You need to be logged in to pull the image)
2. Run the Longhorn Catalog in the Rancher UI and configure the `PRIVATE REGISTRY URL, username, and password` in the PRIVATE REGISTRY SETTINGS
3. Set Customize Default Settings to False
4. Leave the default values for the other parameters
5. need to delete the Longhorn image already on the host （`docker images | grep longhornio | awk '{print $3}' | xargs docker rmi -f`）
6. Finally, click Launch

**results**

1. Unable to pull the image
![image](https://user-images.githubusercontent.com/20707603/89676205-dc359700-d91d-11ea-86c0-d66ef0fb1b2b.png)
2. ImagePullSecrets are not in the Pod YAML file

```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    cni.projectcalico.org/podIP: 10.60.0.75/32
    cni.projectcalico.org/podIPs: 10.60.0.75/32
  creationTimestamp: "2020-08-07T18:18:40Z"
  generateName: longhorn-manager-
  labels:
    app: longhorn-manager
    app.kubernetes.io/instance: longhorn-system
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: longhorn
    app.kubernetes.io/version: v1.0.1
    controller-revision-hash: ff5464f98
    helm.sh/chart: longhorn-1.0.1
    io.cattle.field/appId: longhorn-system
    pod-template-generation: "1"
  managedFields:
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:generateName: {}
        f:labels:
          .: {}
          f:app: {}
          f:app.kubernetes.io/instance: {}
          f:app.kubernetes.io/managed-by: {}
          f:app.kubernetes.io/name: {}
          f:app.kubernetes.io/version: {}
          f:controller-revision-hash: {}
          f:helm.sh/chart: {}
          f:io.cattle.field/appId: {}
          f:pod-template-generation: {}
        f:ownerReferences:
          .: {}
          k:{"uid":"66a30ba2-c59e-4a3e-8a75-bb682f00d46d"}:
[...]

PR #2007, opened 2020-11-21
change registry secret location

PhanLe1010:
To avoid confusing user, we move the registry secret setting outside of default setting section. Now, user will set registry secret setting in `privateRegistry.registrySecret`

Also, we change `imagesPullPolicy` of `longhorn-post-upgrade` job and `longhorn-uninstall` job to `IfNotPresent` so that user can upgrade/uninstall Longhorn using Helm in airgap environment. 

longhorn/longhorn#1670
longhorn/longhorn#1986

### DIFF
--- a/chart/questions.yml
+++ b/chart/questions.yml
@@ -130,6 +130,12 @@ questions:
   group: "Private Registry Settings"
   type: password
   default: ""
+- variable: privateRegistry.registrySecret
+  label: Private registry secret name
+  description: "Longhorn will automatically generate a Kubernetes secret with this name and use it to pull images from your private registry."
+  group: "Private Registry Settings"
+  type: string
+  default: ""
 - variable: longhorn.default_setting
   default: "false"
   description: "Customize the default settings before installing Longhorn for the first time. This option will only work if the cluster hasn't installed Longhorn."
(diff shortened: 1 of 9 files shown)


chart/questions.yml
```diff
@@ -130,6 +130,12 @@ questions:
   group: "Private Registry Settings"
   type: password
   default: ""
+- variable: privateRegistry.registrySecret
+  label: Private registry secret name
+  description: "The Kubernetes Secret name"
```
yasker, line 135: If we will generate it automatically, we need to call it out.

chart/templates/postupgrade-job.yaml
```diff
@@ -18,7 +18,7 @@ spec:
       containers:
       - name: longhorn-post-upgrade
         image: {{ template "registry_url" . }}{{ .Values.image.longhorn.manager.repository }}:{{ .Values.image.longhorn.manager.tag }}
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
```
yasker, line 21: I remember we have related issues regarding the imagepullpolicy of the uninstaller? Should refer it.

yasker (changes requested):

chart/templates/postupgrade-job.yaml
PhanLe1010: @yasker I couldn't find the related issue. Do you have a link?

yasker: https://github.com/longhorn/longhorn/issues/1986

PhanLe1010: Thank you. I was searching for `uninstall` keyword so I couldn't find the issue.

PhanLe1010: Linked

chart/questions.yml
PhanLe1010: Thank you. Fixed

yasker (approved):

Issue #2031, opened 2020-12-03
[FEATURE] allow specify namespace when install longhorn as sub-chart


guangbochen:
**Is your feature request related to a problem? Please describe.**
When using the longhorn as a sub-chart the release namespace is not always going to be the `longhorn-system`, we should allow the user to specify override namespace in this scenario.

**Describe the solution you'd like**
Add `namespaceOverride` value that allows the user to override the `Release.Namespace` when it is not using the `longhorn-system`

meldafrawi:
Validation: **PASSED**

PR #2032, opened 2020-12-03
Add override namespace value

guangbochen:
**Description:**
When using the longhorn as a sub-chart the release namespace is not always going to be the longhorn-system, we should allow the user to specify override namespace in this scenario.

**Solution:**
Add `namespaceOverride` value that allows the user to override the Release.Namespace when it is not using the `longhorn-system`

**related issue:**
https://github.com/longhorn/longhorn/issues/2031


meldafrawi:
LGTM

meldafrawi (approved):

Issue #1741, opened 2020-08-27
Improve documentation for installing open-iscsi


dbpolito:
I was trying to install Longhorn on EKS and faced the problem with missing open-iscsi, i see it's documented here: https://longhorn.io/docs/1.0.2/deploy/install/#installing-open-iscsi

But for EKS how to ssh into the node, deal with autoscaling, etc.

I was searching a better way for handling this and found out that the approach on openebs is pretty straight forward:

https://openebs.github.io/charts/openebs-ubuntu-setup.yaml
https://openebs.github.io/charts/openebs-amazonlinux-setup.yaml

I think it would be very useful having a more friendly way of installing this on EKS for new comers like me.

Installing on DOKS was smooth.

Thank you for your hard work ❤️

yasker:
Thanks @dbpolito , that's a good idea. We will see if we can include something similar in our next release.

dbpolito:
Just in case someone need, i deployed this and everything is working now:

```yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: longhorn-amazonlinux-init
  labels:
    app: longhorn-amazonlinux-init
  annotations:
    command: &cmd yum install iscsi-initiator-utils -y && sudo systemctl enable iscsid && sudo systemctl start iscsid
spec:
  selector:
    matchLabels:
      app: longhorn-amazonlinux-init
  template:
    metadata:
      labels:
        app: longhorn-amazonlinux-init
    spec:
      hostNetwork: true
      initContainers:
      - name: init-node
        command:
          - nsenter
          - --mount=/proc/1/ns/mnt
          - --
          - sh
          - -c
          - *cmd
        image: alpine:3.7
        securityContext:
          privileged: true
      hostPID: true
      containers:
      - name: wait
        image: k8s.gcr.io/pause:3.1
      hostPID: true
      hostNetwork: true
      tolerations:
      - effect: NoSchedule
        key: node-role.kubernetes.io/master
  updateStrategy:
    type: RollingUpdate
```

dbpolito:
I'm glad to see this going,  good approach you guys found making use of current daemonset and not increase number of pod on cluster... 👍 

Are there plans to add into the `chart` as an option? https://github.com/longhorn/charts/blob/master/charts/longhorn/templates/daemonset-sa.yaml

If not planned are you guys open for a PR, i can definitely contribute to that...

boknowswiki:
[...]

PR #2036, opened 2020-12-04
Fix crash loop error in longhorn-iscsi-installation

PhanLe1010:
We install iscsi in the init container then sleep in the main container. This avoids crash loop after finishing installing iscsi

longhorn/longhorn#1741

### DIFF
--- a/deploy/iscsi/longhorn-iscsi-installation.yaml
+++ b/deploy/iscsi/longhorn-iscsi-installation.yaml
@@ -16,7 +16,8 @@ spec:
         app: longhorn-iscsi-installation
     spec:
       hostNetwork: true
-      containers:
+      hostPID: true
+      initContainers:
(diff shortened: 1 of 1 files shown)


deploy/iscsi/longhorn-iscsi-installation.yaml
```diff
@@ -16,7 +16,8 @@ spec:
         app: longhorn-iscsi-installation
     spec:
       hostNetwork: true
-      containers:
+      hostPID: true
```
yasker, line 19: Why do we need host PID namespace? Thought the previous one works fine.

yasker: Oh, it was moved only.

yasker (approved):

yasker (approved):

Issue #2061, opened 2020-12-08
task: Longhorn v1.1.0 release chart


yasker:
Need to update Rancher charts for Longhorn v1.1.0 release.

shuo-wu:
The test Chart for Cluster Manager: `https://github.com/shuo-wu/charts/` branch `master`
The test Chart for Cluster Explorer: `https://github.com/shuo-wu/charts/` branch `dev-v2.5`

meldafrawi:
Validated Using the Provided charts: **LGTM**

PR #2065, opened 2020-12-09
chart: Update Kubernetes version requirement for chart

shuo-wu:
Since we update the CRD apiVersion to apiextensions.k8s.io/v1, the
minimal Kubernetes version requirement is v1.16 now.

Longhorn #2061

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 name: longhorn
 version: 1.1.0-rc1
 appVersion: v1.1.0-rc1
-kubeVersion: ">=v1.14.0-r0"
+kubeVersion: ">=v1.16.0-r0"
 description: Longhorn is a distributed block storage system for Kubernetes.
 keywords:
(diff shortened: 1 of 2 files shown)


chart/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 name: longhorn
 version: 1.1.0-rc1
 appVersion: v1.1.0-rc1
-kubeVersion: ">=v1.14.0-r0"
+kubeVersion: ">=v1.16.0"
```
yasker, line 5: follow the same pattern as before, otherwise, anything with a different suffix will be blocked.

shuo-wu: Updated. BTW, what does this suffix `r0` mean?

yasker: You added it.. https://github.com/longhorn/longhorn/commit/cac53ad3ba980a6136c6e919a6ba83dd7d17952d 

I remembered it's because of suffix limitation.

Issue #2006, opened 2020-11-21
[FEATURE] Enable backing image feature in Longhorn


yasker:
Longhorn Engine has the ability to use a qcow2 file as the backing image for the Longhorn volume. It was designed for VM but the feature was never implemented in Longhorn manager nor UI. We need to enable this feature for Harvester to use Longhorn with existing root images.

Btw, the field was called `base image` but it's better to rename everything into `backing image` because `base image` has become a standard Docker term.

Ref:
https://github.com/longhorn/longhorn/issues/1989#issuecomment-731458474
https://github.com/rancher/harvester/issues/227

yasker:
Move to Icebox. Pending merge of https://github.com/longhorn/longhorn/issues/2295

innobead:
The new version of backing image implementation is actually be based on #2295 . To avoid confusing QA for these similar tickets, close this and move focus on #2295.

PR #2133, opened 2020-12-23
enhancement: Add a new enhancement 'enable backing image'

shuo-wu:
#2006

### DIFF
--- a/enhancements/20201228-enable-backing-image.md
+++ b/enhancements/20201228-enable-backing-image.md
@@ -1,0 +1,188 @@
+# Enable Backing Image
+
+## Summary
+Longhorn can set a backing image of a Longhorn volume, which is designed for VM usage.
+
+### Related Issues
+https://github.com/longhorn/longhorn/issues/2006
+
+## Motivation
+### Goals
+1. A qcow2 or raw image file can be used as the backing image of a volume.
+2. The backing image works fine with backup or restore.  
+3. Multiple replicas in the same disk can share one backing image.   
+
+## Proposal
+1. For `longhorn-manager`:
+    1. The backing image of a Longhorn volume should be downloaded by someone before starting the related volume replicas.
+        - Instance manager pods are only responsible for managing process, which implicitly means they are isolated from the data layer. Hence, Longhorn should not rely on instance manager to fetch backing images.  
+        - Since a backing image can be shared by multiple replicas, using replica processes to download backing images is not a good idea. Otherwise, a backing image may be downloaded duplicately by multiple replica processes.
+        - As a result, the only choice here is longhorn manager.
+    2. Since a backing image is a kind of resource, it's better to use a new CRD to abstract it.
+        - Similar to the CRD `engineimages.longhorn.io`, there is one CR for each backing image.
+        - Longhorn can present this kind of resource with a separate UI page so that users are capable of operating the CRs manually. Then when users want to create a volume with a backing image, they need to choose one from the existing backing image list.
+        - Since some disks won't be chosen by replicas with backing images, there should be a disk map in the CR spec to indicate in which disk/node a backing image should be downloaded.
+        - Considering the disk migration and failed replica reusage features, there will be an actual backing image file for each disk rather than each node.
+        - There will be a new controller to launch pods to download backing images based on the disk maps, as well as handling the backing image download status for each disk.
+    3. Before sending requests to launch replica progresses, replica controller will check and wait for the backing image ready if a backing image is set for the related volumes.
+        - This is a common logic that is available for not only normal volumes but also restoring/DR volumes. 
+        - We need to make sure the backing image names as well as the URLs are stored in backup volumes. So users are able to re-specify the backing image when restoring a volume in case of the original image becoming invalid.
+    4. There should be a cleanup timeout setting and a related timestamp that indicates when a backing image can be removed from a disk when no replica in the disk is using the backing image.  
+2. For `longhorn-engine`:
+    - Most of the backing image related logic is already in there.
+    - The raw image support will be introduced.
+    - Make sure the backing file path will be updated each time when the replica starts.
+
+### User Stories
+#### Rebuild replica for a large volume after network fluctuation/node reboot
+Before the enhancement, users need to manually copy the backing image data to the volume in advance.
+
+After the enhancement, users can directly specify the backing image during volume creation/restore with a click.
+
+### User Experience In Detail
+1. Users can modify the backing image cleanup timeout setting so that all non-used backing images will be cleaned up automatically from disks.
+2. Create a volume with a backing image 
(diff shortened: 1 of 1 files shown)


enhancements/20201223-enable-backing-file.md
```diff
@@ -0,0 +1,98 @@
...
+2. The backing file works fine with backup or restore.  
+3. Multiple replicas on the same disk can share a backing file if the backing file URLs are identical.   
+
+## Proposal
+1. The backing file of a Longhorn volume should be downloaded by someone before starting the related volume replicas.
+    - Instance manager pods are only responsible for managing process, which implicitly means they are isolated from the data layer. Hence Longhorn should not rely on instance manager to fetch backing files.  
+    - Since a backing file can be shared by multiple replicas, using replica processes to download replicas is not a good idea. Otherwise, a backing file may be duplicately downloaded by multiple replica processes.
+    - As a result, the only choice here is longhorn manager.
+2. Since a backing file is downloaded separately by longhorn manager and it may be shared by multiple replicas, it's better to use a new CRD to abstract it.
+    - Considering the disk migration and failed replica reusage features, for each backing file, the related CR is per disk rather than per node. In brief, there will be a CR for each backing file URL in each disk.
```
yasker, line 21: One CR for each file on each disk? There can be tons of CRs then. How about one CR per image? Where the image should be downloaded to can be set in the spec of the CR as a map. The status of the CR can contain the status e.g. `downloaded`/`failed`/`in progress`.

enhancements/20201223-enable-backing-file.md
```diff
@@ -0,0 +1,98 @@
...
+    1. The related backing file CR status. (reference count, timestamp)
+    2. The only backing file is not removed when the cleanup timeout is not passed.
+5. Set `BackingFileCleanupTimeout` to a small value. Then verify:
+    1. The only backing file is removed when the cleanup timeout is passed.
+    2. Other backing files keep unchanges. And all volumes work fine.
+6. Create a backup for a volume.
+7. Wait for the backup creation then restore a new volume from the backup. Verify:
+    1. The backing file is not included in the backup.
+    2. The data content in the restore volume is correct. 
+8. Clean up all volumes.
```
yasker, line 86: Also try deleting the backing file before restore. Make sure the restore process downloads the backing file again.

enhancements/20201223-enable-backing-file.md
```diff
@@ -0,0 +1,98 @@
...
+
+After the enhancement, users can create/restore a Longhorn volume with a backing file.
+
+### User Experience In Detail
+1. Users can set a cleanup timeout for no-used backing files so that they will be cleaned up automatically.
+2. Users create/restore a volume with backing file URL set.
+3. When backing up a volume with a backing file, the backing file content won't be upload to the backstore. Instead, the backing file will be re-downloaded from the original URL once it's required.
+
+### API Changes
+No API change is required.
```
yasker, line 41: Can user remove backing file directly? We can provide a way to do that before the file reaches `CleanupTimeout`. Since those backing files are likely taking spaces, and it become invisible to the user in the current design. We do need a way to manage this resource.

Also, I am thinking instead of asking for URL everytime when a volume is created, can we ask user to `create/import` backing file, then choose the backing file during the volume creation. Though it will add additional burden to users if they want to restore a volume with backing file - or we can automatically add the backing file from specified URL to the system.

yasker (changes requested):

yasker:
Also, can we change the feature name to `backing image`? it sounds better than `backing file`.

enhancements/20201228-enable-backing-image.md
```diff
@@ -0,0 +1,115 @@
...
+## Design
+### Implementation Overview
+#### longhorn-manager:
+1. Add a setting `BackingImageCleanupTimeout`.
+2. Add a new CRD `backingimages.longhorn.io`.
+    - Field `Spec.Disks` is designed for recording the disks that the backing images need to be downloaded to. And the map is empty when a backing image CR is just created.
+    - A disk/corresponding download progress map should be in `status`. So that a replica can directly find and verify the corresponding backing image before launching the replica process. If the status is not `downloaded`, the replica controller needs to wait for the state before launching the replica process. 
+3. Add a new controller `Backing Image Controller`.
+    - By comparing field `Spec.Disks` with field `Status.DiskStatusMap`, the controller knows if a pod for downloading the backing image should be created.  
+    - Field `BackingImageDownloadState` is the key of map `Status.DiskStatusMap`. It can be `downloaded`/`downloading`/`failed`/`terminating`.
```
joshimoo, line 58: should this be value? key I am assuming is the disk/disk.name

enhancements/20201228-enable-backing-image.md
```diff
@@ -0,0 +1,115 @@
...
+    - Once a pod successfully downloads the file, a file that is used as the download success flag will be created, and the pod readiness probe will rely on the file existence to determine if the pod is available. 
+4. In Replica Controller:
+    - Check and wait for backing image disk map in the status before sending requests to replica instance managers.
+5. In Node Controller:
+    - Set a timestamp in backing image`Spec.DiskLastRefAtMap` when the backing image in the disk is used by no replica. Once the timeout `BackingImageCleanupTimeout` is passed, the backing image in the disk should be deleted.
```
joshimoo, line 63: Do we unset this field if after a while the backing image is used again?

shuo-wu: Yeah. The field will be unset in the node controller when there is a replica reusing the backing image in the disk.

enhancements/20201228-enable-backing-image.md
shuo-wu: Yeah. This is a typo. The key of this map is the disk UUID

enhancements/20201228-enable-backing-image.md
```diff
@@ -0,0 +1,184 @@
...
+#### longhorn-ui:
+1. Launch a new page to present and operate backing images.
+    1. Show the image (download) status for each disk based on `Status.DiskStatusMap` when users expand one backing image.
+    2. Add the following operating list:
+        - `Create`: The required field is `image`.
+        - `Delete`: No field is required. It should be disabled when there is one replica using the backing image.
+        - `CleanupDiskImages`: This allows users to manually clean up the images in some disks in advance. It's a batch operation.
+2. Allow choosing a backing image for volume creation.
+3. Allow choosing/re-specifying a new backing image for restore/DR volume creation.
+
```
yasker, line 92: Have we implemented automatic image loading from CSI? If so, we can mention it here.

yasker (commented):
LGTM in general. @joshimoo can you take a look as well?

joshimoo (approved):
LGTM, added some typos can be ignored.
Go ahead and merge this no need for me to rereview.

Issue #2152, opened 2021-01-05
[FEATURE] Upgrade longhorn engine automatically


gitlawr:
**Is your feature request related to a problem? Please describe.**
Currently, users need to upgrade the engine image manually in the UI after the longhorn manager is updated.

**Describe the solution you'd like**
Be able to upgrade engines without UI interactions. For example, provide an option. When it is enabled, upgrade engine images once the longhorn manager is updated.

**Describe alternatives you've considered**
Call the API to upgrade engines for automation.

PhanLe1010:
**Manual test plan:**
1. Install Longhorn using `longhorn-manager:master`
1. Turn on the setting `Automatically upgrade engine for volumes after upgrading Longhorn`
1. Create a volume attached to a pod write some data. Create a DR volume. Create volume and leave it in detached state.
1. For each test scenario below, try to upgrade Longhorn with longhorn engine image from `From` column to `To` column

   Test scenarios:

   |  From | To  | Expected  |  
   |---|---|---|
   | lower valid lower semantic version (`longhorn-engine/v1.0.2`)|  higher valid semantic version (`longhorn-engine/v1.1.0`)| Auto upgrade  |
   | higher valid semantic version (`longhorn-engine/v1.1.0`)  | lower valid lower semantic version (`longhorn-engine/v1.0.2`)  | Don't upgrade  |
   | valid semantic version (`longhorn-engine/v1.0.2`)  | invalid semantic version (`longhorn-engine/dev-01`) | Don't upgrade  |
   | invalid semantic version (`longhorn-engine/dev-01`) | valid semantic version (`longhorn-engine/v1.1.0`)| Don't upgrade |
   | invalid semantic version (`longhorn-engine/dev-01`) |  invalid semantic version (`longhorn-engine/dev-02`)| Don't upgrade |
   | valid semantic version (`longhorn-engine/v1.1.0`) | master (`longhorn-engine/master`) | Auto upgrade |

// Test concurrent upgrade

// Test deploy new image using Longhorn UI

// Test DR volume with healthy/unhealthy state

// Test expanding volume

yasker:
Remember to update the chart too. The automation bot has been updated to contain the questionnaire regarding the chart and yaml update.

PhanLe1010:
Updated the chart at https://github.com/longhorn/longhorn/pull/2220

anupama2501:
Test scenarios verified:

[...]

Issue #1544, opened 2020-07-02
[FEATURE]Doc for API version mechanism


shuo-wu:
**Is your feature request related to a problem? Please describe.**
It's better to have a doc that indicates if the modification breaks the backward compatibility when the gRPC definition is updated, and what we need to do after the modification.

**Describe the solution you'd like**
A doc indicates that:
1. When the backward compatibility is broken.
2. How to update the API version number.
3. What we need to check and what we need to do if the compatibility is/isn't broken.
4. Provide some conventions/suggestions for the gRPC modification, which can help avoid the compatibility break in the feature.

PR #2172, opened 2021-01-11
Add LEP for feature automatically upgrade engine for volumes

PhanLe1010:
longhorn/longhorn#2152

### DIFF
--- a/enhancements/20210111-upgrade-engine-automatically.md
+++ b/enhancements/20210111-upgrade-engine-automatically.md
@@ -1,0 +1,173 @@
+# Upgrade Volume's Engine Image Automatically
+
+## Summary
+
+Currently, users need to upgrade the engine image manually in the UI after upgrading Longhorn.
+We should provide an option. When it is enabled, automatically upgrade engine images for volumes when it is applicable.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/2152
+
+## Motivation
+
+### Goals
+
+Reduce the amount of manual work users have to do when upgrading Longhorn by automatically upgrade engine images 
+for volumes when they are ok to upgrade. E.g. when we can do engine live upgrade or when volume is in detaching state.
+
+## Proposal
+
+Add a new boolean setting, `Concurrent Automatic Engine Upgrade Per Node Limit`, so users can control how Longhorn upgrade engines.
+The value of this setting specifies the maximum number of engines per node that are allowed to upgrade to the default engine image at the same time.
+If the value is 0, Longhorn will not automatically upgrade volumes' engines to default version.
+
+After user upgrading Longhorn to a new version, there will be a new default engine image and possibly a new default instance manager image.
+
+The proposal has 2 parts:
+1. In which component we will do the upgrade.
+2. Identify when is it ok to upgrade engine for volume to the new default engine image.
+
+For part 1, we will do the upgrade inside `engine image controller`. 
+The controller will constantly watch and upgrade the engine images for volumes when it is ok to upgrade.
+
+For part 2, we upgrade engine image for a volume when the following conditions are met:
+1. The new default engine image is ready
+1. Volume is not upgrading engine image
+1. The volume condition is one of the following:
+   1. Volume is in detached state.
+   1. Volume is in attached state (live upgrade).
+      And volume is healthy.
+      And The current volume's engine image is compatible with the new default engine image.
+      And If volume is not a DR volume.
+      And Volume is not expanding.
+
+### User Stories
+
+Before this enhancement, users have to manually upgrade engine images for volume after upgrading Longhorn system to a newer version.
+If there are thoudsands of volumes in the system, this is a significant manual work.
+
+After this enhancement users either have to do nothing (in case live upgrade is possible) 
+or they only have to scale down/up the workload (in case there is a new default IM image)
+
+### User Experience In Detail
+
+1. User upgrade Longhorn to a newer version. 
+The new Longhorn version is compatible with the volume's current engine image. 
+Longhorn automatically do live engine image upgrade for volumes
+
+2. User upgrade Longhorn to a newer version. 
+The new Longhorn version is not compatible with the volume's current engine image. 
+Users only have to scale the workload down and up.
+This experience is similar to restart Google Chrome to use a new version. 
+
+3. Note that users need to disable this feature if they want to update the engine image to a specific version for the volumes.
+If `Concurrent Automatic Engine Upgrade Per Node Limit` setting is bigger than 0, Longhorn will not allow user to manually upgrade engine to a version other than the default version.
+
+### API changes
+
+No API change is needed.
+
+## Design
+
+### Implementation Overview
+
+1. Inside `engine image controller` sync function, get the value of the setting `Concurrent Automatic Engine Upgrade Per Node Limit` and assign it to concurrentAutomaticEngineUpgradePerNodeLimit variable.
+  If concurrentAutomaticEngineUpgradePerNodeLimit <= 0, we skip upgrading.
+1. Find the new default engine image. Check if the new default engine image is ready. If it is not we skip the upgrade.
+
+1. List all volumes in Longhorn system. 
+   Select a set of volume candidate for upgrading.
+   We select candidates that has the condition is one of the following case:
+   1. Volume is in detached state.
+   1. Volume is in attached state (live upgrade).
+      And volume is healthy.
+      And Volume is not upgrading engine image.
+      And The current volume's engine image is compatible with the new default engine image.
+      And the volume is not a DR volume.
+      And volume is not expanding.
+     
+1. Make sure not to upgrade too many volumes on the same node at the same time.
+   Filter the upgrading candidate set so that total number of upgrading volumes and candidates per node is not over `concurrentAutomaticEngineUpgradePerNodeLimit`.
+1. For each volume candidate, set `v.Spec.EngineImage = new default engine image` to update the engine for the volume.
+1. If the engine upgrade failed to complete (e.g. the v.Spec.EngineImage != v.Status.CurrentImage), 
+   we just consider it is the same as volume is in upgrading process and skip it.
+   Volume controller will handle the reconciliation when it is possible.
+
+### Test plan
+
+
+Integration test plan.
+
+Preparation:
+1. set up a backup store
+2. Deploy a compatible new engine image
+
+Case 1: Concurrent engine upgrade
+1. Create 10 volumes each of 1Gb.
+2. Attach 5 volumes vol-0 to vol-4. Write data to it
+3. Upgrade all volumes to the new engine image
+4. Wait until the upgrades are completed (volumes' engine image changed,
+   replicas' mode change to RW for attached volumes, reference count of the
+   new engine image changed, all engine and replicas' engine image changed)
+5. Set concurrent-automatic-engine-upgrade-per-node-limit setting to 3
+6. In a retry loop, verify that the number of volumes who
+   is upgrading engine is always smaller or equal to 3
+7. Wait until the upgrades are completed (volumes' engine image changed,
+   replica mode change to RW for attached volumes, reference count of the
+   new engine image changed, all engine and replicas' engine image changed,
+   etc ...)
+8. verify the volumes' data
+
+Case 2: Dr volume
+1. Create a backup for vol-0. Create a DR volume from the backup
+2. Try to upgrade the DR volume engine's image to the new engine image
+3. Verify that the Longhorn API returns error. Upgrade fails.
+4. Set concurrent-automatic-engine-upgrade-per-node-limit setting to 0
+5. Try to upgrade the DR volume engine's image to the new engine image
+6. Wait until the upgrade are completed (volumes' engine image changed,
+   replicas' mode change to RW, reference count of the new engine image
+   changed, engine and replicas' engine image changed)
+7. Wait for the DR volume to finish restoring
+8. Set concurrent-automatic-engine-upgrade-per-node-limit setting to 3
+9. In a 2-min retry loop, verify that Longhorn doesn't automatically
+   upgrade engine image for DR volume.
+
+Case 3: Expanding volume
+1. set concurrent-automatic-engine-upgrade-per-node-limit setting to 0
+2. Upgrade vol-0 to the new engine image
+3. Wait until the upgrade are completed (volumes' engine image changed,
+   replicas' mode change to RW, reference count of the new engine image
+   changed, engine and replicas' engine image changed)
+4. Detach vol-0
+5. Expand the vol-0 from 1Gb to 5GB
+6. Wait for the vol-0 to start expanding
+7. Set concurrent-automatic-engine-upgrade-per-node-limit setting to 3
+8. While vol-0 is expanding, verify that its engine is not upgraded to
+   the default engine image
+9. Wait for the expansion to finish and vol-0 is detached
+10. Verify that Longhorn upgrades vol-0's engine to the default version
+
+Case 4: Degraded volume
+1. set concurrent-automatic-engine-upgrade-per-node-limit setting to 0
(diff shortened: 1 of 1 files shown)


PhanLe1010:
Please help to review the logic.
I will correct spelling and add test case later.

enhancements/20210111-upgrade-engine-automatically.md
```diff
@@ -0,0 +1,114 @@
...
+engineImageTag:
+  - compatibleIMImageTags: # List of compatible IM image tags for the engineImageTag
+  - upgradableFromEngineIMTags: # The engineImageTag can be upgraded from a engine IM in this list of engine IM tags
+```
+
+For example:
+```
+"v1.1.0": 
+  - compatibleIMImageTags: ["v1_20201216"]
+  - upgradableFromEngineIMTags: ["v1.0.2", "v1.0.1", "v1.0.0", "v0.8.1", "v0.8.0"]
```
PhanLe1010, line 79: It seems that hard coding is not a good approach because we will have to modify/build manager code for every Longhorn release. I am thinking about using semantic version comparison to make sure that we only upgrade to newer engine image version and never downgrade.

enhancements/20210111-upgrade-engine-automatically.md
```diff
@@ -0,0 +1,114 @@
...
+The controller will constantly watch and upgrade the engine images for volumes when it is ok to upgrade.
+
+For part 2, we upgrade engine image for a volume when the following conditions are met:
+1. The new default engine image is ready
+1. Volume is not upgrading engine image
+1. The volume condition is one of the following:
+   1. Volume is in detached state.
+   1. Volume is in attached state (live upgrade).
+      The current volume's instance manager image is the same as the new default instance manager image.
+   1. If volume is a DR volume, the volume's robustness must be `healthy`
```
shuo-wu, line 39: Notice that Longhorn will directly apply a full restore after upgrading a DR volume. I am not sure if we need to do the automatic live upgrade for DR/restore volumes.

Besides, we need to disable this for expanding volumes.

enhancements/20210111-upgrade-engine-automatically.md
shuo-wu: Why do we need these 2 lists? 
For detached volumes, we can directly do the offline upgrade without compatibility issues.
For attached volumes, if the latest engine image is compatible with the current instance manager an old volume is running on, Longhorn will directly do a live upgrade for it. And we can determine the compatibility by checking the version numbers in `InstanceManagerStatus` and `EngineImageStatus`.

shuo-wu (changes requested):

enhancements/20210111-upgrade-engine-automatically.md
PhanLe1010: Thank you! That makes sense. I was planned to use the list to prevent downgrading from a higher version to lower version incase something went wrong and user rollback to the previous deployment. However, hard coding is not a good idea. I will use [semantic version comparison](https://github.com/Masterminds/semver) instead

enhancements/20210111-upgrade-engine-automatically.md
```diff
@@ -0,0 +1,114 @@
...
+## Motivation
+
+### Goals
+
+Reduce the amount of manual work users have to do when upgrading Longhorn by automatically upgrade engine images 
+for volumes when they are ok to upgrade. E.g. when we can do engine live upgrade or when volume is in detaching state.
+
+## Proposal
+
+Add a new boolean setting, `Automatically Upgrade Engine When Possible`, so users can turn on/off this feature.
```
yasker, line 21: `Automatically Upgrade Engine to Default Version`

enhancements/20210111-upgrade-engine-automatically.md
```diff
@@ -0,0 +1,114 @@
...
+
+### User Experience In Detail
+
+1. User upgrade Longhorn to a newer version. 
+The new Longhorn version comes with the same default IM image. 
+Longhorn automatically do live engine image upgrade for volumes
+
+2. User upgrade Longhorn to a newer version. 
+The new Longhorn version comes with the new default IM image. 
+Users only have to scale the workload down and up.
```
yasker, line 57: User shouldn't need to scale workload down and up. We will set the spec and it should be upgraded automatically once it's detached.

yasker (changes requested):

enhancements/20210111-upgrade-engine-automatically.md
PhanLe1010: Thank you! 

I meant the same. Longhorn will retry upgrade engine when volume is detached (i.e. scale the workload down and up.)

We cannot set the spec because it will upgrade engine immediately.

enhancements/20210111-upgrade-engine-automatically.md
PhanLe1010: Thank you! Fixed

enhancements/20210111-upgrade-engine-automatically.md
PhanLe1010: > Besides, we need to disable this for expanding volumes.

Thank you. Fixed

> Notice that Longhorn will directly apply a full restore after upgrading a DR volume. I am not sure if we need to do the automatic live upgrade for DR/restore volumes.

I think we should keep every thing up to date

enhancements/20210111-upgrade-engine-automatically.md
```diff
@@ -0,0 +1,103 @@
...
+For part 1, we will do the upgrade inside `engine image controller`. 
+The controller will constantly watch and upgrade the engine images for volumes when it is ok to upgrade.
+
+For part 2, we upgrade engine image for a volume when the following conditions are met:
+1. The new default engine image is ready
+1. Volume is not upgrading engine image
+1. The volume condition is one of the following:
+   1. Volume is in detached state.
+   1. Volume is in attached state (live upgrade).
+      The current volume's instance manager image is the same as the new default instance manager image.
```
yasker, line 38: Does this mean that automatic live upgrade won't work if the instance manage image is upgraded as well? Not sure why it's needed.

enhancements/20210111-upgrade-engine-automatically.md
```diff
@@ -0,0 +1,103 @@
...
+   1. Volume is in detached state.
+      And the current volume's engine image tag is smaller than default engine image tag in term of semantic version comparison.
+   1. Volume is in attached state (live upgrade).
+      And the current volume's IM image tag is smaller than default engine image tag in term of semantic version comparison.
+      And The current volume's engine IM is compatible with the new default engine image.
+      And the volume's robustness must be `healthy` if volume is a DR volume.
+      And volume is not expanding
+      
+1. If the above checks pass, set `v.Spec.EngineImage = new default engine image` to update the engine for the volume.
+1. We will load balance the upgrading. Make sure not to upgrade too many volumes on the same node at the same time.
```
yasker, line 86: It's not called load balance. 

Also, need more details on how will we throttle the automated upgrade process (e.g. is there a customizable parameter or it's hardcoded. Sound like the parameter can be set to 0 to disable the automated engine upgrade process), and what will happen if the engine upgrade failed to complete.

yasker (changes requested):

enhancements/20210111-upgrade-engine-automatically.md
PhanLe1010: Yeah, it is not needed. We didn't do this in the code PR. Updated the LEP

enhancements/20210111-upgrade-engine-automatically.md
PhanLe1010: Thank you. Fixed the name and added more details

enhancements/20210111-upgrade-engine-automatically.md
```diff
@@ -0,0 +1,109 @@
...
+The proposal has 2 parts:
+1. In which component we will do the upgrade.
+2. Identify when is it ok to upgrade engine for volume to the new default engine image.
+
+For part 1, we will do the upgrade inside `engine image controller`. 
+The controller will constantly watch and upgrade the engine images for volumes when it is ok to upgrade.
+
+For part 2, we upgrade engine image for a volume when the following conditions are met:
+1. The new default engine image is ready
+1. The new default engine image is newer than the current volume's engine image in term of semantic version comparision
```
yasker, line 34: As we discussed, don't need version comparsion. As long as API version is compatible, we can do live upgrade. Otherwise we need to do offline upgrade. 

@shuo-wu API version doc is still on you: https://github.com/longhorn/longhorn/issues/1544 :)

enhancements/20210111-upgrade-engine-automatically.md
```diff
@@ -0,0 +1,109 @@
...
+For engine enhancement, also requires engine integration test plan.
+
+### Upgrade strategy
+
+No upgrade strategy is needed.
+
+
+### Addtional Context
+
+This enhancement is designed to for end users who use office image tags.
```
yasker, line 109: Shouldn't be necessary if we don't compare the image tag for version.

enhancements/20210111-upgrade-engine-automatically.md
```diff
@@ -0,0 +1,109 @@
...
+      And The current volume's engine image is compatible with the new default engine image.
+      And the volume's robustness must be `healthy` if volume is a DR volume.
+      And volume is not expanding
+     
+1. Make sure not to upgrade too many volumes on the same node at the same time.
+   We define a constant `maxConcurrentEngineUpgradingPerNode` which specifies the maximum number of upgrading engines on a node at a time.
+   Filter the upgrading candidate set so that total number of upgrading volumes and candidates per node is smaller than `maxConcurrentEngineUpgradingPerNode`.
+1. For each volume candidate, set `v.Spec.EngineImage = new default engine image` to update the engine for the volume.
+1. if the engine upgrade failed to complete (e.g. the v.Spec.EngineImage != v.Status.CurrentImage), 
+   we just consider it is the same as volume is in upgrading process and skip it.
```
yasker, line 91: Should we have a fail-safe to stop auto-upgrade in case too many engine upgrades failed?

enhancements/20210111-upgrade-engine-automatically.md
```diff
@@ -0,0 +1,109 @@
...
+      And the current volume's engine image tag is smaller than default engine image tag in term of semantic version comparison.
+   1. Volume is in attached state (live upgrade).
+      Volume is not upgrading engine image.
+      And the current volume's IM image tag is smaller than default engine image tag in term of semantic version comparison.
+      And The current volume's engine image is compatible with the new default engine image.
+      And the volume's robustness must be `healthy` if volume is a DR volume.
+      And volume is not expanding
+     
+1. Make sure not to upgrade too many volumes on the same node at the same time.
+   We define a constant `maxConcurrentEngineUpgradingPerNode` which specifies the maximum number of upgrading engines on a node at a time.
```
yasker, line 87: `ConcurrentEngineUpgradePerNodeLimit`. And is this apply to the manual upgrade as well? Otherwise, we should call it `ConcurrentAutomaticEngineUpgradePerNodeLimit`.

yasker (changes requested):

enhancements/20210111-upgrade-engine-automatically.md
PhanLe1010: In case there are many upgrades fails, the `concurrentAutomaticEngineUpgradePerNodeLimit` will server as a gate to stop the additional upgrade

enhancements/20210111-upgrade-engine-automatically.md
PhanLe1010: Thank you. Fixed

enhancements/20210111-upgrade-engine-automatically.md
PhanLe1010: Thank you! Fixed

enhancements/20210111-upgrade-engine-automatically.md
PhanLe1010: Thank you! Fixed

enhancements/20210111-upgrade-engine-automatically.md
```diff
@@ -0,0 +1,105 @@
...
+
+For part 2, we upgrade engine image for a volume when the following conditions are met:
+1. The new default engine image is ready
+1. Volume is not upgrading engine image
+1. The volume condition is one of the following:
+   1. Volume is in detached state.
+   1. Volume is in attached state (live upgrade).
+      The current volume's engine image is compatible with the new default engine image.
+   1. If volume is a DR volume, the volume's robustness must be `healthy`
+   1. Volume is not expanding
```
joshimoo, line 40: If I understood @shuo-wu correctly, this is an exclusion criteria.
So if the volume is attached (one of the cases) but it's expanding we should not do the upgrade.

So maybe something like this:
1. The volume condition is one of the following:
- XXX
- YYY
- ZZZ

2. And not one of the following:
- the volume is expanding

PhanLe1010: Sorry, I was not clear on this part. Fixed it to indicate that the volume condition is one of the 2 cases:
```
   1. Volume is in detached state.
   2. Volume is in attached state (live upgrade).
      And The current volume's engine image is compatible with the new default engine image.
      And If volume is a DR volume, the volume's robustness must be `healthy`
      And Volume is not expanding.
```
How do you think?

enhancements/20210111-upgrade-engine-automatically.md
```diff
@@ -0,0 +1,168 @@
...
+
+For part 1, we will do the upgrade inside `engine image controller`. 
+The controller will constantly watch and upgrade the engine images for volumes when it is ok to upgrade.
+
+For part 2, we upgrade engine image for a volume when the following conditions are met:
+1. The new default engine image is ready
+1. Volume is not upgrading engine image
+1. The volume condition is one of the following:
+   1. Volume is in detached state.
+   1. Volume is in attached state (live upgrade).
```
shuo-wu, line 39: Add the condition: volume should be `healthy`

enhancements/20210111-upgrade-engine-automatically.md
```diff
@@ -0,0 +1,168 @@
...
+
+### User Stories
+
+Before this enhancement, users have to manually upgrade engine images for volume after upgrading Longhorn system to a newer version.
+If there are thoudsands of volumes in the system, this is a significant manual work.
+
+After this enhancement users either have to do nothing (in case live upgrade is possible) 
+or they only have to scale down/up the workload (in case there is a new default IM image)
+
+### User Experience In Detail
```
shuo-wu, line 52: Do we need to mention that users need to disable this if they want to update the engine image to a specific version for the volumes?

enhancements/20210111-upgrade-engine-automatically.md
```diff
@@ -0,0 +1,168 @@
...
+
+1. List all volumes in Longhorn system. 
+   Select a set of volume candidate for upgrading.
+   We select candidates that has the condition is one of the following case:
+   1. Volume is in detached state.
+   1. Volume is in attached state (live upgrade).
+      And Volume is not upgrading engine image.
+      And The current volume's engine image is compatible with the new default engine image.
+      And the volume is not a DR volume.
+      And volume is not expanding.
```
shuo-wu, line 83: And volume is healthy

shuo-wu (changes requested):

enhancements/20210111-upgrade-engine-automatically.md
PhanLe1010: Thank you. Fixed

enhancements/20210111-upgrade-engine-automatically.md
PhanLe1010: Thank you. Fixed

enhancements/20210111-upgrade-engine-automatically.md
PhanLe1010: Thank you. Fixed

shuo-wu (approved):

yasker (approved):

Issue #2111, opened 2020-12-17
task: switch the image base for Longhorn share manager to ubuntu


yasker:
Trivy doesn't support scanning Fedora for vulnerabilities

joshimoo:
Need to test arm images.

khushboo-rancher:
Verified with Longhorn-master `01/21/2021`

Validation - **Fail**

With image `longhornio/longhorn-share-manager:v1_20210106`, on deploying https://raw.githubusercontent.com/longhorn/longhorn-manager/master/examples/rwx/rwx-nginx-deployment.yaml, the deployment doesn't come up and stuck with the below error.
```
7m47s       Warning   FailedMount              pod/rwx-test-5d98bd9fff-sg9xw   MountVolume.SetUp failed for volume "pvc-3d9bd71e-e7ab-4420-bffa-b158e7de9186" : rpc error: code = Internal desc = mount failed: exit status 32
Mounting command: /usr/bin/systemd-run
Mounting arguments: [--description=Kubernetes transient mount for /var/lib/kubelet/pods/17a30a9f-0b82-4500-91bb-211b8b2345b9/volumes/kubernetes.io~csi/pvc-3d9bd71e-e7ab-4420-bffa-b158e7de9186/mount --scope -- /bin/mount -t nfs -o vers=4.1,noresvport,soft,sync,intr,timeo=7,retrans=3 10.43.228.161:/pvc-3d9bd71e-e7ab-4420-bffa-b158e7de9186 /var/lib/kubelet/pods/17a30a9f-0b82-4500-91bb-211b8b2345b9/volumes/kubernetes.io~csi/pvc-3d9bd71e-e7ab-4420-bffa-b158e7de9186/mount]
Output: Running scope as unit: run-ra89ae0577d5f4d7f86e12343de582104.scope
mount.nfs: mounting 10.43.228.161:/pvc-3d9bd71e-e7ab-4420-bffa-b158e7de9186 failed, reason given by server: No such file or directory
```

joshimoo:
Leaving a note here, when testing the new image on arm ubuntu 20.04 host.
It looks /etc/mtab is not available when starting ganesha so it's unable to find all mounted filesystems.
You can investigate this easily by deploying a share-manager image, but replacing the start command with /bin/bash then manually start `./longhorn-share-manager` with an attached volume.

This will try to start ganesha and you can investigate the ganesha.log in /tmp/ganesha.log.
[...]

PR #2173, opened 2021-01-13
Update default share-manager image to v1_20210106

joshimoo:
We switched from fedora to an ubuntu base image, to support the security
scanner.

longhorn/longhorn#2111

### DIFF
--- a/chart/questions.yml
+++ b/chart/questions.yml
@@ -65,7 +65,7 @@ questions:
     label: Longhorn Share Manager Image Repository
     group: "Longhorn Images Settings"
   - variable: image.longhorn.shareManager.tag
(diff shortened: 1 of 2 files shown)


yasker (changes requested):
Also need to change https://github.com/longhorn/longhorn/blob/master/chart/questions.yml#L68

joshimoo:
Updated to pr and switched the base to master, so it's available on all future dev builds.

yasker:
This should be for v1.1.1 branch, otherwise, all the users using master will start using the image immediately.

joshimoo:
changed to v1.1.1

yasker (approved):

Issue #2207, opened 2021-01-23
[FEATURE] Enhanced resource reservation mechanism


yasker:
**Is your feature request related to a problem? Please describe.**
To make sure users have a stable storage system, we need to be more specific on reserving resources for Longhorn, as well as make it clear in the case that the reserved resources might not be enough for Longhorn to function when there are too many engines/replicas running on the node.

**Describe the solution you'd like**
Make the Guaranteed Engine CPU to be node-specific. We can give the user two options to choose how many CPUs to reserve by default for engine instance manager, and replica instance manager. Both will be set to 1/8 of the total available CPU number on the node.

Be careful since this might result in instance manager fail to start during installation or upgrade. For upgrade, we might require double the resource due the upgrade of the instance managers.

**Describe alternatives you've considered**
As we currently do, ask user to use GuaranteedEngineCPU manually as a global option, but it's not good enough compared to the solution above.

**Additional context**
The other part of this problem is to warn user about CPU starving, which will be covered by https://github.com/longhorn/longhorn/issues/1930

https://github.com/longhorn/longhorn/issues/1691 has more information regarding the resource requirement.

khushboo-rancher:
Verified with Longhorn master - `04/01/2021`

@shuo-wu Can you confirm the below results are fine.
For below config node the default cpu request with 12 in the setting `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU` -  The request doesn't fit with the formula -  `total cpu * Guaranteed Manager CPU%`

RAM | CPU Request
--- |---
32 GB | 960m
8 GB | 480m
4 GB | 240m

Also, have we explicitly mentioned in the doc anywhere that for Longhorn previous version upgraded to v1.1.1, one has to set the existing cpu request to 0 to have this setting applicable?

shuo-wu:
The requests in the table fit the formula. e.g., 8 CPU * 12% = 0.96 CPU = 960 mCPU. It's not related to RAM...

In the setting description, there is one related note `This global setting will be ignored for a node if the field "ReplicaManagerCPURequest" on the node is set`. But we didn't mention that the node fields will be set when upgrading from old versions. We may need to mention that somewhere.

khushboo-rancher:
Validation - **Pass**

Test cases from https://github.com/longhorn/longhorn-tests/pull/532 are passing.

shuo-wu:
The upgrade note PR related to this enhancement: https://github.com/longhorn/website/pull/281

[...]

PR #2213, opened 2021-01-25
enhancement: Add a new LEP 'enhanced-cpu-reservation'

shuo-wu:
#2207

### DIFF
--- a/enhancements/20210125-enhanced-cpu-reservation.md
+++ b/enhancements/20210125-enhanced-cpu-reservation.md
@@ -1,0 +1,98 @@
+# Enhanced CPU Reservation
+
+## Summary
+To make sure Longhorn system stable enough, we can apply the node specific CPU resource reservation mechanism to avoid the engine/replica engine crash due to the CPU resource exhaustion. 
+
+### Related Issues
+https://github.com/longhorn/longhorn/issues/2207
+
+## Motivation
+### Goals
+1. Reserve CPU resource separately for engine manager pods and replica manager pods.
+2. The reserved CPU count is node specific: 
+    1. The more allocable CPU resource a node has, the more CPUs be will reserved for the instance managers in general.
+    2. Allow reserving CPU resource for a specific node. This setting will override the global setting.
+
+### Non-goals
+1. Guarantee that the CPU resource is always enough, or the CPU resource reservation is always reasonable based on the volume numbers of a node.
+2. Notify/Warn users CPU resource exhaustion on a node: https://github.com/longhorn/longhorn/issues/1930
+
+## Proposal
+1. Add new fields `node.Spec.EngineManagerCPURequest` and `node.Spec.ReplicaManagerCPURequest`. This allows reserving a different amount of CPU resource for a specific node.
+2. Add two settings `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU`:
+    1. It indicates how many percentages of CPU on a node will be reserved for one engine/replica manager pod.
+    2. The old settings `Guaranteed Engine CPU` will be deprecated:
+        - This setting will be unset and readonly in the new version.
+        - For the old Longhorn system upgrade, Longhorn will automatically set the node fields based on the old setting then clean up the old setting, so that users don't need to do anything manually as well as not affect existing instance manager pods. 
+
+### User Stories
+Before the enhancement, users rely on the setting `Guaranteed Engine CPU` to reserve the same amount of CPU resource for all engine managers and all replica managers on all nodes. There is no way to reserve more CPUs for the instance managers on node having more allocable CPUs.
+
+After the enhancement, users can:
+    1. Modify the global settings `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU` to reserve how many percentage of CPUs for engine manager pods and replica manager pods, respectively.
+    2. Set a different CPU value for the engine/replica manager pods on some particular nodes by `Node Edit`.
+
+### API Changes
+Add a new field `EngineManagerCPURequest` and `ReplicaManagerCPURequest` for node objects.
+
+## Design
+### Implementation Overview
+#### longhorn-manager:
+1. Add 2 new settings `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU`.
+    - These 2 setting values are integers ranging from 0 to 40.
+    - The sum of the 2 setting values should be smaller than 40 (%) as well.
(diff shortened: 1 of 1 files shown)


enhancements/20210125-enhanced-cpu-reservation.md
```diff
@@ -0,0 +1,69 @@
...
+### Non-goals
+1. Guarantee that the CPU resource is always enough, or the CPU resource reservation is always reasonable based on the volume numbers of a node.
+2. Notify/Warn users CPU resource exhaustion on a node: https://github.com/longhorn/longhorn/issues/1930
+
+## Proposal
+Add two settings `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU`:
+    1. It indicates how many percentages of CPU on a node will be reserved for one engine/replica manager pod.
+    2. The old settings `Guaranteed Engine CPU` will be deprecated:
+        - For the newly launched Longhorn cluster, the value will be empty by default.
+        - For the existing/upgraded Longhorn cluster, both the old setting value and the new setting values will be set. However, the new settings will be applied only if the old setting value becomes empty, so that all existing volumes will be crashed unexpectedly during the system upgrade.
```
yasker, line 23: `will be crashed unexpectedly?`

How does the upgrade process work exactly?

enhancements/20210125-enhanced-cpu-reservation.md
```diff
@@ -0,0 +1,69 @@
...
+
+After the enhancement, users can control how many percentages of CPU will be reserved for one engine or replica manager pod on a node.
+
+### API Changes
+N/A
+
+## Design
+### Implementation Overview
+#### longhorn-manager:
+1. Add 2 new settings `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU`.
```
yasker, line 36: `Guaranteed Engine Instance Manager CPU` and so on.

Are we going to have a node-specific setting?

What's going to happen if the percentage is too large for certain nodes that have too many CPU reserved workloads? Kubernetes's scheduling will fail.

yasker (changes requested):

yasker (changes requested):
I have some doubts regarding the upgrade path.

enhancements/20210125-enhanced-cpu-reservation.md
```diff
@@ -0,0 +1,100 @@
...
+2. Add two settings `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU`:
+    1. It indicates how many percentages of CPU on a node will be reserved for one engine/replica manager pod.
+    2. The old settings `Guaranteed Engine CPU` will be deprecated:
+        - For the newly launched Longhorn cluster, the value will be empty by default.
+        - For the existing/upgraded Longhorn cluster, both the old setting value and the new setting values will be set. However, the new settings will be applied only if the old setting value becomes empty, so that all existing volumes will be crashed unexpectedly during the system upgrade.
+          In other words, the correct steps for the upgrade are:
+              1. Upgrade the system to the new version.
+              2. Find a chance that allows detaching all volumes. If there is no chance to detach all volumes, please modify the new settings or the node fields so that the requested CPUs for the instance manager pods remain the same when the deprecated setting no longer works.
```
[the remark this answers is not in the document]
shuo-wu: OK. Considering the worker node CPUs may be different, Longhorn can set the new fields of the node spec automatically to the old value of `Guarantee Engine CPU`. Then users don't need to worry about the old setting at all after the upgrade.

yasker (approved):
LGTM. @joshimoo can you review as well?

enhancements/20210125-enhanced-cpu-reservation.md
```diff
@@ -0,0 +1,97 @@
...
+    2. Set a different CPU value for the engine/replica manager pods on some particular nodes by `Node Edit`.
+
+### API Changes
+Add a new field `EngineManagerCPURequest` and `ReplicaManagerCPURequest` for node objects.
+
+## Design
+### Implementation Overview
+#### longhorn-manager:
+1. Add 2 new settings `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU`.
+    - Make sure the sum of the 2 setting values should be smaller than 40 (%).
```
joshimoo, line 42: Mentioned this in the PR review https://github.com/longhorn/longhorn-manager/pull/828
Should we treat them independent of each other instead of validating the sum of both?

Thinking about the dedicated storage node setup, where it should be totally valid to request 40% replicas on the storage nodes.

And 40% engine on the workload nodes.

enhancements/20210125-enhanced-cpu-reservation.md
```diff
@@ -0,0 +1,97 @@
...
+    1. Modify the global settings `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU` to reserve how many percentage of CPUs for engine manager pods and replica manager pods, respectively.
+    2. Set a different CPU value for the engine/replica manager pods on some particular nodes by `Node Edit`.
+
+### API Changes
+Add a new field `EngineManagerCPURequest` and `ReplicaManagerCPURequest` for node objects.
+
+## Design
+### Implementation Overview
+#### longhorn-manager:
+1. Add 2 new settings `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU`.
```
joshimoo, line 41: Mention that these are in decimal percentages in the range 0 to 40

enhancements/20210125-enhanced-cpu-reservation.md
```diff
@@ -0,0 +1,97 @@
...
+
+### User Stories
+Before the enhancement, users rely on the setting `Guaranteed Engine CPU` to reserve the same amount of CPU resource for all engine managers and all replica managers on all nodes. There is no way to reserve more CPUs for the instance managers on node having more allocable CPUs.
+
+After the enhancement, users can:
+    1. Modify the global settings `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU` to reserve how many percentage of CPUs for engine manager pods and replica manager pods, respectively.
+    2. Set a different CPU value for the engine/replica manager pods on some particular nodes by `Node Edit`.
+
+### API Changes
+Add a new field `EngineManagerCPURequest` and `ReplicaManagerCPURequest` for node objects.
```
joshimoo, line 36: Mention that these are milli values, can be down in the implementation details.

joshimoo (approved):
LGTM, we should consider treating the two values independent of each other instead of validating the sum of both.

innobead (approved):
LGTM

Issue #2033, opened 2020-12-03
[FEATURE] Help user to install dependencies when installing Longhorn


yasker:
If we decide to install open-iscsi dependency, we need to do more investigation to ensure we have a way to workaround the potential Helm chart installation failure due to initContainer failed to install open-iscsi package.

For RWX volumes, we also require NFS client on the node. We can include it here as well.

Ref: https://github.com/longhorn/longhorn/issues/1741#issuecomment-724183156

yasker:
I am still concerned about automated installation for dependencies. I don't want Helm installation to fail unnecessarily. For the first step, let's update the [yaml](https://github.com/longhorn/longhorn/tree/master/deploy/iscsi) to install `nfs-common`(NFS client) as well. Need to update the doc as well.

cclhsu:
Have update/add manifest for iscsi/nfs (#2210, #2211).  I am currently exploring using helm installation.

yasker:
@cclhsu Can you file out the checklist before moving the issue to Merged? Thanks.

khushboo-rancher:
The Yaml files for `iscsi` and `nfs` installation are validated with https://github.com/longhorn/longhorn/issues/2058 and https://github.com/longhorn/longhorn/issues/1741.

We should be good to close this.
@yasker @cclhsu anything specifically needs to be validated with this one?

PR #2215, opened 2021-01-26
update helm to install iscsi nfs using daemonset (#2033)

cclhsu:
#### Proposed Changes ####

1. Update helm for daemonset-sa.yaml `longhorn-manager` to install `iscsi` and `nfs` using deamonset/initcontainer.
2. Add flag in values.yaml to enable/disable package host installation.

#### Types of Changes ####

Improvement.

#### Verification ####

1. Deploy RKE/K3S on Centos/Ubuntu/SLES15SP2 and verify if containers run successfully on each node and installed package run successfully.

#### Linked Issues ####

Refs: https://github.com/longhorn/longhorn/issues/2033

#### Further Comments ####

Additional verification is required.

1. Different K8S deployments.
2. End-to-End testing.

NOTE: If initcontainer deployment failed,  `longhorn-manager` will not deploy.

Signed-off-by: cclhsu <clark.hsu@suse.com>

### DIFF
--- a/chart/templates/daemonset-sa.yaml
+++ b/chart/templates/daemonset-sa.yaml
@@ -5,6 +5,15 @@ metadata:
     app: longhorn-manager
   name: longhorn-manager
   namespace: {{ include "release_namespace" . }}
+  {{ if or .Values.iscsi.install .Values.nfs.install }}
+  annotations:
+    {{ if .Values.iscsi.install }}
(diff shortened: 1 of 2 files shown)


yasker:
We need to be very careful regarding the automatic installation of those components. Since it might cause the helm install failure, and we need to always make sure the user can recover from the installation/upgrade failure (which is not an easy thing to do with Helm). That's also why I was hesitant to include the init container installation for v1.1.0 release.

For https://github.com/longhorn/longhorn/issues/2033, I consider we're good once we have the NFS client installed as well. Automated installation through Helm is a stretch goal, and I still think it might bring in more trouble than it's worth. 

In short, if we want to do this, we need to figure out all the failure scenarios and the way to recover from them, then decide if it's worth doing it.

cclhsu:
I agreed. I will close the pull request for now.

Issue #2127, opened 2020-12-21
[FEATURE] Live migration support for KubeVirt


yasker:
**Is your feature request related to a problem? Please describe.**
Harvester is using Longhorn as the underlying storage. However, live migration in Harvester requests Longhorn volume to be simultaneously accessible from two different nodes, when the VM is migrating from one node to another.

**Describe the solution you'd like**
We need to support KubeVirt live migration in someway to enable the feature in Harvester.

**Describe alternatives you've considered**
N/A

**Additional context**
This was implemented for RancherVM (https://github.com/longhorn/longhorn/issues/87) but we removed the feature later.
Also, this depends on https://github.com/longhorn/longhorn/issues/2006 since normally VM requires a backing file for the root device, and Longhorn should be able to do live migration with the backing file.

yasker:
Remove `require/doc` due to this feature is not directly user-facing.

khushboo-rancher:
@joshimoo I tried the below scenario,
```
1. Creates a new RWX migratable volume.
2. Attach to test node to write some test data on it.
3. Detach from test node.
4. Get set of nodes excluding the test node
5. Attach volume to node 1 (initial node)
6. Attach volume to node 2 (migration target)
7. Wait for migration ready (engine running on node 2)
8. Detach volume from node 1
```
 I'm using Longhorn UI to detach the volume and it get stuck at step 8.
```
[longhorn-manager-rvcqb] E0329 17:03:47.721072       1 volume_controller.go:202] fail to sync longhorn-system/pvc-85f358c1-f683-458f-a1a8-e33692d25153: BUG: multiple engines detected when volume pvc-85f358c1-f683-458f-a1a8-e33692d25153 is detached 
[longhorn-manager-rvcqb] time="2021-03-29T17:03:47Z" level=warning msg="Dropping Longhorn volume longhorn-system/pvc-85f358c1-f683-458f-a1a8-e33692d25153 out of the queue" controller=longhorn-volume error="fail to sync longhorn-system/pvc-85f358c1-f683-458f-a1a8-e33692d25153: BUG: multiple engines detected when volume pvc-85f358c1-f683-458f-a1a8-e33692d25153 is detached" node=khush-test-lh-wk1 
[...]

PR #2266, opened 2021-02-19
Add LEP for live migration feature

joshimoo:
Longhorn #2127

### DIFF
--- a/enhancements/20210216-volume-live-migration.md
+++ b/enhancements/20210216-volume-live-migration.md
@@ -1,0 +1,235 @@
+# Volume live migration
+
+## Summary
+
+To enable Harvester to utilize Kubevirts live migration support, we need to allow for volume live migration,
+so that a Kubevirt triggered migration will lead to a volume migration from the old node to the new node.
+
+### Related Issues
+- https://github.com/longhorn/longhorn/issues/2127
+- https://github.com/rancher/harvester/issues/384
+- https://github.com/longhorn/longhorn/issues/87
+
+## Motivation
+
+### Goals
+- Support Harvester VM live migration
+
+### Non-goals
+- Using multiple engines for faster volume failover for other scenarios than live migration
+
+## Proposal
+We want to add volume migration support so that we can use the VM live migration support of Kubevirt via Harvester.
+By limiting this feature to that specific use case we can use the csi drivers attach / detach flow to implement migration interactions.
+To do this, we need to be able to start a second engine for a volume on a different node that uses matching replicas of the first engine.
+We only support this for a volume while it is used with `volumeMode=BLOCK`, since we don't support concurrent writes and having kubernetes mount a filesystem even in read only
+mode can potentially lead to a modification of the filesystem (metadata, access time, journal replay, etc).
+
+
+### User Stories
+Previously the only way to support live migration in Harvester was using a Longhorn RWX volume that meant dealing with NFS and it's problems, 
+instead we want to add support for live migration for a traditional Longhorn volume this was previously implemented for the old RancherVM.
+After this enhancement Longhorn will support a special `migratable` flag that allows for a Longhorn volume to be live migrated from one node to another.
+The assumption here is that the initial consumer will never write again to the block device once the new consumer takes over.
+
+
+### User Experience In Detail
+
+#### Creating a migratable storage class
+To test one needs to create a storage class with `migratable: "true"` set as a parameter.
+Afterwards an RWX PVC is necessary since migratable volumes need to be able to be attached to multiple nodes.
+```yaml
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: longhorn-migratable
+provisioner: driver.longhorn.io
+allowVolumeExpansion: true
+parameters:
+  numberOfReplicas: "3"
+  staleReplicaTimeout: "2880" # 48 hours in minutes
+  fromBackup: ""
+  migratable: "true"
+```
+
+#### Testing Kubevirt VM live migration
+We use CirOS as our test image for the live migration.
+The login account is `cirros` and the password is `gocubsgo`.
+To test with Harvester one can use the below example yamls as a quick start.
+
+Deploy the below yaml so that Harvester will download the CirrOS image into the local Minio store.
+NOTE: The CirrOS servers don't support the range request which Kubevirt importer uses, which is why we let harvester download the image first.
+```yaml
+apiVersion: harvester.cattle.io/v1alpha1
+kind: VirtualMachineImage
+metadata:
+  name: image-jxpnq
+  namespace: default
+spec:
+  displayName: cirros-0.4.0-x86_64-disk.img
+  url: https://download.cirros-cloud.net/0.4.0/cirros-0.4.0-x86_64-disk.img
+```
+
+Afterwards deploy the `cirros-rwx-blk.yaml` to create a live migratabale virtual machine.
+```yaml
+apiVersion: kubevirt.io/v1alpha3
+kind: VirtualMachine
+metadata:
+  labels:
+    harvester.cattle.io/creator: harvester
+  name: cirros-rwx-blk
+spec:
+  dataVolumeTemplates:
+  - apiVersion: cdi.kubevirt.io/v1alpha1
+    kind: DataVolume
+    metadata:
+      annotations:
+        cdi.kubevirt.io/storage.import.requiresScratch: "true"
+      name: cirros-rwx-blk
+    spec:
+      pvc:
+        accessModes:
+        - ReadWriteMany
+        resources:
+          requests:
+            storage: 8Gi
+        storageClassName: longhorn-migratable
+        volumeMode: Block 
+      source:
+        http:
+          certConfigMap: importer-ca-none
+          url: http://minio.harvester-system:9000/vm-images/image-jxpnq # locally downloaded cirros image
+  running: true
+  template:
+    metadata:
+      annotations:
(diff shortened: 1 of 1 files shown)


enhancements/20210216-volume-live-migration.md
```diff
@@ -0,0 +1,193 @@
+# Volume live migration
+
+## Summary
+
+The Summary section is incredibly important for producing high quality user-focused documentation such as release notes or a development roadmap.
```
yasker, line 5: ?

enhancements/20210216-volume-live-migration.md
```diff
@@ -0,0 +1,193 @@
...
+
+https://github.com/longhorn/longhorn/issues/2127
+https://github.com/longhorn/longhorn-manager/pull/822
+https://github.com/rancher/harvester/issues/384
+https://github.com/longhorn/longhorn/issues/87
+
+## Motivation
+
+### Goals
+- use attach/detach flow to implement migration interactions
```
yasker, line 18: The goal is to support the VM live migration. Then we can start talking details a little more.

enhancements/20210216-volume-live-migration.md
```diff
@@ -0,0 +1,193 @@
...
+- run a migration engine for a volume in RWX access mode that is marked as `migratable`
```
yasker, line 19: This is implementation details, not goal.

enhancements/20210216-volume-live-migration.md
```diff
@@ -0,0 +1,193 @@
...
+- only allow volumes during migration to be mounted in block device mode on a pod
```
yasker, line 20: Raw block device mode you meant? or are you referring to `RWO` vs `RWX`?

enhancements/20210216-volume-live-migration.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Non-goals
+- support migration for RWO access mode
+- support mounting a filesystem during migration
+- HA support for NFS shared volumes
```
yasker, line 25: ? Why we want to mention it here?

enhancements/20210216-volume-live-migration.md
```diff
@@ -0,0 +1,193 @@
...
+- support concurrent writes on both engines
+
+## Proposal
+We want to add volume migration support so that we can use the VM live migration support of Kubevirt via Harvester.
+To do this, we need to be able to start a second engine for a volume on a different node that uses matching replicas of the first engine.
+We only want to support this for block device mode, since we don't support concurrent writes and having kubernetes mount a filesystem even in read only
```
yasker, line 31: So you meant `raw block device` mode. 
```
You can set the value of volumeMode to Block to use a volume as a raw block device.
```
https://kubernetes.io/docs/concepts/storage/persistent-volumes/#volume-mode

yasker (changes requested):
The implementation details part looks good. Though need a better summarization at the beginning.

enhancements/20210216-volume-live-migration.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Summary
+
+The Summary section is incredibly important for producing high quality user-focused documentation such as release notes or a development roadmap.
+A good summary is probably at least a paragraph in length.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/2127
+https://github.com/longhorn/longhorn-manager/pull/822
```
shuo-wu, line 11: Remove this line

enhancements/20210216-volume-live-migration.md
shuo-wu: Or `BlockVolume` mode vs `Filesystem` mode?

enhancements/20210216-volume-live-migration.md
```diff
@@ -0,0 +1,193 @@
...
+provisioner: driver.longhorn.io
+allowVolumeExpansion: true
+parameters:
+  numberOfReplicas: "3"
+  staleReplicaTimeout: "2880" # 48 hours in minutes
+  fromBackup: ""
+  migratable: "true"
+```
+
+#### Testing Kubevirt VM live migration
```
shuo-wu, line 61: It seems that the test steps and the user experience are mixed up... It's good to tell us how to test this feature with Harvester in detail. But I think it's better to put it in the test part

enhancements/20210216-volume-live-migration.md
```diff
@@ -0,0 +1,193 @@
...
+apiVersion: kubevirt.io/v1alpha3
+kind: VirtualMachineInstanceMigration
+metadata:
+  name: cirros-rwx-blk
+spec:
+  vmiName: cirros-rwx-blk
+```
+
+### API changes
+- volume detach call now expects a `detachInput { hostId: "" }` if `hostId==""` it will be treated as detach from all nodes same behavior as before.
```
shuo-wu, line 159: Add more explanations about the changed API `attach` and `detach`. e.g.:

For `attach`,
  1. If a volume is mode `RWO`, ......
  2.If a volume is mode `RWX` (NFS), ......
  3. If a volume is mode `RWX` (Migratable), migration means this attached volume receives a attach request to a different node.

For detach,
  1. If a volume is mode `RWO`, ......
  2.If a volume is mode `RWX` (NFS), ......
  3. If a volume is mode `RWX` (Migratable), migration confirmation means this attached volume receives a detach request that requires detaching from the 1st node. And migration rollback means this attached volume receives a detach request that requires detaching from the 2nd node

enhancements/20210216-volume-live-migration.md
```diff
@@ -0,0 +1,193 @@
...
+- csi driver now calls volume attach/detach for all volume types: RWO, RWX (NFS), RWX (Migratable).
+- the api volume-manager now determines, whether attach/detach is necessary and valid instead of the csi driver.
+
+## Design
+
+### Implementation Overview
+
+#### Volume migration flow
```
shuo-wu, line 167: This part maybe not straightforward enough if the reader hasn't reviewed the actual implementation.
Can you check if the following description is better?
1. Users launch a new VM with a new migratable Longhorn volume -> A migratable volume is created then attached to node1. Similar to regular attachment, Longhorn will set `v.spec.nodeID` to `node1` here.
2. Users launch the 2nd VM (pod) with the same Longhorn volume -> 
    1. Kubernetes requests the (attached) volume attaching to node2. Then Longhorn receive the attach call and set `v.spec.migrationNodeID` to `node2` with `v.spec.nodeID = node1`.
    2. Longhorn volume-controller brings up the new engine on node2, with inactive matching replicas (same as live engine upgrade)
    3. Longhorn CSI driver polls for the existence of the second engine on node2 before acknowledging attachment success.
3. Once the migration (actually the 2nd attachment succeeded/failed), the following detach decides whether migration is completed successfully, or a migration rollback is desired:
    1.  If succeeded:...
    2. If failed:....
4. Longhorn volume controller then cleans up the second engine and switches the active replicas to be the current engine ones.

shuo-wu (changes requested):

enhancements/20210216-volume-live-migration.md
joshimoo: The only user of this feature should be Harvester, there is to many things that can go wrong if it's used outside of that context.

enhancements/20210216-volume-live-migration.md
joshimoo: changed to
### Non-goals
- using multiple engines for faster volume failover

enhancements/20210216-volume-live-migration.md
joshimoo: Changed to
We only support this for a volume while it is used with `volumeMode=BLOCK`

enhancements/20210216-volume-live-migration.md
joshimoo: changed to
### Goals
- support harvester VM live migration

enhancements/20210216-volume-live-migration.md
joshimoo: removed

enhancements/20210216-volume-live-migration.md
joshimoo: done

enhancements/20210216-volume-live-migration.md
joshimoo: replaced with:
To enable Harvester to utilize Kubevirts live migration support, we need to allow for volume live migration,
so that a Kubevirt triggered migration will lead to a volume migration from the old node to the new node.

enhancements/20210216-volume-live-migration.md
joshimoo: Implementation detail, that is subject to change down the line.
Not sure if it makes sense to include that detail in a document that will not be updated.

enhancements/20210216-volume-live-migration.md
joshimoo: thanks, included this form of a description.

enhancements/20210216-volume-live-migration.md
yasker: `Support Harvester VM live migration`.

enhancements/20210216-volume-live-migration.md
yasker: `using multiple engines for faster volume failover for other scenarios than live migration`.

joshimoo:
updated the LEP to adress review feedback.

yasker (approved):
LGTM. @shuo-wu ?

shuo-wu (approved):

shuo-wu:
vv `Merging is blocked` vv

Issue #2197, opened 2021-01-19
[BUG] ganesha in the share-manager pod requires on IPv6 on the host


bentastic27:
There are some users that disable IPv6 on their nodes via kernel flags (`sysctl -w net.ipv6.conf.all.disable_ipv6=1`). When users try to create an RWX volume genesha in the share-manager pod fails with this error:

```
level=error msg=“nfs server exited with error” error=“ganesha.nfsd failed with error: exit status 2, output: VFS module failed to unregister”
```

joshimoo:
Just an additional note, this is only a problem when ipv6 is disable via the bootloader.
GRUB_CMDLINE_LINUX_DEFAULT="ipv6.disable=1"

Dynamically disabling it via sysctl will allow for the system to operate correctly.

c3y1huang:
Seem it is required to have host ipv6 enabled because the nfs-ganesha used in share-manager automatically binds to tcp6 socket unless on FreeBSD.

When ipv6 is disabled using sysctl, the change does not pass through to Kubernetes pods hence share-manager manages to survives.
```
# Execute in pod:

/ # cat /sys/module/ipv6/parameters/disable
0

/ # sysctl -a 2>/dev/null | grep disable_ipv
net.ipv6.conf.all.disable_ipv6 = 0
net.ipv6.conf.default.disable_ipv6 = 0
net.ipv6.conf.eth0.disable_ipv6 = 0
net.ipv6.conf.lo.disable_ipv6 = 0

/ # ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
3: eth0@if34: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1450 qdisc noqueue
    link/ether b2:df:ab:8d:aa:3d brd ff:ff:ff:ff:ff:ff
    inet 10.42.0.87/24 brd 10.42.0.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::b0df:abff:fe8d:aa3d/64 scope link
       valid_lft forever preferred_lft forever
```
[...]

PR #2283, opened 2021-03-02
Update default share-manager image to v1_20210302

c3y1huang:
Include
- Patched nfs-ganesha v3.3 to fallback to ipv4 when ipv6 is not available.
  https://github.com/longhorn/longhorn/issues/2197

- Fix for pod hanging at ganasha server error due to missing /etc/mtab file in Ubuntu image.
  https://github.com/longhorn/longhorn/issues/2111

Test: https://github.com/longhorn/longhorn/issues/2197#issuecomment-788595045


joshimoo (approved):
LGTM

Issue #2295, opened 2021-03-04
[FEATURE] Enhance the backing image feature


shuo-wu:
**Is your feature request related to a problem? Please describe.**
Right now when a backing image is requested, Longhorn will fetch the backing image file from the remote for each node. In other words, if a backing image is required for N nodes, each node will take bandwidth to pull the data and total bandwidth consumed is N * backing image size.

To save the bandwidth as well as make some fancy features possible, we need to refactor the backing image feature. 

**Describe the solution you'd like**
One possible solution is that Longhorn would launch a backing image manager for each node/disk. So that one backing image manager is responsible for pull the backing image from the remote first, then other backing image managers can fetching the data from the 1st backing image manager.

**Describe alternatives you've considered**
N/A

shuo-wu:
Will add drone and unit tests during or after the review.

khushboo-rancher:
Validated with Longhorn- master v1.1.1-rc1

Validation - **Pass**

The below test cases have been validated.
1. Backing image with RWO
2. Backing image with RWX
3. Deletion of backing image.
4. Creation of volume with backing image using storage class.
5. Create backup of a volume with backing image
6. Restore backup of a volume with backing image
7. DR volume with a volume with backing image
8. Expansion of volume with backing image.
8. Evict disk with backing image
9. Backing image feature on PSP enabled cluster

Thank you @shuo-wu for helping me with upgrade scenarios testing, this is can be closed once @shuo-wu confirms the upgrade scenarios.

shuo-wu:
The upgrade tests are passed:
1. system upgrade with compatible backing image manager
2. system upgrade with incompatible backing image manager

Issue #1948, opened 2020-11-03
[ENHANCEMENT] Speed up rebuilding by getting checksum simultaneously


shuo-wu:
**Is your feature request related to a problem? Please describe.**
We can speed up the rebuilding by [requesting the server block checksum](https://github.com/longhorn/sparse-tools/blob/bb457e12173d2073ce3f8193cf5e3208c828ac43/sparse/client.go#L248) and [calculating the local block checksum](https://github.com/longhorn/sparse-tools/blob/bb457e12173d2073ce3f8193cf5e3208c828ac43/sparse/client.go#L262) simultaneously.

**Describe the solution you'd like**
See above.

shuo-wu:
The reason that failed replica reuse time and new replica rebuilding time are the same is:
1. Since there is actually no data on the receiver side for the new replica rebuilding case, `client.getServerChecksum` takes less time.
```
[instance-manager-r-3dab8865] [vol-r-793b61f8] time="2020-11-10T06:19:14Z" level=info msg="Sync batchInterval length 131072, total time 3.668229ms, checksum time 1.359858ms, percentage 37.0712406450088%"
[instance-manager-r-3dab8865] [vol-r-793b61f8] time="2020-11-10T06:19:14Z" level=info msg="Sync batchInterval length 131072, total time 2.857143ms, checksum time 741.813µs, percentage 25.963453701827316%"
[instance-manager-r-3dab8865] [vol-r-793b61f8] time="2020-11-10T06:19:14Z" level=info msg="Sync batchInterval length 131072, total time 2.603783ms, checksum time 760.654µs, percentage 29.213417554381454%"
[instance-manager-r-3dab8865] [vol-r-793b61f8] time="2020-11-10T06:19:14Z" level=info msg="Sync batchInterval length 131072, total time 3.549488ms, checksum time 1.706659ms, percentage 48.08183602818209%"
```
2. Though there is no writing for the failed replica reuse case, `client.getServerChecksum` takes a longer time and the client needs to calculate the local checksum. 
```
[instance-manager-r-b148c9fb] [vol-r-839c225b] time="2020-11-10T06:33:56Z" level=info msg="Sync batchInterval length 131072, total time 1.479149ms, checksum time 1.478969ms, percentage 99.98783084057116%"
[...]

PR #2326, opened 2021-03-11
enhancement: Enhance the backing image feature

shuo-wu:
Longhorn #2295


enhancements/20210305-enable-backing-image.md
```diff
@@ -23,33 +25,38 @@ https://github.com/longhorn/longhorn/issues/2006
...
 
-After the enhancement, users can directly specify the backing image during volume creation/restore with a click.
+After the enhancement, users can directly specify the backing image during volume creation/restore with a click. And one backing image can be shared among all replicas in the same disk.
 
 ### User Experience In Detail
 1. Users can modify the backing image cleanup timeout setting so that all non-used backing images will be cleaned up automatically from disks.
 2. Create a volume with a backing image 
     2.1. via Longhorn UI 
-        1. Users add a backing image, which is similar to add an engine image in the system.
+        1. Users add a backing image, which is similar to add an engine image or set up the backup target in the system.
```
yasker, line 53: Set up a backup target is a global setting action. How is that similar to adding a backing image?

enhancements/20210305-enable-backing-image.md
```diff
@@ -77,22 +130,50 @@ After the enhancement, users can directly specify the backing image during volum
...
+    - API Send: Send a backing image file to a receiver. This should be similar to replica rebuilding.
+    - API Delete: Delete a backing image directory.
+    - API List: Collect the status of all backing images in the work directory. 
+    - API Upgrade:
+        - For the new backing image manager, it should start to monitor the backing images, as well as handle all subsequent requests.
+        - For the old backing image manager, it should stop handling all requests and wait for the in progress requests complete, then exit without cleanup.
+- The work directory of the backing image manager is like:
+  ```
+  <Disk Path>/backing-images/
+  <Disk Path>/backing-images/<Downloading backing image 1>/backing
```
yasker, line 153: Why do you decide to use directories instead of a metadata file?

enhancements/20210305-enable-backing-image.md
```diff
@@ -23,33 +25,38 @@ https://github.com/longhorn/longhorn/issues/2006
...
 ### User Experience In Detail
 1. Users can modify the backing image cleanup timeout setting so that all non-used backing images will be cleaned up automatically from disks.
 2. Create a volume with a backing image 
     2.1. via Longhorn UI 
-        1. Users add a backing image, which is similar to add an engine image in the system.
+        1. Users add a backing image, which is similar to add an engine image or set up the backup target in the system.
         2. Users create/restore a volume with the backing image specified from the backing image list.
     2.2. via CSI (StorageClass)
-       1. Users specify `backingImageName` and `backingImageURL` in a StorageClass.
+       1. Users specify `backingImageName`, `backingImageAddress`, and `backingImageSecret` in a StorageClass.
```
yasker, line 56: We can deal with S3/NFS download later, I don't know if we have enough time for it in this release.

enhancements/20210305-enable-backing-image.md
shuo-wu: I meant how we download files for a backing image is similar to retrieve data from the backup target.

enhancements/20210305-enable-backing-image.md
shuo-wu: OK. I will continue using the existing download logic

enhancements/20210305-enable-backing-image.md
shuo-wu: 1. All backing images are using the fixed name `backing`
2. It may be better for isolation.

yasker:
Btw, if we're using the existing sync agent to for transferring images across the disks/nodes, there is a downside that we will run checksum against the whole file and trying to copy it once (since there is no hole in the file).

We do need to limit how much data we can transfer in one batch for the sync, like I mentioned in https://github.com/longhorn/longhorn/issues/1948#issuecomment-791641308

shuo-wu:
> Btw, if we're using the existing sync agent to for transferring images across the disks/nodes, there is a downside that we will run checksum against the whole file and trying to copy it once (since there is no hole in the file).
> 
> We do need to limit how much data we can transfer in one batch for the sync, like I mentioned in [#1948 (comment)](https://github.com/longhorn/longhorn/issues/1948#issuecomment-791641308)

I won't modify/improve this part in this PR. Let's wait for [#1948 (comment)](https://github.com/longhorn/longhorn/issues/1948) fixing the issue then I will directly update the vendor.

enhancements/20210305-enable-backing-image.md
yasker: What if you saw two different directories under an image?

enhancements/20210305-backing-image.md
```diff
@@ -77,22 +133,50 @@ After the enhancement, users can directly specify the backing image during volum
...
+    - API Send: Send a backing image file to a receiver. This should be similar to replica rebuilding.
+    - API Delete: Delete a backing image directory.
+    - API List: Collect the status of all backing images in the work directory. 
+    - API Upgrade:
+        - For the new backing image manager, it should start to monitor the backing images, as well as handle all subsequent requests.
+        - For the old backing image manager, it should stop handling all requests and wait for the in progress requests complete, then exit without cleanup.
+- The work directory of the backing image manager is like:
+  ```
+  <Disk Path>/backing-images/
+  <Disk Path>/backing-images/<Downloading backing image 1>/backing
```
yasker, line 156: `backing` and `sending` seem like the same. It indicates there is something referring to it.

`downloaded` means there is no reference.

You can avoid multi-state problem by using reference instead of the state.

shuo-wu: Got it. Then it's better to create a meta file containing a reference counter here.

yasker: A reference list instead of a counter is better. Otherwise, you don't know what's wrong if a miscount happened, e.g. due to simultaneous write to the file. We need some locking mechanism to protect the file still.

shuo-wu: Do you mean a file lock like backup operations? I am not sure if it's too complicated...

yasker: As long as you can prevent race conditions from happening, any mechanism will be fine.

enhancements/20210305-backing-image.md
```diff
@@ -0,0 +1,209 @@
...
+- Add raw backing file support.
+- Update the backing file info for replicas when a replica is created/opened.
+
+#### backing-image-manager:
+- The work directory of the backing image manager is like:
+  ```
+  <Disk path in container>/backing-images/
+  <Disk path in container>/backing-images/<Downloading backing image1 name>-<Downloading backing image1 UUID>/backing.tmp
+  <Disk path in container>/backing-images/<Downloaded backing image1 name>-<Downloaded backing image UUID>/backing
+  <Disk path in container>/backing-images/<Downloaded backing image1 name>-<Downloaded backing image UUID>/backing.cfg
```
innobead, line 160: Q: Because the backing image will be sent to another manager for serving the replica on the node (including backing.cfg? or it will be created later), how do we keep the backing.cfg consistent? I assume It will be handled by each backing image manager individually to have the backing.cfg finally consistent across the disks. 

Q: Is `backing.cfg` only used for the local record purpose for reference? or it will also be reflected into the backing image status in some cases like backing image controller restart? (I assume the info should be respected by the info saved in K8s because the manager also has a goroutine to check downloaded file status, and then finally update to backup.cfg accordingly)

innobead (approved):
In general, LGTM. 

Some comments to resolve/answer.

enhancements/20210305-backing-image.md
shuo-wu: A1: The most important fields in the cfg file are `UUID` and `Size`. The sender won't sync the cfg file. Instead, it's the caller's responsibility to make sure the meta info is consistent. Before syncing or downloading files, the backing image manager controller will verify the backing image UUID. If there is mismatching in an existing downloaded file, the image will be deleted. Then there is no way to use it as the sender. As for the size, right now there is no extra verification and correction. If the related backing image finds the size is different among different managers, it will print out error logs.

A2: The backing image manager will check and match the fields in the cfg file before trying to reuse an existing file. This is the main purpose of this cfg file. And later we can store the checksum of the image file in the config size.

Issue #2120, opened 2020-12-18
[BUG] annotation last-applied-tolerations magically added in 1.1.0


Bobonium:
**Describe the bug**
I manage my deployments with terraform. As a result I can see changes being done to the managed resources. I just upgraded to 1.1.0 and I did not ever configure any tolerations through the longhorn UI.
After the upgrade the following change is visible for the driver, the manager and the ui:
```
          ~ annotations      = {
              - "longhorn.io/last-applied-tolerations" = jsonencode([]) -> null
            }
```
The change shows us, that terraform wants to remove the annotation, as it's not part of the definition. It doesn't work though as longhorn always magically re-appends the value.
This is at least in theory also a problem for non terraform users. Re-applying the same deployment yamls should not result in a change. With this behavior it will always result in a change

**To Reproduce**
- Deploy longhorn 1.1.0 from deployment yaml
- Check annotations, the `longhorn.io/last-applied-tolerations` was magically added

**Expected behavior**
Please don't dynamically adjust resources that are being created by the deployment yamls. This has negative impacts for anyone that manages their true state in git

**Log**
irrelevant for the bug

**Environment:**
irrelevant for the bug

**Additional context**

PhanLe1010:
@Bobonium 

How does Terraform react when Longhorn automatically adds the annotation? E.g Does it deploy/redeploy Longhorn repeatedly? Doesn't Longhorn's normal functionalities work ok?

In another word, how badly is the problem?

Bobonium:
Hey @PhanLe1010 ,

Terraform will always show and apply a change, because longhorn automatically re-adds the annotation if it does not exist. I can (and I am currently) use a workaround, by simply adding the automated annotation to my terraform code, so it doesn't display a diff anymore. 

This is less a technical problem as it is a design flaw, that in my opinion shows, that Longhorn as a Project is not as mature as claims to be (>1). To be honest I really like this project, more than any of the other alternatives that I've tried so far and I'd like it to succeed and keep on using it in the future. But seeing changes like these let me worry, because the storage layer is the single most important part within a Cluster. If it breaks, most of the running workloads break and as such every design decision should be well thought through. Frankly, I don't think that that this was the case here.

I had a deeper look at how this system works, and from the point of view of someone who has to deploy and manage this, it's as unusable as it gets.
Beforehand you had a hardcoded check on tolerations that started with `kubernetes.io` and simply kept all tolerations that started like that, I can totally understand why you guys were unhappy with this.
[...]

PR #2348, opened 2021-03-15
Allow user to specify priority class and tolerations for Longhorn UI, Driver, Manager

PhanLe1010:
Now that we only watch/update managed components, we should allow user to specify Helm values for user deployed components (manager, driver, UI).

longhorn/longhorn#2120

### DIFF
--- a/chart/templates/daemonset-sa.yaml
+++ b/chart/templates/daemonset-sa.yaml
@@ -86,7 +86,14 @@ spec:
       imagePullSecrets:
       - name: {{ .Values.privateRegistry.registrySecret }}
       {{- end }}
+      {{- if .Values.longhornManager.priorityClass }}
+      priorityClassName: {{ .Values.longhornManager.priorityClass | quote}}
+      {{- end }}
       serviceAccountName: longhorn-service-account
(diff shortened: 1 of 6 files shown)


shuo-wu:
One quick question, do we allow users to set these when using Rancher APP to deploy Longhorn?

PhanLe1010:
@shuo-wu 

Good question. The answer is not directly but there is a workaround. 
Sice toleration is an array input, Rancher's `question.yaml` doesn't support array input. Users can workaround by doing `Edit as YAML` in Rancher UI instead. More details at https://github.com/rancher/rancher/issues/24918. I learn this info from you :)

shuo-wu (approved):

PhanLe1010:
ping @shuo-wu 
Can we merge this one

shuo-wu (approved):

shuo-wu:
Sorry. Can you directly rebase your branch? If I rebase the branch via Github UI, there will be an extra commit.

innobead (approved):

PhanLe1010:
@shuo-wu  
Rebased

Issue #2292, opened 2021-03-03
[BUG] longhorn-post-upgrade job fails due to "Error: container has runAsNonRoot and image will run as root"


Martin-Weiss:
**Describe the bug**
During config change (helm chart deployed via rancher and then changed values and with activated PSP setting) - we see a pod "longhorn-post-upgrade-knbp2" failing to start with the error "Error: container has runAsNonRoot and image will run as root"

Assumption is that this POD does not fulfill the PSP.

**To Reproduce**
Deploy longhorn 1.1.0 and then change a setting in values for the helm based deployment.
Do this on RKE2 with cis-1.5 profile in default namespace/project.
Deployment was done via Rancher 2.5.5

**Expected behavior**
The pod should run with the proper service account and the service account should have the proper PSP access.

**Log**
Error: container has runAsNonRoot and image will run as root"

**Environment:**
 - Longhorn version: 1.1.0
 - Kubernetes distro RKE2 
      rke2 version v1.19.7+rke2r1 (d1430ad44dc162f15a54d093439c20ffb91f559b)
     go version go1.15.5b5
   - Number of management node in the cluster: 3 masters
   - Number of worker node in the cluster: 3 worers
 - Node config
   - OS type and version: SLES 15 SP2
   - CPU per node: 2
   - Memory per node: 8 GB
   - Disk type(e.g. SSD/NVMe): NVMe
   - Network bandwidth between the nodes: 10G
 - Underlying Infrastructure: vmware esxi
 - Number of Longhorn volumes in the cluster: 3

**Additional context**

Assumption is that this is related to the security settings in RKE2 with cis-1.5 profile

joshimoo:
@khushboo-rancher @meldafrawi are we able to reproduce this internally?

Martin-Weiss:
FYI - same issue with longhorn-uninstall POD..

```
  Normal   Scheduled  2m37s                 default-scheduler  Successfully assigned longhorn-system/longhorn-uninstall-7mtld to s-vz-lev-kt022
  Warning  Failed     30s (x12 over 2m36s)  kubelet            Error: container has runAsNonRoot and image will run as root

```

jenting:
@Martin-Weiss,
To clarify, you only change the helm setting by disabled the PSP?

![image](https://user-images.githubusercontent.com/49380831/110760259-7df0de00-8289-11eb-936c-35d2ec3404a1.png)

Martin-Weiss:
[...]

PR #2358, opened 2021-03-16
Eanble post-upgrade and uninstall job security context privileged

jenting:
#### Symptom
Deploy an RKE2 + CIS-1.5 profile enabled cluster:
```shell
# 1st control plane node
sudo su -

systemctl disable firewalld
systemctl stop firewalld

curl -sfL https://get.rke2.io | sh -

systemctl enable rke2-server.service

cat << EOF > /etc/sysctl.d/60-rke2-cis.conf
vm.panic_on_oom=0
vm.overcommit_memory=1
kernel.panic=10
kernel.panic_on_oops=1
EOF
sudo systemctl restart systemd-sysctl

useradd -r -c "etcd user" -s /sbin/nologin -M etcd

mkdir -p /etc/rancher/rke2/
cat << EOF > /etc/rancher/rke2/config.yaml
profile: "cis-1.5"
EOF

systemctl start rke2-server.service

SERVER_NODE_TOKEN=`cat /var/lib/rancher/rke2/server/node-token`

# worker nodes
sudo su -

systemctl disable firewalld
systemctl stop firewalld

curl -sfL https://get.rke2.io | INSTALL_RKE2_TYPE="agent" sh -

systemctl enable rke2-agent.service

cat << EOF > /etc/sysctl.d/60-rke2-cis.conf
vm.panic_on_oom=0
vm.overcommit_memory=1
kernel.panic=10
kernel.panic_on_oops=1
EOF
sudo systemctl restart systemd-sysctl

mkdir -p /etc/rancher/rke2/
cat << EOF > /etc/rancher/rke2/config.yaml
server: https://<LB_IP>:9345
token: $SERVER_NODE_TOKEN
profile: "cis-1.5"
EOF

systemctl start rke2-agent.service
```

After bootstrap, the cluster PSPs are:
```shell
NAME                      PRIV    CAPS   SELINUX    RUNASUSER          FSGROUP     SUPGROUP    READONLYROOTFS   VOLUMES
global-restricted-psp     false          RunAsAny   MustRunAsNonRoot   MustRunAs   MustRunAs   false            configMap,emptyDir,projected,secret,downwardAPI,persistentVolumeClaim
system-unrestricted-psp   true    *      RunAsAny   RunAsAny           RunAsAny    RunAsAny    false            *
```

Fresh install LH onto longhorn-system namespace, the PSP `longhorn-psp` created:
```shell
NAME                      PRIV    CAPS        SELINUX    RUNASUSER          FSGROUP     SUPGROUP    READONLYROOTFS   VOLUMES
global-restricted-psp     false               RunAsAny   MustRunAsNonRoot   MustRunAs   MustRunAs   false            configMap,emptyDir,projected,secret,downwardAPI,persistentVolumeClaim
longhorn-psp              true    SYS_ADMIN   RunAsAny   RunAsAny           RunAsAny    RunAsAny    false            configMap,downwardAPI,emptyDir,secret,projected,hostPath
system-unrestricted-psp   true    *           RunAsAny   RunAsAny           RunAsAny    RunAsAny    false            *
```

If the user modified the `values.yaml` such as change ingress as:
```yaml
ingress:
  enabled: true
  annotations:
    foo: bar
```
and performs `helm upgrade longhorn longhorn/longhorn -f values.yaml -n longhorn-system`, the post-upgrade and uninstall Pod PSP is `global-restricted-psp`, not we expected `longhorn-psp`.

#### Root Cause
According to [PSP Policy Order](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#policy-order):
> 1. PodSecurityPolicies which allow the pod as-is, without changing defaults or mutating the pod, are preferred. The order of these non-mutating PodSecurityPolicies doesn't matter.
> 2. If the pod must be defaulted or mutated, the first PodSecurityPolicy (ordered by name) to allow the pod is selected.

since post-upgrade and uninstall Job did not set any `securityContext`, it causes it would falls into `global-restricted-psp` first since the pod must be defaulted or mutated, the first PodSecurityPolicy (ordered by name) to allow the pod is selected.

#### Solution
Configure the `privileged: true` to post-upgrade and uninstall Job, to match against https://github.com/longhorn/longhorn/blob/v1.1.0/chart/templates/daemonset-sa.yaml#L21-L22

#### Issues
https://github.com/longhorn/longhorn/issues/2292


innobead (approved):
LGTM. 

Waiting for merge unfreeze.

shuo-wu:
It seems that we need to update the uninstaller in `longhorn-manager` as well:
https://github.com/longhorn/longhorn-manager/blob/master/deploy/uninstall/uninstall.yaml

shuo-wu (approved):

innobead:
> It seems that we need to update the uninstaller in `longhorn-manager` as well:
> https://github.com/longhorn/longhorn-manager/blob/master/deploy/uninstall/uninstall.yaml

@jenting please check if we need to update the manifest in longhorn-manager. Thanks.

Issue #2387, opened 2021-03-19
[FEATURE] Add annotations support for resources in helm chart


b225ccc:
**Is your feature request related to a problem? Please describe.**
Adding custom annotations is a common use-case.  The Longhorn helm chart should provide users a way to do this.

**Describe the solution you'd like**
Support for annotations in helm chart.

My specific use-case is that I need to add annotations to the longhorn-manager pods to send the [longhorn prometheus metrics](https://longhorn.io/docs/1.1.0/monitoring/prometheus_and_grafana_setup/) to datadog.  ([Datadog reference](https://docs.datadoghq.com/agent/kubernetes/prometheus/))

**Describe alternatives you've considered**
I can patch the helm resource using Kustomize, but this is not a desirable long-term solution.

**Additional context**
n/a

jenting:
@b225ccc If I understand correctly, the annotation is added to the longhorn-manager Pods (i.e. longhorn-manager DaemonSet template), right?

khushboo-rancher:
Verified with Longhorn helm changes from https://github.com/longhorn/longhorn/pull/2391/files

@jenting The annotation is getting added to `spec->template->metadata->annotations`, shouldn't be this add to `metadata->annotations` ?

```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  annotations:
    deprecated.daemonset.template.generation: "1"
    longhorn.io/last-applied-tolerations: '[]'
  creationTimestamp: "2021-03-26T19:23:41Z"
  generation: 1
  labels:
    app: longhorn-manager
    app.kubernetes.io/instance: longhorn
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: longhorn
    app.kubernetes.io/version: v1.1.0
    helm.sh/chart: longhorn-1.1.0
  managedFields:
  - apiVersion: apps/v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .: {}
          f:deprecated.daemonset.template.generation: {}
        f:labels:
          .: {}
          f:app: {}
          f:app.kubernetes.io/instance: {}
          f:app.kubernetes.io/managed-by: {}
          f:app.kubernetes.io/name: {}
          f:app.kubernetes.io/version: {}
          f:helm.sh/chart: {}
      f:spec:
        f:revisionHistoryLimit: {}
        f:selector: {}
        f:template:
          f:metadata:
            f:annotations:
              .: {}
              f:key1: {}
[...]

PR #2391, opened 2021-03-21
Add annotations to longhorn manager DaemonSet

jenting:
#### Propose Change
Add annotations to longhorn-manager DaemonSet template to support add annotations to the longhorn-manager Pods.

#### Issues
https://github.com/longhorn/longhorn/issues/2387


innobead (approved):
LGTM. 

note: in this case, this setting should be fine to not include in questions.yml.

PR #2414, opened 2021-03-29
Updates wording to specify the annotations only added to longhorn-manager Pods

jenting:
Updates wording according to the comment https://github.com/longhorn/longhorn/issues/2387#issuecomment-808469012

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -164,5 +164,5 @@ enablePSP: true
 ## and its release namespace is not the `longhorn-system`
 namespaceOverride: ""
 
-# Annotations to add to the Longhorn DaemonSet's. Optional.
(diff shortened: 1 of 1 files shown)


chart/values.yaml
```diff
@@ -164,5 +164,5 @@ enablePSP: true
 ## and its release namespace is not the `longhorn-system`
 namespaceOverride: ""
 
-# Annotations to add to the Longhorn DaemonSet's. Optional.
+# Annotations to add to the Longhorn Manager DaemonSet's. Optional.
```
innobead, line 167:
```suggestion
# Annotations to add to the Longhorn Manager DaemonsSet Pods. Optional.
```

innobead (changes requested):
Minor change

innobead (approved):

PR #2419, opened 2021-03-29
Fixed typo in readme

iwankgb:
I think it should be "primitives" rather than "primatives".


innobead (approved):
LGTM. Thanks for the contribution!

PR #2443, opened 2021-04-05
chart: Update chart for the enhanced backing image feature

shuo-wu:
Longhorn #2295

### DIFF
--- a/chart/questions.yml
+++ b/chart/questions.yml
@@ -70,6 +70,18 @@ questions:
     type: string
     label: Longhorn Share Manager Image Tag
     group: "Longhorn Images Settings"
+  - variable: image.longhorn.backingImageManager.repository
+    default: longhornio/backing-image-manager
+    description: "Specify Longhorn Backing Image Manager Image Repository"
+    type: string
+    label: Longhorn Backing Image Manager Image Repository
+    group: "Longhorn Images Settings"
+  - variable: image.longhorn.backingImageManager.tag
+    default: v1_20210407
+    description: "Specify Longhorn Backing Image Manager Image Tag"
+    type: string
+    label: Longhorn Backing Image Manager Image Tag
+    group: "Longhorn Images Settings"
   - variable: image.csi.attacher.repository
     default: longhornio/csi-attacher
     description: "Specify CSI attacher image repository. Leave blank to autodetect."
--- a/chart/templates/crds.yaml
+++ b/chart/templates/crds.yaml
@@ -392,7 +392,7 @@ spec:
     listKind: BackingImageList
     plural: backingimages
     shortNames:
-      - lhbm
+      - lhbi
     singular: backingimage
   scope: Namespaced
   versions:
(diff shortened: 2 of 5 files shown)


shuo-wu:
Will update this PR when a tag like `v20210405` is created for `longhornio/backing-image-manager`

innobead:
@shuo-wu please rebase the code and remove the draft for ready-for-review. Thanks.

shuo-wu:
Updated. I have tested it. It's ready for review.

chart/values.yaml
```diff
@@ -22,6 +22,9 @@ image:
     shareManager:
       repository: longhornio/longhorn-share-manager
       tag: v1_20210406
+    backingImageManager:
+      repository: longhornio/backing-image-manager
+      tag: master
```
innobead, line 27: need to have a tag for the image instead of master

chart/questions.yml
```diff
@@ -70,6 +70,18 @@ questions:
...
     label: Longhorn Share Manager Image Tag
     group: "Longhorn Images Settings"
+  - variable: image.longhorn.backingImageManager.repository
+    default: longhornio/backing-image-manager
+    description: "Specify Longhorn Backing Image Manager Image Repository"
+    type: string
+    label: Longhorn Backing Image Manager Image Repository
+    group: "Longhorn Images Settings"
+  - variable: image.longhorn.backingImageManager.tag
+    default: master
```
innobead, line 80: same here

innobead (changes requested):
LGTM, just need to update the image tag.

innobead (approved):

chart/templates/crds.yaml
```diff
@@ -417,3 +417,54 @@ spec:
...
+      openAPIV3Schema:
+        type: object
+        properties:
+          spec:
+            x-kubernetes-preserve-unknown-fields: true
+          status:
+            x-kubernetes-preserve-unknown-fields: true
+    subresources:
+      status: {}
+    additionalPrinterColumns:
```
PhanLe1010, line 451: Do we want to include the  `.status.currentState` as other CRD?

PhanLe1010 (commented):
LGTM. 
Left a comment

chart/templates/crds.yaml
shuo-wu: It's nice to have. Let me add it with a new PR.

chart/templates/crds.yaml
```diff
@@ -417,3 +417,54 @@ spec:
...
+    - name: Image
+      type: string
+      description: The image the manager pod will use
+      jsonPath: .spec.image
```
PhanLe1010, line 455: A quick quesition, any reason that we prefer `.spec.image` instead of `.status.currentImage` as other CRDs?

shuo-wu: when I use a newer version of the backing image manager, we will directly create new CRs for the new version. In other words, we won't modify `spec.image` then we don't need `status.currentImage`. Same for InstanceManager and ShareManager.

PhanLe1010: I see. Thanks

Issue #2357, opened 2021-03-16
[BUG] RWX mount share ownership is being reset to user nobody


hendra-2020:
**Describe the bug**
I created a few RWX share volumes and mounted them on pods with the following securityContext:

        runAsUser: 0
        runAsGroup: 0
        fsGroup: 0
        fsGroupChangePolicy: "OnRootMismatch"

The first time pod is created, the share volume has correct ownership of root:root. After around 30min, the ownership of share and all files within it changed to nobody:4294967294. This always happens to all my RWX share across several pods. There's no extra error log in share-manager log.

**To Reproduce**
1. Create RWX volume
2. Mount on pod with root fsgroup
3. Wait around 30min
4. All ownership changed to nobody

**Expected behavior**
Ownership of share and all files within in should not changed to nobody.

**Log**
If applicable, add the Longhorn managers' log when the issue happens.
[ganesha.log](https://github.com/longhorn/longhorn/files/6148830/ganesha.log)

You can also attach a *Support Bundle* here. You can generate a Support Bundle using the link at the footer of the Longhorn UI.
[longhorn-support-bundle_799133ef-1954-4bd5-a1c5-bd40d5c61c5f_2021-03-16T12-24-29Z.zip](https://github.com/longhorn/longhorn/files/6148883/longhorn-support-bundle_799133ef-1954-4bd5-a1c5-bd40d5c61c5f_2021-03-16T12-24-29Z.zip)

**Environment:**
 - Longhorn version: 1.1.0
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: RKE2 1.20.4+rke2r1
   - Number of management node in the cluster: 1
   - Number of worker node in the cluster: 1
 - Node config
   - OS type and version: Ubuntu 20.04.2
   - CPU per node: 32
   - Memory per node: 252GB
   - Disk type(e.g. SSD/NVMe): SSD
   - Network bandwidth between the nodes: 10Gbps
 - Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal): Baremetal
 - Number of Longhorn volumes in the cluster: 11

**Additional context**

cclhsu:
Currently, Longhorn workload do not change the mount share ownership   Is possible that you have workload in cluster done it?  Can you share the manifests used to mount RWX volume?

joshimoo:
[...]

PR #2447, opened 2021-04-06
Update default share-manager image to v1_20210406

c3y1huang:
Include:
* Fix rwx mount ownership reset to nobody (https://github.com/longhorn/longhorn/issues/2357)


joshimoo (approved):
LGTM, the images and yaml will get updated before release by a sync of the longhorn-manager repo. But I don't think there should be an issue to manually update them before that sync.

PR #2448, opened 2021-04-07
chart: Show state in the print columns for backing image manager CR

shuo-wu:
Signed-off-by: Shuo Wu <shuo@rancher.com>

### DIFF
--- a/chart/templates/crds.yaml
+++ b/chart/templates/crds.yaml
@@ -449,6 +449,10 @@ spec:
     subresources:
       status: {}
     additionalPrinterColumns:
+    - name: State
+      type: string
+      description: The current state of the manager
+      jsonPath: .status.currentState
     - name: Image
       type: string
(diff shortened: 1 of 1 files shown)


innobead (approved):

chart/templates/crds.yaml
```diff
@@ -449,6 +449,10 @@ spec:
     subresources:
       status: {}
     additionalPrinterColumns:
+    - name: State
+      type: string
+      description: The manager current state
```
PhanLe1010, line 454: Since this is user facing, how about `The current state of the manager` or `The manager's current state` ?

shuo-wu: Sure. Updated

PhanLe1010: Thanks

PhanLe1010 (approved):

Issue #2199, opened 2021-01-20
[FEATURE] Set tolerations and node selectors for dynamically provisioned Longhorn workloads


shuo-wu:
**Is your feature request related to a problem? Please describe.**
It's required by the Windows cluster support.

https://github.com/rancher/rancher/issues/29220

cc https://github.com/rancher/charts/pull/994/files

meldafrawi:
Validation: **PASSED**

PR #2478, opened 2021-04-13
Allow user to set node selector for system managed components, manager, UI, driver deployer

PhanLe1010:
This feature allow user to deploy Longhorn only on a specific
set of nodes

Modification:
1. Remove the `toleration` and `priority class` settings from `question.yaml`. 
 User cannot set `toleration` for the user-deployed components  (manager, UI, Driver) from the `question.yaml` because Rancher chart [doesn't allow question of type `array`](https://github.com/rancher/rancher/issues/24918). Therefore, user must choose to `Edit as YAML` in Rancher Ui when install Longhorn . On the other hand, if user wants set `toleration` and `priority class` for Longhorn, they must set them for both system managed components and user-deployed components. Therefore, to avoid the mistakes, we remove everything from the `question.yaml` and user can edit the yaml to specify value for both system managed components and user-deployed components. 
2. Allow user to specify node selector for  system managed components and user-deployed components. Again, user must chose `edit as YAML` because the limitation of Rancher chart.

longhorn/longhorn#2199

### DIFF
--- a/chart/questions.yml
+++ b/chart/questions.yml
@@ -304,26 +304,6 @@ The available modes are:
     type: int
     min: 0
     default: 300
-  - variable: defaultSettings.taintToleration
-    label: Kubernetes Taint Toleration
-    description: "To dedicate nodes to store Longhorn replicas and reject other general workloads, set tolerations for Longhorn and add taints for the storage nodes.
-All Longhorn volumes should be detached before modifying toleration settings.
-We recommend setting tolerations during Longhorn deployment because the Longhorn system cannot be operated during the update.
-Multiple tolerations can be set here, and these tolerations are separated by semicolon. For example:
-* `key1=value1:NoSchedule; key2:NoExecute`
-* `:` this toleration tolerates everything because an empty key with operator `Exists` matches all keys, values and effects
-* `key1=value1:`  this toleration has empty effect. It matches all effects with key `key1`
-Because `kubernetes.io` is used as the key of all Kubernetes default tolerations, it should not be used in the toleration settings.
-WARNING: DO NOT CHANGE THIS SETTING WITH ATTACHED VOLUMES!"
-    group: "Longhorn Default Settings"
-    type: string
-    default: ""
-  - variable: defaultSettings.priorityClass
-    label: Priority Class
-    description: "The name of the Priority Class to set on the Longhorn workloads. This can help prevent Longhorn workloads from being evicted under Node Pressure. WARNING: DO NOT CHANGE THIS SETTING WITH ATTACHED VOLUMES."
-    group: "Longhorn Default Settings"
-    type: string
-    default: ""
   - variable: defaultSettings.autoSalvage
     label: Automatic salvage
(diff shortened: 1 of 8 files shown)


innobead (approved):
LGTM

PhanLe1010:
@shuo-wu  How do you think about the removal of the `toleration` and `priority class` settings from question.yaml ?
Right now, the user experience is not great when user instal Longhorn using Rancher UI and want to set  `toleration `,  `node selector`.  I have asked Rancher team at [here](https://rancher.slack.com/archives/CFL25P8RH/p1618356527165800). Waiting for their feedback

shuo-wu:
@PhanLe1010 
Now tolerations, priorityclass, and node selector settings are applied to system deployed workloads only, right? Maybe we can directly convert the array/map to a string and applied them to the settings as well.

e.g., for the toleration setting, we can use the following:
```
taint-toleration: {{ .Values.longhornManager.tolerations | toJson }}
```
Then the rendered result is:
```
taint-toleration: [{"effect":"NoSchedule","key":"key","operator":"Equal","value":"value"}]
```

Here is the doc:
https://helm.sh/docs/chart_template_guide/function_list/#type-conversion-functions

Notice that the tolerations/priorityclass/nodeselector are set separately for longhornManager/longhornDriver.... Why cannot use a single setting for all workloads instead?

innobead:
> Notice that the tolerations/priorityclass/nodeselector are set separately for longhornManager/longhornDriver.... Why cannot use a single setting for all workloads instead?

This is a good point, I missed this part. 

One more question: do we allow users to deploy longhorn workloads and longhorn managed workloads separated in different nodes? I don't think so, so probably we just need to make sure the node selector for longhorn both workloads managed via the default setting.

PhanLe1010:
@shuo-wu 
1. for the system managed components. User will specified a string value for `taint-toleration`, `system-managed-components-node-selector`, and `priority-class` in the `default-setting` section. The string values will have similar structure as seen in UI setting page. For `taint-toleration`, the value should look like `key=value:NoExecute` instead of `[{"effect":"NoSchedule","key":"key","operator":"Equal","value":"value"}]`
2. We don't want to use a single tolerations/priorityclass/nodeselector setting for all longhornManager/longhornDriver/UI because they doesn't always have to be the same. For example, it makes sense that users have high priority class for longhorn manager but low priority class for UI. Same goes for toleration and node selector, user may want to run Longhorn UI on a specific node that is different than where Longhorn manager is running. In general, I think we should let user customize the components that they controller (manager, driver, UI) however they want. We should only controller the system managed components

@innobead 
>so probably we just need to make sure the node selector for longhorn both workloads managed via the default setting.

This is our previous design and it create a headache for user who is using GitOps tools: Longhorn continuously modified the  user-deployed components and GitOps try to revert it back to the initial state. You can see more detailed in this user's feedback https://github.com/longhorn/longhorn/issues/2120#issuecomment-748473152

innobead:
Thanks @PhanLe1010 explanation. Let's keep it for now and see user feedback.

PR #2489, opened 2021-04-14
Update chart icon

jenting:
#### Propose Change
Use .png as a chart icon to make it visible on the rancher server.

#### Demo
My forked longhorn charts repo
image

Rancher server icon
image

#### Issues
https://github.com/rancher/charts/issues/1058

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -25,4 +25,4 @@ maintainers:
   email: maintainers@longhorn.io
 - name: Sheng Yang
   email: sheng@yasker.org
-icon: https://raw.githubusercontent.com/cncf/artwork/master/projects/longhorn/icon/color/longhorn-icon-color.svg?sanitize=true
+icon: https://raw.githubusercontent.com/cncf/artwork/master/projects/longhorn/icon/color/longhorn-icon-color.png


innobead (approved):

khushboo-rancher:
Should the change be merged for Rancher charts also with this id?

innobead:
> Should the change be merged for Rancher charts also with this id?

@jenting do you have a PR for the existing rancher chart update?

jenting:
I thought it should be updated along with v1.1.1 chart release.

jenting:
Since the rancher chart https://github.com/rancher/charts/tree/dev-v2.5-source depends on longhorn/charts.

If we want to make a change to the existing released v1.1.0 chart, this means longhorn has to release a new v1.1.0 chart release first and then update the rancher chart.
The v1.1.1 chart will be released soon, I'd prefer to not update the icon link in existed v1.1.0 chart.

innobead:
> Since the rancher chart https://github.com/rancher/charts/tree/dev-v2.5-source depends on longhorn/charts.
> 
> If we want to make a change to the existing released v1.1.0 chart, this means longhorn has to release a new v1.1.0 chart release first and then update the rancher chart.
> The v1.1.1 chart will be released soon, I'd prefer to not update the icon link in existed v1.1.0 chart.

agree.

innobead:
cc @cclhsu

PR #2504, opened 2021-04-16
[Chart Helm] Allow to specify custom labels/annotations for managers pods

Lunik:
# Resume
The goal of this PR is to allow users to set custom labels/annotations on longhorn managers pods.

# Use case
According to [longhorn prometheus integration documentation](https://longhorn.io/docs/1.1.0/monitoring/prometheus_and_grafana_setup/), longhorn managers expose `/metrics` endpoint that can be scraped by Prometheus.

# Usage

Specifying thoses options in chart values should enable Prometheus auto-discovery : 
```yaml
managers:
  # Additional pod annotations (e.g. for mesh injection or prometheus scraping)
  podAnnotations:
    prometheus.io/path: "/metrics"
    prometheus.io/port: "9500"
    prometheus.io/scheme: "http"
    prometheus.io/scrape: "true"
```

# References

- [Prometheus service auto discovery through pod annotations](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config)
- [Prometheus Chart auto discovery config](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/values.yaml#L1575)

### DIFF
--- a/chart/templates/daemonset-sa.yaml
+++ b/chart/templates/daemonset-sa.yaml
@@ -11,8 +11,15 @@ spec:
       app: longhorn-manager
   template:
     metadata:
+      annotations:
+      {{- with .Values.managers.podAnnotations }}
+      {{- toYaml . | nindent 8 }}
+      {{- end }}
       labels: {{- include "longhorn.labels" . | nindent 8 }}
         app: longhorn-manager
+      {{- with .Values.managers.podLabels }}
+      {{- toYaml . | nindent 8 }}
+      {{- end }}
     spec:
       containers:
       - name: longhorn-manager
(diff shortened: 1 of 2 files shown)


jenting:
@Lunik Thank you for the pull request.
However, we addressed it in [PR](https://github.com/longhorn/longhorn/pull/2391) already (on v1.1.1 branch) and we'll include it in the recently v1.1.1 release.

The only difference compares to the PR is that you add the _label_ also, but from my understanding, the label is not necessary for the Prometheus auto-discovery. Right?

Lunik:
I didn't see that one !
This PR is a duplicate then.

Labels are not required for prometheus auto discovery.

/close

PR #2548, opened 2021-04-29
Add backing image manager info & build status

shuo-wu:


innobead (changes requested):
Let's do this aginst 1.2.0 branch instead of master.

shuo-wu:
The merge base is changed. I think it is required for 1.1.2 as well.

innobead:
> The merge base is changed. I think it is required for 1.1.2 as well.

Yes, correct.

innobead (approved):

Issue #587, opened 2019-06-03
[FEATURE] Support automatic rebalance the replicas


yasker:
Add a global setting to enable `Automatic replica rebalance`.

In the case of there is a limited or no redundancy for the volume (e.g. due to node offline temporarily), if this option is enabled, we can automatically check if a new replica can be scheduled on the new node to restore the redundancy. Notice this can have some system level impact since it will result in more network traffic in the system.

Should be implemented with `Maximum Concurrent Replica Rebuilding Count` option. 

from: https://github.com/rancher/longhorn/issues/570

ironsalsa:
Adding my vote here for this feature - it would be great to not need manual intervention after node failures.

michaelpearce-gain:
Is there any update on this? When would it possibly be available. Whilst hard affinity is ok, in AZ failure scenarios it can hurt, you obviously want replicas in separate AZ's so that should a DC fail you know data is in another AZ, but once that occurs, you kind of want your operations to continue in the remaining, including able to create new PVs, as well as have resilience from node loss whilst the other AZ is out,, and simply once the other AZ is back, you'd want to re-balance automatically data so your back to having safety from a future AZ failure, with no manual steps.

joshimoo:
@michaelpearce-gain we do have the option for sofft anti affinity on the zone.
https://longhorn.io/docs/1.1.0/volumes-and-nodes/scheduling/#the-node-and-zone-selection-stage

This would allow a rebuild to happen in Zone `A` when Zone `B` fails.
But we currently don't have auto rebalancing after Zone `B` comes back.

As a workaround a user can manually evict a single replica from Zone `A` once Zone `B` is available again, this will make Longhorn respect the soft anti affinity of the zones and prioritize the rebuild into the now available Zone `B`.

michaelpearce-gain:
@joshimoo  we would need it that theres 100% guarantee that after zone B comes back that replicas will be there again, so should zone A go down weeks later, youre not left with both replicas in zone A. Soft affinity as i understand it is a best effort, not a guarentee, like hard affinity gives as i understand

c3y1huang:
Discussed with @innobead and will remove `require/manual-test-plan` tag. We can refer to e2e since already covered the cases.

innobead:
> Discussed with @innobead and will remove `require/manual-test-plan` tag. We can refer to e2e since already covered the cases.

cc @khushboo-rancher @meldafrawi @kaxing

yasker:
[...]

Issue #685, opened 2019-08-17
[FEATURE] Clean up orphaned unused volume replicas after node failure


vistalba:
I run into a node failure agina... because of a connectivity issue (not related to longhorn).

Longhorn rebuilds the replica on the two remaining nodes.
I then manually rebalanced it after third node was avaiable again.

After all rebuilds I noticed a big bump at disk-space on node3 (that one that has failed before).

```
drwxr-xr-x  2 root root 4.0K Aug 17 04:45 k8s-vol1-data-1bb645c8
drwxr-xr-x  2 root root 4.0K Aug 16 21:18 k8s-vol2-data-ea61e796
drwxr-xr-x  2 root root 4.0K Aug 17 04:30 k8s-vol2-data-fc6ad090
drwxr-xr-x  2 root root 4.0K Aug 16 21:05 k8s-vol3-e72fde22
drwxr-xr-x  2 root root 4.0K Aug 17 07:35 k8s-vol3-ef30aa7f
drwxr-xr-x  2 root root 4.0K Aug 17 04:30 k8s-vol4-data-5d17716b
drwxr-xr-x  2 root root 4.0K Aug 17 04:15 k8s-vol5-data-150d08a3
drwxr-xr-x  2 root root 4.0K Aug 17 04:00 k8s-vol6-data-72eb9a67
drwxr-xr-x  2 root root 4.0K Aug 17 03:45 k8s-vol8-data-f8f3c7be
drwxr-xr-x  2 root root 4.0K Aug 16 23:05 pvc-cd235079-b057-11e9-a7a3-960000289ce9-b4dd5aab
```

As you can see, longhorn leave old folders on the node which are consuming the space. So I just deleted the old from `aug 16`. Longhorn should do that automaticly. Longhorn did not show any failed replica in any of the above volumes in the UI.

yasker:
Currently, the faulted replica will be cleaned up after 30 minutes, and if the node isn't available for clean up, it will be simply skipped since it cannot wait for the node forever. Though it seems the default `staleReplicaTimeout` is too low (which is set for each volume now) and users may want to keep it a bit longer, and it also means if the node came back in time, the replicas would be cleaned up properly.

If we're going to integrate a tool to help clean up the disk space, we must be more careful to make sure it's indeed leftover of deleted replicas, since we don't have any information regarding this directory may be useful or not. I would want this tool to be manually triggered instead of automatically triggered then.

vistalba:
[...]

PR #2581, opened 2021-05-10
Add LEP - automatic rebalance replica

c3y1huang:
https://github.com/longhorn/longhorn/issues/587

### DIFF
--- a/enhancements/20210510-automatic-rebalance-replica.md
+++ b/enhancements/20210510-automatic-rebalance-replica.md
@@ -1,0 +1,396 @@
+# Automatic Replica Rebalance
+
+Have a setting for global and volume-specific settings to enable/disable automatic distribution of the off-balanced replicas when a node is newly available to the cluster.
+
+## Summary
+
+When nodes are offline with setting `Replica Zone Level Soft Anti-Affinity` enabled, or `Replica Node Level Soft Anti-Affinity` enabled, the replicas could be duplicated and retained on the same zone/node if the remain cluster zone/node number is less than the replica number.
+
+Currently, the user needs to be aware when nodes are offline and a new node is added back to the cluster to manually delete the replicas to rebalance onto the newly available node.
+
+This enhancement proposes is to add a new Longhorn global setting `Replica Auto Balance` to enable detection and deletion of the unbalanced replica to achieve automatic rebalancing. And this enhancement also proposes to add a new setting in volume spec `replicaAutoBalance` to enable/disable automatic replica rebalancing for individual volume.
+
+### Related Issues
+
+- https://github.com/longhorn/longhorn/issues/587
+- https://github.com/longhorn/longhorn/issues/570
+
+## Motivation
+
+### Goals
+
+- Support global replica automatic balancing.
+- Support individual volume replica automatic balancing.
+
+### Non-goals [optional]
+[Clean up old Data after node failure](https://github.com/longhorn/longhorn/issues/685#issuecomment-817121946).
+
+## Proposal
+
+Add a new global setting `Replica Auto Balance` to enable replica automatic balancing when a node is newly available to the cluster.
+
+Add a new setting in volume spec `replicaAutoBalance` to enable/disable automatic replica rebalancing for individual volume.
+
+### User Stories
+
+Before this enhancement, the user needs to check and manually delete replicas to have deleted replicas reschedule to newly available nodes. If the user does not take action, this could lead to no redundancy volumes.
+
+After this enhancement, the user does not need to worry about manually balancing replicas when there are newly available nodes.
+
+#### Story 1 - node temporary offline
+As a system administrator,
+
+When a cluster node is offline and comes back online after some time, I want Longhorn to automatically detect and reschedule replicas evenly to all nodes.
+
+So I do not have to worry and standby 24/7 to check and manually rebalance replicas when there are newly available nodes.
+
+#### Story 2 - add new node
+
+As a system administrator,
+
+When a cluster node is offline and a new node is added, I want Longhorn to automatically detect and reschedule replicas evenly to all nodes
+
+So I do not have to worry and standby 24/7 to check and manually rebalance replicas when there are newly available nodes.
+
+#### Story 3 - zone temporary offline
+
+As a system administrator,
+
+When a cluster zone is offline and comes back online after some time, I want Longhorn to automatically detect and reschedule replicas evenly across zones.
+
+So I do not have to worry and standby 24/7 to check and manually rebalance replicas when there are newly available nodes.
+
+#### Story 4 - replica automatic rebalance for individual volume
+
+As a system administrator,
+
+When a cluster node is offline and a new node is added, I want Longhorn to automatically detect and reschedule replicas evenly to some volumes only.
+
+So I do not have to worry and standby 24/7 to check and manually rebalance replicas when there are newly available nodes.
+
+#### Story 5 - replica automatic rebalance for minimal redundancy
+
+As a system administrator,
+
+When multiple cluster node are offline and a new node is added, I want have an option to have Longhorn to automatically detect and reschedule only 1 replica to the new node to achieve minimal redundancy.
+
+So I do not have to worry about over resource consumption during rebalancing and I do not have to worry and standby 24/7 to check and manually rebalance replicas when there are newly available nodes.
+
+#### Story 6 - replica automatic rebalance for even redundancy
+
+As a system administrator,
+
+When multiple cluster node are offline and a new node is added, I want have an option to have Longhorn to automatically detect and reschedule replica to the new node to achieve even replica number redundancy.
+
+So I do not end up with uneven number of replicas across cluster nodes and I do not have to worry and standby 24/7 to check and manually rebalance replicas when there are newly available nodes.
+
+### User Experience In Detail
+
+#### Story 1 - Replica Node Level Unbalanced
+
+With an example of 3 nodes in cluster and default of 3 replicas volume:
+1. When the user enables `replica-soft-anti-affinity` and deploys PVC and Pod on node-1. The replica is distributed evenly to all 3 nodes at this point.
+2. In case of node-2 is offline, the user will see replica was on node-2 gets re-scheduled to node-1 or node-3. And also a warning icon and note will appear on UI `Limited node redundancy: at least one healthy replica is running at the same node as another`.
+
+**Before enhancement**
+3. Bring node-2 online, no change will be done automatically by Longhorn. The user will still see the same warning on UI.
+4. To rebalance replicas, the user needs to find and delete the duplicated replica to trigger the schedule onto node-2. 
+
+**After enhancement**
+3. Longhorn automatically detects and deletes the duplicated replica so the replica will be scheduled onto node-2. User will see the duplicated replica get rescheduled back to node-2 and UI will see volume `Healthy` with the note `Limited node redundancy: at least one healthy replica is running at the same node as another` removed.
+
+#### Story 2 - Replica Zone Level Unbalanced
+
+With an example of cluster set for 2 zones and default of 2 replicas volume:
+1. When the user enables `replica-soft-anti-affinity` and `replica-zone-soft-anti-affinity` and deploys PVC and deployment. The replica is distributed evenly to all 2 zones at this point.
+2. In case of zone-2 is offline, the user will see replica was on zone-2 gets re-scheduled to zone-1.
+
+**Before enhancement**
+3. Bring zone-2 online, no change will be done automatically by Longhorn. The user will still see replicas all running in zone-1.
+4. To rebalance replicas, the user needs to find and delete the duplicated replica to trigger the schedule to zone-2.
+
+**After enhancement**
+3. Longhorn automatically detects and deletes the duplicated replica so the replica will be scheduled to zone-2. The user will see the duplicated replica get rescheduled back to zone-2.
+
+### API changes
+- The new global setting `Replica Auto Balance` will use the same /v1/settings API.
+- When creating a new volume, the body of the request sent to /v1/volumes has a new field `replicaAutoBalance` set to `ignored`, `disabled`, `least-effort`, or `best-effort`.
+- Implement a new API for users to update `replicaAutoBalance` setting for individual volume. The new API could be /v1/volumes/<VOLUME_NAME>?action=updateReplicaAutoBalance. This API expects the request's body to have the form {replicaAutoBalance:<options>}.
+
+## Design
+
+### Implementation Overview
+
+#### longhorn-manager
+- Add new global setting `Replica Auto Balance`.
+  - The setting is `string`.
+  - Available values are: `disabled`, `least-effort`, `best-effort`.
+      - `disabled`: no replica auto balance will be done.
+      - `least-effort`: replica will be balanced to achieve minimal redundancy. For example, after adding node-2, a volume with 4 off-balanced replicas will only rebalance 1 replicas.
+         ```
+         node-1
+         +-- replica-a
+         +-- replica-b
+         +-- replica-c
+         node-2
+         +-- replica-d
+         ```
+      - `best-effort`: replica will be balanced to achieve similar number of replicas redundancy. For example, after adding node-2, a volume with 4 off-balanced replicas will rebalance 2 replicas.
+         ```
+         node-1
+         +-- replica-a
+         +-- replica-b
+         node-2
+         +-- replica-c
+         +-- replica-d
+         ```
+  - The default value is `disabled`.
+
+- Add new volume spec `replicaAutoBalance`.
+  - Available values are: `ignored`, `disabled`, `least-effort`, `best-effort`.
+      - `ignored`: This will adopt to the value from global setting.
+      - `disabled`: Same as global setting value `disabled`.
+      - `least-effort`: Same as global setting value `least-effort`.
+      - `best-effort`: Same as global setting value `best-effort`.
+  - The default value is `ignored`.
+
+- In Volume Controller `syncVolume` -> `ReconcileEngineReplicaState` -> `replenishReplicas`, calculate and add number of replicas to be rebalanced to `replenishCount`.
+  > The logic ignores all `soft-anti-affinity` settings. This will always try to achieve zone balance then node balance. And creating for replicas will leave for ReplicaScheduler to determine for the canidates.
+  1. Skip volume replica rebalance when volume spec `replicaAutoBalance` is `disabled`.
+  2. Skip if volume `Robustness` is not `healthy`.
+  3. For `least-effort`, try to get the replica rebalance count.
+      1. For `zone` duplicates, get the replenish number.
+            1. List all the occupied node zones with volume replicas running.
+                  - The zone is balanced when this is equal to volume spec `NumberOfReplicas`.
+            2. List all available and schedulabled nodes in non-occupied zones.
+                  - The zone is balanced when no available nodes are found.
+            3. Get the number of replicas off-balanced:
+                  - number of replicas in volume spec - number of occupied node zones.
+            4. Return the number to replenish.
+                  - number of non-occupied zones if less than off-balanced, or 
+                  - number off-balanced.
+      2. For `node` duplicates, try to balance `zone` first. Get the replica replenish number.
+            1. List all occupied node IDs with volume replicas running.
+                  - The node is balanced when this is equal to volume spec `NumberOfReplicas`.
+            2. List all available and schedulable nodes.
+                  - The nodes is balanced when number of occupied nodes equal to the number of nodes. This is to determine if balanced when the number of nodes is less then the volume spec `NumberOfReplicas`.
+            3. Get the number of replicas off-balanced:
+                  - number of replicas in volume spec - number of occupied node IDs.
+            4. Return the number to replenish.
+                  - number of non-occupied nodes if less than off-balanced, or
+                  - number off-balanced.
+  4. For `best-effort`, try `least-effort` first to achieve minimal redundancy, then,
+      1. Try to get zone duplicates mapped by zoneID, continue to find duplicates on nodes if no duplicated found here.
+      2. Try to get node duplicates mapped by nodeID.
+      3. Return number to replenish when maximum replica names in duplicates mapping is 2 greater than the minimum replica names in duplicates mapping.
+  5. Add the number to rebalance to `replenishCount` in `replenishReplicas`.
+
+- Cleanup extra replicas for auto-balance in `cleanupExtraHealthyReplicas`.
+  1. Get replica names.
+      - For `best-effort`, use the replica names from duplicates in the most duplicated zones/nodes. 
+      - For `least-effort`, use the replicas names from `getPreferredReplicaCandidatesForDeletion`.
+  3. Delete one replicas from the replica names.
+
+
+### Test plan
+
+#### Integration tests - test_replica_auto_balance_node_least_effort
+
+Scenario: replica auto-balance nodes with `least_effort`.
+
+      Given set `replica-soft-anti-affinity` to `true`.
+      And set `replica-auto-balance` to `least_effort`.
+      And create a volume with 6 replicas.
+      And attach the volume to node-1.
+      And wait for the volume to be healthy.
+      And write some data to the volume.
+      And disable scheduling for node-2.
+          disable scheduling for node-3.
+      And And count replicas running on each nodes.
+      And 6 replicas running on node-1.
+          0 replicas running on node-2.
+          0 replicas running on node-3.
+
+      When enable scheduling for node-2.
+      Then count replicas running on each nodes.
+      And node-1 replica count != node-2 replica count.
+          node-2 replica count != 0.
+          node-3 replica count == 0.
+      And sleep 10 seconds, to ensure no addition scheduling is happening.
+      And count replicas running on each nodes.
+      And number of replicas running should be the same.
+
+      When enable scheduling for node-3.
+      And count replicas running on each nodes.
+      And node-1 replica count != node-3 replica count.
+          node-2 replica count != 0.
+          node-3 replica count != 0.
+      And sleep 10 seconds, to ensure no addition scheduling is happening.
+      And count replicas running on each nodes.
+      And number of replicas running should be the same.
+
+      When check the volume data.
+      And volume data should be the same as written.
+
+#### Integration tests - test_replica_auto_balance_node_best_effort
+
+Scenario: replica auto-balance nodes with `best_effort`.
+
+      Given set `replica-soft-anti-affinity` to `true`.
+      And set `replica-auto-balance` to `best_effort`.
+      And create a volume with 6 replicas.
+      And attach the volume to node-1.
+      And wait for the volume to be healthy.
+      And write some data to the volume.
+      And disable scheduling for node-2.
+          disable scheduling for node-3.
+      And And count replicas running on each node.
+      And 6 replicas running on node-1.
+          0 replicas running on node-2.
+          0 replicas running on node-3.
+
+      When enable scheduling for node-2.
+      And count replicas running on each node.
+      Then 3 replicas running on node-1.
+           3 replicas running on node-2.
+           0 replicas running on node-3.
+      And sleep 10 seconds, to ensure no addition scheduling is  happening.
+      And count replicas running on each node.
+      And 3 replicas running on node-1.
+          3 replicas running on node-2.
+          0 replicas running on node-3.
+
+      When enable scheduling for node-3.
+      And count replicas running on each node.
+      Then 2 replicas running on node-1.
+           2 replicas running on node-2.
+           2 replicas running on node-3.
+      And sleep 10 seconds, to ensure no addition scheduling is  happening.
+      And count replicas running on each node.
+      And 2 replicas running on node-1.
+          2 replicas running on node-2.
+          2 replicas running on node-3.
+
+      When check the volume data.
+      And volume data should be the same as written.
+
+#### Integration tests - test_replica_auto_balance_disabled_volume_spec_enabled
+
+Scenario: replica should auto-balance individual volume when global setting `replica-auto-balance` is `disabled` and volume spec `replicaAutoBalance` is `least_effort`.
+
+      Given set `replica-soft-anti-affinity` to `true`.
+      And set `replica-auto-balance` to `least_effort`.
+      And disable scheduling for node-2.
+          disable scheduling for node-3.
+      And create volume-1 with 3 replicas.
+          create volume-2 with 3 replicas.
+      And set volume-2 spec `replicaAutoBalance` to `least-effort`.
+      And attach volume-1 to node-1.
+          attach volume-2 to node-1.
+      And wait for volume-1 to be healthy.
+          wait for volume-2 to be healthy.
+      And volume-1 replicas should be running on node-1.
+          volume-2 replicas should be running on node-1.
+      And write some data to volume-1.
+          write some data to volume-2.
+
+      When enable scheduling for node-2.
+           enable scheduling for node-3.
+      And count replicas running on each nodes for volume-1.
+          count replicas running on each nodes for volume-2.
+
+      Then volume-1 replicas should be running on node-1.
+      And volume-1 should have 3 replicas running.
+      And volume-2 replicas should be running on node-1, node-2, node-3.
+      And volume-2 should have 3 replicas running.
+      And volume-1 data should be the same as written.
+      And volume-2 data should be the same as written.
+
+#### Integration tests - test_replica_auto_balance_zone_least_effort
+
+Scenario: replica auto-balance zones with least-effort.
+
+      Given set `replica-soft-anti-affinity` to `true`.
+      And set `replica-zone-soft-anti-affinity` to `true`.
+      And set volume spec `replicaAutoBalance` to `least-effort`.
+      And set node-1 to zone-1.
+          set node-2 to zone-2.
+          set node-3 to zone-3.
+      And disable scheduling for node-2.
+          disable scheduling for node-3.
+      And create a volume with 6 replicas.
+      And attach the volume to node-1.
+      And 6 replicas running in zone-1.
+          0 replicas running in zone-2.
+          0 replicas running in zone-3.
+
+      When enable scheduling for node-2.
+      And count replicas running on each node.
+      And zone-1 replica count != zone-2 replica count.
+          zone-2 replica count != 0.
+          zone-3 replica count == 0.
+
+      When enable scheduling for node-3.
+      And count replicas running on each node.
+      And zone-1 replica count != zone-3 replica count.
+          zone-2 replica count != 0.
+          zone-3 replica count != 0.
+
+#### Integration tests - test_replica_auto_balance_zone_best_effort
+
+Scenario: replica auto-balance zones with best-effort.
+
+      Given set `replica-soft-anti-affinity` to `true`.
+      And set `replica-zone-soft-anti-affinity` to `true`.
+      And set volume spec `replicaAutoBalance` to `best-effort`.
+      And set node-1 to zone-1.
+          set node-2 to zone-2.
+          set node-3 to zone-3.
+      And disable scheduling for node-2.
+          disable scheduling for node-3.
+      And create a volume with 6 replicas.
+      And attach the volume to node-1.
+      And 6 replicas running in zone-1.
+          0 replicas running in zone-2.
+          0 replicas running in zone-3.
+
+      When enable scheduling for node-2.
+      And count replicas running on each node.
+      And 3 replicas running in zone-1.
+          3 replicas running in zone-2.
+          0 replicas running in zone-3.
+
+      When enable scheduling for node-3.
+      And count replicas running on each node.
+      And 2 replicas running in zone-1.
+          2 replicas running in zone-2.
+          2 replicas running in zone-3.
+
+#### Integration tests - test_replica_auto_balance_node_duplicates_in_multiple_zones
+
+Scenario: replica auto-balance to nodes with duplicated replicas in the zone.
+
+      Given set `replica-soft-anti-affinity` to `true`.
+      And set `replica-zone-soft-anti-affinity` to `true`.
+      And set volume spec `replicaAutoBalance` to `least-effort`.
+      And set node-1 to zone-1.
+          set node-2 to zone-2.
+      And disable scheduling for node-3.
+      And create a volume with 3 replicas.
+      And attach the volume to node-1.
+      And zone-1 and zone-2 should contain 3 replica in total.
+
+      When set node-3 to the zone with duplicated replicas.
+      And enable scheduling for node-3.
+      Then count replicas running on each node.
+      And 1 replica running on node-1
+          1 replica running on node-2
+          1 replica running on node-3.
+      And count replicas running in each zone.
+      And total of 3 replicas running in zone-1 and zone-2.
+
+### Upgrade strategy
+There is no upgrade needed.
+
+## Note [optional]
+`None`


enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,186 @@
+# Automatic Replica Rebalance
+
+Have a Longhorn setting to enable automatic distribution of the off-balanced replicas when a new node is added to the cluster.
```
innobead, line 3:
```suggestion
Have a Longhorn setting to enable automatic distribution of the off-balanced replicas when a newly available node to the cluster
```

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Summary
+
+When nodes are offline with setting `Replica Node Level Soft Anti-Affinity` enabled, the replicas could be duplicated on the same node if the remain cluster node number is less than the replica number.
+
+Currently, the user needs to be aware when nodes are offline and a new node is added back to the cluster to manually delete the replicas to rebalance onto the new node.
```
innobead, line 9:
```suggestion
Currently, the user needs to be aware when nodes are offline and a new node is added back to the cluster to manually delete the replicas to rebalance onto the newly available node.
```

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,186 @@
...
+
+This enhancement proposes to add a new Longhorn setting `Replica Node Level Auto Balancing` to enable detection and deletion of the unbalanced replica to achieve automatic rebalancing.
```
innobead, line 11:
```suggestion
This enhancement proposes is to add a new Longhorn setting `Replica Node Level Auto Balancing` to enable detection and deletion of the unbalanced replica to achieve automatic rebalancing.
```

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,186 @@
...
+### Related Issues
+
+- https://github.com/longhorn/longhorn/issues/587
+- https://github.com/longhorn/longhorn/issues/570
+
+## Motivation
+
+### Goals
+
+Have a Support replica auto balance.
```
innobead, line 22:
```suggestion
Support replica automatic balancing.
```

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,186 @@
...
+
+### Non-goals [optional]
+`n/a`
+
+## Proposal
+
+Add new setting `Replica Node Level Auto Balancing` to trigger replica deletion when a new node is available to cluster.
```
innobead, line 29:
```suggestion
Add new setting `Replica Node Level Auto Balancing` to enable replica automatic balancing when a newly available node to the cluster.
```

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,186 @@
...
+
+### User Stories
+
+Before this enhancement user needs to check and manually delete replicas to have replicas schedule to the new node. If the user does not take action, this could lead to no redundancy volumes.
```
innobead, line 33:
```suggestion
Before this enhancement, the user needs to check and manually delete replicas to have deleted replicas reschedule to newly available nodes. If the user does not take action, this could lead to no redundancy volumes.
```

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,186 @@
...
+
+After this enhancement user does not need to worry about manually balancing replicas when nodes are offline.
```
innobead, line 35:
```suggestion
After this enhancement, the user does not need to worry about manually balancing replicas when there are newly available nodes.
```

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,186 @@
...
+
+#### Story 1 - node temporary offline
+As a system administrator, 
+
+When a cluster node is offline and comes back online after some time, I want Longhorn to automatically detect and schedule replicas evenly to all nodes.
```
innobead, line 40:
```suggestion
When a cluster node is offline and comes back online after some time, I want Longhorn to automatically detect and reschedule replicas evenly to all nodes.
```

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,186 @@
...
+
+So I do not have to worry and standby 24/7 to check and manually rebalance replicas when nodes are down.
```
innobead, line 42:
```suggestion
So I do not have to worry and standby 24/7 to check and manually rebalance replicas when there are newly available nodes.
```

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,186 @@
...
+
+#### Story 2 - add new node
+
+As a system administrator, 
+
+When a cluster node is offline and a new cluster is added, I want Longhorn to automatically detect and schedule replicas evenly to all nodes
```
innobead, line 48:
```suggestion
When a cluster node is offline and a new cluster is added, I want Longhorn to automatically detect and reschedule replicas evenly to all nodes
```

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,186 @@
...
+
+So I do not have to worry and standby 24/7 to check and manually rebalance replicas when nodes are down.
```
innobead, line 50:
```suggestion
So I do not have to worry and standby 24/7 to check and manually rebalance replicas when there are newly available nodes.
```

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,186 @@
...
+
+**Before enhancement**
+3. Bring node-2 online, no change will be done automatically by Longhorn. The user will still see the same warning on UI.
+4. To rebalance replicas, the user needs to find and delete the duplicated replica to trigger the schedule onto node-2. 
+
+**After enhancement**
+3. Longhorn automatically detects and deletes the duplicated replica so the replica will be scheduled onto node-2. User will see the duplicated replica get rescheduled back to node-2 and UI will see volume `Healthy` with the note `Limited node redundancy: at least one healthy replica is running at the same node as another` removed.
+
+### API changes
+There is no API change
```
innobead, line 66: Actually, we have the contract change for the new setting.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Design
+
+### Implementation Overview
+
+#### longhorn-manager
+1. Add new setting `Replica Node Level Auto Balancing`.
```
innobead, line 73: What's the default value? If enabled, any side effects or notice users need to be aware of.

innobead (changes requested):
LGTM, some comments waiting to resolve.

Another concern is when the replicas get deleted for automatic rebalancing, does it mean the snapshots of the deleted replicas will be left over in the original nodes?

If yes, this probably will cause some extra resource usage overhead, probably we need to tackle it in another ticket. I remember users have some requests for that to clean up the leftover snapshots.

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: Fixed.

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: Fixed.

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: Fixed.

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: Fixed.

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: Fixed.

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: Fixed.

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: Fixed.

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: Fixed.

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: Thank you for point out. I think we can use /v1/settings, updated the PR.

c3y1huang:
> I previously left a comment on how to improve replica deletion, to ensure that the data is cleaned up by utilizing a finalizer on the replica CR. ref [#685 (comment)](https://github.com/longhorn/longhorn/issues/685#issuecomment-817121946)

Thank you @joshimoo for pointing out, I've added to non-goal to be tackled separately.

shuo-wu (changes requested):
1. This feature is a little bit similar to the data locality feature. The main difference is that the data locality requires the new replica to be scheduled to a specific node but this feature requires the new replica to be scheduled to a list of nodes on which there is no replica for the volume.
2. Notice that you plan to delete the redundant replica before the new replica rebuilding. This will lead to the volume becoming `degraded`, which is not recommended. Even though most of the time the redundant replica is not meaningful for the HA.
3. The modification for ReplicaController is not right:
    1. ReplicaController should not update volume status. A controller's responsibility is to update the **status** of its own CRs and the **spec** of other CRs. 
    2. The scheduling-related checking happens in VolumeController. Since Longhorn knows all replicas in the VolumeController sync loop, there is no need to add a new field `volume.status.rebalanceReplicas`.
    3. I am not sure why you use `HealthyAt` to determine which replica should be deleted. I think we can randomly remove a replicate one.
4. Besides node offline and new node adding, there is one more case that would trigger this feature: node becoming schedulable/unschedulable.
5. Do we need to introduce the per volume spec for this feature? I mean, we can allow users to enable this feature for specific volumes by adding a new volume spec.

In general, I think you can refer to the design of the data locality feature.

c3y1huang:
Thank you @shuo-wu for the review and comment.

Moving this to draft to work on.

shuo-wu:
BTW, the current design for the data locality feature maybe not perfect since the flow is a little bit complicated IIRC. If you have any better ideas, please share with us.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,249 @@
...
+**Before enhancement**
+3. Bring node-2 online, no change will be done automatically by Longhorn. The user will still see the same warning on UI.
+4. To rebalance replicas, the user needs to find and delete the duplicated replica to trigger the schedule onto node-2. 
+
+**After enhancement**
+3. Longhorn automatically detects and deletes the duplicated replica so the replica will be scheduled onto node-2. User will see the duplicated replica get rescheduled back to node-2 and UI will see volume `Healthy` with the note `Limited node redundancy: at least one healthy replica is running at the same node as another` removed.
+
+### API changes
+There is no API change. The new global setting `Replica Node Level Auto Balancing` will use the same /v1/settings API.
+
```
shuo-wu, line 75: Besides, there is a new field in volume resources.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,249 @@
...
+### Implementation Overview
+
+#### longhorn-manager
+1. Add new global setting `Replica Node Level Auto Balancing`.
+  - The setting is boolean.
+  - The default value is `False`.
+
+2. Add new volume spec `AutoBalance`.
+  - The default value is nil.
```
shuo-wu, line 86: The default value is `false`

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,249 @@
...
+
+2. In Volume Controller, check and rebalance replica during `syncVolume`, `replenishReplicas`.
+  1. Skip replica rebalance if setting `Replica Node Level Auto Balancing` is `False` or,
+     if volume spec `AutoBalance` is `False`.
+  2. Create a function to check and return: the number of replicas to rebalance and a list of duplicated replica names.
+    1. Record replica names if DiskID is duplicated for the volume.
```
shuo-wu, line 92: `DiskID`? or you mean the related Node?

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,249 @@
...
+    2. Find the number of volume owners.
```
shuo-wu, line 93: What are `volume owners`? The nodes that contain the running volume replicas?

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,249 @@
...
+    1. Do nothing if the number of healthy replicas is less than the volume spec number of replicas.
+    2. Do nothing if the replica is not balanced.
+    3. Delete duplicated replicas one at a time.
+
+
+### Test plan
+
+#### Integration tests - test_replica_auto_balance
+
+```
```
shuo-wu, line 109: Why using ``` to wrap the test steps... Same below.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,249 @@
...
+When node-2 and node 3 up
+Then replica scheduled to node-1, node-2, node-3 for volume-1
+And replica not scheduled to node-1 for volume-2
+And cleanup settings
+```
+
+#### Manual tests - 1 node offline, 2 node online
+
+```
+Given cluster with 3 nodes
```
shuo-wu, line 154: Add step number for manual tests. e.g.:

1. Given cluster with 3 nodes and Longhorn setting
```
kubectl -n longhorn-system --type=merge patch settings/replica-soft-anti-affinity -p '{"value": "true"}'
kubectl -n longhorn-system get settings/replica-soft-anti-affinity
kubectl -n longhorn-system --type=merge patch settings/replica-replenishment-wait-interval -p '{"value": "1"}'
kubectl -n longhorn-system get settings/replica-replenishment-wait-interval
kubectl -n longhorn-system --type=merge patch settings/replica-auto-rebalance -p '{"value": "true"}'
kubectl -n longhorn-system get settings/replica-auto-rebalance
```
2. Create a Pod using a 3-replica Longhorn volume on `node-1`
......

shuo-wu (changes requested):

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: I think we would benefit to adopt to BDD-style Given/When/Then when writing tests. This gives a clearer way to show the prepare/test/result phases. It's currently looking a bit confusing because integration tests are using steps, so I will change them as well.

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: Addressed.

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: Addressed.

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: Thank you for the hint. Updated to use Node ID instead.

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: Thank you for the hint. Updated to use Node ID instead.

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: Thanks, fixed.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,387 @@
...
+  2. List all available nodes with conditions:
+     - node is schedulable.
+     - node zone not duplicated with any of the current running replicas.
+     - node has no volume replica running.
+  3. Do nothing:
+     - When the number of running replica is less than the number of replicas in volume spec.
+     - When available nodes list is empty.
+     - When the number of replicas in volume spec is equal to occupied node zones.
+  4. Calculate the number of replicas to rebalance:
+     - number of replicas in volume spec - number of occupied node zones.
```
shuo-wu, line 136: Min{number of replicas in volume spec - number of occupied node zones, number of zones contains schedulable nodes - number of occupied node zones}

Similar below

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,387 @@
...
+  5. Return the number of available nodes if less than the number to rebalance, or return the number to rebalance.
```
shuo-wu, line 137: The auto rebalance should happen during the calculation. Is returning this number meaningful?

Similar below

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,387 @@
...
+  5. Add the number to rebalance to `replenishCount` in `replenishReplicas`.
+
+- Calculate number to rebalance when there are duplicates in zones.
+  1. List all the occupied node zones with volume replicas running.
+  2. List all available nodes with conditions:
+     - node is schedulable.
+     - node zone not duplicated with any of the current running replicas.
+     - node has no volume replica running.
+  3. Do nothing:
+     - When the number of running replica is less than the number of replicas in volume spec.
```
shuo-wu, line 132: You can consider this (volume `healthy`) as a prerequisite rather than checking it during the auto rebalance.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,387 @@
...
+  - The default value is `False`.
+
+- In Volume Controller, check and rebalance replica during `syncVolume`, `replenishReplicas`.
+  1. Skip volume replica rebalance when global setting `Replica Auto Balancing` is `False`
+  2. Skip volume replica rebalance when volume spec `DisableReplicaRebalance` is `True`.
+  3. Calculate number to rebalance when there are duplicates in zones, if,
+      - global setting `Replica Zone Level Soft Anti-Affinity` is `True`.
+  4. Calculate number to rebalance when there are duplicates on nodes, if,
+      - global setting `Replica Zone Level Soft Anti-Affinity` is `True` and cannot find duplicates in zones.
+      - global setting `Replica Zone Level Soft Anti-Affinity` is `False`.
```
shuo-wu, line 122: With this limitation, can Longhorn do auto rebalance for the following case:

1. `Replica Zone Level Soft Anti-Affinity` is `True`. 
2. There are 3 replicas and 2 zones. And 2 replicas are in the same zone and even on the same node.
3. Add a new node to the above zone that contains 2 replicas.

shuo-wu (changes requested):

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,387 @@
...
+
+### Implementation Overview
+
+#### longhorn-manager
+- Add new global setting `Replica Auto Balancing`.
+  - The setting is boolean.
+  - The default value is `False`.
+
+- Add new volume spec `DisableReplicaRebalance`.
+  - The default value is `False`.
```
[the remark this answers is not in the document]
c3y1huang: Thanks for noticing, I think it's better to use `replicaAutoBalance` with the available value of `enabled` and `disabled` to avoid confusion and the default value will be the same as the global setting.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,387 @@
...
+
+- In Volume Controller, check and rebalance replica during `syncVolume`, `replenishReplicas`.
+  1. Skip volume replica rebalance when global setting `Replica Auto Balancing` is `False`
+  2. Skip volume replica rebalance when volume spec `DisableReplicaRebalance` is `True`.
```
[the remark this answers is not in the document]
c3y1huang: Thanks, removed the global setting check, I think it's better to check only volume spec.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,387 @@
...
+  3. Calculate number to rebalance when there are duplicates in zones, if,
+      - global setting `Replica Zone Level Soft Anti-Affinity` is `True`.
```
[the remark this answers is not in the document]
c3y1huang: Thanks for catching this, I think it should be safe to remove the condition, so it will always check for zone first.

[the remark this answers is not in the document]
c3y1huang: I've also included this scenario in integration tests.
```
Scenario: replica auto-balance to different zone when zone anti-affinity is disabled.
```

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: I do not understand this. Can you elaborate?

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: The plan is to add the returned as part of 
[replenishCount](https://github.com/longhorn/longhorn-manager/blob/master/controller/volume_controller.go#L1501).

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: Replica auto-balance will look for available nodes if cannot find an available zone.

I've included this scenario in the integration tests.
```
Scenario: replica auto-balance to same zones when zone anti-affinity is enabled.
```

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: Good suggestion. Addressed.

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: [4.] is only part of the formula. 

Take your case with the volume with 5 replicas:

Case-1:
If there is no new zone added, it will return 0 to rebalance, because both zones contain running replicas, and nodes within zone-A and zone-B will not be counted as the available nodes.
```
2. List all available nodes with conditions: 
- node zone not duplicated with any of the current running replicas.
3. Do nothing:
     - When available nodes list is empty.
```

Case-2:
If zone-C is then added with 2 replicas then the replicas to rebalance would return 1.

Case-3:
if 4 zones are added, then the replicas would return 3.

enhancements/20210510-automatic-rebalance-replica.md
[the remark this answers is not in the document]
shuo-wu: To totally avoid confusion or misunderstanding, this field can be an enum with 3 options: `enabled`, `disabled`, and `follow the global setting`/`ignored`.

enhancements/20210510-automatic-rebalance-replica.md
shuo-wu: OK. This calculation is complicated. I am thinking if we can directly use the following logic:

1. List available zones (contains node schedulable for the volume replicas):
    1. Reschedule the node-level duplicates replica to one node of each available zone
    2. If the node-level duplicates replica count is 0, then reschedule the zone-level duplicates replica to one node of each available zone.
2. List available nodes (that doesn't contain the volume replica but is schedulable for the replicas):
    1. Reschedule the node-level duplicates replica to one node of each available zone.

This avoids the calculation while getting the final rebalance result.

innobead (changes requested):

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
...
+**Before enhancement**
+3. Bring zone-2 online, no change will be done automatically by Longhorn. The user will still see replicas all running in zone-1.
+4. To rebalance replicas, the user needs to find and delete the duplicated replica to trigger the schedule to zone-2.
+
+**After enhancement**
+3. Longhorn automatically detects and deletes the duplicated replica so the replica will be scheduled to zone-2. The user will see the duplicated replica get rescheduled back to zone-2.
+
+### API changes
+- The new global setting `Replica Auto Balance` will use the same /v1/settings API.
+- When creating a new volume, the body of the request sent to /v1/volumes has a new field `replicaAutoBalance` set to `disabled` or `enabled`.
```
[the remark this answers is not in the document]
innobead: After the explanation by @c3y1huang , we need a way to determine if the value set by the user or not.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
...
+And zone-1 and zone-2 should contain 3 replica in total.
+
+When set node-2 to the zone with duplicated replicas.  
+And enable scheduling for node-2.
+
+Then volume replicas should be running on node-1, node-2, node-3.  
+And zone-1 should contain 2 replica.  
+And zone-2 should contain 1 replica.
+
+#### Manual tests - 1 node offline, 2 node online
```
innobead, line 313: Probably we don't need manual test cases here, because they are similar to the test cases described in the integration tests. (you can add some manual test cases for QA if need)

c3y1huang: Removed.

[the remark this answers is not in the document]
c3y1huang: Manual tests removed.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
...
+
+Given set `replica-soft-anti-affinity` to `true`.  
+And set `replica-zone-soft-anti-affinity` to `true`.  
+And set volume spec `replicaAutoBalance` to `enabled`.  
+And set node-1 to zone-1.  
+And set node-3 to zone-2.  
+And disable scheduling for node-2.  
+And create a volume with 3 replicas.  
+And attach the volume to node-1.  
+And zone-1 and zone-2 should contain 3 replica in total.
```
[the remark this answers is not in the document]
c3y1huang: Fixed.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
...
+And set volume spec `replicaAutoBalance` to `enabled`.  
+And enable scheduling for node-3.
+
+Then wait for replicas to re-schedule to zone-1 and zone-2.  
+And zone-1 should contain 1 replica.  
+And zone-2 should contain 1 replica.
+
+#### Integration tests - test_replica_auto_balance_to_same_zone_anti_affinity_enabled
+
+Scenario: replica auto-balance to same zones when zone anti-affinity is enabled.
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
...
+When set volume spec `replicaAutoBalance` to `enabled`.  
+And enable scheduling for node-3.
+
+Then wait for replicas to re-schedule to zone-1 and zone-2.  
+And zone-1 should contain 1 replica.  
+And zone-2 should contain 1 replica.
+
+#### Integration tests - test_replica_auto_balance_to_different_zone_anti_affinity_disabled
+
+Scenario: replica auto-balance to different zone when zone anti-affinity is disabled.
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
...
+Then volume-1 replicas should be running on node-1.  
+And volume-1 should have 3 replicas running.  
+And volume-2 replicas should be running on node-1, node-2, node-3.  
+And volume-2 should have 3 replicas running.  
+And volume-1 data should be the same as written.  
+And volume-2 data should be the same as written.
+
+#### Integration tests - test_replica_auto_balance_zone_anti_affinity_enabled
+
+Scenario: replica auto-balance with zone anti-affinity enabled.
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
...
+Support individual volume replica automatic balancing.
+
+### Non-goals [optional]
+[Clean up old Data after node failure](https://github.com/longhorn/longhorn/issues/685#issuecomment-817121946).
+
+## Proposal
+
+Add new global setting `Replica Auto Balance` to enable replica automatic balancing when a node is newly available to the cluster.
+
+Add new volume spec `replicaAutoBalance` to enable/disable automatic replica rebalancing for individual volume.
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
...
+Given set `replica-soft-anti-affinity` to `true`.  
+And set `replica-auto-balance` to `enabled`.  
+And create a volume with 3 replicas.  
+And attach the volume to node-1.  
+And wait for the volume to be healthy.  
+And volume replicas should be running on node-1, node-2, node-3.  
+And write some data to the volume.
+
+When disable scheduling and evict node-3.  
+And Wait for replicas to re-schedule to node-1, node-2.
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
...
+3. Bring node-2 online, no change will be done automatically by Longhorn. The user will still see the same warning on UI.
+4. To rebalance replicas, the user needs to find and delete the duplicated replica to trigger the schedule onto node-2. 
+
+**After enhancement**
+3. Longhorn automatically detects and deletes the duplicated replica so the replica will be scheduled onto node-2. User will see the duplicated replica get rescheduled back to node-2 and UI will see volume `Healthy` with the note `Limited node redundancy: at least one healthy replica is running at the same node as another` removed.
+
+#### Story 2 - Replica Zone Level Soft Anti-Affinity
+
+With an example of cluster set for 2 zones and default of 2 replicas volume:
+1. When the user enabled `replica-soft-anti-affinity` and `replica-zone-soft-anti-affinity` and deployed PVC and deployment. The replica is distributed evenly to all 2 zones at this point.
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
...
+
+So I do not have to worry and standby 24/7 to check and manually rebalance replicas when there are newly available nodes.
+
+### User Experience In Detail
+
+#### Story 1 - Replica Node Level Soft Anti-Affinity
+
+With an example of 3 nodes in cluster and default of 3 replicas volume:
+1. When the user enabled `replica-soft-anti-affinity` and deployed PVC and Pod on node-1. The replica is distributed evenly to all 3 nodes at this point.
+2. In case of node-2 is offline, the user will see replica was on node2 gets re-scheduled to node-1 or node-3. And also icon and note will appear on UI `Limited node redundancy: at least one healthy replica is running at the same node as another`.
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
...
+When a cluster node is offline and a new cluster is added, I want Longhorn to automatically detect and reschedule replicas evenly to some volumes only.
+
+So I do not have to worry and standby 24/7 to check and manually rebalance replicas when there are newly available nodes.
+
+### User Experience In Detail
+
+#### Story 1 - Replica Node Level Soft Anti-Affinity
+
+With an example of 3 nodes in cluster and default of 3 replicas volume:
+1. When the user enabled `replica-soft-anti-affinity` and deployed PVC and Pod on node-1. The replica is distributed evenly to all 3 nodes at this point.
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
...
+
+When a cluster zone is offline and comes back online after some time, I want Longhorn to automatically detect and reschedule replicas evenly across zones.
+
+So I do not have to worry and standby 24/7 to check and manually rebalance replicas when there are newly available nodes.
+
+#### Story 3 - replica automatic rebalance for individual volume
+
+As a system administrator,
+
+When a cluster node is offline and a new cluster is added, I want Longhorn to automatically detect and reschedule replicas evenly to some volumes only.
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
...
+
+When a cluster node is offline and comes back online after some time, I want Longhorn to automatically detect and reschedule replicas evenly to all nodes.
+
+So I do not have to worry and standby 24/7 to check and manually rebalance replicas when there are newly available nodes.
+
+#### Story 2 - add new node
+
+As a system administrator,
+
+When a cluster node is offline and a new cluster is added, I want Longhorn to automatically detect and reschedule replicas evenly to all nodes
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
...
+
+#### Story 2 - zone offline
+
+As a system administrator,
+
+When a cluster zone is offline and comes back online after some time, I want Longhorn to automatically detect and reschedule replicas evenly across zones.
+
+So I do not have to worry and standby 24/7 to check and manually rebalance replicas when there are newly available nodes.
+
+#### Story 3 - replica automatic rebalance for individual volume
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
...
+
+#### Story 2 - add new node
+
+As a system administrator,
+
+When a cluster node is offline and a new cluster is added, I want Longhorn to automatically detect and reschedule replicas evenly to all nodes
+
+So I do not have to worry and standby 24/7 to check and manually rebalance replicas when there are newly available nodes.
+
+#### Story 2 - zone offline
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
...
+
+Support global replica automatic balancing.  
+Support individual volume replica automatic balancing.
+
+### Non-goals [optional]
+[Clean up old Data after node failure](https://github.com/longhorn/longhorn/issues/685#issuecomment-817121946).
+
+## Proposal
+
+Add new global setting `Replica Auto Balance` to enable replica automatic balancing when a node is newly available to the cluster.
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
...
+
+- https://github.com/longhorn/longhorn/issues/587
+- https://github.com/longhorn/longhorn/issues/570
+
+## Motivation
+
+### Goals
+
+Support global replica automatic balancing.  
+Support individual volume replica automatic balancing.
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
+# Automatic Replica Rebalance
+
+Have a Longhorn global setting and volume spec to enable/disable automatic distribution of the off-balanced replicas when a node is newly available to the cluster.
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,412 @@
...
+
+## Summary
+
+When nodes are offline with setting `Replica Zone Level Soft Anti-Affinity` enabled, or `Replica Node Level Soft Anti-Affinity` enabled, the replicas could be duplicated on the same zone/node if the remain cluster zone/node number is less than the replica number.
+
+Currently, the user needs to be aware when nodes are offline and a new node is added back to the cluster to manually delete the replicas to rebalance onto the newly available node.
+
+This enhancement proposes is to add a new Longhorn global setting `Replica Auto Balance` to enable detection and deletion of the unbalanced replica to achieve automatic rebalancing. And this enhancement also proposes to add a new volume spec `replicaAutoBalance` to enable/disable automatic replica rebalancing for individual volume.
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: Removed affinity settings conditions.

shuo-wu: I mean, with the above simple flow, we can avoid considering so many conditions and calculating the rebalance number separately. Not sure if it make sense for you.

enhancements/20210510-automatic-rebalance-replica.md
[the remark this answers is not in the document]
c3y1huang: The `ignore` option sounds good to me. And will also try to replace `enable` with `least-effort` and `best-effort`.

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,289 @@
...
+
+#### Story 2 - add new node
+
+As a system administrator,
+
+When a cluster node is offline and a new node is added, I want Longhorn to automatically detect and reschedule replicas evenly to all nodes
+
+So I do not have to worry and standby 24/7 to check and manually rebalance replicas when there are newly available nodes.
+
+#### Story 3 - zone offline
```
innobead, line 55:
```suggestion
#### Story 3 - zone temporary offline
```

c3y1huang: Fixed.

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: To put it simply, rebalance is done in 2 phases, rebalance zones then rebalance nodes. This from my understanding is similar to your statement. I've reworded the implementation so hopefully takes away the distractions of the conditions.

As for affinity settings, it will be ignored to leave for ReplicaScheduler to handle.

shuo-wu (approved):
LGTM

enhancements/20210510-automatic-rebalance-replica.md
shuo-wu: Nice rewording. It's much clear now.

innobead (approved):

enhancements/20210510-automatic-rebalance-replica.md
```diff
@@ -0,0 +1,220 @@
...
+
+#### Integration tests - test_replica_auto_balance_best_effort
+
+Scenario: replica auto-balance with `best_effort`.
+
+#### Integration tests - test_replica_auto_balance_disabled_volume_spec_enabled
+
+Scenario: replica should auto-balance individual volume when global setting `replica-auto-balance` is `disabled` and volume spec `replicaAutoBalance` is `least_effort`.
+
+#### Integration tests - test_replica_auto_balance_least_effort
```
innobead, line 208: zone_least_effort

innobead (approved):
LGTM, just one minor change needed

enhancements/20210510-automatic-rebalance-replica.md
c3y1huang: Fixed.

c3y1huang:
Updated the integration test steps.

Issue #2403, opened 2021-03-25
[FEATURE] Create backing image from volume


yasker:
**Is your feature request related to a problem? Please describe.**
For VM users, it's often that the user wants to update the content of the volume then create more VM with the updated content. As the storage layer, Longhorn needs to implement creating backing images from volumes in order to support this use case.

**Describe the solution you'd like**
Support creating a backing image from a volume.

**Describe alternatives you've considered**
Download the volume content as a raw image file, then upload it somewhere, and create a Longhorn backing image using the remote URL.

**Additional context**

innobead:
Waiting for UI integration. cc @smallteeths

PR #2609, opened 2021-05-20
enhancement: Add a new enhancement 'create-backing-image-from-volume'

shuo-wu:
#2403

### DIFF
--- a/enhancements/20210520-create-backing-image-from-volume.md
+++ b/enhancements/20210520-create-backing-image-from-volume.md
@@ -1,0 +1,85 @@
+# Create Backing Image From Volume
+
+## Summary
+Longhorn can create a backing image from an existing Longhorn volume.
+
+### Related Issues
+https://github.com/longhorn/longhorn/issues/2403
+
+## Motivation
+### Goals
+1. Longhorn should allow users to directly generate a backing image based on an existing volume.
+2. We need to consider the compatibility with [the volume clone feature](https://github.com/longhorn/longhorn/issues/1815).   
+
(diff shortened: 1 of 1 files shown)


enhancements/20210520-create-backing-image-from-volume.md
```diff
@@ -0,0 +1,85 @@
...
+2. We need to consider the compatibility with [the volume clone feature](https://github.com/longhorn/longhorn/issues/1815).   
+
+### Non-goals:
+1. A full design or implementation for feature "volume clone".
+2. Download the volume content to a file.
+
+## Proposal
+1. Add a new gRPC API for the backing image that allow backing image managers fetch the backing file inside the disk.
+2. Add a field for the backing image creation HTTP API that indicates the data sources is a volume.
+    1. Automatically attach & detach the volume in the API call.
```
yasker, line 21: What if the volume is attached already? Since it's similar to the backup, we should be able to do it when the normal traffic is still going:
1. need to create a system snapshot though.
2. if detach happens, then creation will fail. Can try again from the detached state though.

The volume can also be just attached in the maintenance mode. Since it's the replica who will do the job.

yasker:
cc @joshimoo

enhancements/20210520-create-backing-image-from-volume.md
shuo-wu: OK. Then the backing image needs to record the snapshot ID. And the clone processing should be moved to the backing image controller instead of doing it in the api package.

Issue #2404, opened 2021-03-25
[FEATURE] Backing image upload


yasker:
Currently, Longhorn can only retrieve backing image from somewhere outside of the Longhorn system. But since Longhorn already stores copies of the backing image, we should allow users to upload the backing image if the user wants to do so locally.

kaxing:
Validating in progress with Longhorn v1.2.0-preview1

Found a minor bug, when upload got interrupted, the checksum status will be shown as 'Verified'
![image](https://user-images.githubusercontent.com/602540/129181279-8398dd01-1e81-4706-b394-96720326d037.png)

shuo-wu:
@smallteeths ^^ Can you take a look at this bug?

kaxing:
Validation - **Passed**

Verified with master images (0819, ui: 69ebf61) and following combinations:
- File uploading with different file size, including 0m, 1mb to 6gb.
  - With or without checksum
  - Wait until finished or interrupt the upload process

A few memos for the future reference:
- Empty checksum got displayed as Verified now in now fixed
- Upload From Local with interruption will display as Failed in Status column: ![image](https://user-images.githubusercontent.com/602540/130060955-a202688c-d2df-49ac-9d54-3c8e927c4929.png)
- Current default pagination is 10 backing images per page

Progress of this issue :)
![image](https://user-images.githubusercontent.com/602540/130060327-c87b9b16-7797-4dc1-b18b-badae72ee52c.png)

PR #2617, opened 2021-05-22
enhancement: Add a new enhancement 'backing-image-upload'

shuo-wu:
#2404

### DIFF
--- a/enhancements/20210521-backing-image-upload.md
+++ b/enhancements/20210521-backing-image-upload.md
@@ -1,0 +1,87 @@
+# Backing Image Uplad
+
+## Summary
+Besides downloading backing image from remote URLs, Longhorn should allow users to upload a local file as a backing image.
+
+### Related Issues
+https://github.com/longhorn/longhorn/issues/2404
+
+## Motivation
+### Goals
+Support upload large files as backing images.
+
+## Proposal
+1. During backing image creation, Longhorn can provide another option as the backing image data source: `Upload From Remote`.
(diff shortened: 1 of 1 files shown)


joshimoo:
Had a good discussion with @shuo-wu on the design of the backing image manager.
We realized that the core competencies of the manager should be ensure availability of the file + ability to sync.
Therefore I asked shuo if it's possible to strip out the one of actions from the backing image manager, to simplify and focus on the core competencies. 

One Idea of accomplishing this, might be the introduction of a BackingImageSource crd, that is responsible for the initial download/user upload after which the backing image manager can just move over the completed file from a shared tmp directory. 

@shuo-wu will discuss with @yasker and @innobead and investigate the feasibility of this approach.

--------------------------------------------------------
Below are some example/reading materials on how to deal with http multipart data.

This is example for multipart client uploads (client side)
https://medium.com/@owlwalks/sending-big-file-with-minimal-memory-in-golang-8f3fc280d2c

This is example for multipart server uploads (server side) - This is more for learning about validation and stuff, in our case we would want to be able to process the full request and let the multipart writer page to disk, but it gives a good overview of how do validation and partial processing. (we don't need this part)
https://medium.com/@owlwalks/dont-parse-everything-from-client-multipart-post-golang-9280d23cd4ad

Example of the multipart reader:
https://stackoverflow.com/questions/59121113/how-to-deal-with-large-file-upload-via-http-without-storing-the-request-in-tempo
https://golang.org/pkg/net/http/#Request.ParseMultipartForm
https://pkg.go.dev/net/http?utm_source=godoc#Request.MultipartReader
https://pkg.go.dev/mime/multipart?utm_source=godoc#example-NewReader

shuo-wu:
Close this since the design of the upload is already integrated in https://github.com/longhorn/longhorn/pull/2760

Issue #2543, opened 2021-04-27
[BUG] Failed to list backups after upgrade from v1.1.0 to v1.1.1


LuizBuffa25:
**Describe the bug**
When listing backups after upgrade to 1.1.1: 

error listing backups for volume 'longhorn-vol-pv-mongo': error listing backups: Failed to execute: /var/lib/longhorn/engine-binaries/longhornio-longhorn-engine-v1.1.1/longhorn [backup ls --volume longhorn-vol-pv-mongo s3://<my bucket name>@< my region >/], output context deadline exceeded , stderr, time="2021-04-27T11:37:05Z" level=error msg="context deadline exceeded" , error exit status 1

I test the same command on the terminal and got the same error:
context deadline exceeded" , error exit status 1

When I try the same command but using old version 1.1.0 ( /var/lib/longhorn/engine-binaries/longhornio-longhorn-engine-v1.1.0/longhorn ) works fine.

* I have 2 volumes:
One with 10Gi total size, I can list backups for that.  
Another one with 150Gi that I have the problem mentioned in this issue.

**To Reproduce**
1. Upgrade from v1.1.0 to v1.1.1
2. List backups for volumes with 150Gi+ size
3. See error

**Expected behavior**
Should list backups.

**Log**
If applicable, add the Longhorn managers' log when the issue happens.

You can also attach a *Support Bundle* here. You can generate a Support Bundle using the link at the footer of the Longhorn UI.

**Environment:**
 - Longhorn version: 1.1.1
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl): kubectl

**Additional context**

innobead:
cc @jenting

jenting:
@LuizBuffa25 Thanks for your reporting.
1. Which S3 provider you're using? AWS _or_ GCP _or_ others?
2. Could you please check inside the 150Gi+ volume's folder in the S3 bucket, are there any other folders that are not related to these volumes?

LuizBuffa25:
@jenting Thanks for answering.
1) I will confirm but I think it is AWS;
2) There is only volumes related folder inside it;

I think is something related to timeout:
- Using v1.1.0: $ /var/lib/longhorn/engine-binaries/longhornio-longhorn-engine-v1.1.0/longhorn 
   takes between 60-90 seconds to command return all my available backups;

- Using v1.1.1: $ /var/lib/longhorn/engine-binaries/longhornio-longhorn-engine-v1.1.0/longhorn 
[...]

Issue #2536, opened 2021-04-25
[BUG] Restore failed for large volume


PPCM:
**Describe the bug**
Restore failed for large volume with error code 504

**To Reproduce**
1. Create a volume of 200Gi
2. Write some data to increase used size
3. Make backups on NFS drive
4. Restoring the volume generate 'Request failed with status code 504'

**Expected behavior**
My backups are on a NFS drive, backups works fine. I can restore volume with sizes of 20Gi or 60Gi, but impossible with a volume of 200Gi. It raise the error 'Request failed with status code 504'
I guess it is a timeout, but no way to modify the delay

**Environment:**
 - Longhorn version: 1.1.0 and 1.1.1
 - Installation method: Longhorn Catalog in Rancher
 - Kubernetes distro: K3s v1.20.5+k3s1
   - Number of management node in the cluster: 3
   - Number of worker node in the cluster: 4
 - Node config
   - OS type and version: Ubuntu 64bits
   - CPU: ARM on Raspberry Pi 4B
   - Memory per node: 8Go
   - Disk type(e.g. SSD/NVMe): SSD
   - Network bandwidth between the nodes: 1Gb
 - Number of Longhorn volumes in the cluster: 7

shuo-wu:
The error is like a timeout error from an S3 server rather than an NFS server... Can you generate a support bundle and send it to longhorn-support-bundle@Suse.com?

PPCM:
I think too, it is a timeout
For the backup , I confirm, it is a NFS link to a HS-210 with the last firmware
I made some tests. When I make a `du`  on large partition from a longhorn-manager where the NFS is mounted, it takes from 8mn to 13mn to complete, but it takes less than 1mn on small partitions.

The freshly support bundle was sent.

shuo-wu:
Typically one volume creation request is followed by some result log, e.g.,
```
2021-04-26T22:06:44.238270541+02:00 10.42.6.159 - - [26/Apr/2021:20:06:44 +0000] "POST /v1/volumes HTTP/1.1" 200 1285 "" "Go-http-client/1.1"
[...]

Issue #1955, opened 2020-11-05
[BUG] Backup - S3 Timeout


timmy59100:
**Describe the bug**
In the backup UI, I get an error and the backups are not listed.

`error listing backups: error listing backup volumes: Timeout executing: /var/lib/longhorn/engine-binaries/longhornio-longhorn-engine-v1.0.0/longhorn [backup ls --volume-only s3://PATH/], output , stderr, , error <nil>`

Number of Volumes: 39
Number of Volumes with backup enabled: 32

S3 Bucket Size: 2TB

**Expected behavior**
Backups of Volumes are showed

**Log**
`time="2020-11-05T12:55:15Z" level=warning msg="backup store monitor: failed to list backup volumes in s3://S3PATH/: error listing backup volumes: Timeout executing: /var/lib/longhorn/engine-binaries/longhornio-longhorn-engine-v1.0.0/longhorn [backup ls --volume-only s3://longhorn-production@ch-dk-2/], output , stderr, , error <nil>"
`

**Environment:**
 - Longhorn version: 1.0.0
 - Kubernetes version: 1.18.3
 - Node OS type and version: Centos 7.7

**Additional context**
On an empty bucket the backups are listed normally.

S3 Provider https://exoscale.com

yasker:
32 volumes is not a big number. And this command only checks the volume itself, so 2TB of data should be out of concern. We need to check how we can reproduce this issue in our lab.

jenting:
From the AWS SDK, the default time is 30 seconds https://github.com/longhorn/backupstore/blob/18aa87a/vendor/github.com/aws/aws-sdk-go/aws/session/cabundle_transport.go#L17
From the issue reported, the longhorn version is `1.0.0` which means that the list operation can't be fulfilled with 30 seconds.

At longhorn release `1.1.0`, we add retry count to `3` https://github.com/longhorn/longhorn-engine/blob/v1.1.0/vendor/github.com/longhorn/backupstore/s3/s3_service.go#L28 which means the list operation tolerance time is 90 seconds.

@timmy59100 Have you tried to upgrade your longhorn cluster to version `1.1.0` to see the error is still happening?

timmy59100:
@jenting Still the same after upgrade to 1.1.0.
It's critical if we can't restore any volumes.

jenting:
@timmy59100 could you please provide the ping latency between your LH cluster to the S3 provider?

jenting:
In my testing
[...]

Issue #1761, opened 2020-09-04
[REFACTORING] Longhorn Backup refactoring


joshimoo:
While working on the csi snapshot support #304 I noticed more and more issues with our current backup interaction patterns.
As part of that work, I started looking into the addition of a `VolumeBackup` crd, that would be exported to the backupstore
with the actual backup metadata. This would allow us to add a backup-sync controller that would keep the local cluster state in
sync with the backupstore state, from that point onwards all the backup interaction would be happening declaratively via the `VolumeBackup` custom resource.

This would add a persistence layer to the backup operations and would solve issues like the below:
- https://github.com/longhorn/longhorn/issues/1249
- https://github.com/longhorn/longhorn/issues/1311

Together with the backup-sync controller it would be very easy to support multiple backup targets by introducing a `BackupTarget` crd. Which would allow for user defined backup strategies for example: 
(daily -> Target A, weekly -> Target B, monthly -> Target C) 

At the same time, this would allow for the removal of all Backup information from the Longhorn `Volume` custom resource.
Which should make the volume-controller and other controller logic more simple and straightforward. currently, there are a lot of updates to the `Volume` resource anytime a new block gets written the volume status would be updated, etc.
Which means the volume-controller needs to process the whole volume object again.
- https://github.com/longhorn/longhorn/issues/844

For user experience, we also need to make sure all events/status of backup should be tracked, and users can understand what's the status of backups. ref: https://github.com/longhorn/longhorn/issues/627

joshimoo:
I started writing some code for this while working on the csi snapshotter implementation, before realizing that this effort would risk the 1.1 release timeline. Therefore I moved this out of scope of the csi snapshot #304 implementation. So that we can prioritize separately, for 1.2 the roadmap currently includes application backups, so that might be a good time to continue work on this.

There seems to have been a poc of something like this previously https://github.com/longhorn/longhorn/issues/945
Need to sync up about the results of that POC.

jenting:
> Together with the backup-sync controller it would be very easy to support multiple backup targets by introducing a `BackupTarget` crd. Which would allow for user defined backup strategies for example:
> (daily -> Target A, weekly -> Target B, monthly -> Target C)

Nice! I had the same idea when designing the `VolumeBackup` CRD.

innobead:
This will be actually tackled in https://github.com/longhorn/longhorn/issues/2543.

innobead:
[...]

PR #2623, opened 2021-05-25
[LEP] Asynchronous pull backup volumes into cluster CR

jenting:
#### Issues

#1761
#1955
#2536
#2543

### DIFF
--- a/enhancements/20210525-async-pull-backups.md
+++ b/enhancements/20210525-async-pull-backups.md
@@ -320,7 +320,7 @@ None.
       1. Updates the BackupTarget CR `status.available=false` and `status.lastSyncedAt=time.Now()`.
       2. Skip the current reconcile process.
    4. List in cluster BackupVolume CRs `clusterBackupVolumes`.
-   5. Find the difference backup volumes `backupVolumesToPull = backupStoreBackupVolumes - clusterBackupVolumes` and create BackupVolume CR `metadata.name` + `metadata.finalizers["longhorn.io"]`.
+   5. Find the difference backup volumes `backupVolumesToPull = backupStoreBackupVolumes - clusterBackupVolumes` and create BackupVolume CR `metadata.name`.
    6. Find the difference backup volumes `backupVolumesToDelete = clusterBackupVolumes - backupStoreBackupVolumes` and delete BackupVolume CR.
    7. List in cluster BackupVolume CRs `clusterBackupVolumes` again and updates the BackupVolume CR `spec.syncRequestAt = time.Now()`.
    8. Updates the BackupTarget CR status:
@@ -345,7 +345,7 @@ None.
    3. Check if the `status.lastSyncedAt < spec.syncRequestAt`. If no, skip the reconcile process.
    4. Call the longhorn engine to list all the backups `backup ls --volume <volume-name>` from the remote backup target `backupStoreBackups`.
    5.  List in cluster Backup CRs `clusterBackups`.
-   6.  Find the difference backups `backupsToPull = backupStoreBackups - clusterBackups` and create Backup CR `metadata.name` + `metadata.labels["longhornvolume"]=<backup-volume-name>` + `metadata.finalizers["longhorn.io"]`.
+   6.  Find the difference backups `backupsToPull = backupStoreBackups - clusterBackups` and create Backup CR `metadata.name` + `metadata.labels["longhornvolume"]=<backup-volume-name>`.
    7.  Find the difference backups `backupsToDelete = clusterBackups - backupStoreBackups` and delete Backup CR.
    8.  Call the longhorn engine to get the backup volume config's last modification time `backup head <volume-config>` and compares to `status.lastModificationTime`. If the config last modification time not changed, abort the current reconcile process.
    9.  Call the longhorn engine to read the backup volumes' config `backup inspect-volume <volume-name>`.
@@ -387,7 +387,7 @@ None.
         2.  updates Backup CR `status.backupCreationIsStart = true`.
         3.  fork a go routine to monitor the backup creation progress. After backup creation finished (progress = 100):
             1.  update the BackupVolume CR `spec.syncRequestAt = time.Now()` if BackupVolume CR exist.
-            2.  create the BackupVolume CR `metadata.name` + `metadata.finalizers["longhorn.io"]` if BackupVolume CR not exist.
+            2.  create the BackupVolume CR `metadata.name` if BackupVolume CR not exist.
     4. If Backup CR `status.lastSyncedAt != nil`, the backup config had be synced, abort the current reconcile process
     5. Call the longhorn engine to read the backup config `backup inspect <backup-url>`.
     6. Updates the Backup CR status field according to the backup config.
--- a/enhancements/20210525-async-pull-backups.md
+++ b/enhancements/20210525-async-pull-backups.md
@@ -257,7 +257,7 @@ None.
      pollInterval: the backup target poll interval. (metav1.Duration)
      syncRequestAt: the time to request run sync the remote backup target. (*metav1.Time)
    status:
-     ownerID: the node ID which should responsible for run sync the remote backup target. (string)
+     ownerID: the node ID which is responsible for running operations of the backup target controller. (string)
      available: records if the remote backup target is available or not. (bool)
      lastSyncedAt: records the last time the backup target was running the reconcile process. (*metav1.Time)
    ```
@@ -271,7 +271,7 @@ None.
      syncRequestAt: the time to request run sync the remote backup volume. (*metav1.Time)
      fileCleanupRequired: indicate to delete the remote backup volume config or not. (bool)
    status:
-     ownerID: the node ID which should responsible for run sync the remote backup target. (string)
+     ownerID: the node ID which is responsible for running operations of the backup volume controller. (string)
      lastModificationTime: the backup volume config last modification time. (Time)
      size: the backup volume size. (string)
      labels: the backup volume labels. (map[string]string)
@@ -297,7 +297,7 @@ None.
      backingImage: the backing image. (string)
      backingImageURL: the backing image URL. (string)
    status:
-     ownerID: the node ID which should responsible for run sync the remote backup target. (string)
+     ownerID: the node ID which is responsible for running operations of the backup controller. (string)
      backupCreationIsStart: to indicate the snapshot backup creation is start or not. (bool)
      url: the snapshot backup URL. (string)
      snapshotName: the snapshot name. (string)
--- a/enhancements/20210525-async-pull-backups.md
+++ b/enhancements/20210525-async-pull-backups.md
@@ -269,7 +269,7 @@ None.
      name: the backup volume name.
    spec:
      syncRequestAt: the time to request run sync the remote backup volume. (*metav1.Time)
-     deleteRemoteConfig: indicate to delete the remote backup volume config or not. (bool)
+     fileCleanupRequired: indicate to delete the remote backup volume config or not. (bool)
    status:
      lastModificationTime: the backup volume config last modification time. (Time)
      size: the backup volume size. (string)
@@ -290,7 +290,7 @@ None.
      labels:
        longhornvolume=<backup-volume-name>`: this label indicates which backup volume the backup belongs to.
    spec:
-     deleteRemoteConfig: indicate to delete the remote backup config or not (and the related block files if needed). (bool)
+     fileCleanupRequired: indicate to delete the remote backup config or not (and the related block files if needed). (bool)
      snapshotName: the snapshot name. (string)
      labels: the labels of snapshot backup. (map[string]string)
      backingImage: the backing image. (string)
@@ -330,7 +330,7 @@ None.
 6. For the Longhorn manager HTTP endpoints:
 
    - **DELETE** `/v1/backupvolumes/{volName}`:
-     1. Update the BackupVolume CR `spec.deleteRemoteConfig=true` with the given volume name.
+     1. Update the BackupVolume CR `spec.fileCleanupRequired=true` with the given volume name.
      2. Delete a BackupVolume CR with the given volume name.
 
 7. Create a new controller `backup_volume_controller`.
@@ -338,9 +338,9 @@ None.
    Watches the change of BackupVolume CR. The backup volume controller is responsible for deleting Backup CR and deleting backup volume from remote backup target if delete BackupVolume CR event comes in, and updating BackupVolume CR status field, and creating/deleting Backup CR. The reconcile loop steps are:
    1. Check if the current node ID == BackupTarget CR spec.responsibleNodeID. If no, skip the reconcile process.
    2. If the delete BackupVolume CR event comes in:
-      1. updates Backup CRs `spec.deleteRemoteConfig=true` if BackupVolume CR `spec.deleteRemoteConfig=true`.
+      1. updates Backup CRs `spec.fileCleanupRequired=true` if BackupVolume CR `spec.fileCleanupRequired=true`.
       2. deletes Backup CR with the given volume name.
-      3. deletes the backup volume from the remote backup target `backup rm --volume <volume-name> <url>` if `spec.deleteRemoteConfig=true`.
+      3. deletes the backup volume from the remote backup target `backup rm --volume <volume-name> <url>` if `spec.fileCleanupRequired=true`.
       4. remove the finalizer.
    3. Check if the `status.lastSyncedAt < spec.syncRequestAt`. If no, skip the reconcile process.
    4. Call the longhorn engine to list all the backups `backup ls --volume <volume-name>` from the remote backup target `backupStoreBackups`.
@@ -371,7 +371,7 @@ None.
           backingImageURL: <backing-image-URL>
         ```
    - **DELETE** `/v1/backupvolumes/{volName}?action=backupDelete`:
-     1. Update the Backup CR `spec.deleteRemoteConfig=true` with the given volume name.
+     1. Update the Backup CR `spec.fileCleanupRequired=true` with the given volume name.
      2. Delete a Backup CR with the given backup name.
 
 9.  Create a new controller `backup_controller`.
@@ -379,7 +379,7 @@ None.
     Watches the change of Backup CR. The backup controller is responsible for updating the Backup CR status field and creating/deleting backup to/from the remote backup target. The reconcile loop steps are:
     1. Check if the current node ID == BackupTarget CR spec.responsibleNodeID. If no, skip the reconcile process.
     2. If the delete Backup CR event comes in:
-        1.  delete the backup from the remote backup target `backup rm <url>` if Backup CR `spec.deleteRemoteConfig=true`.
+        1.  delete the backup from the remote backup target `backup rm <url>` if Backup CR `spec.fileCleanupRequired=true`.
         2.  update the BackupVolume CR `spec.syncRequestAt=time.Now()`.
         3.  remove the finalizer.
     3. Check if the Backup CR `spec.snapshotName != ""` and `status.backupCreationIsStart == false`. If yes:
--- a/enhancements/20210525-async-pull-backups.md
+++ b/enhancements/20210525-async-pull-backups.md
@@ -1,0 +1,448 @@
+# Async Pull/Push Backups From/To The Remote Backup Target
+
+## Summary
+
+Currently, Longhorn uses a blocking way for communication with the remote backup target, so there will be some potential voluntary or involuntary factors (ex: network latency) impacting the functions relying on remote backup target like listing backups or even causing further [cascading problems](###related-issues) after the backup target operation.
+
+This enhancement is to propose an asynchronous way to pull backup volumes and backups from the remote backup target (S3/NFS) then persistently saved via cluster custom resources.
+
+This can resolve the problems above mentioned by asynchronously querying the list of backup volumes and backups from the remote backup target for final consistent available results. It's also scalable for the costly resources created by the original blocking query operations.
+
+### Related Issues
+
+- https://github.com/longhorn/longhorn/issues/1761
+- https://github.com/longhorn/longhorn/issues/1955
+- https://github.com/longhorn/longhorn/issues/2536
+- https://github.com/longhorn/longhorn/issues/2543
+
+## Motivation
+
+### Goals
+
+Decrease the query latency when listing backup volumes _or_ backups in the circumstances like lots of backup volumes, lots of backups, or the network latency between the Longhorn cluster and the remote backup target.
+
+### Non-goals
+
+- Automatically adjust the backup target poll interval.
+- Supports multiple backup targets.
+- Supports API pagination for listing backup volumes and backups.
+
+## Proposal
+
+1. Change the [longhorn/backupstore](https://github.com/longhorn/backupstore) _list_ command behavior, and add _inspect-volume_ and _head_ command.
+   - The `backup list` command includes listing all backup volumes and the backups and read these config.
+     We'll change the `backup list` behavior to perform list only, but not read the config.
+   - Add a new `backup inspect-volume` command to support read backup volume config.
+   - Add a new `backup head` command to support get config metadata.
+2. Create a BackupTarget CRD _backuptargets.longhorn.io_ to save the backup target URL, credential secret, and poll interval.
+3. Create a BackupVolume CRD _backupvolumes.longhorn.io_ to save to backup volume config.
+4. Create a Backup CRD _backups.longhorn.io_ to save the backup config.
+5. At the existed `setting_controller`, which is responsible creating/updating the default BackupTarget CR with settings `backup-target`, `backup-target-credential-secret`, and `backupstore-poll-interval`.
+6. Create a new controller `backup_target_controller`, which is responsible for creating/deleting the BackupVolume CR.
+7. Create a new controller `backup_volume_controller`, which is responsible:
+   1. deleting Backup CR and deleting backup volume from remote backup target if delete BackupVolume CR event comes in.
+   2. updating BackupVolume CR status.
+   3. creating/deleting Backup CR.
+8. Create a new controller `backup_controller`, which is responsible:
+   1. deleting backup from remote backup target if delete Backup CR event comes in.
+   2. calling longhorn engine/replica to perform snapshot backup to the remote backup target.
+   3. updating the Backup CR status, and deleting backup from the remote backup target.
+9.  The HTTP endpoints CRUD methods related to backup volume and backup will interact with BackupVolume CR and Backup CR instead interact with the remote backup target.
+
+### User Stories
+
+Before this enhancement, when the user's environment under the circumstances that the remote backup target has lots of backup volumes _or_ backups, and the latency between the longhorn manager to the remote backup target is high.
+Then when the user clicks the `Backup` on the GUI, the user might hit list backup volumes _or_ list backups timeout issue (the default timeout is 1 minute).
+
+We choose to not create a new setting for the user to increase the list timeout value is because the browser has its timeout value also.
+Let's say the list backups needs 5 minutes to finish. Even we allow the user to increase the longhorn manager list timeout value, we can't change the browser default timeout value. Furthermore, some browser doesn't allow the user to change default timeout value like Google Chrome.
+
+After this enhancement, when the user's environment under the circumstances that the remote backup target has lots of backup volumes _or_ backups, and the latency between the longhorn manager to the remote backup target is high.
+Then when the user clicks the `Backup` on the GUI, the user can eventually list backup volumes _or_ list backups without a timeout issue.
+
+#### Story 1
+
+The user environment is under the circumstances that the remote backup target has lots of backup volumes and the latency between the longhorn manager to the remote backup target is high. Then, the user can list all backup volumes on the GUI.
+
+#### Story 2
+
+The user environment is under the circumstances that the remote backup target has lots of backups and the latency between the longhorn manager to the remote backup target is high. Then, the user can list all backups on the GUI.
+
+#### Story 3
+
+The user creates a backup on the Longhorn GUI. Now the backup will create a Backup CR, then the backup_controller reconciles it to call Longhorn engine/replica to perform a backup to the remote backup target.
+
+### User Experience In Detail
+
+None.
+
+### API changes
+
+1. For [longhorn/backupstore](https://github.com/longhorn/backupstore)
+
+   The current [longhorn/backupstore](https://github.com/longhorn/backupstore) list and inspect command behavior are:
+   - `backup ls --volume-only`: List all backup volumes and read it's config (`volume.cfg`). For example:
+     ```json
+     $ backup ls s3://backupbucket@minio/ --volume-only
+     {
+       "pvc-004d8edb-3a8c-4596-a659-3d00122d3f07": {
+         "Name": "pvc-004d8edb-3a8c-4596-a659-3d00122d3f07",
+         "Size": "2147483648",
+         "Labels": {},
+         "Created": "2021-05-12T00:52:01Z",
+         "LastBackupName": "backup-c5f548b7e86b4b56",
+         "LastBackupAt": "2021-05-17T05:31:01Z",
+         "DataStored": "121634816",
+         "Messages": {}
+       },
+       "pvc-7a8ded68-862d-4abb-a08c-8cf9664dab10": {
+         "Name": "pvc-7a8ded68-862d-4abb-a08c-8cf9664dab10",
+         "Size": "10737418240",
+         "Labels": {},
+         "Created": "2021-05-10T02:43:02Z",
+         "LastBackupName": "backup-432f4d6afa31481f",
+         "LastBackupAt": "2021-05-10T06:04:02Z",
+         "DataStored": "140509184",
+         "Messages": {}
+       }
+     }
+     ```
+   - `backup ls --volume <volume-name>`: List all backups and read it's config (`backup_backup_<backup-hash>.cfg`). For example:
+     ```json
+     $ backup ls s3://backupbucket@minio/ --volume pvc-004d8edb-3a8c-4596-a659-3d00122d3f07
+     {
+       "pvc-004d8edb-3a8c-4596-a659-3d00122d3f07": {
+         "Name": "pvc-004d8edb-3a8c-4596-a659-3d00122d3f07",
+         "Size": "2147483648",
+         "Labels": {},
+         "Created": "2021-05-12T00:52:01Z",
+         "LastBackupName": "backup-c5f548b7e86b4b56",
+         "LastBackupAt": "2021-05-17T05:31:01Z",
+         "DataStored": "121634816",
+         "Messages": {},
+         "Backups": {
+           "s3://backupbucket@minio/?backup=backup-02224cb26b794e73\u0026volume=pvc-004d8edb-3a8c-4596-a659-3d00122d3f07": {
+             "Name": "backup-02224cb26b794e73",
+             "URL": "s3://backupbucket@minio/?backup=backup-02224cb26b794e73\u0026volume=pvc-004d8edb-3a8c-4596-a659-3d00122d3f07",
+             "SnapshotName": "backup-23c4fd9a",
+             "SnapshotCreated": "2021-05-17T05:23:01Z",
+             "Created": "2021-05-17T05:23:04Z",
+             "Size": "115343360",
+             "Labels": {},
+             "IsIncremental": true,
+             "Messages": null
+            },
+           ...
+           "s3://backupbucket@minio/?backup=backup-fa78d89827664840\u0026volume=pvc-004d8edb-3a8c-4596-a659-3d00122d3f07": {
+             "Name": "backup-fa78d89827664840",
+             "URL": "s3://backupbucket@minio/?backup=backup-fa78d89827664840\u0026volume=pvc-004d8edb-3a8c-4596-a659-3d00122d3f07",
+             "SnapshotName": "backup-ac364071",
+             "SnapshotCreated": "2021-05-17T04:42:01Z",
+             "Created": "2021-05-17T04:42:03Z",
+             "Size": "115343360",
+             "Labels": {},
+             "IsIncremental": true,
+             "Messages": null
+           }
+         }
+       }
+     }
+     ```
+   - `backup inspect <backup>`: Read a single backup config (`backup_backup_<backup-hash>.cfg`). For example:
+     ```json
+     $ backup inspect s3://backupbucket@minio/?backup=backup-fa78d89827664840\u0026volume=pvc-004d8edb-3a8c-4596-a659-3d00122d3f07
+     {
+       "Name": "backup-fa78d89827664840",
+       "URL": "s3://backupbucket@minio/?backup=backup-fa78d89827664840\u0026volume=pvc-004d8edb-3a8c-4596-a659-3d00122d3f07",
+       "SnapshotName": "backup-ac364071",
+       "SnapshotCreated": "2021-05-17T04:42:01Z",
+       "Created": "2021-05-17T04:42:03Z",
+       "Size": "115343360",
+       "Labels": {},
+       "IsIncremental": true,
+       "VolumeName": "pvc-004d8edb-3a8c-4596-a659-3d00122d3f07",
+       "VolumeSize": "2147483648",
+       "VolumeCreated": "2021-05-12T00:52:01Z",
+       "Messages": null
+     }
+     ```
+
+   After this enhancement, the [longhorn/backupstore](https://github.com/longhorn/backupstore) list and inspect command behavior are:
+   - `backup ls --volume-only`: List all backup volume names. For example:
+     ```json
+     $ backup ls s3://backupbucket@minio/ --volume-only
+     {
+       "pvc-004d8edb-3a8c-4596-a659-3d00122d3f07": {},
+       "pvc-7a8ded68-862d-4abb-a08c-8cf9664dab10": {}
+     }
+     ```
+   - `backup ls --volume <volume-name>`: List all backup names. For example:
+     ```json
+     $ backup ls s3://backupbucket@minio/ --volume pvc-004d8edb-3a8c-4596-a659-3d00122d3f07
+     {
+       "pvc-004d8edb-3a8c-4596-a659-3d00122d3f07": {
+         "Backups": {
+           "backup-02224cb26b794e73": {},
+           ...
+           "backup-fa78d89827664840": {}
+         }
+       }
+     }
+     ```
+   - `backup inspect-volume <volume>`: Read a single backup volume config (`volume.cfg`). For example:
+     ```json
+     $ backup inspect-volume s3://backupbucket@minio/?volume=pvc-004d8edb-3a8c-4596-a659-3d00122d3f07
+     {
+       "Name": "pvc-004d8edb-3a8c-4596-a659-3d00122d3f07",
+       "Size": "2147483648",
+       "Labels": {},
+       "Created": "2021-05-12T00:52:01Z",
+       "LastBackupName": "backup-c5f548b7e86b4b56",
+       "LastBackupAt": "2021-05-17T05:31:01Z",
+       "DataStored": "121634816",
+       "Messages": {}
+     }
+     ```
+   - `backup inspect <backup>`: Read a single backup config (`backup_backup_<backup-hash>.cfg`). For example:
+     ```json
+     $ backup inspect s3://backupbucket@minio/?backup=backup-fa78d89827664840\u0026volume=pvc-004d8edb-3a8c-4596-a659-3d00122d3f07
+     {
+       "Name": "backup-fa78d89827664840",
+       "URL": "s3://backupbucket@minio/?backup=backup-fa78d89827664840\u0026volume=pvc-004d8edb-3a8c-4596-a659-3d00122d3f07",
+       "SnapshotName": "backup-ac364071",
+       "SnapshotCreated": "2021-05-17T04:42:01Z",
+       "Created": "2021-05-17T04:42:03Z",
+       "Size": "115343360",
+       "Labels": {},
+       "IsIncremental": true,
+       "VolumeName": "pvc-004d8edb-3a8c-4596-a659-3d00122d3f07",
+       "VolumeSize": "2147483648",
+       "VolumeCreated": "2021-05-12T00:52:01Z",
+       "Messages": null
+     }
+     ```
+   - `backup head <config>`: Get the config metadata. For example:
+     ```json
+     {
+       "ModificationTime": "2021-05-17T04:42:03Z",
+     }
+     ```
+
+   Generally speaking, we want to separate the **list**, **read**, and **head** commands.
+
+2. The Longhorn manager HTTP endpoints.
+
+   | HTTP Endpoint                                                | Before                                                | After                                                                                                                                                |
+   | ------------------------------------------------------------ | ----------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------- |
+   | **GET** `/v1/backupvolumes`                                  | read all backup volumes from the remote backup target | read all the BackupVolume CRs                                                                                                                        |
+   | **GET** `/v1/backupvolumes/{volName}`                        | read a backup volume from the remote backup target    | read a BackupVolume CR with the given volume name                                                                                                    |
+   | **DELETE** `/v1/backupvolumes/{volName}`                     | delete a backup volume from the remote backup target  | delete the BackupVolume CR with the given volume name, `backup_volume_controller` reconciles to delete a backup volume from the remote backup target |
+   | **POST** `/v1/volumes/{volName}?action=snapshotBackup`       | create a backup to the remote backup target           | create a new Backup, `backup_controller` reconciles to create a backup to the remote backup target                                                   |
+   | **GET** `/v1/backupvolumes/{volName}?action=backupList`      | read a list of  backups from the remote backup target | read a list of Backup CRs with the label filter `volume=<backup-volume-name>`                                                                        |
+   | **GET** `/v1/backupvolumes/{volName}?action=backupGet`       | read a  backup from the remote backup target          | read a Backup CR with the given backup name                                                                                                          |
+   | **DELETE** `/v1/backupvolumes/{volName}?action=backupDelete` | delete a backup from the remote backup target         | delete the Backup CR, `backup_controller` reconciles to delete a backup from reomte backup target                                                    |
+
+## Design
+
+### Implementation Overview
+
+1. Create a new BackupTarget CRD `backuptargets.longhorn.io`.
+
+   ```yaml
+   metadata:
+     name: the backup target name (`default`), since currently we only support one backup target.
+   spec:
+     backupTargetURL: the backup target URL. (string)
+     credentialSecret: the backup target credential secret. (string)
+     pollInterval: the backup target poll interval. (metav1.Duration)
+     responsibleNodeID: the node ID which should responsible for run sync the remote backup target. (string)
+     syncRequestAt: the time to request run sync the remote backup target. (*metav1.Time)
+   status:
+     available: records if the remote backup target is available or not. (bool)
+     lastSyncedAt: records the last time the backup target was running the reconcile process. (*metav1.Time)
+   ```
+
+2. Create a new BackupVolume CRD `backupvolumes.longhorn.io`.
+
+   ```yaml
+   metadata:
+     name: the backup volume name.
+   spec:
+     syncRequestAt: the time to request run sync the remote backup volume. (*metav1.Time)
+     deleteRemoteConfig: indicate to delete the remote backup volume config or not. (bool)
+   status:
+     lastModificationTime: the backup volume config last modification time. (Time)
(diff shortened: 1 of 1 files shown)


joshimoo:
related #1761 I have some code stashed somewhere where I started on creating CR for backups, not sure how useful that code still is today :)

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,134 @@
...
+- `backup ls --volume <volume-name> --volume-only`: Read a single backup volume name.
+- `backup inspect <metadata-path>`: Read a single backup volume metadata (`volume.cfg`) or read a single volume's backup metadata (`backup_backup_<backup-hash>.cfg`).
+
+Generally speaking, we want to separate the **list** and **read** command.
+
+## Design
+
+### Implementation Overview
+
+We'll create a new Custom Resource Definition (CRD) called _backups`_ to save pulled backup volumes and its historical volume backups as the Custom Resource (CR).
```
shuo-wu, line 89: Can you talk more about this CRD?
Will Longhorn create one CR for each backup or backup volume record in the external backupstore? After reading this LEP, I still have no basic picture about this new CRD

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,134 @@
...
+
+There is already a backup store monitor that runs as a timer, the timer period is the setting `backupstore-poll-interval`. Within it, it runs:
```
shuo-wu, line 91: 1. Need to consider the modification of the backup target setting. If the setting is changed, all CRs should be removed, right?
2. Need to think about the CR update conflict, and how Longhorn balances the below workload to all workers of all nodes. Do we need a new controller for the new CRD?

shuo-wu (changes requested):

enhancements/20210525-async-pull-backups.md
jenting: I've updated it.

enhancements/20210525-async-pull-backups.md
jenting: > 1. Need to consider the modification of the backup target setting. If the setting is changed, all CRs should be removed, right?

I'd prefer to remove it when running synchronous in the controller, that is:
- Backup Volume:
1. Pull the backup volumes that are in the backup store and aren't in the cluster CR, create a new CR.
2. Delete the backup volumes that are in the cluster CR and aren't in the backup store, delete the CR.
- Volume Backups:
1. Calculate the difference set of the volume backups that are in the backup store and aren't in the cluster CR, add it to the backup volume CR's `spec.backups`.
2. Calculate the difference set of the volume backups that are in the cluster CR and aren't in the backup store, delete it from the backup volume CR's `spec.backups`
3. Update the backup volume CR

> 2. Need to think about the CR update conflict, and how Longhorn balances the below workload to all workers of all nodes. Do we need a new controller for the new CRD? I think Joshua mentions some ideas in a Github ticket. You can refer to it first.

Now, the backup store monitor running as a goroutine in the setting controller. My current thought is to keep it as-is.
I'm fine to create a new controller _or_ keep it in the current setting controller. WDYT?

I don't fully understand the question `how Longhorn balances the below workload to all workers of all nodes`. But I'll try to answer my understanding. The current [code](https://github.com/longhorn/longhorn-manager/blob/254aa17/controller/setting_controller.go#L645-L677) would run the workload randomly, this is how Joshua enhance it at v1.1.1 to prevent each node access the external backup store.

shuo-wu (changes requested):
1. Putting all info in `spec` rather than status is a little bit weird. By definition, `status` reflects the current situation of a resource, while `spec` reflects the desired situation.
2. The benefit of launching a new controller is that, each controller is responsible for monitoring and updating some backup volumes only. Currently, Longhorn picks up a random node and asks it to monitor all backup volumes, which means we will put all workloads/tasks on a node. Of course, the alternative is, launching a go monitor on each node and ask each monitor to check part of  backup volumes. 
3. Besides, it's better to put Resource creation, deletion, and spec update in one controller, and put the status update in another controller. In other words, you need to explain the life cycle management of new CRs in this LEP

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,173 @@
...
+- `backup inspect <metadata-path>`: Read a single backup volume metadata (`volume.cfg`) or read a single volume's backup metadata (`backup_backup_<backup-hash>.cfg`).
+
+Generally speaking, we want to separate the **list** and **read** command.
+
+## Design
+
+### Implementation Overview
+
+We'll create a new Custom Resource Definition (CRD) called `backups.longhorn.io` to save pulled backup volumes metadata 
+and its historical volume backups metadata as the Custom Resource (CR). For each backup volume, we'll create a new CR.
```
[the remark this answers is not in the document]
jenting: Good suggestion.

jenting:
> 1. Putting all info in `spec` rather than status is a little bit weird. By definition, `status` reflects the current situation of a resource, while `spec` reflects the desired situation.

I have thought this before. The reason why I choose to put it onto `spec` rather than `status` is that the field `spec`  is easier to make a change if we'd like to debug or test it via [kubectl edit CRD](https://github.com/longhorn/longhorn/wiki/dev:-How-to-modify-the-status-subresource-with-%60kubectl-edit%60-(CRD)).

Never mind, I'll move it from `spec` to `status` field.

> 2. The benefit of launching a new controller is that, each controller is responsible for monitoring and updating some backup volumes only. Currently, Longhorn picks up a random node and asks it to monitor all backup volumes, which means we will put all workloads/tasks on a node. Of course, the alternative is, launching a go monitor on each node and ask each monitor to check part of backup volumes.
> 3. Besides, it's better to put Resource creation, deletion, and spec update in one controller, and put the status update in another controller. In other words, you need to explain the life cycle management of new CRs in this LEP

I agree to create a separate control called `backup_volume_sync_controller` to handle synchronize the external backup store to cluster CR instead of putting this in `setting_controller`, this makes each controller is responsible for different task/function and the controller name is easier to know what's it's purpose.

However, I haven't thought the benefit of each node check part of the backup volumes because it'd make the logic inside different nodes more complicated. But I believe you want to achieve that if a backup volume (backup-volume-A) existed in the cluster and located in a node (Node-1), then this node ((Node-1) is responsible for the backup volume (backup-volume-A), right?

I'll update the life cycle management of new CRs but now my idea is spec update and status update all in one controller `backup_volume_sync_controller`.

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,173 @@
...
+- `spec.backingImageName`: the backingImageName field inside the backup volume metadata.
+- `spec.backingImageURL`: the backingImageURL field inside the backup volume metadata.
+- `spec.baseImage`: the baseImage field inside the backup volume metadata.
+- `spec.backups`: a map of volume backups (key is volume backup name, value is the volume backup metadata).
+  - `spec.backups[volumeBackupName].name`: the name field inside a historical volume backup metadata.
+  - `spec.backups[volumeBackupName].url`: the url field inside a historical volume backup metadata.
+  - `spec.backups[volumeBackupName].snapshotName`: the snapshotName field inside a historical volume backup metadata.
+  - `spec.backups[volumeBackupName].snapshotCreated`: the snapshotCreated field inside a historical volume backup metadata.
+  - `spec.backups[volumeBackupName].created`: the created field inside a historical volume backup metadata.
+  - `spec.backups[volumeBackupName].size`: the size field inside a historical volume backup metadata.
```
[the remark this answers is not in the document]
jenting: Yes, the volume backup data size.

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,173 @@
...
+  - `spec.backups[volumeBackupName].labels`: the labels field inside a historical volume backup metadata.
+  - `spec.backups[volumeBackupName].volumeName`: the volumeName field inside a historical volume backup metadata.
+  - `spec.backups[volumeBackupName].volumeSize`: the volumeSize field inside a historical volume backup metadata.
+  - `spec.backups[volumeBackupName].volumeCreated`: the volumeCreated field inside a historical volume backup metadata.
+  - `spec.backups[volumeBackupName].volumeBackingImageName`: the volumeBackingImageName field inside a historical volume backup metadata.
+  - `spec.backups[volumeBackupName].volumeBackingImageURL`: the volumeBackingImageURL field inside a historical volume backup metadata.
```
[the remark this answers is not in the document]
jenting: Okay, I removed all backing image related fields.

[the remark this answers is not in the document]
shuo-wu: Please retain `volumeBackingImageName`. This field is still available

shuo-wu:
If you plan to add a controller, you can call it `backup_volume_controller`. `sync` is redundant. 

Adding field `OwnerID` for the new CR is enough: first-come, first-served. Since there is no preference when choosing the OwnerID for this CR. 

If you want to manually modify the status, you can refer to this doc: https://github.com/longhorn/longhorn/wiki/dev:-How-to-modify-the-status-subresource-with-%60kubectl-edit%60-(CRD)

After moving fields from `spec` to `status`, there is almost nothing in `spec`. If `spec` contains fields like `backupTarget`, you can consider updating the `spec` in the setting controller.

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,169 @@
...
+Generally speaking, we want to separate the **list** and **read** command.
+
+## Design
+
+### Implementation Overview
+
+We'll create a new Custom Resource Definition (CRD) called `backupvolumes.longhorn.io` to save pulled backup volumes metadata 
+and its historical volume backups metadata as the Custom Resource (CR). For each backup volume, we'll create a new CR.
+- `metadata.name`: the backup volume name.
+- `status.name`: the name field inside the backup volume metadata.
```
shuo-wu, line 224: Remove this one. The backup volume name is already in `metadata`

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,169 @@
...
+For the longhorn manager endpoints:
+- **GET** `/v1/backupvolumes`: read all backup volumes from the cluster custom resource (CR).
+- **GET** `/v1/backupvolumes/{volName}`: read a single backup volume from the cluster custom resource (CR).
+- **DELETE** `/v1/backupvolumes/{volName}`: delete a single backup volume from the cluster custom resource (CR) and also from external backup store.
+- **GET** `/v1/backupvolumes/{volName}?action=backupList`: read a single volume backups from the cluster custom resource (CR).
+- **GET** `/v1/backupvolumes/{volName}?action=backupGet`: read a single volume backup from the cluster custom resource (CR).
+- **DELETE** `/v1/backupvolumes/{volName}?action=backupDelete`: delete a single volume backup from the cluster custom resource (CR) and also from external backup store.
+
+### Test plan
+
```
shuo-wu, line 273: Test the case: 2 clusters share the same backupstore

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,169 @@
...
+   After `backupstore-poll-interval` seconds, list backup volumes on the Longhorn GUI, the deleted one does not exist.
+6. When the user deletes a historical volume backup on the external backup store manually.
+   After `backupstore-poll-interval` seconds, list historical volume backups on the Longhorn GUI, the deleted one does not exist immediately.
+
+### Upgrade strategy
+
+Before this enhancement, the user might set `backupstore-poll-interval` to `0` to reduce the request to the external backup store.
+This is majorly for the user who wants to reduce the cost on access AWS S3.
+
+After this enhancement, if the user ever configured the setting `backupstore-poll-interval` to `0`, then the user is unable to list backup volumes
```
shuo-wu, line 294: Can we just add a warning in UI rather than directly forbidding listing (historical) backup volumes?

shuo-wu (changes requested):
1. Some fields in the backup volume/backup cfg are confusing. Can we rename it in this CR? e.g.,
`status.created`, `status.backups[volumeBackupName].url`, `status.backups[volumeBackupName].created`...
2. Remember to consider that there can be multiple clusters using the same backupstore. This means the backup/backup volumes would be suddenly create/removed without HTTP API calls

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,300 @@
+# Title
+
+Asynchronous pull backup volumes and historical volume backups into cluster custom resource (CR).
```
innobead, line 3:
```suggestion
Final consistent backup volumes with historical volume backups from the remote backup store
```

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,300 @@
...
+
+## Summary
+
+Asynchronous way to pull backup volumes and historical volume backups from the external backup store (S3/NFS) into cluster custom resource (CR).
+This enhancement is to fix the timeout issue when listing backup volumes from the external backup store _or_ listing historical volume backups 
+from the external backup store.
```
innobead, line 9:
```suggestion
Currently, Longhorn uses a blocking way for communication with the remote backup store, so there will be some potential voluntary or involuntary factors (ex:  network latency) impacting the functions relying on remote backup stores like listing volume backups or even causing further [cascading problems](###related-issues) after the backup store operation.

This enhancement is to propose an asynchronous way to pull backup volumes and historical volume backups from the external backup store (S3/NFS) then persistently saved via cluster custom resources. 

This can resolve the problems above mentioned by asynchronously querying the list of backup volumes and historical volume backups from the remote backup store for final consistent available results. It's also scalable for the costly resources created by the original blocking query operations.
```

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,300 @@
...
+- https://github.com/longhorn/longhorn/issues/2536
+- https://github.com/longhorn/longhorn/issues/2543
+
+## Motivation
+
+### Goals
+
+Decrease the time when the user clicks list backup _or_ list historical volume backups under the circumstances that lots of volume backups _or_ 
+lots of historical volume backups in the external backup store **and** the latency from Longhorn cluster to the external backup store is high.
+
```
innobead, line 24:
```suggestion
Decrease the query latency when listing volume backups _or_ historical volume backups in the circumstances like lots of volume backups, lots of historical volume backups, or the network latency between the Longhorn cluster and the remote backup store.
```

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,300 @@
...
+### Non-goals [optional]
+
+Automatically adjust the external backup store pull period.
+
+## Proposal
+
+Currently, we have a setting `backupstore-poll-interval` to periodically pull the external backup store lastest volume backups within a controller.
+We want to leverage it to pull the external backup volumes and historical volume backups from the external backup store and saves them into the cluster 
+custom resource (CR). Therefore, we'll create a new Custom Resource Definition (CRD) called _backups_ to save the backup volumes and historical volume backups.
```
innobead, line 33:
```suggestion
custom resource (CR). Therefore, we'll create a new Custom Resource Definition (CRD) called _backupvolumes_ to save the backup volumes and historical volume backups.
```

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,300 @@
...
+
+With this enhancement, we want to:
+1. Pull the backup volumess and historical volume backups that **are in** the external backup store and **aren't in** the cluster CR only.
```
innobead, line 36:
```suggestion
1. Pull the backup volumes and historical volume backups that **are in** the external backup store and **aren't in** the cluster CR only.
```

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,300 @@
...
+Then if the user clicks the `Backup` on the GUI, the user might hit list backup volumes _or_ list historical volume backups timeout issue 
+(the default timeout is 1 min).
+
+We choose to not create a new setting for the user to increase the list timeout value is because the browser has its timeout value also.
+Let's say the list backups needs 5 minutes to finish. Even we allow the user to increase the longhorn manager list timeout value, we can't change 
+the browser default timeout value. Furthermore, some browser doesn't allow the user to change default timeout value like Google Chrome.
+
+After this enhancement, when the user's environment under the circumstances that the external backup store has lots of backup volumes _or_ 
+historical volume backups, and the latency between the longhorn manager to the external backup store is high.
+Then if the user clicks the `Backup` on the GUI, the user can list backup volumes _or_ list historical volume backups without timeout issue.
```
innobead, line 54:
```suggestion
Then if the user clicks the `Backup` on the GUI, the user can eventually list backup volumes _or_ list historical volume backups without timeout issue.
```

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,300 @@
...
+
+Before this enhancement, the user might set `backupstore-poll-interval` to `0` to reduce the request to the external backup store.
+This is majorly for the user who wants to reduce the cost on access AWS S3.
+
+After this enhancement, if the user ever configured the setting `backupstore-poll-interval` to `0`, then the user is unable to list backup volumes
+_or_ historical volume backups. We should add this note to the documentation and the release note.
+
+## Note
+
+With this enhancement, the CR _backups.longhorn.io_ is updated when the asynchronous timer be triggered.
```
innobead, line 299:
```suggestion
With this enhancement, the CR _backupvolumes.longhorn.io_ is updated when the asynchronous timer be triggered.
```

innobead (changes requested):
In general, LGTM but some comments required to resolve.

P.S. except for pulling, I would suggest when doing the implementation, you can also check if there are some possible solutions for `push mode` like watch/notify when the backup info changes in the remote backup store, so we can efficiently do the pulling when there is something really changed.

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,300 @@
...
+      "Created": "2021-05-12T00:52:01Z",
+      "LastBackupName": "backup-c5f548b7e86b4b56",
+      "LastBackupAt": "2021-05-17T05:31:01Z",
+      "DataStored": "121634816",
+      "Messages": {}
+    }
+    ```
+  - Read a single volume's backup metadata (`backup_backup_<backup-hash>.cfg`). For example:
+    ```shell
+    $ backup inspect s3://backupbucket@minio/?backup=backup-fa78d89827664840\u0026volume=pvc-004d8edb-3a8c-4596-a659-3d00122d3f07
```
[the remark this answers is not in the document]
jenting: How about this? Since the URL and volume name is mandatory, but the backup name is optional.
```
backup inspect <URL> <volume-name> --backup <backup-name>
```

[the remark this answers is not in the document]
innobead: Good to me. Actually, the mandatory option is better to use `argument` :+1:

enhancements/20210525-async-pull-backups.md
jenting: Good suggestion, thank you.

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,348 @@
...
+
+2. At the existed `setting_controller`.
+   
+   We'll remove the existed code on watches the change of CR `settings.longorn.io` field `backup-target`, and `backup-target-credential-secret`, and `backupstore-poll-interval`. Also, we'll remove the backup store monitor in the setting controller. The backup store monitor run as a goroutine to 
+   periodically pull the latest volume backup name and volume backup timestamp, and updates the `status.LastBackup` and `status.LastBackupAt` to the 
+   CR `volumes.longhorn.io` if it's a DR volume.
+   
+3. Create a new controller called `backup_volume_controller`.
+   
+   It watches the change of the settings `backup-target`, and `backup-target-credential-secret`, and `backupstore-poll-interval`. Within the controller, it runs:
```
innobead, line 289: I assume AssumRole logic should take effect automatically.a

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,348 @@
...
+from longhorn manager to the remote backup store:
+1. The user can list backup volumes and list historical volume backups on the Longhorn GUI.
+2. When the user deletes a backup volume on the Longhorn GUI:
+   1. list backup volumes on the Longhorn GUI, the deleted one does not exist immediately.
+   2. check the remote backup store, the backup volume will be deleted after a while.
+3. When the user deletes a historical volume backup on the Longhorn GUI:
+   1. list historical volume backups on the Longhorn GUI, the deleted one does not exist immediately.
+   2. check the remote backup store, the historical backup will be deleted after a while.
+4. When the user deletes a backup volume or deletes a historical volume backup on the remote backup store manually.
+   After `backupstore-poll-interval` seconds, list backup volumes or list historical volume backups on the Longhorn GUI, the deleted one does not exist anymore.
```
innobead, line 326: Before this period, just make sure that any operation against to the missing volume backup should be failed w/ error message.

innobead (approved):
In general, LGTM. 

Some comments need to be noted. Leaving to @shuo-wu or @joshimoo to review before the merge.

shuo-wu (changes requested):
If possible, I still prefer to separate backupvolume creation & deletion from backup_volume_controller. Here is the possible solution:
1. For setting_controller, it can take care of:
    1. Update `spec.backupStore` and `spec.pollInterval` if the settings change
    2. List backup volumes (volume only), then create or delete CRs.
2. For backup_volume_controller, it should consider:
    1. inspect all backups for **each** backup volume then update **status**.
    2. how to archive **periodic** inspection based on the poll interval? There may be 3 possible options: 
        1. Somehow modifying [the resync period](https://github.com/longhorn/longhorn-manager/blob/52db49fce3b9b8f2112404d837fb64eaeb5f67e3/controller/controller_manager.go#L64-L71) for this controller only. But I am not sure if it is possible. 
        2. Launching a goroutine for each backupvolume. But there may be  lots of goroutines if there are hundreds of backupvolumes. It may be a waste of resources.
        3. Add `spec.RequestUpdateAt`. This field will be updated by setting_controller (periodically) or HTTP APIs(immediately). The field update can trigger backup_volume_controller. Then backup_volume_controller can blindly sync with the remote backupstore and update a new field named `status.LastUpdateAt`.

jenting:
> If possible, I still prefer to separate backupvolume creation & deletion from backup_volume_controller. Here is the possible solution:
> 
> 1. For setting_controller, it can take care of:
>    
>    1. Update `spec.backupStore` and `spec.pollInterval` if the settings change
>    2. List backup volumes (volume only), then create or delete CRs.

Hmmm.. if we have two controllers running the access remote backup store, this means it'll duplicate code on check the poll interval. I'm thinking what's the pros of separate sync backup volumes and volume snapshot backups on the different controllers.

> 2. For backup_volume_controller, it should consider:
>    
>    1. inspect all backups for **each** backup volume then update **status**.
>    2. how to archive **periodic** inspection based on the poll interval? There may be 3 possible options:
>       
>       1. Somehow modifying [the resync period](https://github.com/longhorn/longhorn-manager/blob/52db49fce3b9b8f2112404d837fb64eaeb5f67e3/controller/controller_manager.go#L64-L71) for this controller only. But I am not sure if it is possible.
>       2. Launching a goroutine for each backupvolume. But there may be lots of goroutines if there are hundreds of backupvolumes. It may be a waste of resources.

Or we could launch just one goroutine. This is like the current backup store monitor implementation https://github.com/longhorn/longhorn-manager/blob/254aa17/controller/setting_controller.go#L297

>       3. Add `spec.RequestUpdateAt`. This field will be updated by setting_controller (periodically) or HTTP APIs(immediately). The field update can trigger backup_volume_controller. Then backup_volume_controller can blindly sync with the remote backupstore and update a new field named `status.LastUpdateAt`.

In fact, we could implement by checking `(time.Now() - status.LastUpdateAt) > spec.pollInterval`? If yes, run the sync logic; otherwise, skip this reconcile loop. The only drawback is if the `spec.pollInterval < the controller resync period`, the poll interval equals to the controller resync period.

joshimoo:
Here is some idea of a design I would probably go with, if you want to go a different way that's fine.

------
Add a BackupTarget CRD, that defines all necessary information for the target + poll interval.
This way the user can add multiple backup targets easily and we can remove the settings and instead allow defining a default target as part of the installation process or by the creating a CRD.

Add BackupTarget controller, that is responsible for polling the backup target and managing the cluster Backup CRD resources (add / delete) based on new information on the backup target.

-------
Add a Backup CRD, that defines all necessary information for the actual backup and is responsible for the creation / deletion of that backup. Add a finalizer to the backup crds, so that deletion of the CRD can be verified and then removed.

This way the user/we can initiate a backup solely by the creation of a CRD and manage the lifecycle of that backup as well.
The collection of backups with Label vol == X represents a backup volume.

Add a Backup controller that is responsible for managing the lifecycle of a backup.

-----
(optional, but probably nice to have) 
Add a BackupVolume CRD that contains summary information about all the volume backups.
I am not sure we need this, but if we do we can have a BackupVolume controller that updates the BackupVolume summary information based on the existing Backup CRD's. This can also contain the Recurring Job information

--------------------------------
With the above design we can easily add the following user features:

- Add a DR-controller that is responsible for the creation of DR-Volumes / lastest backup updates of all volumes that are present on the BackupTarget when the Target/BackupVolume is marked for this.
This feature would add values to our users, since now they can just mark a single BackupTarget and we will automatically replicate all volumes from a different cluster, so the whole cluster is ready for failover, previously the user needs to manually create each DR volume.

- Allows the user to manage creation/deletion of Backups via creation of Backup CRD's
This makes it easy for the user to manage recurring backups themselves but also creation of exceptional case backups, for example create a backup for each volume before node maintenance. Is easily done by creation of a BackupCRD per volume.

- Simplifies the CSI snapshot mechanism
- Persists the Backup information (1=1 mapping of snapshots-backup per backup target)

-----------------------------

If you want to explore this design further let me know and I will help you flesh it out :)

jenting:
> Here is some idea of a design I would probably go with, if you want to go a different way that's fine.
> 
> Add a **BackupTarget** CRD, that defines all necessary information for the target + poll interval.
> This way the user can add multiple backup targets easily and we can remove the settings and instead allow defining a default target as part of the installation process or by the creating a CRD.

When designing this LEP, I thought this, and this will benefit if we want to support HA backup target in the future.
In fact, when designing this LEP, I'm a bit struggle with should we have the `status.LastUpdateAt` in each BackupVolume CR?
I'd prefer put the `status.LastUpdateAt` to BackupTarget CR.

However, to make this LEP focus on async poll backup volumes, so I drop the design and only focus on the BackupVolume CRD.
But if other @longhorn/maintainer think it's good to have the BackupTarget CRD along with BackupVolume CRD, I'm fine to update this LEP. What does other thinking?

> Add a **Backup** CRD, that defines all necessary information for the actual backup and is responsible for the creation / deletion of that backup. Add a finalizer to the backup crds, so that deletion of the CRD can be verified and then removed.
> 
> This way the user/we can initiate a backup solely by the creation of a CRD and manage the lifecycle of that backup as well.
> The collection of backups with Label vol == X represents a backup volume.
> 
> Add a Backup controller that is responsible for managing the lifecycle of a backup.
> (optional, but probably nice to have)
> Add a **BackupVolume** CRD that contains summary information about all the volume backups.
> I am not sure we need this, but if we do we can have a BackupVolume controller that updates the BackupVolume summary information based on the existing Backup CRD's. This can also contain the Recurring Job information
> 
Could you please explain more on separate the Backup CRD and BackupVolume CRD pros and cons?

Furthermore, the BackupVolume CRD is still needed since there is some information contains in the backup volume metadata (`volume.cfg`) like the volume size.

shuo-wu:
Having duplicate poll code is better than asking a controller to create/delete its CRs or update spec for the CRs. The responsibility of a controller is always to push and update the current status to the spec by issuing some operations. It does not involve in updating the CR spec or managing the lifecycle.

---

Launching CRD BackupTarget is a good idea. IMO, the pros are:
    1. It separates the responsibility of creating/deleting missing/redundant BackupVolume (or Backup) CRs from setting_controller. So that setting_controller won't be overwhelmed by the backup polling.
    2. It allows easily defining default backup target during Longhorn installation.
    3. HA backup target? Not sure if this is required

---

Separating the Backup CRD from BackupVolume CRD means, Longhorn doesn't need to send requests to engines/replicas in the API layer when receiving backup-related calls. Instead, backup_controller will sends/retry sending requests to the engines based on the Backup CRs

jenting:
@shuo-wu @joshimoo This is ready to review, my only concern is Note 2.
https://github.com/longhorn/longhorn/blob/da882bb33fcfd066619cd20223bf3ebb12383d37/enhancements/20210525-async-pull-backups.md#note
I can't think of a way from backup_controller to notify backup_volume_controller to update the newly created backup volume status. This would make if the user clicks `Create Backup` -> `Backup`, the user can't see the newly created backup volume until the backup volume controller updates the BackupVolume CR status from the remote backup store.

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,423 @@
...
+   3. Check if BackupVolume CR `spec.forceSync == false` and if the default BackupTarget CR `spec.PollInterval == 0` _or_ `time.Now() - the default BackupTarget CR status.lastSyncedAt < spec.pollInterval`, abort the current reconcile process.
+   4. Call the longhorn engine to list all the backups `backup ls --volume <volume-name>` from the remote backup target `backupStoreBackups`.
+   5.  List in cluster Backup CRs `clusterBackups`.
+   6.  Find the difference backups `backupsToPull = backupStoreBackups - clusterBackups` and create Backup CR `metadata.name` + `metadata.labels["longhornvolume"]=<backup-volume-name>`.
+   7.  Find the difference backups `backupsToDelete = clusterBackups - backupStoreBackups` and delete Backup CR.
+   8.  Call the longhorn engine to get the backup volume config's modification time `backup config-metadata <volume-config>`. If the config metadata not changed, abort the current reconcile process.
+   9. Call the longhorn engine to read the backup volumes' config `backup inspect-volume <volume-name>`.
+   10. Updates the BackupVolume CR status field according to the backup volumes' config, and also updates the BackupVolume CR `status.lastSyncedAt`.
```
shuo-wu, line 346: Should these 2 steps stand behind step 3? If there are backup updates, there must be modifications happening on the config file. 
Besides, I am thinking about the necessity of API `backup config-metadata`. If we can directly identify if there are modifications by checking the content of the config file, why do we want an extra API call? There are too much data in the config file, and checking the last modified time of the config file would save some bandwidth cost?

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,423 @@
...
+   11. Updates the BackupVolume CR spec field `spec.forceSync` and `spec.configModificationTime`.
+   12. Updates the Volume CR `status.lastBackup` and `status.lastBackupAt`.
+
+8.  For the Longhorn manager HTTP endpoints:
+
+   - **POST** `/v1/volumes/{volName}?action=snapshotBackup`:
+     1. Generate the backup name <backup-name>.
```
shuo-wu, line 353: Need to clarify if Longhorn should create a BackupVolume CR in this API when it doesn't exist.

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,423 @@
...
+          labels: 
+            longhornvolume: <backup-volume-name>
+        spec:
+          snapshotName: <snapshot-name>
+          labels: <snapshot-backup-labels>
+          backingImage: <backing-image>
+          backingImageURL: <backing-image-URL>
+        ```
+   - **DELETE** `/v1/backupvolumes/{volName}?action=backupDelete`:
+     1. Add the finalizer to the Backup CR with the given backup name.
```
shuo-wu, line 367: Adding the finalizer here? Shouldn't be the finalizer set once the CR is created?

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,423 @@
...
+     2. Delete a Backup CR with the given backup name.
+
+9.  Create a new controller `backup_controller`.
+
+    Watches the change of Backup CR. The backup controller is responsible to update the Backup CR status field and create/delete backup to/from the remote backup target. The reconcile loop steps are:
+    1.  Get the default BackupTarget CR.
+    2.  If the delete Backup CR event comes in and the finalizer configured, delete the backup volume from the remote backup target `backup rm --volume <volume-name> <url>`.Then, updates BackupVolume CR `spec.forceSync=true` to request backup_volume_controller to immediately reconcile the backup volume. After that, remove the finalizer.
+    3.  Check if the Backup CR `spec.snapshotName != ""` and `status.backupCreate == false`. If yes, call longhorn engine/replica for backup creation. Then fork a go routine to monitor the backup creation progress. After that, updates Backup CR `status.backupCreate = true`.
```
shuo-wu, line 375: when will `status.backupCreateAt` be set? Can we use this field to indicate if the backup creation is complete?

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,423 @@
...
+     backupCreateAt: the snapshot backup creation time. (string)
+     size: the snapshot size. (string)
+     labels: the labels of snapshot backup. (map[string]string)
+     messages: the error messages when calling longhorn engine on listing or inspecting backups. (map[string]string)
+     lastSyncedAt: records the last time the backup was synced into the cluster. (*metav1.Time)
+   ```
+
+4. At the existed `setting_controller`.
+
+   Watches the changes of Setting CR `settings.longorn.io` field `backup-target`, `backup-target-credential-secret`, and `backupstore-poll-interval`. The setting controller is responsible to create/update the default BackupTarget CR.
```
shuo-wu, line 311: What will happen if the backup target changes? Add it's better to add a test for this case.

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,423 @@
...
+   - **GET** `/v1/backupvolumes`: read all the BackupVolume CRs.
+   - **GET** `/v1/backupvolumes/{volName}`: read a BackupVolume CR with the given volume name.
+   - **GET** `/v1/backupvolumes/{volName}?action=backupList`: read a list of Backup CRs with the label filter `volume=<backup-volume-name>`.
+   - **GET** `/v1/backupvolumes/{volName}?action=backupGet`: read a Backup CR with the given backup name.
+
+### Test plan
+
+With over 1k backup volumes and over 1k backups under pretty high network latency (700-800ms per operation)
+from longhorn manager to the remote backup target:
+- Create a single cluster with 2 volumes (vol-A and vol-B).
```
shuo-wu, line 393: It's better to mention the main purpose of each test.

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,423 @@
...
+   ```
+
+2. Create a new BackupVolume CRD `backupvolumes.longhorn.io`.
+
+   ```yaml
+   metadata:
+     name: the backup volume name.
+   spec:
+     forceSync: to force sync the backup volume or not. (bool)
+     configModificationTime: the backup volume config modification time. (Time)
```
shuo-wu, line 271: 1. Do we really need this? I don't find when the controllers will use it.
2. If YES, do not put this field in `spec`.

shuo-wu (changes requested):

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,423 @@
...
+   1. Get the default BackupTarget CR.
+   2. Check if the default BackupTarget CR `spec.PollInterval == 0` _or_ `spec.forceSync == false`. If yes, abort the current reconcile process. It no, continue the reconcile process.
+   3. Check if `time.Now() - the default BackupTarget CR status.lastSyncedAt >= spec.pollInterval`.
+     - If no, skip the current reconcile process.
+     - If yes:
+       1. Call the longhorn engine to list all the backup volumes `backup ls --volume-only` from the remote backup target `backupStoreBackups`. If the remote backup target is not available, updates the BackupTarget CR `status.available=false` and `status.lastSyncedAt=time.Now()`, then skip the current reconcile process.
+       2. List in cluster BackupVolume CRs `clusterBackupVolumes`.
+       3. Find the difference backup volumes `backupVolumesToPull = backupStoreBackups - clusterBackupVolumes` and create BackupVolume CR `metadata.name` + `spec.forceSync = true`.
+       4. Find the difference backup volumes `backupVolumesToDelete = clusterBackupVolumes - backupStoreBackups` and delete BackupVolume CR.
+       5. Updates the BackupTarget CR `spec.forceSync = false` if `spec.forceSync = true`.
```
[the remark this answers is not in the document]
jenting: okay!

enhancements/20210525-async-pull-backups.md
jenting: Before backup refactoring,
- Longhorn list and read BackupVolume and/or Backup config from the remote backup store when the user clicks the `Backup` at the GUI.
- Longhorn list and read BackupVolume config within the setting_controller timer (the timer period is the poll interval).

After backup refactoring, the backup_volume_controller and backup_controller be triggered every 30 secs. It means every 30 secs, the read API be triggered. Moreover, the Read API is a time-consuming API than the HEAD API.

I was thinking that if the Backup config has been synced, do not read the Backup config again. Because in reality, the backup config should be created once.
However, when I look at the integration test, Longhorn did a test that the Backup config might be corrupted after a successful backup config was created.

enhancements/20210525-async-pull-backups.md
jenting: It's a bit tricky for the volume that has never run snapshot backup before.

When the user creates volume backup, the Backup CR is created, and the BackupVolume CR is created after the BackupVolume Config be pulled from the remote backup store. (It depends on the backup store poll interval).

If the backup store poll interval is 1 hour, it means that the user could see the BackupVolume on the GUI after 1 hour. Therefore, the BackupVolume CR is created for a better user experience.

enhancements/20210525-async-pull-backups.md
jenting: If the Backup CR is created with the finalizer. When the user deletes the remote backup store Backup config, the backup_volume_controller will delete the Backup CR, then the backup_controller reconciles the Delete event to delete the remote backup store Backup even if has been deleted already.

To add the finalizer is when the remote backup store config also needs to do clean up. That's why I only add the finalizer when the HTTP DELETE API be called.

enhancements/20210525-async-pull-backups.md
jenting: Yes, when the backup creation complete.

enhancements/20210525-async-pull-backups.md
jenting: The backup target changes did nothing, the backup_target_controller reconciles to create/delete BackupVolume CRs and back_volume_controller reconciles to create/delete Backup CRs.

The only to note is when backup target is not accessible, the BackupVolume CRs and Backup CRs will be deleted.

shuo-wu: If Longhorn decides to create BackupVolume CR in this HTTP API, it can set `backuptarget.spec.forceSync` to avoid waiting for the pull interval. 
If you want to keep the current design unchanged, it's better to explain this case in the doc.

enhancements/20210525-async-pull-backups.md
shuo-wu: If Longhorn decides to create BackupVolume CR in this HTTP API, it can set `backuptarget.spec.forceSync` to avoid waiting for the pull interval. 
If you want to keep the current design unchanged, it's better to explain this case in the doc.

enhancements/20210525-async-pull-backups.md
shuo-wu: Got it. After syncing with the new backup target, the controllers will delete all expired CRs automatically as well as creating new CRs based on the backup target contents.

enhancements/20210525-async-pull-backups.md
shuo-wu: I prefer the assumption you raised here. 
When backup-controller sends a delete call to the remote backup store for a non-existing backup file, it will receive either a not-found error or a success response. Then backup-controller knows the cleanup is done and it's safe to delete the CR itself by upsetting the finalizer. With this design, the deletion workflow is consistent with other kinds of CRs.

enhancements/20210525-async-pull-backups.md
shuo-wu: Then do we still need `status.backupCreate`?

enhancements/20210525-async-pull-backups.md
joshimoo: The finalizers should be added as part of CR creation.
And only removed after validation that the necessary resources have been cleaned up.

In this case, that the backup.cfg is truly gone on the backupstore.
As @shuo-wu mentioned a missing (NotFound) backup.cfg can be treated as a successful delete.

Just to elaborate a finalizer should be used to prevent deletion, till all necessary validation has been done. 

Some examples below:
(dependent resources have been removed, volume is safe to delete, replica data has been deleted on the file system, node is gone, etc)

enhancements/20210525-async-pull-backups.md
jenting: Yes, if the backup is in progress, the `status.BackupCreateAt` is empty. Then when the backup_controller reconciles again and sees the `status.BackupCreateAt` is empty, the back_controller calls the snapshot backup process again.

enhancements/20210525-async-pull-backups.md
jenting: Oops, I had a bad memory of this yesterday.

After re-read the code, Longhorn would not create BackupVolume CR in this HTTP API call.
Instead, after backup completion and the BackupVolume CR does not exist, it creates the BackupVolume CR and set `backupvolume.spec.forceSync=true` in the `backup_controller``.

But I like your approach that set `backuptarget.spec.forceSync` to avoid waiting for the pull interval :+1:

enhancements/20210525-async-pull-backups.md
jenting: Got it. I'll add the finalizer when the CR is created.

enhancements/20210525-async-pull-backups.md
shuo-wu: If Longhorn creates BackupVolume CR in the backup-controller, there is no need to set `backuptarget.spec.forceSync`.

Notice that Creating a BackupVolume CR at the beginning matches the current behavior: Before the 1st backup is created, users can find an empty BackupVolume on the UI backup page.

enhancements/20210525-async-pull-backups.md
shuo-wu: Oh, OK. This field actually indicates if the backup creation is **started** or not. Can you rename this field to avoid misunderstanding?

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,423 @@
...
+
+   Watches the change of BackupTarget CR. The backup target controller is responsible for creating/deleting BackupVolume CR metadata+spec. The reconcile loop steps are:
+   1. Get the default BackupTarget CR.
+   2. Check if the default BackupTarget CR `spec.PollInterval == 0` _or_ `spec.forceSync == false`. If yes, abort the current reconcile process. It no, continue the reconcile process.
+   3. Check if `time.Now() - the default BackupTarget CR status.lastSyncedAt >= spec.pollInterval`.
+     - If no, skip the current reconcile process.
+     - If yes:
+       1. Call the longhorn engine to list all the backup volumes `backup ls --volume-only` from the remote backup target `backupStoreBackups`. If the remote backup target is not available, updates the BackupTarget CR `status.available=false` and `status.lastSyncedAt=time.Now()`, then skip the current reconcile process.
+       2. List in cluster BackupVolume CRs `clusterBackupVolumes`.
+       3. Find the difference backup volumes `backupVolumesToPull = backupStoreBackups - clusterBackupVolumes` and create BackupVolume CR `metadata.name` + `spec.forceSync = true`.
```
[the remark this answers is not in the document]
jenting: Fix the typo.

enhancements/20210525-async-pull-backups.md
jenting: Okay, now I implement it as when the backup completion (progress 100%) then creating the BackupVolume CR.

enhancements/20210525-async-pull-backups.md
jenting: Change the wording to `status.backupCreationIsStarted`.

enhancements/20210525-async-pull-backups.md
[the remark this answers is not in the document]
jenting: Use `spec.forceSync` and `status.forceSynced` would be easier for the user to control directly patch the BackupTarget CR _or_ BackupVolume CR to immediately force sync the CRs.

For example:
```shell
kubectl patch lhbt default --type merge -p '{"spec":{"forceSync":true}}'

kubectl patch lhbv <backup-volume-name> --type merge -p '{"spec":{"forceSync":true}}'
```

enhancements/20210525-async-pull-backups.md
jenting: After recap my memory. The reason _why_ I only add finalizer on the HTTP Delete API is because I want to prevent if the user deletes the BackupVolume/Backup CR by accident. The controller reconciles to delete the remote config file.
Although we could add a field `spec.toDelete`, however, I realized that adding a finalizer would be much easier and simple to understand. WDYT? @shuo-wu @joshimoo

enhancements/20210525-async-pull-backups.md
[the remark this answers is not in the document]
jenting: I'm thinking that I could use the `status.LastSyncedAt` to controller whether to force sync the backup_volume_controller.

For example, in the backup_controller, after snapshot backup completion, set the BackupVolume CR `status.LastSyncedAt = nil`. At backup_volume_controller, since the `status.LastSyncedAt = nil`, it will run sync immediately.

But I'm not sure at your point, is it okay/safe for the backup_controller to update the BackupVolume CR `status` field?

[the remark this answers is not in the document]
shuo-wu: This is not recommended. The user should be responsible for updating the spec only. Then Kubernetes/Longhorn can do something based on the spec.
IMO, if users want to force sync a Backup or BackupVolume via CR update, they should set the timestamp correctly. we don't need to consider if the field can be manually set **easily** here.

enhancements/20210525-async-pull-backups.md
shuo-wu: When will the BackupVolume/Backup CRs be deleted by accident? During Longhorn uninstallation? For the uninstallation, Longhorn can remove/unset the backup target first then BackupVolume/Backup CR deletion won't lead to actual backup file cleanup.

enhancements/20210525-async-pull-backups.md
shuo-wu: As we discussed in the session, after the backup creation complete, the controller doesn't need to check the backup cfg file anymore. But the for the backup volume cfg file, I think the backup-volume controller would call the `head` API first then determine if we need to inspect the cfg file and list all backups when the sync is required.

enhancements/20210525-async-pull-backups.md
jenting: The users delete the BackupVolume/Backup CRs manually :(
Since the external backup is the last mile to protect the data, we should make sure the backup data won't be deleted if the user just accidentally deletes CRs.

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,448 @@
...
+    2. If the delete Backup CR event comes in:
+        1.  delete the backup from the remote backup target `backup rm <url>` if Backup CR `spec.deleteRemoteConfig=true`.
+        2.  update the BackupVolume CR `spec.syncRequestAt=time.Now()`.
+        3.  remove the finalizer.
+    3. Check if the Backup CR `spec.snapshotName != ""` and `status.backupCreationIsStart == false`. If yes:
+        1.  call longhorn engine/replica for backup creation.
+        2.  updates Backup CR `status.backupCreationIsStart = true`.
+        3.  fork a go routine to monitor the backup creation progress. After backup creation finished (progress = 100):
+            1.  update the BackupVolume CR `spec.syncRequestAt = time.Now()` if BackupVolume CR exist.
+            2.  create the BackupVolume CR `metadata.name` + `metadata.finalizers["longhorn.io"]` if BackupVolume CR not exist.
```
shuo-wu, line 390: There is no need to mention the finalizers during creation. 

Similar above.

shuo-wu (changes requested):

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,448 @@
...
+
+3. Create a new Backup CRD `backups.longhorn.io`.
+
+   ```yaml
+   metadata:
+     name: the backup name.
+     labels:
+       longhornvolume=<backup-volume-name>`: this label indicates which backup volume the backup belongs to.
+   spec:
+     deleteRemoteConfig: indicate to delete the remote backup config or not. (bool)
```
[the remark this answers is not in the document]
jenting: backup_controller calls the longhorn/backstore to delete remote backup config (and the related block files if needed) when delete CR event comes in and `spec.deleteRemoteConfig=true`.

enhancements/20210525-async-pull-backups.md
jenting: updated!

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,448 @@
...
+      2. deletes Backup CR with the given volume name.
+      3. deletes the backup volume from the remote backup target `backup rm --volume <volume-name> <url>` if `spec.deleteRemoteConfig=true`.
+      4. remove the finalizer.
+   3. Check if the `status.lastSyncedAt < spec.syncRequestAt`. If no, skip the reconcile process.
+   4. Call the longhorn engine to list all the backups `backup ls --volume <volume-name>` from the remote backup target `backupStoreBackups`.
+   5.  List in cluster Backup CRs `clusterBackups`.
+   6.  Find the difference backups `backupsToPull = backupStoreBackups - clusterBackups` and create Backup CR `metadata.name` + `metadata.labels["longhornvolume"]=<backup-volume-name>` + `metadata.finalizers["longhorn.io"]`.
+   7.  Find the difference backups `backupsToDelete = clusterBackups - backupStoreBackups` and delete Backup CR.
+   8.  Call the longhorn engine to get the backup volume config's last modification time `backup head <volume-config>` and compares to `status.lastModificationTime`. If the config last modification time not changed, abort the current reconcile process.
```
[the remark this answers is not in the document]
jenting: updated!

enhancements/20210525-async-pull-backups.md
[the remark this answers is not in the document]
shuo-wu: Then the naming of this field is confusing. How about naming it to something like `fileCleanupRequired`?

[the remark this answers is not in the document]
jenting: Sure~

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,476 @@
...
+   ```yaml
+   metadata:
+     name: the backup target name (`default`), since currently we only support one backup target.
+   spec:
+     backupTargetURL: the backup target URL. (string)
+     credentialSecret: the backup target credential secret. (string)
+     pollInterval: the backup target poll interval. (metav1.Duration)
+     syncRequestAt: the time to request run sync the remote backup target. (*metav1.Time)
+   status:
+     ownerID: the node ID which should responsible for run sync the remote backup target. (string)
```
innobead, line 260:
```suggestion
     ownerID: the node ID which is responsible for running operations of backup target controller (string)
```

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,476 @@
...
+2. Create a new BackupVolume CRD `backupvolumes.longhorn.io`.
+
+   ```yaml
+   metadata:
+     name: the backup volume name.
+   spec:
+     syncRequestAt: the time to request run sync the remote backup volume. (*metav1.Time)
+     fileCleanupRequired: indicate to delete the remote backup volume config or not. (bool)
+   status:
+     ownerID: the node ID which should responsible for run sync the remote backup target. (string)
```
innobead, line 274:
```suggestion
     ownerID: the node ID which is responsible for running operations of backup volume controller (string)
```

enhancements/20210525-async-pull-backups.md
```diff
@@ -0,0 +1,476 @@
...
+     labels:
+       longhornvolume=<backup-volume-name>`: this label indicates which backup volume the backup belongs to.
+   spec:
+     fileCleanupRequired: indicate to delete the remote backup config or not (and the related block files if needed). (bool)
+     snapshotName: the snapshot name. (string)
+     labels: the labels of snapshot backup. (map[string]string)
+     backingImage: the backing image. (string)
+     backingImageURL: the backing image URL. (string)
+   status:
+     ownerID: the node ID which should responsible for run sync the remote backup target. (string)
```
innobead, line 300:
```suggestion
     ownerID: the node ID which is responsible for running operations of the backup controller (string)
```

innobead (commented):
In general, LGTM.

@shuo-wu and @joshimoo do you have any concerns about this proposal?

shuo-wu (approved):
LGTM

PR #2656, opened 2021-06-03
bugfix: the environment check script should check all pods

laverya:
instead of just the first pod that it finds

"get desired number scheduled" logic is copied from `wait_ds_ready`.


innobead (approved):
LGTM

innobead:
@laverya thanks for the contribution :+1:

Issue #467, opened 2019-04-05
Default recurring backup policy


yasker:
We're need recurring backup policies.

It would be easier to apply backup policies to volumes instead of hand fill volumes' recurring schedule. Also, one of them can be set as default, which will be applied to all the volumes.

If the default policy was set, any newly created volume will automatically have this policy when created. The policy will be applied to all the existing and new volumes. Change of policy will result in a change of the existing volumes' schedulings as well.

One problem with the default backup policy is if we want to run that job exactly as users want, we will backup all the volumes at the exact same moment e.g. every day. That's not desired since it will cause high load and high bandwidth consumption at an arbitrary moment. Ideally, a range should be provided.

The policy can be set even when the backup target is empty, though the volumes won't be backed up if the backup target is empty.

Related to https://github.com/rancher/longhorn/issues/83

yasker:
Keep it open

c3y1huang:
At the moment the rough idea is to have a new backup-policy CRD and allows to add the volume name to each policy. When the volume name is updated to the policy the volume-controller will then make updates to the .spec.recurringJobs which then can be created/updated by the existing recurring job creation flow.

PR #2737, opened 2021-06-25
Add LEP - label-driven recurring job

c3y1huang:
https://github.com/longhorn/longhorn/issues/467

### DIFF
--- a/enhancements/20210624-label-driven-recurring-job.md
+++ b/enhancements/20210624-label-driven-recurring-job.md
@@ -1,0 +1,1043 @@
+# Label-driven Recurring Job
+
+## Summary
+
+Replace volume spec `recurringJobs` with the label-driven model. Abstract volume recurring jobs to a new CRD named "RecurringJob". The names or groups of recurring jobs can be referenced in volume labels.
+
+Users can set a recurring job to the `Default` group and Longhorn will automatically apply when the volume has no job labels.
+
+Only one cron job will be created per recurring job. Users can also update the recurring job, and Longhorn reflects the changes to the associated cron job.
+
+When instruct Longhorn to delete a recurring job, this will also remove the associated cron job.
+
+StorageClass now should use `recurringJobSelector` instead to refer to the recurring job names.
+
+During the version upgrade, existing volume spec `recurringJobs` and storageClass `recurringJobs` will automatically translate to volume labels, and recurring job CRs will get created.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/467
+
+## Motivation
+
+### Goals
+
+Phase 1:
+- Each recurring job can be in single or multiple groups.
+- Each group can have single or multiple recurring jobs.
+- The jobs and groups can be reference with the volume label.
+- The recurring job in `default` group will automatically apply to a volume that has no job labels.
+- Can create multiple recurring jobs with UI or YAML.
+- The recurring job should include settings; `name`, `groups`, `task`, `cron`, `retain`, `concurrency`, `labels`.
+
+Phase2:
+The StorageClass and upgrade migration are still dependent on volume spec, thus complete removal of volume spec should be done in phase 2.
+
+### Non-goals [optional]
+
+1. Does the snapshot/backup operation one by one. The operation order can be defined as sequential, consistent (for volume group snapshot), or throttled (with a concurrent number as a parameter) in the future.
+
+> https://github.com/longhorn/longhorn/pull/2737#issuecomment-887985811
+
+## Proposal
+
+#### Story 1 - set recurring jobs and groups by volume labels
+
+As a Longhorn user / System administrator.
+
+I want to directly update recurring jobs referenced in multiple volumes.
+
+So I do not need to update each volume with cron job definition.
+
+#### Story 2 - automatically applies for the default recurring jobs.
+
+As a Longhorn user / System administrator.
+
+I want the ability to set one or multiple `backup` and `snapshot` recurring jobs as default. All volumes without any recurring job label should automatically apply with the default recurring jobs.
+
+So I can be assured that all volumes without any recurring job label will automatically apply with default.
+
+#### Story 3 - automatically upgrade migration
+
+As a Longhorn user / System administrator
+
+I want Longhorn to automatically convert existing volume spec `recurringJobs` to volume labels, and create associate recurring job CRs.
+
+So I don't have to manually create recurring job CRs and patch the labels.
+
+### User Experience In Detail
+
+#### Story 1 - set recurring job in volume
+
+1. Create a recurring job on UI `Recurring Job` page or via `kubectl`.
+2. In UI, Navigate to Volume, `Recurring Jobs Schedule`.
+3. User can choose from `Job` or job `Group` from the tab.
+  - On `Job` tab,
+    1. User sees existing recurring jobs that volume had labeled.
+    1. User able to select `Backup` or `Snapshot` for the `Type` from the drop-down list.
+    2. User able to edit the `Schedule`, `Retain`, `Concurrency` and `Labels`.
+  - On the job `Group` tab.
+    1. User sees all existing recurring job groups from the `Name` drop-down list.
+    2. User selects the job from the drop-down list.
+    3. User sees all recurring `jobs` under the `group`.
+4. Click `Save` updates to the volume label.
+5. Update the recurring job CRs also reflect on the cron job and UI `Recurring Jobs Schedule`.
+
+**Before enhancement**
+
+Recurring jobs can only be added and updated per volume spec.
+
+Create cron job definitions for each volume causing duplicated setup effort.
+
+The recurring job can only be updated per volume.
+
+**After enhancement**
+
+Recurring jobs can be added and update as the volume label.
+
+Can select a recurring job from the UI drop-down menu and will automatically show the information from the recurring job CRs.
+
+Update the recurring job definition will automatically apply to all volumes with the job label.
+
+#### Story 2 - automatically apply to default recurring jobs
+
+1. Add `default` to one or multiple recurring jobs `Groups` in UI or `kubectl`.
+2. Longhorn automatically applies the `default` group recurring jobs to all volumes without job labels.
+
+**Before enhancement**
+
+Default recurring jobs are set via StorageClass at PVC creation only. No default recurring job can be set up for UI-created volumes.
+
+Updating StorageClass does not reflect on the existing volumes.
+
+**After enhancement**
+
+Have the option to set default recurring jobs via `StorageClass` or `RecurringJob`.
+
+Longhorn recurring job controller automatically applies default recurring jobs to all volumes without the job labels.
+
+Longhorn adds the default recurring jobs when all job labels are removed from the volume.
+
+When the `RecurringJobSelector` is set in the `StorageClass`, it will be used as default instead.
+
+#### Story 3 - automatically upgrade migration
+
+1. Perform upgrade.
+2. StorageClass `recurringJobs` will get convert to `recurringJobSelector`.
+3. Recurring job CRs will get created from `recurringJobs`.
+4. Volume will be labeled with `recurring-job.longhorn.io/<jobTask>-<jobRetain>-<hash(jobCron)>-<hash(jobLabelJSON)>: enabled` from volume spec `recurringJobs`.
+5. Recurring job CRs will get created from volume spec `recurringJobs`. When the config is identical among multiple volumes, only one will get created and volumes will share this recurring job CR.
+6. Volume spec `recurringJobs` will get removed.
+
+### API Changes
+
+- Add new HTTP endpoints:
+  - GET `/v1/recurringjobs` to list of recurring jobs.
+  - GET `/v1/recurringjobs/{name}` to get specific recurring job.
+  - DELETE `/v1/recurringjobs/{name}` to delete specific recurring job.
+  - POST `/v1/recurringjobs` to create a recurring job.
+  - PUT `/v1/recurringjobs/{name}` to update specific recurring job.
+  - `/v1/ws/recurringjobs` and `/v1/ws/{period}/recurringjobs` for websocket stream.
+- Add new RESTful APIs for the new `RecurringJob` CRD:
+  - `Create`
+  - `Update`
+  - `List`
+  - `Get`
+  - `Delete`
+- Add new APIs for users to update recurring jobs for individual volume. The :
+  - `/v1/volumes/<VOLUME_NAME>?action=recurringJobAdd`, expect request's body in form {name:<name>, isGroup:<bool>}.
+  - `/v1/volumes/<VOLUME_NAME>?action=recurringJobList`.
+  - `/v1/volumes/<VOLUME_NAME>?action=recurringJobDelete`, expect request's body in form {name:<name>, isGroup:<bool>}.
+
+## Design
+
+### Implementation Overview
+
+#### Add Recurring Job CRD.
+  - Update the ClusterRole to include `recurringjob`.
+  - Printer column should include `Name`, `Groups`, `Task`, `Cron`, `Retain`, `Concurrency`, `Age`, `Labels`.
+     ```
+     NAME        GROUPS                 TASK       CRON        RETAIN   CONCURRENCY   AGE   LABELS
+     snapshot1   ["default","group1"]   snapshot   * * * * *   1        2             14m   {"label/1":"a","label/2":"b"}
+
+     ```
+     - The `Name`: String used to reference the recurring job in volume with the label `recurring-job.longhorn.io/<Name>: enabled`.
+     - The `Groups`: Array of strings that set groupings to the recurring job. This is used to reference the recurring job group in volume with the label `recurring-job-group.longhorn.io/<Name>: enabled`. When including `default`, the recurring job will be added to the volume label if no other job exists in the volume label.
+     - The `Task`: String of either one of `backup` or `snapshot`.
+       Also, add validation in the CRD YAML with pattern regex match.
+     - The `Cron`: String in cron expression represents recurring job scheduling.
+     - The `Retain`: Integer of the number of snapshots/backups to keep for the volume.
+     - The `Concurrency`: Integer of the concurrent job to run by each cron job.
+     - The `Age`: Date of the CR creation timestamp.
+     - The `Labels`: Dictionary of the labels.
+
+#### Add Command `recurring-job` To `longhorn-manager` Binary
+
+  1. Add new command `recurring-job <job.name> --longhorn-manager <URL>` and remove old command `snapshot`.
+     > Get the `recurringJob.Spec` on execution using Kubernetes API.
+  2. Get volumes by label selector `recurring-job.longhorn.io/<job.name>: enabled` to filter out volumes.
+  3. Get volumes by label selector `recurring-job-group.longhorn.io/<job.group>: enabled` to filter out volumes if the job is set with a group.
+  4. Filter and create a list of the volumes in the state `attached` or setting `allow-recurring-job-while-volume-detached`.
+  5. Use the concurrent number parameter to throttle goroutine with channel. Each goroutine creates `NewJob()` and `job.run()` for the volumes.
+     The job `snapshotName` format will be `<job.name>-c-<RandomID>`.
+
+#### Changes In The Volume Controller
+
+  1. The `updateRecurringJobs` method is responsible to add the default label if not other labels exist.
+
+  > Since the storage class and upgrade migration contains recurringJobs spec. So we will keep the `VolumeSpec.RecurringJobs` in code to create the recurring jobs for volumes from the `storageClass`.
+
+  > In case names are duplicated between different `storageClasses`, only one recurring job CR will be created.
+
+#### Changes In The VolumeManager `CreateVolume`
+
+  - Add new method input `recurringJobSelector`:
+     1. Convert `Volume.Spec.RecurringJobs` to `recurringJobSelector`.
+     2. Add recurring job label if `recurringJobSelector` method input is not empty.
+
+#### Changes In The Datastore
+
+  - For `CreateVolume` and `UpdateVolume` add a function similar to `fixupMetadata` that handles recurring jobs:
+     1. Add recurring job labels if `Volume.Spec.RecurringJobs` is not empty. Then unset `Volume.Spec.RecurringJobs`.
+     2. Label with `default` job-group if no other recurring job label exists.
+
+#### Introduce `recurringJobSelector` As Part Of StorageClass Parameters.
+  - The CSI controller can use `recurringJobSelector` for volume creation.
+
+#### Changes In CSI Controller Server
+
+  1. Put `recurringJobSelector` to `vol.RecurringJobSelector` at HTTP API layer to use for adding volume recurring job label in `VolumeManager.CreateVolume`. The `CreateVolume` method will have a new input `recurringJobSelector`.
+  2. Get `recurringJobs` from parameters, validate and create recurring job CRs via API if not already exist.
+
+#### Add Recurring Job Controller
+
+  - The code structure will be the same as other controllers.
+  - Add the finalizer to the recurring job CRs if not exist.
+  - The controller will be informed by `recurringJobInformer` and `enqueueRecurringJob`.
+  - Create and update CronJob per recurring job.
+    1. Generate a new cron job object.
+       - Include labels `recurring-job.longhorn.io`.
+         ```
+         recurring-job.longhorn.io: <Name>
+         ```
+       - Compose command,
+         ```
+         longhorn-manager -d\
+           recurring-job <job.name>\
+           --manager-url <url>
+         ```
+    2. Create new cron job with annotation `last-applied-cronjob-spec` or update cron job if the new cron job spec is different from the `last-applied-cronjob-spec`.
+  - Use defer to clean up CronJob.
+    1. When a recurring job gets deleted.
+      1. Delete the cron job with selected labels: `recurring-job.longhorn.io/<Name>`.
+      2. Remove the finalizer.
+
+
+#### UI
+
+##### Add New Page `Recurring Job` In UI
+
+A new page for `Recurring Job` to create/update/delete recurring jobs.
+
+```
+Recurring Job                                                                                        [Custom Column]
+====================================================================================================================
+[Create] [Delete]                                                                  [Search Box   v ][__________][Go]
+                                                                                   | Name
+                                                                                   | Group
+                                                                                   | Type
+                                                                                   | Schedule
+                                                                                   | Labels
+                                                                                   | Retain
+                                                                                   | Concurrency
+===================================================================================================================
+[] | Name  | Group  | Type   | Schedule        | Labels       | Retain | Concurrency | Operation   |
+---+-------+--------+--------+-----------------+--------------+--------+-------------+-------------+--------------|
+[] | dummy | aa, bb | backup | 00:00 every day | k1:v1, k2:v2 | 20     | 10          | [Icon]    v |
+                                                                                       | Update
+                                                                                       | Delete
+===================================================================================================================
+                                                 [<] [1] [>]
+```
+
+**Scenario: Add Recurring Job**
+
+*Given* user sees `Create` on top left of the page.
+
+*When* user click `Create`.  
+*Then* user sees a pop-up form.
+
+```
+* Name
+[      ]
+
+Groups +
+
+* Task
+[Backup]
+
+* Schedule
+[00:00 every day]
+
+* Retain
+[20]
+
+* Concurrency
+[10]
+
+* Labels +
+```
+
+- Field with `*` is mendatory
+- User can click on `+` next to `Group` to add more groups.
+- User can click on the `Schedule` field and a window will pop-up for `Cron` and `Generate Cron`.
+- `Retain` cannot be `0`.
+- `Concurrency` cannot be `0`.
+- User can click on `+` next to `Labels` to add more labels.
+
+*When* user click `OK`.  
+*Then* frontend **POST** `/v1/recurringjobs` to create a recurring job.
+```
+❯ curl -X POST -H "Content-Type: application/json" \
+-d '{"name": "sample", "groups": ["group-1", "group-2"], "task": "snapshot", "cron": "* * * * *", "retain": 2, "concurrency": 1, "labels": {"label/1": "a"}}' \
+http://54.251.150.85:30944/v1/recurringjobs | jq
+{
+  "actions": {},
+  "concurrency": 1,
+  "cron": "* * * * *",
+  "groups": [
+    "group-1",
+    "group-2"
+  ],
+  "id": "sample",
+  "labels": {
+    "label/1": "a"
+  },
+  "links": {
+    "self": "http://54.251.150.85:30944/v1/recurringjobs/sample"
+  },
+  "name": "sample",
+  "retain": 2,
+  "task": "snapshot",
+  "type": "recurringJob"
+}
+```
+
+**Scenario: Update Recurring Job**
+
+*Given* an `Operation` drop-down list next to the recurring job. 
+
+*When* user click `Edit`.  
+*Then* user sees a pop-up form.
+```
+Name
+[sample]
+
+Groups
+[group-1]
+[group-2]
+
+Task
+[Backup]
+
+Schedule
+[00:00 every day]
+
+Retain
+[20]
+
+Concurrency
+[10]
+
+Labels
+[labels/1]: [a]
+[labels/2]: [b]
+```
+- `Name` field should be immutable.
+- `Task` field should be imuutable.
+
+*And* user edit the fields in the form.
+
+*When* user click `Save`.  
+*Then* frontend **PUT** `/v1/recurringjobs/{name}` to update specific recurring job.
+```
+❯ curl -X PUT -H "Content-Type: application/json" \
+-d '{"name": "sample", "groups": ["group-1", "group-2"], "task": "snapshot", "cron": "* * * * *", "retain": 2, "concurrency": 1, "labels": {"label/1": "a", "label/2": "b"}}' \
+http://54.251.150.85:30944/v1/recurringjobs/sample | jq
+{
+  "actions": {},
+  "concurrency": 1,
+  "cron": "* * * * *",
+  "groups": [
+    "group-1",
+    "group-2"
+  ],
+  "id": "sample",
+  "labels": {
+    "label/1": "a",
+    "label/2": "b"
+  },
+  "links": {
+    "self": "http://54.251.150.85:30944/v1/recurringjobs/sample"
+  },
+  "name": "sample",
+  "retain": 2,
+  "task": "snapshot",
+  "type": "recurringJob"
+}
+```
+
+**Scenario: Delete Recurring Job**
+
+*Given* an `Operation` drop-down list next to the recurring job.  
+
+*When* user click `Delete`.  
+*Then* user should see a pop-up window for confirmation.
+
+*When* user click `OK`.  
+*Then* frontend **DELETE** `/v1/recurringjobs/{name}` to delete specific recurring job.
+```
+❯ curl -X DELETE http://54.251.150.85:30944/v1/recurringjobs/sample | jq
+```
+
+> Also need a button for batch deletion on top left of the table.
+
+##### Updates `Volume` Page On UI
+
+**Scenario: Select From Recurring Job or Job Group**
+
+*When* user should be able to choose if want to add recurring job as `Job` or `Group` from the tab.
+
+**Scenario: Add Recurring Job Group On Volume Page**
+
+*Given* user go to job `Group` tab.  
+*When* user click `+ New`.  
+*And* Frontend can **GET** `/v1/recurringjobs` to list of recurring jobs.  
+*And* Frontend need to gather all `groups` from data.  
+```
+❯ curl -X GET http://54.251.150.85:30783/v1/recurringjobs | jq
+{
+  "data": [
+    {
+      "actions": {},
+      "concurrency": 2,
+      "cron": "* * * * *",
+      "groups": [
+        "group2",
+        "group3"
+      ],
+      "id": "backup1",
+      "labels": null,
+      "links": {
+        "self": "http://54.251.150.85:30783/v1/recurringjobs/backup1"
+      },
+      "name": "backup1",
+      "retain": 1,
+      "task": "backup",
+      "type": "recurringJob"
+    },
+    {
+      "actions": {},
+      "concurrency": 2,
+      "cron": "* * * * *",
+      "groups": [
+        "default",
+        "group1"
+      ],
+      "id": "snapshot1",
+      "labels": {
+        "label/1": "a",
+        "label/2": "b"
+      },
+      "links": {
+        "self": "http://54.251.150.85:30783/v1/recurringjobs/snapshot1"
+      },
+      "name": "snapshot1",
+      "retain": 1,
+      "task": "snapshot",
+      "type": "recurringJob"
+    }
+  ],
+  "links": {
+    "self": "http://54.251.150.85:30783/v1/recurringjobs"
+  },
+  "resourceType": "recurringJob",
+  "type": "collection"
+}
+```
+*Then* the user selects the group from the drop-down list.
+
+*When* user click on `Save`.  
+*Then* frontend **POST** `/v1/volumes/<VOLUME_NAME>?action=recurringJobAdd` with request body `{name: <group-name>, isGroup: true}`.  
+```
+❯ curl -X POST -H "Content-Type: application/json" \
+-d '{"name": "test3", "isGroup": true}' \
+http://54.251.150.85:30783/v1/volumes/pvc-4011f9a6-bae3-43e3-a2a1-893997d0aa63\?action\=recurringJobAdd | jq
+{
+  "data": [
+    {
+      "actions": {},
+      "id": "default",
+      "isGroup": true,
+      "links": {
+        "self": "http://54.251.150.85:30783/v1/volumerecurringjobs/default"
+      },
+      "name": "default",
+      "type": "volumeRecurringJob"
+    },
+    {
+      "actions": {},
+      "id": "test3",
+      "isGroup": true,
+      "links": {
+        "self": "http://54.251.150.85:30783/v1/volumerecurringjobs/test3"
+      },
+      "name": "test3",
+      "type": "volumeRecurringJob"
+    }
+  ],
+  "links": {
+    "self": "http://54.251.150.85:30783/v1/volumes/pvc-4011f9a6-bae3-43e3-a2a1-893997d0aa63"
+  },
+  "resourceType": "volumeRecurringJob",
+  "type": "collection"
+}
+```
+*And* user sees all `jobs` with the `group`.  
+
+**Scenario: Remove Recurring Job Group On Volume Page**
+
+*Given* user go to job `Group` tab.  
+*When* user click the `bin` icon of the recurring job group.  
+*Then* frontend `/v1/volumes/<VOLUME_NAME>?action=recurringJobDelete` with request body `{name: <group-name>, isGroup: true}`.
+```
+❯ curl -X POST -H "Content-Type: application/json" \
+-d '{"name": "test3", "isGroup": true}' \
+http://54.251.150.85:30783/v1/volumes/pvc-4011f9a6-bae3-43e3-a2a1-893997d0aa63\?action\=recurringJobDelete | jq
+{
+  "data": [
+    {
+      "actions": {},
+      "id": "default",
+      "isGroup": true,
+      "links": {
+        "self": "http://54.251.150.85:30783/v1/volumerecurringjobs/default"
+      },
+      "name": "default",
+      "type": "volumeRecurringJob"
+    }
+  ],
+  "links": {
+    "self": "http://54.251.150.85:30783/v1/volumes/pvc-4011f9a6-bae3-43e3-a2a1-893997d0aa63"
+  },
+  "resourceType": "volumeRecurringJob",
+  "type": "collection"
+}
+```
+
+**Scenario: Add Recurring Job On Volume Page**
+
+*Given* user go to `Job` tab.  
+*When* user click `+ New`.  
+*And* user sees the name is auto-generated.  
+*And* user can select `Backup` or `Snapshot` from the drop-down list.  
+*And* user can edit `Schedule`, `Labels`, `Retain` and `Concurrency`.
+
+*When* user click on `Save`.  
+*Then* frontend **POST** /v1/recurringjobs to create a recurring job.
+```
+❯ curl -X POST -H "Content-Type: application/json" \
+-d '{"name": "backup1", "groups": [], "task": "backup", "cron": "* * * * *", "retain": 2, "concurrency": 1, "labels": {"label/1": "a"}}' \
+http://54.251.150.85:30944/v1/recurringjobs | jq
+{
+  "actions": {},
+  "concurrency": 1,
+  "cron": "* * * * *",
+  "groups": [],
+  "id": "backup1",
+  "labels": {
+    "label/1": "a"
+  },
+  "links": {
+    "self": "http://54.251.150.85:30783/v1/recurringjobs/backup1"
+  },
+  "name": "backup1",
+  "retain": 2,
+  "task": "backup",
+  "type": "recurringJob"
+}
+```
+*And* frontend **POST** `/v1/volumes/<VOLUME_NAME>?action=recurringJobAdd` with request body `{name: <job-name>, isGroup: false}`.
+```
+❯ curl -X POST -H "Content-Type: application/json" \
+-d '{"name": "backup1", "isGroup": false}' \
+http://54.251.150.85:30783/v1/volumes/pvc-4011f9a6-bae3-43e3-a2a1-893997d0aa63\?action\=recurringJobAdd | jq
+{
+  "data": [
+    {
+      "actions": {},
+      "id": "default",
+      "isGroup": true,
+      "links": {
+        "self": "http://54.251.150.85:30783/v1/volumerecurringjobs/default"
+      },
+      "name": "default",
+      "type": "volumeRecurringJob"
+    },
+    {
+      "actions": {},
+      "id": "backup1",
+      "isGroup": false,
+      "links": {
+        "self": "http://54.251.150.85:30783/v1/volumerecurringjobs/backup1"
+      },
+      "name": "backup1",
+      "type": "volumeRecurringJob"
+    }
+  ],
+  "links": {
+    "self": "http://54.251.150.85:30783/v1/volumes/pvc-4011f9a6-bae3-43e3-a2a1-893997d0aa63"
+  },
+  "resourceType": "volumeRecurringJob",
+  "type": "collection"
+}
+```
+
+**Scenario: Delete Recurring Job On Volume Page**
+
+Same as **Scenario: Remove Recurring Job Group in Volume Page** with request body `{name: <group-name>, isGroup: false}`.
+```
+❯ curl -X POST -H "Content-Type: application/json" \
+-d '{"name": "backup1", "isGroup": false}' \
+http://54.251.150.85:30783/v1/volumes/pvc-4011f9a6-bae3-43e3-a2a1-893997d0aa63\?action\=recurringJobDelete | jq
+{
+  "data": [
+    {
+      "actions": {},
+      "id": "default",
+      "isGroup": true,
+      "links": {
+        "self": "http://54.251.150.85:30783/v1/volumerecurringjobs/default"
+      },
+      "name": "default",
+      "type": "volumeRecurringJob"
+    }
+  ],
+  "links": {
+    "self": "http://54.251.150.85:30783/v1/volumes/pvc-4011f9a6-bae3-43e3-a2a1-893997d0aa63"
+  },
+  "resourceType": "volumeRecurringJob",
+  "type": "collection"
+}
+```
+
+**Scenario: Keep Recurring Job Details Updated On Volume Page**
+
+- Frontend can monitor new websocket `/v1/ws/recurringjobs` and `/v1/ws/{period}/recurringjobs`.
+- When a volume is labeled with a none-existing recurring job or job-group. UI should show warning icon.
+
+### Test plan
+
+> The existing recurring job test cases need to be fixed or replaced.
+
+#### Integration test - test_recurring_job_group
+
+Scenario: test recurring job groups (S3/NFS)
+
+    Given create `snapshot1` recurring job with `group-1, group-2` in groups.
+               set cron job to run every 2 minutes.
+               set retain to 1.
+          create `backup1`   recurring job with `group-1`          in groups.
+               set cron job to run every 3 minutes.
+               set retain to 1
+    And volume `test-job-1` created, attached, and healthy.
+        volume `test-job-2` created, attached, and healthy.
+
+    When set `group1` recurring job in volume `test-job-1` label.
+         set `group2` recurring job in volume `test-job-2` label.
+    And write some data to volume `test-job-1`.
+        write some data to volume `test-job-2`.
+    And wait for 2 minutes.
+    And write some data to volume `test-job-1`.
+        write some data to volume `test-job-2`.
+    And wait for 1 minute.
+
+    Then volume `test-job-1` should have 3 snapshots after scheduled time.
+         volume `test-job-2` should have 2 snapshots after scheduled time.
+     And volume `test-job-1` should have 1 backup after scheduled time.
+         volume `test-job-2` should have 0 backup after scheduled time.
+
+#### Integration test - test_recurring_job_default
+
+Scenario: test recurring job set with default in groups
+
+    Given 1 volume created, attached, and healthy.
+
+    When create `snapshot1` recurring job with `default, group-1` in groups.
+         create `snapshot2` recurring job with `default`          in groups..
+         create `snapshot3` recurring job with ``                 in groups.
+         create `backup1`   recurring job with `default, group-1` in groups.
+         create `backup2`   recurring job with `default`          in groups.
+         create `backup3`   recurring job with ``                 in groups.
+    Then default `snapshot1` cron job should     exist.
+         default `snapshot2` cron job should     exist.
+                 `snapshot3` cron job should not exist.
+         default `backup1`   cron job should     exist.
+         default `backup2`   cron job should     exist.
+                 `backup3`   cron job should not exist.
+
+    # Setting recurring job in volume label should not remove the defaults.
+    When set `snapshot3` recurring job in volume label.
+    Then should contain `default`   job-group in volume labels.
+         should contain `snapshot3` job       in volume labels.
+    And  default `snapshot1` cron job should     exist.
+         default `snapshot2` cron job should     exist.
+                 `snapshot3` cron job should     exist.
+         default `backup1`   cron job should     exist.
+         default `backup2`   cron job should     exist.
+                 `backup3`   cron job should not exist.
+
+    # Should be able to remove the default.
+    When delete recurring job-group `default` in volume label.
+    And  default `snapshot1` cron job should not exist.
+         default `snapshot2` cron job should not exist.
+                 `snapshot3` cron job should     exist.
+         default `backup1`   cron job should not exist.
+         default `backup2`   cron job should not exist.
+                 `backup3`   cron job should not exist.
+
+    # Remove all volume recurring job labels should bring in default
+    When delete all recurring jobs in volume label.
+    Then default `snapshot1` cron job should     exist.
+         default `snapshot2` cron job should     exist.
+                 `snapshot3` cron job should not exist.
+         default `backup1`   cron job should     exist.
+         default `backup2`   cron job should     exist.
+                 `backup3`   cron job should not exist.
+
+    # Add `default` to snapshot3 and backup3 recurring job `Group`.
+    # should also reflect on the cron jobs
+    When add `snapshot3` recurring job with `default` in groups.
+         add `backup3`   recurring job with `default` in groups.
+    Then default `snapshot1` cron job should exist.
+         default `snapshot2` cron job should exist.
+         default `snapshot3` cron job should exist.
+         default `backup1`   cron job should exist.
+         default `backup2`   cron job should exist.
+         default `backup3`   cron job should exist.
+
+    # Remove `default` in recurring job `Group` should also
+    # reflect on the cron jobs
+    When remove `default` from `snapshot3` recurring job groups.
+    Then default `snapshot1` cron job should     exist.
+         default `snapshot2` cron job should     exist.
+                 `snapshot3` cron job should not exist.
+         default `backup1`   cron job should     exist.
+         default `backup2`   cron job should     exist.
+         default `backup3`   cron job should     exist.
+
+    # Remove `default` in all recurring job `Group` should also
+    # reflect on the cron jobs
+    When remove `default` from all recurring jobs groups.
+    Then `snapshot1` cron job should not exist.
+         `snapshot2` cron job should not exist.
+         `snapshot3` cron job should not exist.
+         `backup1`   cron job should not exist.
+         `backup2`   cron job should not exist.
+         `backup3`   cron job should not exist.
+
+
+#### Integration test - test_recurring_job_delete
+
+Scenario: test delete recurring job
+
+    Given 1 volume created, attached, and healthy.
+
+    When create `snapshot1` recurring job with `default, group-1` in groups.
+         create `snapshot2` recurring job with `default`          in groups..
+         create `snapshot3` recurring job with ``                 in groups.
+         create `backup1`   recurring job with `default, group-1` in groups.
+         create `backup2`   recurring job with `default`          in groups.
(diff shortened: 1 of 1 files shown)


enhancements/20210624-backup-policy.md
```diff
@@ -0,0 +1,123 @@
...
+As a Longhorn user / System administrator.
+
+I want the ability to set Longhorn volume name to backup policies for automatic recurring job creation.
+
+So I do not need to manually create recurring jobs for each volume. And I will be able to see volume assignments based on the backup policy.
+
+#### Story 2 - cleanup none-exist volume names
+As a Longhorn user / System administrator.
+
+I want the ability to automatically remove volume names in the backup policy when volumes don't exist in the cluster.
```
yasker, line 38: It should be only volume referring to the backup policy.

We can select volumes for certain backup policies using labels. But it's unnecessary to record who is using which backup policy.

enhancements/20210624-backup-policy.md
```diff
@@ -0,0 +1,123 @@
...
+As a Longhorn user / System administrator.
+
+When the default backup policy is set, new volumes should be automatically be assigned.
+
+So I do not need to manually create the same recurring job to all volumes.
+
+#### Story 3 - Re-use volume assignment
+As a Longhorn user / System administrator.
+
+When a volume is removed and restored to the cluster, the backup policy should automatically pick up the volume assignment and re-create the recurring job.
```
yasker, line 52: Do you mean we're storing the backup policy in the backup's metadata? It doesn't make sense to have the local backup policy tie to a volume.

enhancements/20210624-backup-policy.md
```diff
@@ -0,0 +1,123 @@
...
+When a volume is removed and restored to the cluster, the backup policy should automatically pick up the volume assignment and re-create the recurring job. 
+
+So I do not need to re-create recurring jobs or re-assign a backup policy.
+
+### User Experience In Detail
+
+#### Story 1 - backup policy assignment
+
+1. Create a backup policy with `kubectl` or UI.
+2. Once the volume is created, update the backup policy spec with `kubectl` or UI to include volume name. This will automatically create the recurring job to the assigned volume.
```
yasker, line 61: Update the volume instead of updating the backup policy.

It's like PVC might have an associated storage class, but storage class doesn't need to know about PVC.

enhancements/20210624-backup-policy.md
```diff
@@ -0,0 +1,123 @@
...
+> TBU.
+
+## Design
+
+### Implementation Overview
+
+- Add backup policy CRD.
+  - Update the ClusterRole to include `backuppolices` and `backuppolicies/status`
+  - The volume recurring job name will inherit from the backup policy name. It should be 8 characters or less, thus we are likely not going to have suffix or prefix to indicate this is from the backup policy. If the backup policy is named with more than 8 characters, an error shows on UI. This can be validated at new backup policy creation with specifying validation in the CRD YAML.
+  - Printer Colume should include `Name`, `Default`, `Recurring`, `Retain`, `Age`.
```
yasker, line 91: Need type. Can be snapshot or backup.

enhancements/20210624-backup-policy.md
```diff
@@ -0,0 +1,123 @@
...
+### API changes
+
+> TBU.
+
+## Design
+
+### Implementation Overview
+
+- Add backup policy CRD.
+  - Update the ClusterRole to include `backuppolices` and `backuppolicies/status`
```
yasker, line 89: I don't think we need `status`.

yasker (changes requested):
Volume to backup policy should be a one-way relationship instead of a two-way relationship. 

I don't think it's necessary to store any information regarding which volume has the policy in the backup policy itself. That information's life cycle should be with volume, rather than the policy itself.

Also, need to change the name of the policy from `backup policy` since it can contain a recurring snapshot job as well.

Probably call it `Recurring Job Policy`, or something covers both snapshot and backup.

I can understand the motivation for moving the tracking into the policy controller. It might be to reduce the complexity of the volume controller, but we don't need to do anything other than writing the policy name into the volume (which should be next to the existing customized recurring job, basically a new type of recurring job), and update labels(there can be multiple policies tie to a volume). 

If the policy was deleted, we can warn the user that the policy no longer exists for the volume. That can be in a condition. We don't need to remove the policy name from the volume (and we cannot because it's in the spec). The latter reminds me that, how will we clean up the volume name from the policy CR? It should be in the spec but the controller cannot change the spec.

Also, the backup job related to the recurring job policy is better to be handled in one place, maybe by the recurring job policy controller, in case the policy itself was changed during the volume's life cycle. Essentially, for the cronjob running for certain recurring job policy, it should be able to find all the volumes that are using the policy at the moment (by using the label), then do the job. It can even potentially throttle the operations in parallel, make sure we're not going to overwhelm the system.

And CRD definition/update needs to be in the LEP as well.

enhancements/20210624-backup-policy.md
```diff
@@ -0,0 +1,123 @@
+# Backup Policy
+
+## Summary
+
+Users can assign volumes to backup policies. The assigned volumes inherent the recurring job settings from the backup policy and will automatically create/update/delete the recurring job.
```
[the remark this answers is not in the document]
c3y1huang: Updated the LEP to have backup policy reference by the volume recurring job.

enhancements/20210624-backup-policy.md
```diff
@@ -0,0 +1,123 @@
...
+https://github.com/longhorn/longhorn/issues/467
+
+## Motivation
+
+### Goals
+
+- Ability to create multiple backup policies with UI or YAML.
+- Ability to do volume assignment.
+- Ability to assign default backup policy when a new volume is created.
+- Ability to enable cleanup backup policy volume list when volumes get removed from the cluster.
```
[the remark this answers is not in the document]
c3y1huang: No longer part of the goal because the life cycle now goes with the volume.

enhancements/20210624-backup-policy.md
```diff
@@ -0,0 +1,123 @@
...
+- Include recurring job settings; `name`, `cron`, `retain`.
+- Should have backup policy status for each volume assigned, including `error`, `lastUpdateTime`.
```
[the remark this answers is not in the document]
c3y1huang: No longer part of the goal because the life cycle now goes with the volume and no point in storing the update status in the policy.

enhancements/20210624-backup-policy.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Motivation
+
+### Goals
+
+- Ability to create multiple backup policies with UI or YAML.
+- Ability to do volume assignment.
+- Ability to assign default backup policy when a new volume is created.
+- Ability to enable cleanup backup policy volume list when volumes get removed from the cluster.
+- Include recurring job settings; `name`, `cron`, `retain`.
```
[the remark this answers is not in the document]
c3y1huang: Added.

enhancements/20210624-backup-policy.md
c3y1huang: Fixed.

enhancements/20210624-backup-policy.md
c3y1huang: Removed the user story because we will do policy reference with volume's recurring job.

enhancements/20210624-backup-policy.md
c3y1huang: Fixed.

enhancements/20210624-backup-policy.md
c3y1huang: Added type.

enhancements/20210624-backup-policy.md
c3y1huang: Removed.

yasker (changes requested):

enhancements/20210624-recurring-job-policy.md
```diff
@@ -0,0 +1,117 @@
...
+
+**After enhancement**
+Create a recurring job with type `Policy` automatically reference to the recurring job policy.
+
+#### Story 2 - default recurring job policy reference
+
+1. Create a recurring job policy with `default` set to `true` in UI or `kubectl`.
+2. Create a new volume and it should automatically refer to default backup/snapshot policies.
+
+> Question: Should the default recurring job policy overrule the storageClass recurring backup jobs? Or should they co-exist?
```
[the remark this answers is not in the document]
c3y1huang: Reflecting on [comment](https://github.com/longhorn/longhorn/pull/2737#discussion_r660053424). Does it make sense to update the volume spec with the default recurring job policy on first-time volume creation? 
```
spec:
  recurringJobs:
  - name: test
    task: policy
```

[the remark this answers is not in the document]
yasker: There are two ways of doing it:
1. Don't add them in. So the controller needs to scan for the volumes without a schedule and run them on schedule.
    * Pros: No modification to the volume. And once the user decides to set a schedule for the volume, the default schedule will be automatically overridden.
    * Cons: It's not clear from a volume perspective that if the volume will be backed up based on the schedule.
2. Add the reference in.
    * Exactly the opposite as the first option.

I don't have a clear preference over one or the other. What do you think?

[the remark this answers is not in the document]
c3y1huang: ATM, I am favoring adding the default recurring job to the volume spec on first-time volume creation, for reasons:
1. The behavior is consistent with storageClass. When the recurring job is applied in the SC it will also get added to the volume spec.
2. I've not found a good way that can identify the first-time volume is created to apply the default with the recurring job controller. But we can leverage the volume controller to identify this and add to the volume spec and the recurring job controller can then scan and reference for cronjob creation.

enhancements/20210624-recurring-job-policy.md
```diff
@@ -0,0 +1,117 @@
...
+https://github.com/longhorn/longhorn/issues/467
+
+## Motivation
+
+### Goals
+
+- Ability to create multiple recurring backup/snapshot job policies with UI or YAML.
+- Ability to use recurring job type - `Policy` for policy reference.
+- Ability to set for automatic reference to default recurring backup/snapshot policy with new volume.
+- Recurring job policy should include recurring job settings; `name`, `type`. `cron`, `retain`.
```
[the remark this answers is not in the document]
c3y1huang: Fixed.

enhancements/20210624-recurring-job-policy.md
```diff
@@ -0,0 +1,117 @@
...
+
+So I do not need to manually create the same recurring job for all volumes.
+
+
+### User Experience In Detail
+
+#### Story 1 - refer volume to recurring job policy
+
+1. Create a recurring job policy in UI or `kubectl`.
+2. Create `Recurring Snapshot and Backup Schedule` with UI and input recurring job name as `Name` and type as `Policy`. Then click `Save` should see automatically filled in `Schedule`, `Retain`.
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210624-recurring-job-policy.md
```diff
@@ -0,0 +1,117 @@
...
+
+**Before enhancement**
+Manually create recurring jobs for each volume.
+
+**After enhancement**
+Create a recurring job with type `Policy` automatically reference to the recurring job policy.
+
+#### Story 2 - default recurring job policy reference
+
+1. Create a recurring job policy with `default` set to `true` in UI or `kubectl`.
```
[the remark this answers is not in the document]
c3y1huang: Fixed wording.

enhancements/20210624-recurring-job-policy.md
```diff
@@ -0,0 +1,117 @@
...
+
+- Integrate recurring job policy to volume controller when volume spec recurring job has type - `Policy`
+  1. Before `createCronJob`, get the recurring job policy by name and update the volume jobs `.Cron` and `.Retain`.
+  2. append cmd `--backup` if the matching recurring job policy has type - `backup`.
+
+- Add recurring job policy controller.
+  - The code structure will be the same as other controllers.
+  - The controller will be informed by `recurringJobPolicyInformer`, `volumeInformer`.
+  - The `syncRecurringJobPolicy` will,
+    1. Update the ownership of `Default` backup and snapshot policy when a backup or snapshot policy is updated/created with `.Spec.Default`.
```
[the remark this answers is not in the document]
c3y1huang: Not required anymore. Removed.

enhancements/20210624-recurring-job-policy.md
```diff
@@ -0,0 +1,117 @@
...
+    2. Update the associated volume recurring job `.Cron`, `.Retain` when the recurring job policy is updated.
```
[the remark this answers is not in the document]
c3y1huang: Removed.

enhancements/20210624-recurring-job-policy.md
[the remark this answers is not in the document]
yasker: In fact, we will apply the policy to all the volumes without any backup policy, not only for the first-time volume creation. Since it's the `default` as a `fail-safe`, it should act as a catch-all.

For example, if a volume is created with a recurring job, and later the job was removed. The default policy should still cover it.

After you describe the process, I realized there is another issue with this approach: we shouldn't modify the volume's spec, since it's user-created and defined.

enhancements/20210624-recurring-job-policy.md
```diff
@@ -0,0 +1,117 @@
...
+### API changes
+
+> TBU.
+
+## Design
+
+### Implementation Overview
+
+- Add recurring job policy CRD.
+  - Update the ClusterRole to include `recurringjobpolices`.
```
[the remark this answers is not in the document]
c3y1huang: Discussed with @yasker and we will keep the same as the [others](https://github.com/longhorn/longhorn/blob/master/deploy/longhorn.yaml#L47-L51).

enhancements/20210624-recurring-job-policy.md
```diff
@@ -0,0 +1,117 @@
...
+  - The volume recurring job name will reference the recurring job policy name. It should be 8 characters or less. If the policy is named with more than 8 characters, an error shows on UI. This can be validated at new recurring job policy creation with specifying validation in the CRD YAML.
```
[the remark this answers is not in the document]
c3y1huang: Discussed with @yasker and we will name the cronjob the same as the policy name by introducing new longhorn-manager command to use per policy.

enhancements/20210624-recurring-job-policy.md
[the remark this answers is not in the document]
c3y1huang: Thank you for clarifying the expectation. Fixed.

enhancements/20210624-recurring-job-policy.md
```diff
@@ -0,0 +1,117 @@
...
+  - The `Type` should be one of `backup`, `snapshot` and can be validated in the CRD YAML pattern regex match.
+  - Printer Colume should include `Name`, `Type`, `Default`, `Recurring`, `Retain`, `Age`.
+
+- Integrate recurring job policy to volume controller when volume spec recurring job has type - `Policy`
+  1. Before `createCronJob`, get the recurring job policy by name and update the volume jobs `.Cron` and `.Retain`.
```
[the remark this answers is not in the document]
c3y1huang: Fixed.

enhancements/20210624-recurring-job-policy.md
```diff
@@ -0,0 +1,422 @@
...
+       - Compose command,
+         ```
+         longhorn-manager -d\
+           snapshot-policy <policy-name> <volume-names>\
+           --manager-url <url>\
+           --labeles <recurring-job-label>=<policy-name>\
+           --retain <policy.spec.retain>\
+           <--backup>
+         ```
+    3. Delete the cron job if annotation `last-applied-cronjob-volume` is different from the desire volume list.
```
c3y1huang, line 148: Discussed with @yasker and will remove the cron job annotation and add labels in volume instead.

c3y1huang: Because the [K8s limits label key and value to 63 chars](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set), and Longhorn does not limit the number of policy can apply for each volume, so we will probably need to have one label for each policy.
```
- apiVersion: longhorn.io/v1beta1
  kind: Volume
  metadata:
    creationTimestamp: "2021-07-27T07:06:35Z"
    finalizers:
    - longhorn.io
    generation: 2
    labels:
      longhorn.io/job-policy-backup1: backup
      longhorn.io/job-policy-backup2: backup
      longhorn.io/job-policy-backup3: backup
      longhorn.io/job-policy-snapshot1: snapshot
      longhorn.io/job-policy-snapshot2: snapshot
      longhorn.io/job-policy-snapshot3: snapshot
      longhornvolume: pvc-1bad969c-7f40-466f-9e2f-c7ffc5ea527a
    name: pvc-1bad969c-7f40-466f-9e2f-c7ffc5ea527a
    namespace: longhorn-system
    resourceVersion: "650605"
```
And also need to implement logic to update the labels.

enhancements/20210624-recurring-job-policy.md
```diff
@@ -0,0 +1,422 @@
...
+
+- In the volume controller,
+  1. Skip `DeleteCronJob` if the applied cron job with label `longhorn.io/job-task` is set to `policy`.
+
+- Add recurring job policy controller.
+  - The code structure will be the same as other controllers.
+  - The controller will be informed by `recurringJobPolicyInformer`, `volumeInformer`.
+     - When informed with `volume` changes, the controller enqueues the recurring job policy when set in `spec.recurringJobs` and also enqueues default `snapshot`/`backup` recurring job policy if the `task` type is missing from the volume `spec.recurringJobs`.
+  - Add the finalizer to the recurring job policy CRs if not exist.
+  - Create and update CronJob per policy.
```
shuo-wu, line 128: Here, the controller can: 
    1. List all volumes that require the policy.
        1. If the volume is not attached while the setting is not `true`, remove the volume from the list.
        2. For a default policy, remove all volumes that have recurring jobs or policies set.
    2. Delete the recurring job CR if it doesn't match the spec or no volume is using the policy
    3. Create a recurring job CR if there is a volume using the policy

shuo-wu (changes requested):
After checking this LEP,  I still prefer to ask policy CRs to record the volume references in specs. Here are the cons of the current design:
1. Users need to understand the recurring jobs in volume spec and non-default recurring policies before using default policies correctly. 
2. Both VolumeController and the new PolicyController should be responsible for creating/deleting recurring job CRs, which is not elegant enough. 
3. The policy controller needs to list all volumes that require the related policy anyway. Then why not directly update and record the volume list at the beginning (when users set the recurring jobs for volumes via HTTP APIs)?
4. This adds more complexity for volume CRs.
5. There is no way to separate all recurring-related concepts from volume CRD as the further refactoring. The current design also prevents Longhorn from introducing more fancy features. e.g., Adding tag-selector for volumes and recurring policies, which provide more flexibility of configuring the volume recurring jobs.

cc @yasker @joshimoo

c3y1huang:
Seems like there are different opinions on if Longhorn should reference volume in cronjob/policy or policy should reference in volume.

I've updated to include:
- Option 1: Volume referenced in the cron job annotation.
- Option 2: Policy referenced in the volume label.

shuo-wu:
> * Option 1: Volume referenced in the cron job annotation.

Volume list can be stored in policy spec or label as well. The problem is, which one is responsible for recording the reference.

yasker:
@shuo-wu We cannot keep the `default` policy in the volume spec since the user owns the volume spec, we cannot change it at our will. Think how users would operate it using Kubernetes API instead of Longhorn API, which is the ultimate goal.

The ultimate goal for this design is to transition into a separate recurring job instead of embedded with the volume object, which is way too crowded already.

> 1. Users need to understand the recurring jobs in volume spec and non-default recurring policies before using default policies correctly.

There are only two scenarios here:
1. If there is anything in the `volume.spec.recurringJob`, use it.
2. Otherwise, the default recurring job policy applies.

> 2. Both VolumeController and the new PolicyController should be responsible for creating/deleting recurring job CRs, which is not elegant enough.

I have the same thought. So the ultimate goal is to move recurring jobs completely outside of volume spec to be handled by `recurring job controller` (currently known as `recurring job policy controller`), but that's a bigger work than scoped.

> 3. The policy controller needs to list all volumes that require the related policy anyway. Then why not directly update and record the volume list at the beginning (when users set the recurring jobs for volumes via HTTP APIs)?

The policy controller doesn't need to do it. The job can do it on the fly when it's executing.

Again, we cannot rely on HTTP APIs. We are moving away from it into Kubernetes native API.

> 4. This adds more complexity for volume CRs.

Nothing was added in the volume CRs except for the policy and label, which are ignored by the volume controller. It's not the volume controller who will schedule the recurring job for recurring job policies, it's a separate cronjob scheduled by the recurring job controller.

> 5. There is no way to separate all recurring-related concepts from volume CRD as the further refactoring. The current design also prevents Longhorn from introducing more fancy features. e.g., Adding tag-selector for volumes and recurring policies, which provide more flexibility of configuring the volume recurring jobs.

I don't follow. This design is essentially moving away from using the volume CRD for recurring jobs. Tag selector for the volumes can be defined in the separate recurring job CR later. We can also use labels to select multiple recurring job policies, like what we've done with the `default` recurring job policies.

enhancements/20210624-recurring-job-policy.md
joshimoo: On a quick look I think a 1 way mapping is all we need.
Volumes don't need to know which policies apply to them.
This gets double important once we introduce volume groups.

The Policy/Recuring whatever crd should keep track of the volumes that it applies to.
One idea for this is to specify a volume/group label selector as part of the spec / annotation of the policy crd.

Then it's easy to implement volume groups just by tagging the volumes with the associated group.
It's also easy to offset the backup creation while only doing a sync snapshot.

The policy/job controller can then just create backup crds once the backup refactor is included.
Will have a deeper look at the LEP tomorrow.

joshimoo:
I agree and I think I mentioned this before, the volume really doesn't need any knowledge about anything related to backup policy / recurring jobs definitions.

The Policy CRD can reference an arbitrary set of volumes via the definition of a labelSelector as part of  the policy spec.
The default policy just becomes `Select label == *` the selectors are pseudo code for demonstration only.
This allows for all kinds of flexibility either by just using the volume.name label or by specifying additional labels on the volumes.

Some Example policies below:

Example Policy for a volume group X
Select label volume-group == X

A policy that only selects kafka volumes.
Select label app == kafka

A policy that only applies to 2 volumes in the whole cluster
Select label where volume in (X | Y)

**Have a look at the set based selectors.**
https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement

yasker:
In fact, more I think about this, more I realized we probably should move to use recurring job policy 100%. That means users can only set recurring job policy to the volume, no more separate entry for snapshot/backup is needed. Thinking about it, when would you want your 100 volumes all starts to backup at totally different times with totally different configurations?

What they would prefer instead is:
1. Create a recurring job policy.
2. Set certain volumes to use it.

All of those can work today since we've introduced the recurring job policy controller. We just need to migrate all the previous recurring jobs (including storage class) into new auto-created policies when upgrade.

So:
1. Volume spec will only contain the recurring job policy names.
2. Those names will be added to the volume as labels, as @joshimoo mentioned above. 
    * If there is no recurring job policy for the volume, the `default` policy label will be automatically added to the volume.
3. Recurring job controller will start the cron jobs according to the cron in the policy
    * it doesn't care who are the volumes will be backed up at this time.
4. When the time comes, the cron job will retrieve the content of the recurring job policy by the name, and find who are the volumes have the related label, and perform the snapshot/backup operations:
    * if the recurring job policy is set to default, search for `default` in addition to the `name`.
    * Does the snapshot/backup operation one by one. The operation order can be defined as `sequential`, `consistent` (for volume group snapshot), or `throttled` (with a concurrent number as parameter) in the future.
    * Be careful with the retry mechanism here. We don't want to re-backup previously successfully backed up volumes if a later volume failed.
    * The policy can also use a label selector to the volume, but that would require volume to have some labels. I think we can implement this part later.

yasker:
Had a discussion with @c3y1huang and @shuo-wu , we will move to 100% recurring job policy (now it will be called just `recurring job` since it's the only way to create recurring job.)

Also, `default` will become a group of recurring jobs (instead of a special case). Volume can select one or more recurring jobs, and/or one or more recurring job groups. The recurring job controller will be responsible for label the volume for the `default` recurring job group if no other recurring job was found. The `default` recurring job group will not be removed by the recurring job controller if users add more recurring jobs. Users need to remove the `default` recurring job group manually.

Also, a recurring job can be performed for a volume group with a special `volume-group: xxx` label. It will be evaluated during the runtime of the cronjob. I assume this selector will be kept in the recurring job CR. 

For individual volumes, the recurring job information will be kept in the volume's metadata instead of the recurring job CR itself, since it's easier to use the label to filter volumes instead of keeping a list in the recurring job CR itself.

cc @joshimoo

c3y1huang:
More notes from the discussion with @yasker and @shuo-wu.
- Remove recurringJobs from volume spec.
- Change recurring job policy to recurring job.
- Volume will contain two types of label - `job`, `job-group`.
- Remove the field `Default` and use `Group` instead. The recurring job applies if recurring jobs contain default group.
- Do not need to treat default snapshot and backup separately. Only need to determine if the volume policy label exists.
- Translate existing recurring jobs to the label during the upgrade.
- Add `throttle` to the longhorn-manager binary and recurring job spec. Options `sequential`, `consistent` comes later.

enhancements/20210624-recurring-job-policy.md
```diff
@@ -0,0 +1,422 @@
...
+    1. Get a `snapshot`/`backup` mapping of volumes with no recurring jobs to determine if need to set to the default.
+       1. Skip volume if volume is not `attached` or `allow-recurring-job-while-volume-detached` is not set to `true`.
```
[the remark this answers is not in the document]
c3y1huang: Fixed.

enhancements/20210624-recurring-job-policy.md
```diff
@@ -0,0 +1,422 @@
...
+So I can be assured that all volumes will have both `backup` and `snapshot` recurring jobs after volume spec `recurringJobs` are not set or gets removed.
+
+### User Experience In Detail
+
+#### Story 1 - set recurring job policy in volume spec
+
+1. Create a recurring job policy in UI or `kubectl`.
+2. In UI, Navigate to Volume, `Recurring Snapshot and Backup Schedule`.
+3. Select `Policy` from the `Type` list.
+4. Select existing recurring job policy from the `Name` list.
```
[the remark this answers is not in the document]
c3y1huang: I am thinking `Setting/Recurring Job`. What do you think?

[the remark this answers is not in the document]
shuo-wu: I think we can directly add a new button on the top of the UI like `Volume` or `Setting`. Since this page may be frequently used. It is a core feature.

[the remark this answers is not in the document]
c3y1huang: Make sense. Fixed.

enhancements/20210624-recurring-job-policy.md
c3y1huang: > Here, the controller can:
> 
> 1. List all volumes that require the policy.
>    
>    1. If the volume is not attached while the setting is not `true`, remove the volume from the list.

This will be done in binary on-the-fly with the label selector.

>    2. For a default policy, remove all volumes that have recurring jobs or policies set.

Volume controller will be responsible for adding the default policy.

> 2. Delete the recurring job CR if it doesn't match the spec or no volume is using the policy

I think we should keep the recurring job CR, the recurring job will also be responsible to remove any excess corn jobs.

> 3. Create a recurring job CR if there is a volume using the policy

Users will mostly be creating their own recurring job CRs. I think the only case where Longhorn will be creating for the user is during upgrade migration or for the PVC cases.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,572 @@
...
+    Then moniter the cron job pod log.
+    And should see 2 jobs created concurrently.
+
+    When update `snapshot1` recurring job with `concurrency` set to `3`.
+    Then moniter the cron job pod log.
+    And should see 3 jobs created concurrently.
+
+### Upgrade strategy
+#### Automated Migration
+1. Translate all volume spec recurringJobs to label `longhorn.io/job-<jobTask>-<RandomID>: enabled`.
```
yasker, line 559: We don't need `<RandomID>` for upgrade.

The recurring job is defined by:
1. Task
2. cron
3. Retain

So, as long as those three fields are the same, we can regarding it as the same recurring job. For example, we can name a recurring job: `<task>-<retain>-<hashing of cron>`, e.g. `backup-3-abcdef12` which `abcdef12` is some hashing result from the cron field.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,572 @@
...
+     - The `Groups`: Array of strings that set groupings to the recurring job. When including `default`, the recurring job will be added to the volume label if no other job exists in the volume label.
+     - The `Task`: String of either one of `backup` or `snapshot`.
+       Also add validation in the CRD YAML with pattern regex match.
+     - The `Cron`: String in cron expression represents recurring job scheduling.
+     - The `Retain`: Integer of the number of snapshots/backups to keep for the volume.
+     - The `Concurrency`: Integer of the concurrent job to run by each cron job.
+     - The `Age`: Date of the CR creation timestamp.
+
+- In the longhorn-manager binary,
+  1. Add new command `snapshot-job <job.name>` with flags `--groups`, `--label`, `--retain`, `--concurrent` `--backup`.
```
yasker, line 146: `recurring-job` instead of `snapshot-job`.

Also,`recurring-job` command can get those parameters automatically as well, since it has the access to the Kubernetes API server. So we don't need to recreate the job every time when the spec changes.

So the only necessary command line is
```
longhorn-manager -d\
           recurring-job <job.name>\
           --manager-url <url>
```

yasker (changes requested):

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: Thank you. Fixed.

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: Make sense, fixed.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,962 @@
...
+    "group-2"
+  ],
+  "id": "snap2",
+  "links": {
+    "self": "http://54.251.150.85:32279/v1/recurringjobs/snap2"
+  },
+  "name": "snap2",
+  "retain": 2,
+  "task": "snapshot",
+  "type": "recurringJob"
```
yasker, line 267: Where are the labels? It was a part of recurring job definition in volume spec. https://github.com/longhorn/longhorn-manager/blob/2dfb1487f325f019a003d15db056535fe18c7252/types/resource.go#L170

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,962 @@
...
+I want the ability to set one or multiple `backup` and `snapshot` recurring jobs as default. All volumes without any recurring job label should automatically apply with the default recurring jobs.
+
+So I can be assured that all volumes without any recurring job label will automatically apply with default.
+
+### User Experience In Detail
+
+#### Story 1 - set recurring job in volume
+
+1. Create a recurring job on UI `Recurring Job` page or via `kubectl`.
+2. In UI, Navigate to Volume, `Recurring Snapshot and Backup Schedule`.
```
yasker, line 64: `Recurring Jobs Schedule` is fine.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,962 @@
...
+3. When clicked on the `New` should have the option to choose from `single` or `group`.
```
yasker, line 65: The flow is unclear after choosing `single` or `group`.

E.g.
1. ... `single` or `group`
    1. If `single` is chosen, ...
    2. If `group` is chosen, ...

Though we can also update them in the UI. Doesn't need to be 100% accurate here.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,962 @@
...
+4. When clicked on the drop-down menu - `Name`, it will show a list of existing recurring jobs or a pop-up window to input the recurring job group name.
+5. The selected recurring job(s) will show the `Type`, `Schedule`, `Retain`, `Concurrency` information from the `RecurringJob` CR.
```
yasker, line 67: Missing labels field.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,962 @@
...
+
+Can select a recurring job from the UI drop-down menu and will automatically show the information from the recurring job CRs.
+
+Update the recurring job definition will automatically apply to all volumes with the job label.
+
+#### Story 2 - automatically apply to default recurring jobs
+
+1. Add `default` to one or multiple recurring jobs `Groups` in UI or `kubectl`.
+2. Longhorn automatically applies the `default` group recurring jobs to all volumes without job labels.
+3. Add job label to volume removed the `default` recurring jobs.
```
yasker, line 92: It's a bit hard to implement this once we added the `default` recurring job group label to the volume. I am fine with not removing the `default` group automatically. Users might want to add more jobs on top of the existing `default` group.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,962 @@
...
+     - The `Retain`: Integer of the number of snapshots/backups to keep for the volume.
+     - The `Concurrency`: Integer of the concurrent job to run by each cron job.
+     - The `Age`: Date of the CR creation timestamp.
+
+#### Add Command `recurring-job` to `longhorn-manager` Binary
+
+  1. Add new command `recurring-job <job.name> --longhorn-manager <URL>`.
+     > Get the `recurringJob.Spec` on execution using Kubernetes API.
+  2. Get volumes by label selector `longhorn.io/job-<job.name>: enabled` to filter out volumes.
+  3. Get volumes by label selector `longhorn.io/job-group-<job.group>: enabled` to filter out volumes if the job is set with a group.
```
yasker, line 156: `jobgroup` instead of `job-group`. Since the job group might start with `group-xxx` name then it will be confusion.

In fact it's better to name the selector `recurring-job.longhorn.io/<job-name>` and `recurring-job-group.longhorn.io/<group-name>`.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,962 @@
...
+
+- In the manager/volume, create recurring jobs for volumes from the `storageClass` if not already exist.
+  > Since the storage class and upgrade migration contains recurringJobs spec. So we will keep the `VolumeSpec.RecurringJobs` in code to create the recurring jobs for volumes from the `storageClass`.
+
+  > In case names are duplicated between different `storageClasses`, only one recurring job CR will be created.
+
+#### Changes in the Datastore
+
+  - For `CreateVolume`:
+     1. Add recurring job label if `VolumeSpec.RecurringJobs` is not empty, or label with `default` job-group.
```
yasker, line 173: It's probably better to do the upgrade from `spec.recurringJobs` to `recurring job` in the upgrade path (https://github.com/longhorn/longhorn-manager/tree/master/upgrade ) , not in the datastore.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,962 @@
...
+#### Add Recurring Job Controller
+
+  - The code structure will be the same as other controllers.
+  - Add the finalizer to the recurring job CRs if not exist.
+  - The controller will be informed by `recurringJobInformer` and `volumeInformer`.
+    - `recurringJobInformer` do `enqueueRecurringJob`.
+    - `volumeInformer` get all recurring job from volume recurring job labels and `enqueueRecurringJob`. It will also label and update the volume with `default` job-group if no recurring job label is found.
+  - Create and update CronJob per recurring job.
+    1. Generate a new cron job object.
+       - Include labels `longhorn.io/job`.
```
yasker, line 188: Update the label accordingly.

yasker (changes requested):

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: Fixed.

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: Fixed.

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: Addressed.

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: Fixed.

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: Removed (3).

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: Addressed.

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: Fixed.

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: This is related to the volume created by the storage class, not the upgrade. Discussed with @yasker , will introduce `recurringJobSelectors` and parameter `recurringJobs` will be deprecated in future.

yasker: `recurringJobSelector`, the singular form (not the plural) to keep consistent with other selectors.

c3y1huang: Added the implementation for `recurringJobSelector` also the upgrade migration from `recurringJobs` to `recurringJobSelector` for the storage class parameter.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1087 @@
...
+
+- In the manager/volume, create recurring jobs for volumes from the `storageClass` if not already exist.
+  > Since the storage class and upgrade migration contains recurringJobs spec. So we will keep the `VolumeSpec.RecurringJobs` in code to create the recurring jobs for volumes from the `storageClass`.
+
+  > In case names are duplicated between different `storageClasses`, only one recurring job CR will be created.
+
+#### Changes In The Datastore
+
+  - For `CreateVolume`:
+     1. Add recurring job label if `VolumeSpec.RecurringJobs` is not empty.
```
yasker, line 198: Hmm, once we've done with the upgrade, I assume the recurringJobs should be empty?

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1087 @@
...
+#### Add Recurring Job Controller
+
+  - The code structure will be the same as other controllers.
+  - Add the finalizer to the recurring job CRs if not exist.
+  - The controller will be informed by `recurringJobInformer` and `volumeInformer`.
+    - `recurringJobInformer` do `enqueueRecurringJob`.
+    - `volumeInformer` get all recurring job from volume recurring job labels and `enqueueRecurringJob`. It will also label and update the volume with `default` job-group if no recurring job label is found.
+  - Create and update CronJob per recurring job.
+    1. Generate a new cron job object.
+       - Include labels `recurring-job.longhorn.io` and `recurring-job-task.longhorn.io`.
```
yasker, line 222: What do we need `recurring-job-task.longhorn.io` for?

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1087 @@
...
+
+So I don't have to manually create recurring job CRs and patch the labels.
+
+### User Experience In Detail
+
+#### Story 1 - set recurring job in volume
+
+1. Create a recurring job on UI `Recurring Job` page or via `kubectl`.
+2. In UI, Navigate to Volume, `Recurring Jobs Schedule`.
+3. When clicked on the `New` should have the option to choose from `Job` or `Job Group`.
```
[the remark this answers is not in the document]
yasker: Oh, I see the creation part has been addressed below. So this is the list part.

yasker (changes requested):
Only minor comments. Looks good in general. I will do a final review next time.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1087 @@
...
+  4. Filter and create a list of the volumes in the state `attached` or setting `allow-recurring-job-while-volume-detached`.
+  5. Use the concurrent number parameter to throttle goroutine with channel. Each goroutine creates `NewJob()` and `job.run()` for the volumes.
+     The job `snapshotName` format will be `<job.name>-c-<RandomID>`.
+
+#### Changes In The Volume Controller
+
+  1. Remove `updateRecurringJobs` method.
+
+- In the manager/volume, create recurring jobs for volumes from the `storageClass` if not already exist.
+  > Since the storage class and upgrade migration contains recurringJobs spec. So we will keep the `VolumeSpec.RecurringJobs` in code to create the recurring jobs for volumes from the `storageClass`.
```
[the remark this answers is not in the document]
c3y1huang: https://github.com/longhorn/longhorn/issues/2866
https://github.com/longhorn/longhorn/issues/2865

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: The upgrade will automatically convert the `recurringJobs` to `recurringJobSelector` and also the volume spec `recurringJobs` to labels. The `recurringJobs` will be empty after upgrade, However this field is still functional, for example if the user add the `recurringJobs` into storageClass then it will get converted to labels and a new recurring job CR will get created.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1087 @@
...
+
+### Goals
+
+Phase 1:
+- Each recurring job can be in single or multiple groups.
+- Each group can have single or multiple recurring jobs.
+- The jobs and groups can be reference with the volume label.
+- The recurring job in `default` group will automatically apply to a volume that has no job labels.
+- Can create multiple recurring jobs with UI or YAML.
+- The recurring job should include settings; `name`, `groups`, `task`, `cron`, `retain`, `concurrency`.
```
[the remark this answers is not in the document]
c3y1huang: Fixed.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1087 @@
...
+
+### User Experience In Detail
+
+#### Story 1 - set recurring job in volume
+
+1. Create a recurring job on UI `Recurring Job` page or via `kubectl`.
+2. In UI, Navigate to Volume, `Recurring Jobs Schedule`.
+3. When clicked on the `New` should have the option to choose from `Job` or `Job Group`.
+  - When select `Job`,
+    1. User sees all existing recurring jobs from the `Name` drop-down list.
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: Since the names should be unique so I think this is not needed. Removed.

enhancements/20210624-label-driven-recurring-job.md
yasker: So we still keep the storage class recurring jobs functional for v1.2.0. In that case, this will be fine.

yasker (approved):
LGTM. Feel free to merge if there is no objection from other maintainers.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1026 @@
...
+  3. Get volumes by label selector `recurring-job-group.longhorn.io/<job.group>: enabled` to filter out volumes if the job is set with a group.
+  4. Filter and create a list of the volumes in the state `attached` or setting `allow-recurring-job-while-volume-detached`.
+  5. Use the concurrent number parameter to throttle goroutine with channel. Each goroutine creates `NewJob()` and `job.run()` for the volumes.
+     The job `snapshotName` format will be `<job.name>-c-<RandomID>`.
+
+#### Changes In The Volume Controller
+
+  1. Remove `updateRecurringJobs` method.
+
+- In the manager/volume, create recurring jobs for volumes from the `storageClass` if not already exist.
```
shuo-wu, line 190: Shouldn't this be done by the CSI? In the CSI volume creation function, we can find all matching recurring job CRs based on field `recurringJobs` or `recurringJobSelector`. If there is no matching result, the plugin can send HTTP calls to create new ones.

enhancements/20210624-label-driven-recurring-job.md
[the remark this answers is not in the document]
shuo-wu: We can retain the field itself but remove the code related to it. In the new version, all volumes created by the old StorageClass can directly use labels as well:
1. In the CSI plugin, check the old StorageClass `recurringJobs` and filter out the recurring job CRs based on corn value. If no exist, the plugin can create them.
2. Create a volume and set recurring job labels

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1026 @@
...
+     2. Remove the `recurringJobs` from volume spec.
+
+#### Introduce `recurringJobSelector` As Part Of StorageClass Parameters.
+  - The CSI controller can use `recurringJobSelector` for volume creation.
+
+#### Changes In CSI Controller Server
+
+  1. Wait after the volume is created by API client.
+  2. Get recurring job names from the storageClass parameter `recurringJobSelector`.
+  2. Add volume labels by volume `ActionRecurringJobAdd` API with `input.IsStorageClass`. Volume manager will delete the default label when sees `input.IsStorageClass`.
```
shuo-wu, line 211: Can the volume creation API directly accept the recurring job lists as a new input field?

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1026 @@
...
+
+#### Add Recurring Job Controller
+
+  - The code structure will be the same as other controllers.
+  - Add the finalizer to the recurring job CRs if not exist.
+  - The controller will be informed by `recurringJobInformer` and `volumeInformer`.
+    - `recurringJobInformer` do `enqueueRecurringJob`.
+    - `volumeInformer` get all recurring job from volume recurring job labels and `enqueueRecurringJob`. It will also label and update the volume with `default` job-group if no recurring job label is found.
```
shuo-wu, line 219: I am thinking if we really need this informer. The main purpose of this controller is to create, update, and delete the corn jobs. If we don't delete the corn jobs when there is no volume using a recurring job, this workflow can be simplified and the informer is no longer required.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1026 @@
...
+         longhorn-manager -d\
+           recurring-job <job.name>\
+           --manager-url <url>
+         ```
+    2. Delete the cron job if annotation `last-applied-cronjob-spec` is different from cron job object spec.
+    3. Create cron job with annotation `last-applied-cronjob-spec`.
+    4. Update cron job if the new cron job spec is different from the `last-applied-cronjob-spec`.
+  - Use defer to clean up CronJob.
+    1. When a recurring job gets deleted.
+      1. Delete the cron job with selected labels: `recurring-job.longhorn.io/<Name>`.
```
shuo-wu, line 237: Do you mean removing the labels for all volumes referred to this curring job?

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1026 @@
...
+  "name": "sample",
+  "retain": 2,
+  "task": "snapshot",
+  "type": "recurringJob"
+}
+```
+
+**Scenario: Delete Recurring Job**
+
+*Given* an `Operation` drop-down list next to the recurring job.
```
shuo-wu, line 377: Add a button for batch deletion on the top left.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1026 @@
...
+      1. Delete the cron job with selected labels: `recurring-job.longhorn.io/<Name>`.
+      2. Remove the finalizer.
+    2. When there is no volume with label `recurring-job.longhorn.io/<Name>: enabled` and `recurring-job-group.longhorn.io/<Group>: enabled`.
+
+
+#### UI
+
+##### Add New Page `Recurring Job` In UI
+
+A new page for `Recurring Job` to create/update/delete recurring jobs.
```
shuo-wu, line 246: In the Recurring job list page, which fields need to be shown in that list/table? And can we aggregate recurring jobs by group names?

shuo-wu (changes requested):

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,539 @@
+# Label-driven Recurring Job
+
+## Summary
+
+Replace volume spec `recurringJobs` with the label-driven model. Allow users to create recurring backup/snapshot jobs with new `RecurringJob` CRD. The created recurring jobs can be referenced by adding the recurring job and group to volume with labels.
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1026 @@
...
+- The recurring job in `default` group will automatically apply to a volume that has no job labels.
+- Can create multiple recurring jobs with UI or YAML.
+- The recurring job should include settings; `name`, `groups`, `task`, `cron`, `retain`, `concurrency`, `labels`.
+
+Phase2:
+The StorageClass and upgrade migration are still dependent on volume spec, thus complete removal of volume spec should be done in phase 2.
+
+### Non-goals [optional]
+
+1. When volumes are all using the same recurring backup/snapshot policy, means all volumes will perform backups at the same time. As the original issue suggested, `Ideally, a range should be provided.`. This is currently an existing issue and should also be a problem when the recurring jobs are set at the storage class level.
```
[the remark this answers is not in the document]
c3y1huang: This does look a bit confusing, I've removed this because job concurrency is added and this statement was removed from the original issue.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1026 @@
...
+
+> https://github.com/longhorn/longhorn/pull/2737#issuecomment-887985811
+
+## Proposal
+
+#### Story 1 - set recurring jobs and groups by volume labels
+
+As a Longhorn user / System administrator.
+
+I want to create backup/snapshot recurring jobs that can be referenced in volume by the job name.
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1026 @@
...
+
+**Before enhancement**
+
+Default recurring jobs are set via StorageClass at PVC creation only. No default recurring job can be set up for UI-created volumes.
+
+Updating StorageClass does not reflect on the existing volumes.
+
+**After enhancement**
+
+Have the option to set default recurring jobs via `StorageClass` or `RecurringJob`. The `RecurringJob` respects when the `recurringJobs` are given in the `StorageClass`.
```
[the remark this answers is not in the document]
c3y1huang: Reworded the statement.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1026 @@
...
+Longhorn recurring job controller automatically applies default recurring jobs to all volumes without the job labels.
+
+Longhorn adds the default recurring jobs when all job labels are removed from the volume. 
+
+#### Story 3 - automatically upgrade migration
+
+1. Perform upgrade.
+2. StorageClass `recurringJobs` will get convert to `recurringJobSelector`.
+3. Recurring job CRs will get created from `recurringJobs`.
+4. The `recurringJobs` will get removed from the storageClass.
```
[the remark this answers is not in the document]
c3y1huang: Removed.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1026 @@
...
+5. Volume will be labeled with `recurring-job.longhorn.io/<jobTask>-<jobRetain>-<hash(jobCron)>-<hash(jobLabelJSON)>: enabled` from volume spec `recurringJobs`.
+6. Recurring job CRs will get created from volume spec `recurringJobs`.
```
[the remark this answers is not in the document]
c3y1huang: Yes. Updated the statement.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1026 @@
...
+- Add new APIs for users to update recurring jobs for individual volume. The :
+  - `/v1/volumes/<VOLUME_NAME>?action=recurringJobAdd`, expect request's body in form {name:<name>, isGroup:<bool>}.
+  - `/v1/volumes/<VOLUME_NAME>?action=recurringJobList`.
+  - `/v1/volumes/<VOLUME_NAME>?action=recurringJobDelete`, expect request's body in form {name:<name>, isGroup:<bool>}.
+
+## Design
+
+### Implementation Overview
+
+#### Add Recurring Job CRD.
```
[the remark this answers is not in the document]
c3y1huang: No special labels in the recurring job CRD, the labels and annotation will be set in the cronjob instead.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1026 @@
...
+       Also, add validation in the CRD YAML with pattern regex match.
+     - The `Cron`: String in cron expression represents recurring job scheduling.
+     - The `Retain`: Integer of the number of snapshots/backups to keep for the volume.
+     - The `Concurrency`: Integer of the concurrent job to run by each cron job.
+     - The `Age`: Date of the CR creation timestamp.
+     - The `Labels`: Dictionary of the labels.
+
+#### Add Command `recurring-job` To `longhorn-manager` Binary
+
+  1. Add new command `recurring-job <job.name> --longhorn-manager <URL>`.
```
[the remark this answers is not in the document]
c3y1huang: I think we can remove it. Updated the statement.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1026 @@
...
+     > Get the `recurringJob.Spec` on execution using Kubernetes API.
+  2. Get volumes by label selector `recurring-job.longhorn.io/<job.name>: enabled` to filter out volumes.
+  3. Get volumes by label selector `recurring-job-group.longhorn.io/<job.group>: enabled` to filter out volumes if the job is set with a group.
+  4. Filter and create a list of the volumes in the state `attached` or setting `allow-recurring-job-while-volume-detached`.
+  5. Use the concurrent number parameter to throttle goroutine with channel. Each goroutine creates `NewJob()` and `job.run()` for the volumes.
```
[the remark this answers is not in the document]
c3y1huang: Both.

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: Yes, it makes sense to have this done in CSI. Updated.

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: I am not sure if it makes sense from the user's point of view if the job pod still gets created if no volume is referencing it. No strong opinions on this one...

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: I think we should keep the volume labels in case the recurring job gets deleted by accident, the user can add it back and the recurring job is still referenced for the volume.

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: Scenario updated.

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: > In the Recurring job list page, which fields need to be shown in that list/table? 

I've updated a table in the commit.

> And can we aggregate recurring jobs by group names?

I think we can aggregate with the search box.

enhancements/20210624-label-driven-recurring-job.md
shuo-wu: I don't understand what you mean... If the input `recurringJobs` is not empty, it means Longhorn can add the corresponding labels in the manifest before calling the creation API. And there is no need to use the `default` group for the volume.

enhancements/20210624-label-driven-recurring-job.md
shuo-wu: Yes. The cron job will launch pods that do nothing then directly become `complete`. That's the cost of this idea. I think it's acceptable. But the benefit is, this controller no longer needs to take care of volume. This makes both the enqueuing and the reconcile loop become simpler.

enhancements/20210624-label-driven-recurring-job.md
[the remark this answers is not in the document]
shuo-wu: Since you decide to create recurring jobs in CSI, we don't need to do this anymore, right?

enhancements/20210624-label-driven-recurring-job.md
shuo-wu: Not sure why there are 2 search boxes below. Maybe I need to wait for a UI demo image before discussing this part.

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: I've removed the volume informer and I think it wouldn't hurt if we keep the cron job cleanup.

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: After discussing with @shuo-wu , I've updated to use `Volume.Spec.RecurringJobSelector` for recurring job labeling 
 during `CreateVolume` for storageClass.

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: Sorry, there should be only one. Updated.

enhancements/20210624-label-driven-recurring-job.md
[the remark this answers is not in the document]
c3y1huang: > Since you decide to create recurring jobs in CSI, we don't need to do this anymore, right?

Not entirely sure if you are referring to the deprecation notice or the field removal?

For field removal, the recurringJob parameter will still be functional, but it will be translated to labels and recurring job CRs instead of `recurringJobs` in volume spec. The `recurringJobs` in volume spec will no longer be visible.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1046 @@
...
+         ```
+       - Compose command,
+         ```
+         longhorn-manager -d\
+           recurring-job <job.name>\
+           --manager-url <url>
+         ```
+    2. Delete the cron job if annotation `last-applied-cronjob-spec` is different from cron job object spec.
+    3. Create cron job with annotation `last-applied-cronjob-spec`.
+    4. Update cron job if the new cron job spec is different from the `last-applied-cronjob-spec`.
```
shuo-wu, line 231: Contradict to statement 2

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1046 @@
...
+  - Use defer to clean up CronJob.
+    1. When a recurring job gets deleted.
+      1. Delete the cron job with selected labels: `recurring-job.longhorn.io/<Name>`.
+      2. Remove the finalizer.
+    2. When there is no volume with label `recurring-job.longhorn.io/<Name>: enabled` and `recurring-job-group.longhorn.io/<Group>: enabled`.
```
shuo-wu, line 236: As we discussed before, the controller doesn't need to remove the cron job even if there is no volume using it. Let the cron job create a pod that does nothing. I don't think users care about this.

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1046 @@
...
+       - Compose command,
+         ```
+         longhorn-manager -d\
+           recurring-job <job.name>\
+           --manager-url <url>
+         ```
+    2. Delete the cron job if annotation `last-applied-cronjob-spec` is different from cron job object spec.
+    3. Create cron job with annotation `last-applied-cronjob-spec`.
+    4. Update cron job if the new cron job spec is different from the `last-applied-cronjob-spec`.
+  - Use defer to clean up CronJob.
```
shuo-wu, line 232: `defer`? I think you can do cleanup only when the deletion timestamp of a recurring job is not empty.

BTW, do you consider using cascading deletion here?
1. Add owner reference when creating the cron job for each recurring job.
2. Set option `DeletePropagationForeground` in the recurring job deletion function.

shuo-wu (changes requested):

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: It will sit in the `reconcileRecurringJob` I think we can discuss the detail in the implementation PR.

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: Removed.

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: Fixed.

enhancements/20210624-label-driven-recurring-job.md
shuo-wu: Sure

enhancements/20210624-label-driven-recurring-job.md
shuo-wu: I mean, adding a new field in the [HTTP API layer](https://github.com/longhorn/longhorn-manager/blob/d4ee01d214c95768048294eaadae2595ba6b6f0e/api/model.go#L22) then we can specify the recurring jobs in the HTTP VolumeCreation call. It doesn't mean that this field needs to be added in `VolumeSpec` or `VolumeStatus`. You can refer to [this field as an example](https://github.com/longhorn/longhorn-manager/blob/d4ee01d214c95768048294eaadae2595ba6b6f0e/api/model.go#L979).

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1043 @@
...
+
+  - For `CreateVolume`:
+     1. Add recurring job label if `Volume.Spec.RecurringJobs` is not empty.
+     2. Remove the `recurringJobs` from volume spec.
+     3. Add recurring job label if `Volume.Spec.RecurringJobSelector` is not empty.
+     4. Remove the `recurringJobSelector` from volume spec.
+     5. Label with `default` job-group if no other recurring job label exists.
+  - For `UpdateVolume`:
+     1. Label with `default` job-group if no other recurring job label exists.
+     2. Remove the `recurringJobs` from volume spec.
```
shuo-wu, line 202: For `CreateVolume` and `UpdateVolume`:
1. Add recurring job labels if `Volume.Spec.RecurringJobs` is not empty. Then unset `Volume.Spec.RecurringJobs`.
2. Label with `default` job-group if no other recurring job label exists.

shuo-wu (changes requested):

enhancements/20210624-label-driven-recurring-job.md
c3y1huang: Fixed.

shuo-wu (approved):

enhancements/20210624-label-driven-recurring-job.md
```diff
@@ -0,0 +1,1043 @@
...
+     > Get the `recurringJob.Spec` on execution using Kubernetes API.
+  2. Get volumes by label selector `recurring-job.longhorn.io/<job.name>: enabled` to filter out volumes.
+  3. Get volumes by label selector `recurring-job-group.longhorn.io/<job.group>: enabled` to filter out volumes if the job is set with a group.
+  4. Filter and create a list of the volumes in the state `attached` or setting `allow-recurring-job-while-volume-detached`.
+  5. Use the concurrent number parameter to throttle goroutine with channel. Each goroutine creates `NewJob()` and `job.run()` for the volumes.
+     The job `snapshotName` format will be `<job.name>-c-<RandomID>`.
+
+#### Changes In The Volume Controller
+
+  1. Remove `updateRecurringJobs` method.
```
c3y1huang, line 186: Just realize the `updateRecurringJobs` method cannot get removed since the volume informer is not in the recurring job controller. This method should be responsible to add default labels if no other labels exist.

cc @yasker @shuo-wu

innobead (approved):
LGTM

@c3y1huang thanks for this kind of very detailed design documentation more than a proposal I expect.

Issue #2699, opened 2021-06-17
[BUG] Liveness/Readiness Probe timeoutSeconds at Kubernetes v1.20+


jenting:
**Describe the bug**

I encounter an issue on Kubernetes v1.21.0, the kubelet with log:
```shell
2021-05-12T05:03:40.341728+00:00 jenting-longhorn-master-0 k3s[17498]: E0512 05:03:40.341637   17498 remote_runtime.go:394] "ExecSync cmd from runtime service failed" err="rpc error: code = DeadlineExceeded desc = failed to exec in container: timeout 1s exceeded: context deadline exceeded" containerID="2772def98866331ae76c100ab412ce1961a1e35e19bf4f9f9345077a0d733f9b" cmd=[sh -c ls /data/longhorn && /data/longhorn version --client-only]
```

The workaround to fix it is to increase the `readinessProbe.timeoutSeconds` on engine-image-ei-xxx DaemonSet from 1 sec to 3 secs. According to the https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes, start from Kubernetes 1.20, the `timeoutSeconds` behavior got fixed. Therefore, if the user's Kubernetes cluster that the `readinesssProbe` is over than 1 sec, the Pod can't mount value successfully.  The only way to know the root cause is to dump the kubelet log.
![image](https://user-images.githubusercontent.com/49380831/122930159-4d4b3500-d39e-11eb-9a75-6f85ecae32ac.png)

**To Reproduce**

It's not easy to reproduce, I could only reproduce it on my local desktop and creates 8 nodes (2 master nodes + 6 worker nodes)
and install Longhorn on top of it.

**Expected behavior**

All Longhorn Pods should run successfully and also Pods are able to mount the volume.

**Log**
If applicable, add the Longhorn managers' log when the issue happens.

You can also attach a *Support Bundle* here. You can generate a Support Bundle using the link at the footer of the Longhorn UI.

**Environment:**
 - Longhorn version: v1.1.1
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl): Helm
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: k3s v1.21.0
   - Number of management node in the cluster: 2
   - Number of worker node in the cluster: 6
 - Node config
   - OS type and version: 
   - CPU per node: 
   - Memory per node:
   - Disk type(e.g. SSD/NVMe):
[...]

PR #2773, opened 2021-07-07
[cherry-pick-v1.1.2] increase liveness timeoutseconds

jenting:
#### Related Issue

https://github.com/longhorn/longhorn/issues/2699


innobead:
1.1.2 was released, so no need to cherry-pick, just make sure the code on the master.

Issue #2757, opened 2021-06-30
[TASK] Bump Kubernetes, API version group, CSI component's dependency version


yasker:
Every minor release is a chance for us to bump up the Kubernetes (client-go etc) and CSI component versions.

Currently, our manager depends on Kubernetes v1.16, and CSI components are also a few versions behind.

We also need to take care of the necessary API version group bump on Kubernetes. Refer to https://kubernetes.io/docs/reference/using-api/deprecation-guide/ for details.

For example, `extensions/v1beta1` will be removed in v1.22, which is the next Kubernetes release. `policy/v1beta1` will be removed in v1.25 which means we likely can wait until the next minor release to update it.

We might also decide to keep the external dependency version the same for now, but need a strong reason for that.

joshimoo:
ref for the csi component updates #2388 
Anytime we do the kubernetes updates we should also evaluate what the min go version is going to be.
Since over time that will also be updated as part of the kubernetes release.

Currently kubernetes 1.18 is already end of life and 1.19 will be in a couple of weeks.

joshimoo:
@innobead we should check if the leader election patch set can be removed after this upgrade as well, ref https://github.com/longhorn/longhorn-manager/pull/639

jenting:
https://github.com/longhorn/longhorn/issues/2828

PhanLe1010:
Thanks @joshimoo and @jenting I totally forgot about this patch

But the news is things seems working without the patch when testing. I will double check the code

PhanLe1010:
It seems that Kubernetes and client-go v1.18 fixed the problem and we no longer needs the above patch.
https://github.com/longhorn/longhorn-manager/pull/639#issuecomment-905030885

jenting:
> It seems that Kubernetes and client-go v1.18 fixed the problem and we no longer needs the above patch.
> [longhorn/longhorn-manager#639 (comment)](https://github.com/longhorn/longhorn-manager/pull/639#issuecomment-905030885)

Please help close #2828 :pray:

innobead:
@PhanLe1010 we missed this to backport to 1.1.3 because this issue was closed. I will create another one to track for 1.1.3.

PR #2809, opened 2021-07-19
Bump ingress api version to networking.k8s.io/v1

jenting:
#### Proposal Changes

extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+
- use `networking.k8s.io/v1beta1` if Kubernetes version < 1.19
- use `networking.k8s.io/v1` if Kubernetes version >= 1.19

ref to: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122

#### Issues

https://github.com/longhorn/longhorn/issues/2757


chart/templates/ingress.yaml
```diff
@@ -1,5 +1,9 @@
 {{- if .Values.ingress.enabled }}
-apiVersion: extensions/v1beta1
+{{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion -}}
```
innobead, line 2: Is this compatible for helm v2 and v3?

jenting: - helmv3: https://github.com/helm/helm/blob/6d9e15b/pkg/chartutil/create.go#L223-L228
- helmv2: https://github.com/helm/helm/blob/dev-v2/pkg/chartutil/create.go#L154-L158

innobead (approved):

Issue #2530, opened 2021-04-23
[FEATURE] Backing image enhancement


yasker:
- [x] We need to support checksum to validate the integrity of the downloaded backing image.
~- [ ] After downloaded, backing images needs to be stored across different nodes as well, in case one node went down then we might lose the image in the cluster since the internet is not always available in some edge cases. **_(optional, tackle in https://github.com/longhorn/longhorn/issues/2856)_**~
- [x] We also need to initiate the download when the image is added since there is no guarantee that the internet is available when needed later.

innobead:
And some refactoring efforts should be considered in https://github.com/longhorn/longhorn/pull/2617#issuecomment-846966089.

innobead:
@smallteeths Please help update the backing image modal as below to make the margin/indent of the value of `parameters during creation` looks better.

![image](https://user-images.githubusercontent.com/899492/128802127-5bbc3204-9245-46f9-8e92-353f41207f71.png)

shuo-wu:
Will address the 2nd feautre in #2856

kaxing:
Validating in progress with master image(v1.2.0-rc2)

With following test cases:

Node Down
https://longhorn.github.io/longhorn-tests/manual/pre-release/node/backing-image-on-a-down-node/

Longhorn Upgrade
https://longhorn.github.io/longhorn-tests/manual/pre-release/upgrade/backing-image-during-upgrade/

Error Reporting and Retry
https://longhorn.github.io/longhorn-tests/manual/pre-release/ha/backing-image-error-reporting-and-retry/

Upload, create via CSI and CRD yaml file
https://longhorn.github.io/longhorn-tests/manual/release-specific/v1.2.0/test-backing-image-upload/

kaxing:
Validation: **PASSED**

Tested with Longhorn v1.2.0-rc2

New feature UI highlight for future reference:

| List of Backing Images | Dialog of creating Backing Image | Dialog of backing image detail |
| - | - | - |
| ![image](https://user-images.githubusercontent.com/602540/131284240-12f13730-e53d-49d6-95da-d85d0fb3a253.png) | ![image](https://user-images.githubusercontent.com/602540/131283866-9e051476-dfd7-4e1d-9760-8333c4d9277a.png) | ![image](https://user-images.githubusercontent.com/602540/131283929-9dd610ca-84bd-4826-b4a9-4ea21d101669.png) |

A note on creating backing image via StorageClass, these are latest parameters:
```yaml
[...]

PR #2854, opened 2021-08-10
chart: Update chart for the enhanced backing image feature

shuo-wu:
Longhorn #2530

Will update the backing image manager image tag during the release.


innobead:
@shuo-wu we should do it again at 1.2.0 branch instead of the master branch because of new stuff introduced.

innobead (approved):
LGTM

PR #2890, opened 2021-08-17
Update chart and YAML for Label driven recurring job

c3y1huang:
https://github.com/longhorn/longhorn/issues/467


innobead (approved):
LGTM merge this after removing the draft

innobead (approved):

PR #2930, opened 2021-08-25
chart: Add backing image options for default StorageClass

shuo-wu:
longhorn/longhorn#2403
longhorn/longhorn#2404
longhorn/longhorn#2530


innobead:
@shuo-wu is it ready to review?

Issue #2963, opened 2021-09-01
[BUG] Helm chart kubeVersion requirement doesn't support complex SemVer


chasestech:
**Describe the bug**
When using the Helm chart to upgrade from version 1.1.2 to 1.2.0, I received: `Error: chart requires kubeVersion: >=v1.18.0 which is incompatible with Kubernetes v1.21.3-3+90fd5f3d2aea0a`

The previous version of the Helm chart allowed for complex Kubernetes semantic versions, but the new version of the chart does not. Changing the constraint to `kubeVersion: >=1.18.0-r0` succeeded in my local testing on both a complex and non-complex semver Kubernetes distro.

**To Reproduce**
1. Using a Kubernetes distribution that has a complex semver, such as MicroK8s, attempt to install or upgrade to Helm Chart version 1.2.0. Install/upgrade will fail.

**Expected behavior**
New version of Helm chart should support Kubernetes versions that use complex semver format.

**Log**
If applicable, add the Longhorn managers' log when the issue happens.

You can also attach a *Support Bundle* here. You can generate a Support Bundle using the link at the footer of the Longhorn UI.

**Environment:**
 - Longhorn version: 1.1.2
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl): Helm
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: MicroK8s v1.21.3-3+90fd5f3d2aea0a
   - Number of management node in the cluster: 1
   - Number of worker node in the cluster: 1
 - Node config
   - OS type and version: Ubuntu 20.04
   - CPU per node: 12
   - Memory per node: 64
   - Disk type(e.g. SSD/NVMe): HDD
   - Network bandwidth between the nodes: NA
 - Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal): Baremetal
 - Number of Longhorn volumes in the cluster: 8

**Additional context**
NA

jenting:
Thank you @chasestech We'll fix it
cc @PhanLe1010

innobead:
cc @PhanLe1010

shuo-wu:
Should we modify it to `"kubeVersion: >=1.18.0-0"` rather than `"kubeVersion: >=1.18.0-r0"`? To allow prereleases, typically the constraints suffix `-0` will be used. And I am not sure if we can still use the suffix `-r0`:
https://github.com/rancher/rancher/issues/34588

[...]

PR #2969, opened 2021-09-02
Helm chart doesn't allow `-xxx` as valid version

jenting:
#### Proposal Change

Helm chart doesn't allow `-xxx` as a valid version unless we add `-0` and without prefix `v`.
    
See helm/helm#9371 for details.

Besides that, the semver parser doesn't allow with prefix `v`.
https://github.com/rancher/rancher/blob/release/v2.5.9/pkg/catalog/manager/manager.go#L218-L221

#### Issue

https://github.com/longhorn/longhorn/issues/2963

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -18,7 +18,6 @@ sources:
 - https://github.com/longhorn/longhorn-engine
 - https://github.com/longhorn/longhorn-instance-manager
 - https://github.com/longhorn/longhorn-share-manager
-- https://github.com/longhorn/backing-image-manager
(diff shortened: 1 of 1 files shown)


PhanLe1010 (approved):

jenting:
@innobead What branch should I fix at? v1.2.1 _or_ master branch?

jenting:
> @innobead What branch should I fix at? v1.2.1 _or_ master branch?

Change to v1.2.1 branch.

innobead (approved):
LGTM

Issue #2964, opened 2021-09-01
[BUG] Longhorn 1.2.0 - wrong volume permissions inside container / broken fsGroup


bkupidura:
**Describe the bug**
After upgrade longhorn to 1.20, some container are unable to start corectly (e.g prometheus).

Looks like root cause is wrong Longhorn volume permisions inside container when container is not running as root.

Even with `fsGroup` specified, permissions are not set for volume.

**To Reproduce**
```
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: broken-longhorn
  namespace: default
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
  storageClassName: longhorn

---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: broken-longhorn
  namespace: default
  labels:
    app.kubernetes.io/name: broken-longhorn
spec:
  replicas: 1
  strategy:
    type: Recreate
  selector:
    matchLabels:
      app.kubernetes.io/name: broken-longhorn
  template:
    metadata:
      labels:
        app.kubernetes.io/name: broken-longhorn
    spec:
      containers:
        - name: broken-longhorn
          image: ubuntu:focal-20210723
          command:
          - "/bin/sh"
          - "-ec"
          - |
            tail -f /dev/null
          imagePullPolicy: IfNotPresent
          volumeMounts:
            - mountPath: /data
              name: data
      securityContext:
        runAsUser: 65534
        runAsNonRoot: true
        runAsGroup: 65534
        fsGroup: 65534
        fsGroupChangePolicy: Always
      volumes:
        - name: data
          persistentVolumeClaim:
            claimName: broken-longhorn
```

```
% kubectl get pvc broken-longhorn -o yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"PersistentVolumeClaim","metadata":{"annotations":{},"name":"broken-longhorn","namespace":"default"},"spec":{"accessModes":["ReadWriteOnce"],"resources":{"requests":{"storage":"1Gi"}},"storageClassName":"longhorn"}}
    pv.kubernetes.io/bind-completed: "yes"
    pv.kubernetes.io/bound-by-controller: "yes"
    volume.beta.kubernetes.io/storage-provisioner: driver.longhorn.io
[...]

PR #2974, opened 2021-09-02
Chart: explicitly set default fsType to EXT4

PhanLe1010:
This allow Kubernetes to properly changes the volume ownership
and permissions when user specify fsGroup in workload pod

longhorn/longhorn#2964

### DIFF
--- a/chart/templates/storageclass.yaml
+++ b/chart/templates/storageclass.yaml
@@ -20,6 +20,9 @@ data:
       numberOfReplicas: "{{ .Values.persistence.defaultClassReplicaCount }}"
       staleReplicaTimeout: "30"
       fromBackup: ""
+      {{- if .Values.persistence.defaultFsType }}
+      fsType: "{{.Values.persistence.defaultFsType}}"
+      {{- end }}
       {{- if .Values.persistence.backingImage.enable }}
       backingImage: {{ .Values.persistence.backingImage.name }}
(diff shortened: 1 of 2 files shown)


shuo-wu (commented):
Quick question: If the filesystem type is not set in the CSI provisioner, does adding this field in StorageClass work?

PhanLe1010:
> Quick question: If the filesystem type is not set in the CSI provisioner, does adding this field in StorageClass work?

Yeah. CSI provisioner check and see that the SC has `fsType` so it use it regardless of whether we set the default fs in csi provisioner

shuo-wu:
> > Quick question: If the filesystem type is not set in the CSI provisioner, does adding this field in StorageClass work?
> 
> Yeah. CSI provisioner checks and see that the SC has `fsType` so it uses it regardless of whether we set the default fs in csi provisioner

Great. Then please:
1. mention this as a workaround in both https://github.com/longhorn/longhorn/issues/2964#issuecomment-910969543 and the Slack channel.
2. Update the default StorageClass and the example StorageClass

shuo-wu (approved):

innobead (approved):

PhanLe1010:
@shuo-wu we discussed in Slack that we don't want to recommend user to change longhorn SC so we don't want to mention this workaround. 


I will update the SC example.

Issue #1879, opened 2020-10-14
[FEATURE]Separate the CRD file from the Chart template directory


shuo-wu:
If the CRD file is directly moved out from the directory, Helm will wrongly delete the CRDs. Hence we plan to add annotation `helm.sh/resource-policy: keep` to avoid the unexpected deletion in the v1.1.0, then separate the CRD file in v1.2.0.

The original discussion is in #1820

innobead:
Removing from the sprint first because right now we don't have a good solution to tackle this.

PhanLe1010:
After evaluating the options:

1. Moving the crd in upstream from `./templates` to `./crds` is hard since we have to deal with how to make crd not being deleted, [link](https://github.com/longhorn/longhorn/issues/1879). Also using this approach means that we cannot upgrade or remove the crd: [link](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#install-a-crd-declaration-before-using-the-resource:~:text=There%20is%20no%20support%20at%20this%20time%20for%20upgrading%20or%20deleting%20CRDs%20using%20Helm) ===> even though we can have some kind of pre-hook and post-hook to workaround, this approach is complicated and error-prone
1. Moving the longhorn crd to a separate chart ===> I agree this one is the best approach. There are 2 parts for this approach:
    1. Update the upstream Longhorn chart to separate longhorn and longhorn-crd chart. This will require some annotation and label migration: https://github.com/longhorn/longhorn/pull/3224#issuecomment-986391892. This can be done in the future.
    1. Update the rancher-longhorn chart to separate rancher-longhorn chart and racnher-longhorn-crd chart. Currently, we can set both rancher-longhorn chart and racnher-longhorn-crd chart to track the same longhorn upstream chart. In the future, when part `2.i` is done, we can set racnher-longhorn-crd to track the upstream longhorn-crd chart and rancher-longhorn to track the upstream longhorn chart. This is possible as confirmed by Arvind [here](https://suse.slack.com/archives/C02DR3N5T24/p1661994218940599?thread_ts=1661871640.487689&cid=C02DR3N5T24)

PhanLe1010:
I think we should close this ticket in favor of [the](https://github.com/longhorn/longhorn/pull/3224)

How do you think @shuo-wu @innobead ?

PhanLe1010:
How do you think @shuo-wu @innobead ?

PR #3224, opened 2021-11-09
Separate the CRDs to another chart

jenting:
According to helm chart CRD best practices
https://helm.sh/docs/chart_best_practices/custom_resource_definitions/
We have to ensure CRDs are installed before any CRs created during the
chart installation.

Therefore, we propose to separate the CRDs to another chart called
longhorn-crd, this is how currently Rancher package Longhorn chart.

https://github.com/longhorn/longhorn/issues/1879

Signed-off-by: JenTing Hsiao <jenting.hsiao@suse.com>


innobead:
@jenting this should be for 1.2.3 instead of the master branch directly.

jenting:
Should I change the Version/AppVersion to v1.3.0 as well?

shuo-wu:
After the splitting, can the Longhorn deployed in Rancher cluster manager or cluster explorer be upgraded correctly?

jenting:
> After the splitting, can the Longhorn deployed in Rancher cluster manager or cluster explorer be upgraded correctly?

If you are talking about the chart published by Rancher, it splits into 2 charts already. So, install/upgrade on Rancher with rancher chart is okay. The problem is how the Rancher team packages the Longhorn chart to the Rancher chart, but I think we could tackle it once we file PR to the Rancher chart.
```
rancher/longhorn                   	100.1.0+up1.2.2 	v1.2.2     	Longhorn is a distributed block storage system ...
rancher/longhorn-crd                    100.1.0+up1.2.2 	           	Installs the CRDs for longhorn.                   
```

So the left problem is if the user the chart repo `https://charts.longhorn.io`, and the user must install/upgrade longhorn-crd chart before the longhorn chart. And this will be addressed by documentation.

shuo-wu:
1. For Rancher Cluster Explorer Chart, what we did is move the CRD to the sub-directory `crds` via (a patch) so that Rancher Cluster Explorer can detect the CRDs then create a sub-chart automatically. (The 1st method mentioned in the Helm doc).
2. But for the Rancher Cluster Manager Chart, I don't know how to achieve that. And will upgrading the CRD chart then the main Chart here cause any issues? This is what I want to confirm.

innobead:
> 1. For Rancher Cluster Explorer Chart, what we did is move the CRD to the sub-directory `crds` via (a patch) so that Rancher Cluster Explorer can detect the CRDs then create a sub-chart automatically. (The 1st method mentioned in the Helm doc).
> 2. But for the Rancher Cluster Manager Chart, I don't know how to achieve that. And will upgrading the CRD chart then the main Chart here cause any issues? This is what I want to confirm.

I think both of them follow the same mechanism to use independent charts instead of `crds` folder and use `catalog.cattle.io/auto-install: longhorn-crd=match` to determine which chart should be installed first. 

image

- For cluster explorer, as you know, they are already separate, so that should be fine.
- For cluster manager, we need to follow the mechanism to create another longhorn-crd and make the dependency for longhorn chart via the annotation.

shuo-wu:
> * For cluster manager, we need to follow the mechanism to create another longhorn-crd and make the dependency for longhorn chart via the annotation.

Previously when I moved the CRD YAML file into a separate directory then upgraded the Longhorn system, Helm considered the old CRD YAML are no longer needed and tried to delete them. (I don't remember if I did any other modifications when this is triggered.) Hope the annotation would avoid this kind of potential CRD deletion.

jenting:
Tried helm upgrade from Longhorn chart v1.2.2.

Install/Upgrade the `longhorn-crd` chart first
```shell
helm upgrade --install longhorn-crd <chart-crd-path> -n longhorn-system
```

it failed with an error message:
```shell
Error: rendered manifests contain a resource that already exists. Unable to continue with install: CustomResourceDefinition "engines.longhorn.io" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "longhorn-crd": current value is "longhorn"
```

Need to think about how to migrate the CRD folder...

jenting:
Close it because current we can't move the CRDs into the new chart, it failed on the upgrade case.
This is because the CRDs original owner is the release name `longhorn`.
Once we move the CRDs into a new chart, the owner will be the release name `longhorn-crd`.
However, the helm doesn't allow the owner to change.

PhanLe1010:
We can use a script to migrate annotation similar to what we did for rancher chart migration https://github.com/longhorn/longhorn/issues/3714

After evaluating the options:

1. Moving the crd in upstream from `./templates` to `./crds` is hard since we have to deal with how to make crd not being deleted, [link](https://github.com/longhorn/longhorn/issues/1879). Also using this approach means that we cannot upgrade or remove the crd: [link](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#install-a-crd-declaration-before-using-the-resource:~:text=There%20is%20no%20support%20at%20this%20time%20for%20upgrading%20or%20deleting%20CRDs%20using%20Helm) ===> even though we can have some kind of pre-hook and post-hook to workaround, this approach is complicated and error-prone
1. Moving the longhorn crd to a separate chart ===> I agree this one is the best approach. There are 2 parts for this approach:
    1. Update the upstream Longhorn chart to separate longhorn and longhorn-crd chart. This will require some annotation and label migration: https://github.com/longhorn/longhorn/pull/3224#issuecomment-986391892. This can be done in the future.
    1. Update the rancher-longhorn chart to separate rancher-longhorn chart and racnher-longhorn-crd chart. Currently, we can set both rancher-longhorn chart and racnher-longhorn-crd chart to track the same longhorn upstream chart. In the future, when part `2.i` is done, we can set racnher-longhorn-crd to track the upstream longhorn-crd chart and rancher-longhorn to track the upstream longhorn chart. This is possible as confirmed by Arvind [here](https://suse.slack.com/archives/C02DR3N5T24/p1661994218940599?thread_ts=1661871640.487689&cid=C02DR3N5T24)

PR #3274, opened 2021-11-18
Update CRD with structural schema

jenting:
#### Propose Change

- Copy the CRD yaml manifest from longhorn-manager
- Manually add the `{{- include "longhorn.labels" . | nindent 4 }}` after `metadata.labels`.

#### Issue
Longhorn 791


jenting:
Test basic fresh install and upgrade cases. Both works.

innobead (approved):

Issue #2285, opened 2021-03-02
[FEATURE] Longhorn should support multi-network K8s clusters (storage network)


janeczku:
**Is your feature request related to a problem? Please describe.**

A typical bare-metal server/network architecture involves servers being attached to multiple networks, including a dedicated storage fabric that provides accelerated throughput and IO separation from the data plane.

Longhorn does not support segregate networks and uses the default CNI network for all the storage traffic. That makes it impossible to guarantee network resources available for storage operations because the same network is shared with all in-cluster application traffic.

**Describe the solution you'd like**

![Multi-Network-Storage](https://user-images.githubusercontent.com/3813921/109638614-80975900-7b4e-11eb-9c2e-71a874910e4a.png)

The canonical approach to provide multi-network capabilities in a K8s cluster is using a CNI multiplexer such as Multus which provides the means to attach pods to different underlay networks.

Longhorn should have an integration with Multus.
That may be implemented by exposing a global setting allowing the admin to specify the Multus network that should be used for any storage-related traffic (e.g. engine -> replica). Longhorn would then configure the pods with the Multus specific annotations, e.g. `k8s.v1.cni.cncf.io/networks: storage-net`.

The setup and configuration of the Multus network (e.g. creation of `NetworkAttachmentDefinition`) would be out of the scope of Longhorn.

**Describe alternatives you've considered**
N/A

yasker:
Considering moving this back to v1.2.0.

yasker:
Move out to v1.3.0.

innobead:
cc @jenting

jenting:
Interested.

As I know, Harvester uses Multus, it'd benefit the Harvester project.

Also, make sure the Longhorn engine <-> replica network throughput would not be affected by other applications.

innobead:
Doing the evaluation in 1.2.0 to see if we could come out with experimental support in 1.2.0. Otherwise, 1.3.0 instead.

innobead:
@jenting as we discussed, please sum up where we are and what original ideas you plan to do for the evaluation.

jenting:
For the longhorn-engine replica part, the data server and sync-agent server are the data plane.
Having the setting allows the user to specify which network interface is the data plane.
At the longhorn engine, it gets the network interface IPs from the Pod and creates the data and sync-agent server with that interface IP address. (Need a way to report error if the network interface not found)
[...]

Issue #2821, opened 2021-07-26
[EPIC] Improve communication between components (GRPC)


joshimoo:
We need to improve the overall communication handling between components.

For the GRPC handling there are currently issues like:
- one shot connections, we end up opening/closing a new connection for a single action (costly)
- reuse connections and rely on the grpc multiplexer to bypass connection establishment (reduces latency)
- better error handling and retry mechanisms

For longhorn-manager <-> engine is using binary invocations as part of the controller reconcile loop.
- the binary invocations should be replaced with grpc communication
- the monitoring should run Independent from the the controller reconcile loop

joshimoo:
Basically one refactor at a time will get us there slowly :)
I layed the ground work for grpc connection reuse here:
https://github.com/longhorn/longhorn-engine/pull/644

This also has an example on replacing the task methods.
The goal should be to eventually turn all task methods into free functions where the consumer passes the client in.

innobead:
See if need to have a LEP for this. Assigned to @c3y1huang.

innobead:
Assigned back to @joshimoo who is working on this.

PR #3277, opened 2021-11-18
Add LEP for Storage Network Through gRPC Proxy

c3y1huang:
https://github.com/longhorn/longhorn/issues/2285
https://github.com/longhorn/longhorn/issues/3546

### DIFF
--- a/enhancements/20220428-storage-network-through-grpc-proxy.md
+++ b/enhancements/20220428-storage-network-through-grpc-proxy.md
@@ -1,0 +1,323 @@
+# Storage Network Through gRPC Proxy
+
+## Summary
+
+Currently, Longhorn uses the Kubernetes cluster CNI network and share the network with the entire cluster resources. This makes network availability impossible to control.
+
+We would like to have a global `Storage Network` setting to allow users to input an existing Multus `NetworkAttachmentDefinition` CR network in `<namespace>/<name>` format. Longhorn can use the storage network for in-cluster data traffics.
+
+The segregation can achieve by replacing the engine binary calls in the Longhorn manager with gRPC connections to the instance manager. Then the instance manager will be responsible for handling the requests between the management network and storage network.
+
+---
+**_NOTE:_** There are other possible approaches we have considered to segregating the networks:
+
+- Add Longhorn Manager to the storage network. The Manager needs to restart itself to get the secondary storage network IP, and there is no storage network segregation to the Longhorn data plane (engine & replica).
+
+- Provide Engine/Replica with dual IPs. Code change around this approach is confusing and likely to increase maintenance complexity.
+---
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/2285
+
+https://github.com/longhorn/longhorn/issues/3546
+
+## Motivation
+
+### Goals
+
+- Have a new `Storage Network` setting.
+
+- Replace Manager engine binary calls with gRPC client to the instance manager.
+
+- Keep using the management network for the communication between Manager and Instance Manager.
+
+- Use the storage network for the data traffic of data plane components to the instance processes. Those are the engines and replicas in Instance Manager pods.
+
+- Support backward compatibility of the communication between the new Manager and the old Instance Manager after the upgrade. Ensure existing engine/replicas work without issues.
+
+### Non-goals [optional]
+
+- Setup and configure the Multus `NetworkAttachmentDefinition` CRs.
+
+- Monitor for `NetworkAttachmentDefintition` CRs. The user needs to ensure the traffic is reachable between pods and across different nodes. Without monitoring, Longhorn will not get notified of the update of the `NetworkAttachmentDefinition` CRs. Thus the user should create a new `NetworkAttachmentDefinition` CR and update the `storage-network` setting.
+
+- Out-cluster data traffic. For example, backing image upload and download.
+
+
+## Proposal
+
+### Communication between Manager and Engine/Replica processes via Instance Manager gRPC proxy
+
+- Introduce a new gRPC server in Instance Manager.
+
+- Keep re-usable connections between Manager and Instance Managers.
+
+- Allow Manager to fall back to engine binary call when communicating with old Instance Manager.
+
+### Storage Network
+
+- Add a new `Storage Network` global setting.
+
+- Add `k8s.v1.cni.cncf.io/networks` annotation to pods that involve data transfer. The annotation will use the value from the storage network setting. Multus will attach a secondary network to pods with this annotation.
+  - Engine instance manager pods
+  - Replica instance manager pods
+  - Backing image data source pods
+  - Backing image manager pods
+
+- Add new `storageIP` to `Engine`, `Replica` and `BackingImageManager` CRD status. The storage IP will be use to communicate to the instance processes.
+
+### User Stories
+
+#### Story 1 - set up the storage network
+
+As a Longhorn user / System administrator.
+
+I have set up Multus `NetworkAttachmentDefinition` for additional network management.
+And I want to segregate Longhorn in-cluster data traffic with an additional network interface.
+Longhorn should provide a setting to input the `NetworkAttachmentDefinition` CR name for the storage network.
+
+So I can guarantee network availability for Longhorn in-cluster data traffic.
+
+
+#### Story 2 - upgrade
+
+As a Longhorn user / System administrator.
+
+When I upgrade Longhorn, the changes should support existing attached volumes.
+
+So I can decide when to upgrade the Engine Image.
+
+
+### User Experience In Detail
+
+#### Story 1 - set up the storage network
+
+1. I have a Kubernetes cluster with Multus installed.
+1. I created `NetworkAttachmentDefinition` CR and ensured the configuration is correct.
+1. I Added `<namespace>/<NetworkAttachmentDefinition name>` to Longhorn `Storage Network` setting.
+1. I see setting update failed when volumes are attached.
+1. I detach all volumes.
+1. When updating the setting I see engine/replica instance manager pod and backing image manager pods is restarted.
+1. I attach the volumes.
+1. I describe Engine, Replica, and BackingImageManager, and see the `storageIP` in CR status is in the range of the `NetworkAttachmentDefinition` subnet/CIDR. I also see the `storageIP` is different from the `ip` in CR status.
+1. I describe the Engine and see the `replicaAddressMap` in CR spec and status is using the storage IP. 
+1. I see pod logs indicate the network directions.
+
+#### Story 2 - upgrade
+
+1. I Longhorn v1.2.4 cluster.
+1. I have healthy volumes attached.
+1. I upgrade Longhorn.
+1. I see volumes still attached and healthy with available engine image upgrade.
+1. I cannot upgrade the volume engine image with the volume attached.
+1. After I detach the volume, I can upgrade its engine image.
+1. I attached the volumes.
+1. I see the volumes are healthy.
+
+### API changes
+
+- The new global setting `Storage Network` will use the existing `/v1/settings` API.
+
+## Design
+
+### Overview gRPC Proxy Implementation
+
+#### Instance Manager
+
+- Start the gRPC proxy server with the next port to the process server. The default should be `localhost:8501`.
+- The gRPC proxy service shares the same `imrpc` package name as the process server.
+  ```
+    Ping
+
+    ServerVersionGet
+
+    VolumeGet
+    VolumeExpand
+    VolumeFrontendStart
+    VolumeFrontendShutdown
+
+    VolumeSnapshot
+    SnapshotList
+    SnapshotRevert
+    SnapshotPurge
+    SnapshotPurgeStatus
+    SnapshotClone
+    SnapshotCloneStatus
+    SnapshotRemove
+
+    SnapshotBackup
+    SnapshotBackupStatus
+    BackupRestore
+    BackupRestoreStatus
+    BackupVolumeList
+    BackupVolumeGet
+    BackupGet
+    BackupConfigMetaGet
+    BackupRemove
+
+    ReplicaAdd
+    ReplicaList
+    ReplicaRebuildingStatus
+    ReplicaVerifyRebuild
+    ReplicaRemove
+  ```
+
+#### Manager
+
+- Create a _proxyHandler_ object to map the controller ID to an _EngineClient_ interface. The _proxyHandler_ object is shared between controllers.
+
+- The Instance Manager Controller is responsible for the life cycle of the proxy gRPC client. For every enqueue:
+  - Check for the existing gRPC client in the _proxyHandler_, and check the connection liveness with the `Ping` request.
+  - If the proxy gRPC client connection is dead, stop the proxy gRPC client and error so it will re-queue.
+  - If the proxy gRPC client doesn't exist in the _proxyHandler_, start a new gRPC connection and map it to the current controller ID.
+  - Do not create the proxy gRPC connection when the instance manager version is less than the current version. We will provide the fallback interface caller provided when getting the client.
+
+- The gRPC client will use the _EngineClient_ interface.
+  - Provide a fallback interface caller when getting the gRPC client from the _proxyHandler_. The fallback callers are:
+    -  the existing `Engine` client used for the binary call
+    -  `BackupTargetClient`.
+  - Use the fallback caller when the instance manager version is less than the current version.
+  - Add new `BackupTargetBinaryClient` interface for fallback.
+    ```
+    type BackupTargetBinaryClient interface {
+	    BackupGet(destURL string, credential map[string]string) (*Backup, error)
+	    BackupVolumeGet(destURL string, credential map[string]string) (volume *BackupVolume, err error)
+	    BackupNameList(destURL, volumeName string, credential map[string]string) (names []string, err error)
+	    BackupVolumeNameList(destURL string, credential map[string]string) (names []string, err error)
+	    BackupDelete(destURL string, credential map[string]string) (err error)
+	    BackupVolumeDelete(destURL, volumeName string, credential map[string]string) (err error)
+	    BackupConfigMetaGet(destURL string, credential map[string]string) (*ConfigMetadata, error)
+    }
+    ```
+  - Introduce A new `EngineClientProxy` interface for the Proxy, which includes proxy-specific methods and implementation of the existing `EnglineClient` and `BackupTargetClient` interfaces. This will be adaptive when using the EngineClient interface for the proxy or non-proxy/fallback operations.
+    ```
+    type EngineClientProxy interface {
+      EngineClient
+      BackupTargetBinaryClient
+
+      IsGRPC() bool
+      Start(*longhorn.InstanceManager, logrus.FieldLogger, *datastore.DataStore) error
+      Stop(*longhorn.InstanceManager) error
+      Ping() error
+    }
+    ```
+
+### Overview Storage Network Overview Implementation
+
+#### Setting
+
+Add a new global setting `Storage Network`.
+- The setting is `string`.
+- The default value is `""`.
+- The setting should be in the `danger zone` category.
+- The setting will be validated at admission webhook setting validator.
+  - The setting should be in the form of `< NAMESPACE>/<NETWORK-ATTACHMENT-DEFINITION-NAME>`.
+  - The setting cannot be updated when volumes are attached.
+
+#### CRD
+
+Engine:
+- New `storageIP` in status.
+- Use the replica `status.storageIP` instead of the replica `status.IP` for the replicaAddressMap.
+
+Replica:
+- New `storageIP` in status.
+
+BackingImageManager:
+- New `storageIP` in status.
+
+#### Instance Manager Controller
+
+1. When creating instance manager pods, add `k8s.v1.cni.cncf.io/networks` annotation with `lhnet1` as interface name. Use the `storage-network` setting value for the namespace and name.
+    ```
+    k8s.v1.cni.cncf.io/networks: '
+      [
+        {
+          "namespace": "kube-system",
+          "name": "demo-10-30-0-0",
+          "interface": "lhnet1"
+        }
+      ]
+    '
+    ```
+
+#### Instance Handler
+
+1. Get the IP from instance manager Pod annotation `k8s.v1.cni.cncf.io/network-status`. Use the IP for `Engine` and `Replica` Storage IP. When the `storage-network` setting is empty, The Storage IP will be the pod IP.
+
+#### Backing Image Manager Controller
+
+1. When creating backing image manager pods, add `k8s.v1.cni.cncf.io/networks` annotation with `lhnet1` as interface name. Use the `storage-network` setting value for the namespace and name.
+    ```
+    k8s.v1.cni.cncf.io/networks: '
+      [
+        {
+          "namespace": "kube-system",
+          "name": "demo-10-30-0-0",
+          "interface": "lhnet1"
+        }
+      ]
+    '
+    ```
+1. Get the IP from backing image manager Pod annotation `k8s.v1.cni.cncf.io/network-status`. Use the IP for `BackingImageManager` Storage IP. When the `storage-network` setting is empty, The Storage IP will be the pod IP.
+
+#### Backing Image Data Source Controller
+
+1. When creating backing image data source pods, add `k8s.v1.cni.cncf.io/networks` annotation with `lhnet1` as interface name. Use the `storage-network` setting value for the namespace and name.
+    ```
+    k8s.v1.cni.cncf.io/networks: '
+      [
+        {
+          "namespace": "kube-system",
+          "name": "demo-10-30-0-0",
+          "interface": "lhnet1"
+        }
+      ]
+    '
+    ```
+
+#### Backing Image Manager - Export From volume
+
+1. get the IPv4 of the `lhnet1` interface and use it as the receiver address. Use the pod IP if the interface doesn't exist.
+
+#### Setting Controller
+
+1. Do not update the `storage-network` setting and return an error when `Volumes` are attached.
+1. Delete all backing image manager pods.
+1. Delete all instance manager pods.
+
+
+### Test plan
+
+#### CI Pipeline
+
+All existing tests should pass when the cluster has the storage network configured. We should consider having a new test pipeline for the storage network.
+
+Infra Prerequisites:
+- Secondary network interface added to each cluster instance.
+- Multus deployed.
+- Network-attachment-definition created.
+- Routing is configured in all cluster nodes to ensure the network is accessible between instances.
+- For AWS, disable network source/destination checks for each cloud-provider instance.
+
+#### Test storage-network setting
+
+Scenario: `Engine`, `Replica` and `BackingImageManager` should use IP in `storage-network` `NetworkAttachmentDefinition` subnet/CIDR range after setting update.
+
+### Upgrade strategy
+
+[Some old instance manager pods are still running after upgrade](https://longhorn.io/kb/troubleshooting-some-old-instance-manager-pods-are-still-running-after-upgrade/).
+Old engine instance managers do not have the gRPC proxy server for Manager to communicate.
+Hence, we need to support backward compatibility.
+
+Manager communication:
+- Bump instance manager API version.
+- Manager checks for incompatible version and fall back to requests through the engine binary.
+
+Volume/Engine live upgrade:
+- Keep live upgrade. This will be a soft notice for users to know we will not enforce any change in 1.3, but it will happen in 1.4.
+
+## Note [optional]
+
+`None`


enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,143 @@
...
+- Add a new global setting `Storage Network`.
+  - The setting is `string`.
+  - The setting should be in the form of `<NAMESPACE>/<NETWORK-ATTACHMENT-DEFINITION-NAME>`.
+  - The default value is `""`.
+  - The setting should be in the `danger zone` category. 
+
+#### ClusterRole
+- Add new cluster role for `k8s.cni.cncf.io/network-attachment-definitions`.
+
+#### Instance Manager Controller
```
shuo-wu, line 95: Similar logic should be applied to the backing image part. Notice that Longhorn directly uses PodIP when talking with backing image manager/data source pods. e.g.: https://github.com/longhorn/longhorn-manager/blob/7bd89abdbae3c406ec56094364cf6c2da1d36d9d/api/forwarder.go#L185

If you are not familiar with the backing image part, I can explain more details for you in a separate session.

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Goals
+
+- Can set storage network at deployment.
+- Can update the storage network with global settings on the running cluster.
+- The storage network applies for Longhorn data plan (engine, replica) traffic, segregate from the Longhorn control plan traffic.
+
+### Non-goals [optional]
+
+- Setup and configuration of the Multus and network-attachment-definition.
```
shuo-wu, line 25: Will we have a simple example as a quick hands-on?

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,143 @@
...
+- Monitor the changes for network-attachment-definition.
+
+## Proposal
+
+- Introduce a new global setting `storage-network`.
+- No change will apply if `storage-network` setting is not set.
+- The setting `storage-network` will be used for `k8s.v1.cni.cncf.io/networks` annotation.
```
shuo-wu, line 32: The setting `storage-network` will be the value of annotation `k8s.v1.cni.cncf.io/networks`.

shuo-wu (commented):
I am thinking if the backing image or volume download feature will break the network isolation and lead to any risks when this kind of storage network is set up.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Addressed.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Thank you for the discussion. Part added. And as discussed, we can probably leave out the upload and download of backing image as in my opinion storage network is mainly for inter-cluster.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: I prefer not to expose the example in the doc because it takes some more manual setups on the cluster nodes and could also vary base on platforms. However, I will be adding manual instructions and sample in the test case.

shuo-wu: OK, got it.

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,155 @@
...
+
+https://github.com/longhorn/longhorn/issues/2285
+
+## Motivation
+
+### Goals
+
+- Can set storage network at deployment.
+- Can update the storage network with global settings on the running cluster.
+- The storage network applies for Longhorn data plan (engine, replica) traffic, segregate from the Longhorn control plan traffic.
```
innobead, line 21: data plane: engine, replica and backing image data sources.

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Non-goals [optional]
+
+- Setup and configuration of the Multus and network-attachment-definition.
```
innobead, line 25: Correct. 

It also means we will not focus on what types of the network or ipam being to used for the storage network like overlay, ipvlan/macvlan, etc. Just need to make sure the second interface also fulfill the conditions of the first cluster network. (ref: http://kubernetes.io/docs/concepts/cluster-administration/networking/)

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,155 @@
...
+- Monitor the changes for network-attachment-definition.
+- Upload and download of backing image.
```
innobead, line 27: I assume we only care about the traffic between our data path components. So for backing image, image sync between nodes or volume to/from backing image are what we care.

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Proposal
+
+- Introduce a new global setting `storage-network`.
+- No change will apply if `storage-network` setting is not set.
+- The setting `storage-network` will be the value of annotation `k8s.v1.cni.cncf.io/networks`.
+- The manager DaemonSet `spec.template.metadata` will have `k8s.v1.cni.cncf.io/networks` annotation with the `storage-network` setting value.
```
innobead, line 34: `k8s.v1.cni.cncf.io/networks` should be only to apply to the pods related to data path? why longhorn-manager needs this? I assume only engine/replica instance manager pods or even backing image data source pods need that.

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,155 @@
...
+- The `Engine`, `Replica` and `BackingImageManager` status IP uses the network interface IP from the instance manager's `storage-network` interface.
```
innobead, line 35: What does `status IP` mean here? and the purpose?

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,155 @@
...
+- During setting application, the instance manager status will be marked to error. This should trigger instance manager pod re-creation with annotation from the `storage-network` setting value.
```
innobead, line 36: Or should we only allow this setting taking effect until all volumes detached then recreate these managed pods? Making instnace manager pods error to recreate, it sounds lke a little unsafe to possibly cause some data loss?

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,155 @@
...
+- Include `network-attachment-definition` for all namespaces in the support bundle.
```
innobead, line 37: Do you mean multus workloads? We don't manage multus deployment/provision, this should not be covered in our support bundle.

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,155 @@
...
+  - The setting is `string`.
+  - The setting should be in the form of `<NAMESPACE>/<NETWORK-ATTACHMENT-DEFINITION-NAME>`.
+  - The default value is `""`.
+  - The setting should be in the `danger zone` category. 
+
+#### ClusterRole
+- Add new cluster role for `k8s.cni.cncf.io/network-attachment-definitions`.
+
+#### Instance Manager Controller
+1. Check and return if NetworkAttachmentDefinition does not exist in the cluster.
```
innobead, line 100: Maybe we don't need this, otherwise you will need to operate NAD resources w/ RBAC. 

Just apply the annotation and if NAD and mutlus not there, just nothing happens. (note: because you can do the pre-check, but can't do the post-check to make sure the NAD still there. This condition check is probably a little meaningless)

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,155 @@
...
+1. Update the instance manager's current status to error if pod `network-attachment-definitions` annotation does not match the `storage-network` setting value.
+1. Annotate pod with `k8s.v1.cni.cncf.io/networks` to the `storage-network` setting value when creating instance manager pod.
+
+#### Instance Handler
+1. Use the instance managers `k8s.v1.cni.cncf.io/networks: <STORAGE-NETWORK>` network interface IP for `Engine` and `Replica` `Status.IP` instead of instance manager status IP.
```
innobead, line 105: What's the purpose?

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,155 @@
...
+
+#### Backing Image - Export From volume
+1. Assume the secondary network interface in Pod is from `storage-network`. Check the existence of a secondary network interface and use the IP instead of the Pod IP from the environment variable.
```
innobead, line 108: Do we consider the case of backing image sync among nodes?

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,155 @@
...
+
+#### Backing Image Manager Controller
+1. Check and return if NetworkAttachmentDefinition does not exist in the cluster.
+1. Use the instance managers `k8s.v1.cni.cncf.io/networks: <STORAGE-NETWORK>` network interface IP for `BackingImageManager` `Status.IP` instead of Pod status IP.
+
+#### Setting Controller
+1. Check if NetworkAttachmentDefinition exists.
+1. Delete all backing image manager Pods.
+1. Update manager Daemonset `.Spec.Template.Annotation[k8s.v1.cni.cncf.io/networks]` to the `storage-network` setting value.
```
innobead, line 117: Why manager needs the second interface?

innobead (changes requested):

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Longhorn uses the status IP to commute data between engine -> replica, backing-image-manager <-> backing-image-manager. For example, https://github.com/longhorn/longhorn-manager/blob/master/controller/backing_image_manager_controller.go#L708 and https://github.com/longhorn/longhorn-manager/blob/master/controller/engine_controller.go#L1228.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: I am thinking if we should also add the validation to the HTTP API client so it can be validated when the setting is changed via UI. Or just keep it silent when NAD doesn't exist? WDYT?

enhancements/20211118-storage-network-with-multus.md
c3y1huang: The backing-image-manager <-> backing-image-manager will use the backing-image-manager status IP which will get updated.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Fixed.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: You are right, it's not required. Removed.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: https://github.com/longhorn/longhorn/pull/3277#discussion_r765364137

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Not required, removed.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Sorry, Longhorn manager does need to have the second interface for the controller to update the status. For example, https://github.com/longhorn/longhorn-manager/blob/master/engineapi/engine.go#L92.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: After discussing with @innobead , Longhorn will not check the NAD because we do not want to add the dependency to operate the NAD resources.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: After discussing with @innobead , Longhorn will include the NADs in the support bundle because we do not want to add the dependency to operate the NAD resources.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: https://github.com/longhorn/longhorn/pull/3277#discussion_r765411366

enhancements/20211118-storage-network-with-multus.md
innobead: The status update should be through the controller interface instead?

c3y1huang: Currently, the Longhorn engine uses single IP for control, sync, and data. The easiest way is to add a secondary interface to the manager so the engine controller can get command over replicas. To separate that would take much more effort.

c3y1huang: LEP updated.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Introduce new `StorageIP` status instead. LEP updated.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Addressed.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Removed.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Removed.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: LEP updated.

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,337 @@
...
+- Factory creates should be provided with a cluster address to use for the service URLs. Also a storage address for data connection.
+- When creating a new replica client, use the storage address to get the host. The host in the client will be used for data transfers and sync. Use the cluster address for the service URLs.
+- Similar reason as the command `--replica` argument, most GRPC operation needs to include cluster address and storage address in the requests. For exampe:
+```
+cr, err := controllerServiceClient.ControllerReplicaCreate(ctx, &ptypes.ControllerReplicaCreateRequest{
+-               Address:          address,
++               Address:          address.Cluster + "," + address.Storage,
```
shuo-wu, line 117: Do we need to directly add required inputs for the engine/replica process creation instead? But this means the live upgrade is no longer supported.

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,337 @@
...
+        storage: 10.30.1.29:10000
+    requestedBackupRestore: ""
+    requestedDataSource: ""
+    revisionCounterDisabled: false
+    salvageRequested: false
+    upgradedReplicaAddressMap: {}
+    volumeName: t1
+    volumeSize: "52428800"
+  status:
+    cloneStatus:
```
shuo-wu, line 172: How about the IP of these status fields? Are they all using storage IPs?

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,337 @@
...
+- Replace the single address in `replicaAddressMap.<replica>` with maps of `cluster` and `storage`. Same for the `currentReplicaAddressMap` and `upgradedReplicaAddressMap`.
+
+Replica:
+- New `storageIP` in status.
+
+BackingImage:
+- New `storageIP` in status.
+
+#### Engine
+- For data-related commands must include cluster address and the storage address. Use the existing global `--url` command argument as the service URL for GRPC operations. Provide storage address to the `--replica` command argument for data transfer and sync. If a command does not include a `Pod` host IP  in the`--url` argument,  it should be included in the `--replica` argument in comma-separated string format, for example `--replica <pod IP>,<storage IP>`. This also applies when multiple replicas are provided to the command.
```
shuo-wu, line 110: Sorry, I still don't understand why the cluster IP is required. And what are the data-related commends? I think it's a little bit hard to define if a command is a data-related commend or not.

If I understand correctly:
1. Longhorn manger pods use the storage network to talk with engine processes, replica processes, and the instance manager pods. And this is why you need to add the longhorn manager pods to the storage network.
2. Engine processes always use the storage network talk with replica processes.

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,337 @@
...
+- Factory creates should be provided with a cluster address to use for the service URLs. Also a storage address for data connection.
```
shuo-wu, line 111: Same here. Why does a replica/engine process require the cluster IP?

shuo-wu (commented):
In general, `Implementation Overview` - `Engine` is still confusing. You can have tables to list the CRD changes and API changes.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: The storage-network will only be applied for the storage related data traffic (read-n-write, file sync), for example [here](https://github.com/longhorn/longhorn-engine/blob/master/pkg/sync/rpc/server.go#L439), [here](https://github.com/longhorn/longhorn-engine/blob/master/pkg/backend/remote/remote.go#L267) and [here](https://github.com/longhorn/longhorn-engine/blob/master/pkg/sync/rpc/server.go#L374) .The talking between processes will still be using the cluster-network, hence we don't need to apply storage-network in Longhorn manager.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Yes, they should be showing with the storage IPs.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Let me re-check on the live upgrade.

jenting (commented):
Are the user easily to know the reason that failed with these operations like volume attachment failed, replica replication failed is because of the storage network misconfigured?

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,337 @@
+# Title
```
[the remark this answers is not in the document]
c3y1huang: Thanks, updated.

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,337 @@
...
+#### Story 2 - update the storage network for existing cluster
+1. A Kubernetes cluster with Multus installed.
+1. Create `NetworkAttachmentDefinition` CR.
+1. Create volume-1.
+1. Create a backing image from volume-1.
+1. Create volume-2 from the backing image.
+1. Detach all volumes.
+1. Update Longhorn setting *storage-network* via `kubectl` or UI.
+1. Longhorn instance manager and backing image manager `Pods` will restart.
+1. Attach then volumes.
```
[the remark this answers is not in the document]
c3y1huang: Thanks, updated.

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,337 @@
...
+
+- Setup and configuration of the Multus and `NetworkAttachmentDefinition`.
+- Monitor for `NetworkAttachmentDefintition` CRs. This means the user will need to ensure the traffic is reachable between `Pods` and across different `Nodes`. Without monitoring, Longhorn will not get notified of the `NetworkAttachmentDefintion` CR updates. The user needs to create a new `NetworkAttachmentDefinition` CR and update the `storage-network` setting. 
+- Backing image upload and download.
+
+## Proposal
+
+- Introduce a new *storage-network* in Longhorn `Setting` CRD. The value will use as the `k8s.v1.cni.cncf.io/networks` annotation value so it must be in `<namespace/network-attachment-definition>` format.
+- Replace single string address in `replicaAddressMap` with maps of cluster address and storage address in the `Engine` CRD.
+- Introduce new `storageIP` status in `Engine`, `Replica` and `BackingImageManager` CRDs. It will be network interface IP from the parent manager's `storage-network` interface or use the `status.IP` if no not set for storage-network.
```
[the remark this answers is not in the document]
c3y1huang: Reworded.

c3y1huang:
> Are the user easily to know the reason that failed with these operations like volume attachment failed, replica replication failed is because of the storage network misconfigured?

The user is responsible to get the correct network configurations in. Multus deployment and configuring networks are out of scope. 

Longhorn can probably validate the secondary CNI at the setting controller with the informer. But unless we implement a network-attachment-definition controller, we will not know about any user post-updates that may break the traffics.

When misconfiguration happens, the best we could do is to ask for network-attachment-definition CRs, look for clues in the engine CRs and ask the user to check the ping response.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: https://github.com/longhorn/longhorn/pull/3277#discussion_r772156302

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,322 @@
...
+- Setup and configuration of the Multus and `NetworkAttachmentDefinition`.
+- Monitor for `NetworkAttachmentDefintition` CRs. This means the user will need to ensure the traffic is reachable between `Pods` and across different `Nodes`. Without monitoring, Longhorn will not get notified of the `NetworkAttachmentDefintion` CR updates. The user needs to create a new `NetworkAttachmentDefinition` CR and update the `storage-network` setting. 
+- Backing image upload and download.
+
+## Proposal
+
+- Introduce a new *storage-network* in Longhorn `Setting` CRD. The value will use as the `k8s.v1.cni.cncf.io/networks` annotation value so it must be in `<namespace/network-attachment-definition>` format.
+- Replace single string address in `replicaAddressMap` with maps of cluster address and storage address in the `Engine` CRD.
+- Introduce new `storageIP` status in `Engine`, `Replica` and `BackingImageManager` CRDs. It will be network interface IP from the parent manager's `storage-network` interface. Use the `status.IP` value if storage-network setting is empty.
+- Reflect/Show the correct addresses used in the `Engine`, `Replica`, and `BackingImageManager` CRs.
```
jenting, line 34: If I understand correctly, the correct addresses mean the `status.IP` and `status.storageIP` for cluster control address and storage address, right?

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,322 @@
...
+- Backing image upload and download.
+
+## Proposal
+
+- Introduce a new *storage-network* in Longhorn `Setting` CRD. The value will use as the `k8s.v1.cni.cncf.io/networks` annotation value so it must be in `<namespace/network-attachment-definition>` format.
+- Replace single string address in `replicaAddressMap` with maps of cluster address and storage address in the `Engine` CRD.
+- Introduce new `storageIP` status in `Engine`, `Replica` and `BackingImageManager` CRDs. It will be network interface IP from the parent manager's `storage-network` interface. Use the `status.IP` value if storage-network setting is empty.
+- Reflect/Show the correct addresses used in the `Engine`, `Replica`, and `BackingImageManager` CRs. 
+- Annotate the instance manager `Pods`, backing image data source `Pods` and backing image manager `Pods` with  `k8s.v1.cni.cncf.io/networks: <storage-network>` to apply the *storage-network* CNI.
+- When updating the *storage-network* `Setting`, Longhorn will restart all instance manager `Pods` and backing image manager `Pods` to apply the CNI annotations.
```
jenting, line 36: Does the backing image data source Pods need to restart as well? [1]

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,322 @@
...
+
+#### Story 2 - update the storage network for existing cluster
+1. A Kubernetes cluster with Multus installed.
+1. Create `NetworkAttachmentDefinition` CR.
+1. Create volume-1.
+1. Create a backing image from volume-1.
+1. Create volume-2 from the backing image.
+1. Detach all volumes.
+1. Update Longhorn setting *storage-network* via `kubectl` or UI.
+1. Longhorn instance manager and backing image manager `Pods` will restart.
```
jenting, line 78: [1]

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,322 @@
...
+1. Attach the volumes.
+1. Check the `Engine`, and `Replica` CRs and see `StorageIP` in status and address in the *storage-network* subnet/CIDR range.
```
jenting, line 80: I _guess_ the BackingImageManager CR should contain the `status.storageIP` as well, correct?

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,322 @@
...
+    replicaAddressMap:
+      t1-r-2f096c17: 10.42.1.17:10000,10.30.2.3:10000
+      t1-r-9d49997d: 10.42.2.19:10000,10.30.3.3:10000
+      t1-r-e300fee6: 10.42.0.20:10000,10.30.1.2:10000
+```
+
+Replica:
+- New `storageIP` in status.
+
+BackingImage:
```
jenting, line 112: BackingImageManager?

enhancements/20211118-storage-network-with-multus.md
jenting: Would you mind providing an example of what's the engine's command to launch a new replica before and after this LEP?

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,322 @@
...
+- New `storageIP` in status.
+
+#### Engine
+- For data-related commands must include cluster address and the storage address. Use the existing global `--url` command argument as the service URL for GRPC operations. Provide storage address to the `--replica` command argument for data transfer and sync. If a command does not include a `Pod` host IP  in the`--url` argument,  it should be included in the `--replica` argument in comma-separated string format, for example `--replica <pod IP>,<storage IP>`. This also applies when multiple replicas are provided to the command.
+- Factory creates should be provided with a cluster address to use for the service URLs. Also a storage address for data connection.
+- When creating a new replica client, use the storage address to get the host. The host in the client will be used for data transfers and sync. Use the cluster address for the service URLs.
```
jenting, line 118: Is the host means the host interface?

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,322 @@
...
+    started: true
+    storageIP: 10.30.1.3
+kind: List
+metadata:
+  resourceVersion: ""
+  selfLink: ""
+```
+
+#### Instance Manager Controller
+1. Annotate with `k8s.v1.cni.cncf.io/networks` to the `storage-network` setting value when creating instance manager `Pod` CRs.
```
jenting, line 273:
```suggestion
1. Annotate with `k8s.v1.cni.cncf.io/networks` to the `storage-network` setting value when creating instance manager `Pod`.
```

jenting: If the instance manager Pods already exist, do we need to restart the instance manager Pods? Or we only need to annotate the existing instance manager Pods? [2]

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,322 @@
...
+
+#### Instance Handler
+1. Use the `k8s.v1.cni.cncf.io/networks: <STORAGE-NETWORK>` network interface IP in instance managers `Pod` CR for the `Engine` and `Replica` `StorageIP` in status. When *storage-network* `Setting` is empty, `Status.IP` will be used instead.
```
jenting, line 276: Not quite understand this sentence. If I understand correctly, it means we'll update the `status.storageIP` _or_ `status.IP` from the instance manager Pod's annotation `k8s.v1.cni.cncf.io/networks: <STORAGE-NETWORK>` network interface IP address?

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,322 @@
...
+
+#### Backing Image - Export From volume
+1. Include cluster IP and storage IP in the sender address, comma-separated string.
+1. For receiver address, assume the secondary network interface in Pod is from then *storage-network* `Setting`. Check the existence of a secondary network interface and use this IP instead of the Pod IP from the environment variable.
+
+#### Backing Image Data Source Controller
+1. Annotate with `k8s.v1.cni.cncf.io/networks` to the *storage-network* `Setting` value when creating backing image data source `Pod`.
```
jenting, line 283: [2]

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,322 @@
...
+1. Include the cluster IP and storage IP in the CLI parameters.
+
+#### Backing Image Manager Controller
+1. Use the backing image manager pod `k8s.v1.cni.cncf.io/networks: <storage-network>` network interface IP for the `BackingImageManager` `Status.StorageIP`. If no additional network interface, `Status.IP` will be used instead.
+1. Annotate with `k8s.v1.cni.cncf.io/networks` to the `storage-network` setting value when creating backing image manager `Pod` CRs.
```
jenting, line 288: [2]

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,322 @@
...
+- Disable network source/destination checks for each cloud-provider instance.
+All existing tests should pass.
+
+#### Test create cluster with storage-network setting
+
+Scenario: `Engine`, `Replica` should use IP in `storage-network` `NetworkAttachmentDefinition` subnet/CIDR range after deployment.
+
+#### Test update storage-network setting
+
+Scenario: `Engine`, `Replica` and `BackingImageManager` should use IP in `storage-network` `NetworkAttachmentDefinition` subnet/CIDR range after setting update.
```
jenting, line 314: Do the `BackingImageDataSource` needs as well? Or this part means the CR `status.storageIP`?

enhancements/20211118-storage-network-with-multus.md
c3y1huang: All fields with addresses should reflect. I've updated the LEP.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Hmm.. I don't think we are expecting users to do any operation during storage-network setting updates, and the volume will need to be detached for the setting to apply, hence there shouldn't be any backing image data source Pods.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Yes, it's covered in the `Backing Image Manager Controller` section.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Thanks, fixed.

c3y1huang: Thanks, fixed!

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Thanks, fixed.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: There will be no change in the engine command's arguments, most of the commands already included the cluster IP in `--url`. If I'm correct, the `controller` command is the only one that doesn't include the correct IP in `--url` and `--replica` will include both storage and cluster network addresses.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Sorry, that was hard to understand. What I mean is the host in the `ReplicaClient` object. Reworded.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: > Annotate with `k8s.v1.cni.cncf.io/networks` to the `storage-network` setting

Updated.

c3y1huang: > If the instance manager Pods already exist, do we need to restart the instance manager Pods? Or we only need to annotate the existing instance manager Pods? [2]

Yes, it should be restarted and already covered in the `Setting Controller` section.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Reworded.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: https://github.com/longhorn/longhorn/pull/3277#discussion_r772773296

enhancements/20211118-storage-network-with-multus.md
c3y1huang: I've tested the live upgrade. We are inputting the cluster and storage network to the `controller --replica` argument so it shouldn't break the live upgrade.
https://github.com/longhorn/longhorn/pull/3277#discussion_r772781052

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,330 @@
...
+- Backing image upload and download.
+
+## Proposal
+
+- Introduce a new *storage-network* in Longhorn `Setting` CRD. The value will use as the `k8s.v1.cni.cncf.io/networks` annotation value so it must be in `<namespace/network-attachment-definition>` format.
+- Replace single string address in `replicaAddressMap` with maps of cluster address and storage address in the `Engine` CRD.
+- Introduce new `storageIP` status in `Engine`, `Replica` and `BackingImageManager` CRDs. It will be the network interface IP from the parent manager's `storage-network` interface. The `k8s.v1.cni.cncf.io/networks` annotation will not be applied when the *storage-network* setting is empty and the `status.storageIP` will use the value from the `status.IP`. When the secondary network does not exist in parent `Pods` then `status.IP` value will also be used for the `status.storageIP` instead.
```
jenting, line 33: Could you please elaborate more on who is the parent manager? The longhorn-manager _or_ Multus manager?

enhancements/20211118-storage-network-with-multus.md
jenting: Because I saw the above line would annotate the backing image data source Pods, so I'm wondering should the backing image data source Pods should be restarted or not.

enhancements/20211118-storage-network-with-multus.md
jenting: May I say that the value of `--url` and `<pod IP>` are the same?

jenting (commented):
The only part I don't quite understand now is how we tackle the backing image data source Pods.
- Should we add annotations to the backing image data source Pods?
- Should the baking image data source Pods be restarted after annotations?
- Should we add the `status.storageIP` to the backing image data source CRD?

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Mostly true except for the controller. [here](https://github.com/longhorn/longhorn-manager/blob/master/manager/engine.go#L220) -> [here](https://github.com/longhorn/longhorn-manager/blob/master/engineapi/engine.go#L49) -> [here](https://github.com/longhorn/longhorn-manager/blob/master/engineapi/engine.go#L61-L74)

enhancements/20211118-storage-network-with-multus.md
c3y1huang: The annotation will be applied for the new backing image data Pod. Since the backing image data source Pods requires volume to be attached, users will not be able to update the storage-network Settings.

c3y1huang:
> The only part I don't quite understand now is how we tackle the backing image data source Pods.
> 
> * Should we add annotations to the backing image data source Pods?

Yes, for data fetch from replica.

> * Should the baking image data source Pods be restarted after annotations?

No, user will not be able to update the storage-network setting while volume is attached.

> * Should we add the `status.storageIP` to the backing image data source CRD?

IMO No, the backing image data source Pods will be removed once it finishes data fetch. I couldn't think where Longhorn could use that info.

enhancements/20211118-storage-network-with-multus.md
jenting: I see.

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,330 @@
...
+### Non-goals [optional]
+
+- Setup and configuration of the Multus and `NetworkAttachmentDefinition`.
+- Monitor for `NetworkAttachmentDefintition` CRs. This means the user will need to ensure the traffic is reachable between `Pods` and across different `Nodes`. Without monitoring, Longhorn will not get notified of the `NetworkAttachmentDefintion` CR updates. The user needs to create a new `NetworkAttachmentDefinition` CR and update the `storage-network` setting. 
+- Backing image upload and download.
+
+## Proposal
+
+- Introduce a new *storage-network* in Longhorn `Setting` CRD. The value will use as the `k8s.v1.cni.cncf.io/networks` annotation value so it must be in `<namespace/network-attachment-definition>` format.
+- Replace single string address in `replicaAddressMap` with maps of cluster address and storage address in the `Engine` CRD.
```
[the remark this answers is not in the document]
c3y1huang: Thanks, reworded.

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,330 @@
...
+### Goals
+
+- Can set storage network at deployment and running cluster.
+- The storage network should only be responsible for the Longhorn data-plan traffic. This means the storage data traffic in-between cluster nodes (engine, replica, backing image data source, backing image manager).
+- The responsible CRs should reflect on the network in use.
+
+### Non-goals [optional]
+
+- Setup and configuration of the Multus and `NetworkAttachmentDefinition`.
+- Monitor for `NetworkAttachmentDefintition` CRs. This means the user will need to ensure the traffic is reachable between `Pods` and across different `Nodes`. Without monitoring, Longhorn will not get notified of the `NetworkAttachmentDefintion` CR updates. The user needs to create a new `NetworkAttachmentDefinition` CR and update the `storage-network` setting.
```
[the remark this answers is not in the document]
c3y1huang: Thanks, fixed.

enhancements/20211118-storage-network-with-multus.md
c3y1huang: Reworded.

enhancements/20211118-storage-network-with-multus.md
```diff
@@ -0,0 +1,330 @@
...
+- Backing image upload and download.
+
+## Proposal
+
+- Introduce a new *storage-network* in Longhorn `Setting` CRD. The value will use as the `k8s.v1.cni.cncf.io/networks` annotation value so it must be in `<namespace/network-attachment-definition>` format.
+- Replace the single address in `replicaAddressMap.<replica>` with cluster and storage addresses split by a comma.
+- Introduce new `storageIP` status in `Engine`, `Replica` and `BackingImageManager` CRDs. It will be the network interface IP from the corresponding manager `Pod` `storage-network` interface; the corresponding manager is referring to instance-manager `Pods` for `Engine`, `Replica` and backing-image-manager `Pods` for `BackingImageManager`. The `k8s.v1.cni.cncf.io/networks` annotation will not be applied when the *storage-network* setting is empty and the `status.storageIP` will use the value from the `status.IP`. When the secondary network does not exist in the instance manager's `Pods` then `status.IP` value will also be used for the `status.storageIP` instead.
```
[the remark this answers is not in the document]
c3y1huang: After discussing with @shuo-wu , using `BackingImageDataSource` `storageIP` makes more sense instead of checking the second network interface in Pod. I will update it later.

/cc @jenting

yasker:
One question @shuo-wu raised in https://github.com/longhorn/longhorn/pull/3277#discussion_r773112940 is, is it possible for us to use one IP for engine and replica?

Engine and replica can communicate via the storage IP. And Instance Manager can start engine and replica processes using the storage IP.

However, we need to decide what IP to use for the Longhorn manager to communicate with Instance Manager, Engine process, and implicitly replica process (via e.g. backup command). There are two scenarios:
1. The cluster network is routable to the storage network. Then the Longhorn manager can communicate with IM, engine, and replica without any problem.
    * But then we will require additional routing rules for those networks.
2. The cluster network is NOT routable to the storage network. Then either IM, engine, and replica need to have two IPs (one for each network) to keep the communication, or Longhorn Manager needs to be in the storage network as well. 
    * But it's hard to put Longhorn Manager in the storage network since that means Longhorn Manager needs to modify it's own spec. 
    * Can we mitigate the issue by only putting IM in charge of the communication, and keeping the single IP for engine and replica? That means Longhorn Manager won't have direct access to the engine and replicas, which is hard to implement in the current architecture. 

In the end, I feel the dual IP approach is probably the most flexible solution at the moment.

However, can we run the engine binary for communication in the IM instead of the Manager pod? If we can do that, we can avoid mgr directly communicate with the engine and replica, so keep the single IP for the engine and replica.

enhancements/20211118-storage-network-with-multus.md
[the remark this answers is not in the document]
c3y1huang: I've revisited this logic and am not sure if applying `storageIP` to `BackingImageDataSource` makes sense.

The `BackingImageDataSource` can only get and update its `storageIP` after the backing image data source Pod is created. Currently, the backing image manager used `status.podIP` to export volume. So for this to work somehow we need to update the Pod again so it contains the storageIP from the `BackingImageDataSource`?

I think it's more straightforward if we check for the secondary network when doing volume export. What do you think @shuo-wu ?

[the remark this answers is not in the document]
c3y1huang: After discussing with @shuo-wu , it's not easy to get the `storageIP` into the `BackingImageDataSource` Pod without adding a new API. Another way we agreed is to specify the interface name with the annotation and check for the pod network interface matching the annotated interface name.

yasker:
Talked with @c3y1huang . Since the code is almost ready, let's see how does it look like to implement the dual IP for the engine. It's the more flexible approach. The only concern I have is the complexity.

c3y1huang:
Ran the coretest yesterday and there are still 2 unexpected fails that require time to fix before I wrap up for the draft PR.

Shuo also suggested tackling https://github.com/longhorn/longhorn/issues/2821 with this issue so there is no engine binary invocation so keeping single IP for the engine and replica.

IMO if we decided to include https://github.com/longhorn/longhorn/issues/2821 in v1.3.0, then it would be best to tackle it together as Shuo suggested.

shuo-wu:
> @shuo-wu also suggested tackling #2821 with this issue so there is no engine binary invocation and use IM for proxy so keeping single IP for the engine and replica.
> 
> IMO if we decided to include #2821 in v1.3.0, then it would be best to tackle it together as @shuo-wu suggested.

If allowing a process to use a single (storage network) IP is preferred, I would suggest handling #2821 first.

innobead:
> > @shuo-wu also suggested tackling #2821 with this issue so there is no engine binary invocation and use IM for proxy so keeping single IP for the engine and replica.
> > IMO if we decided to include #2821 in v1.3.0, then it would be best to tackle it together as @shuo-wu suggested.
> 
> If allowing a process to use a single (storage network) IP is preferred, I would suggest handling #2821 first.

It seems we would like to have single IP to avoid complicated dual IP management, but the precondition is to have IM as a proxy to serve the requests from LM. 

If that's the case, what efforts will probably be introduced?
- A new proxy server in IM
- Extra requests processing and dispatching to the downstream engine/replica servers (controller/replica server)
- If the proxy dies, the downstream communications will not work. This is different from the current design that each communication between LM and engine/replica process is independent.

I do care about potential extra costs possibly introduced to IM pods, and then it will probably impact the stability of engine/replica processes. Dual IPs will not introduce extra resource costs instead of complicated IP management. 

This is a trade-off I would say but still encourage you guys to figure out the pros and cons, then update this LEP.

c3y1huang:
After discussing with @innobead and @shuo-wu , we have reached a consensus to add in the gRPC proxy. The communication between MGR and IM will use the cluster network then proxy to engine/replica processes via the storage network.

pros:
The long stream connection to replace MGR calls for engine binary. This should improve manager performance.
The is no direct communication between MGR and engine/replica processes so keeping single IP to the processes.
In case of disconnection, the client should automatically restore, and the proxy is running along with the gGRPC server which life cycle is the same as the IM pod.
Fits better with the SPDK implementation.

cons:
There is no upper bound to the IM pod, hence when the resource consumed reached the host's upper bound might make the node unresponsive. However, the assumption is proxy shouldn't cost too much extra overhead since they are in the same pod. After implementation, we should do some performance tests.

enhancements/20220428-storage-network-through-grpc-proxy.md
```diff
@@ -0,0 +1,345 @@
...
+    ReplicaVerifyRebuild
+    ReplicaRemove
+  ```
+
+#### Manager
+
+- Create a _proxyHandler_ object to map the controller ID to an _EngineClient_ interface. The _proxyHandler_ object is shared between controllers.
+
+- The Instance Manager Controller is responsible for the life cycle of the proxy gRPC client. For every enqueue:
+  - Check for the existing gRPC client in the _proxyHandler_, and check the connection liveness with the `Ping` request.
```
innobead, line 171: Suggest adding the context for this and the following statements only if at least one volume is attached. If no volumes on that node, there should be no any engine client connections.

enhancements/20220428-storage-network-through-grpc-proxy.md
```diff
@@ -0,0 +1,345 @@
...
+  - Do not create the proxy gRPC connection when the instance manager version is less than the current version. We will provide the fallback interface caller provided when getting the client.
+
+- The gRPC client will use the _EngineClient_ interface.
+  - Provide a fallback interface caller when getting the gRPC client from the _proxyHandler_. The fallback callers are:
+    -  the existing `Engine` client used for the binary call
+    -  `BackupTargetClient`.
+  - Use the fallback caller when the instance manager version is less than the current version.
+  - The EngineClient interface is used for `Proxy`, existing `Engine` and existing `BackupTargetClient` methods. Update any inconsistent collection of methods.
+    ```
+    type EngineClient interface {
```
innobead, line 183: I would suggest just using embedded interface w/ reusing the existing interfaces instead if possible. Usually for any exposed interfaces, we will not just update but rather extend w/ another interface, so we are still able to retrieve the corresponding implementation based on a different interface.

```golang
type EngineClientProxy interface {
    EngineClient
    BackupTargetClient
    
    IsGRPC() bool
    ResetAndStopProxyClient(*longhorn.InstanceManager) error
    Start(*longhorn.InstanceManager, logrus.FieldLogger, *datastore.DataStore) error
    Ping() error
}
```

enhancements/20220428-storage-network-through-grpc-proxy.md
```diff
@@ -0,0 +1,345 @@
...
+    	IsGRPC() bool
+    	ResetAndStopProxyClient(*longhorn.InstanceManager) error
```
innobead, line 185: Why not just use `Stop` for the method name?

enhancements/20220428-storage-network-through-grpc-proxy.md
```diff
@@ -0,0 +1,345 @@
...
+
+### Overview Storage Network Overview Implementation
+
+#### Setting
+
+- Add a new global setting `Storage Network`.
+  - The setting is `string`.
+  - The default value is `""`.
+  - The setting should be in the `danger zone` category.
+  - The setting should be in the form of `< NAMESPACE>/<NETWORK-ATTACHMENT-DEFINITION-NAME>`.
```
innobead, line 237: Need to have a validation in validation webhook.

enhancements/20220428-storage-network-through-grpc-proxy.md
```diff
@@ -0,0 +1,345 @@
...
+  - The setting cannot be updated when volumes are attached. And will be validated at admission webhook setting validator.
+
+#### CRD
+
+Engine:
+- New `storageIP` in status.
+- Use the replica `status.storageIP` instead of the replica `status.IP` for the replicaAddressMap.
```
innobead, line 244: Why not using the same variable but have another flag to determine this IP is from the storage network?

enhancements/20220428-storage-network-through-grpc-proxy.md
```diff
@@ -0,0 +1,345 @@
...
+          "name": "demo-10-30-0-0",
+          "interface": "lhnet1"
+        }
+      ]
+    '
+    ```
+
+#### Instance Handler
+
+1. Get the IP from instance manager Pod annotation `k8s.v1.cni.cncf.io/network-status`. Use the IP for `Engine` and `Replica` Storage IP. When the `storage-network` setting is empty, The Storage IP will be the pod IP.
```
innobead, line 269: It seems we can just use the original variable `IP` for all cases, because you still use `StorageIP` adaptively here.

enhancements/20220428-storage-network-through-grpc-proxy.md
```diff
@@ -0,0 +1,345 @@
...
+    '
+    ```
+
+#### Backing Image Manager - Export From volume
+
+1. get the IPv4 of the `lhnet1` interface and use it as the receiver address. Use the pod IP if the interface doesn't exist.
+
+#### Setting Controller
+
+1. Do not update the `storage-network` setting and return an error when `Volumes` are attached.
```
innobead, line 308: I feel there will probably a timing issue for this validation at runtime, and how can you validate this during the upgrade case if users configure the storage network in values.yaml, then upgrade the chart?

If this is not 100% granted validation, I would suggest letting users configure, but it will take effect only if all volumes detached. However, we can have a tip info next to the setting description.

innobead (changes requested):
In general, LGTM.

enhancements/20220428-storage-network-through-grpc-proxy.md
```diff
@@ -0,0 +1,345 @@
...
+Currently, Longhorn uses the Kubernetes cluster CNI network and share the network with the entire cluster resources. This makes network availability impossible to control.
+
+We would like to have a global `Storage Network` setting to allow users to input an existing Multus `NetworkAttachmentDefinition` CR network in `<namespace>/<name>` format. Longhorn can use the storage network for in-cluster data traffics.
+
+The segregation can achieve by replacing the engine binary calls in the Longhorn manager with gRPC connections to the instance manager. Then the instance manager will be responsible for handling the requests between the management network and storage network.
+
+---
+**_NOTE:_** There are other possible approaches we have considered to segregating the networks:
+
+- Add Longhorn Manager to the storage network. The Manager needs to restart itself to get the secondary storage network IP.
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20220428-storage-network-through-grpc-proxy.md
```diff
@@ -0,0 +1,345 @@
...
+
+## Motivation
+
+### Goals
+
+- Have a new `Storage Network` setting.
+
+- Replace Manager engine binary calls with gRPC client to the instance manager.
+
+- Use the management network for Manager to Instance Manager communication.
```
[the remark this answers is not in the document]
c3y1huang: Updated.

enhancements/20220428-storage-network-through-grpc-proxy.md
```diff
@@ -0,0 +1,345 @@
...
+
+- Use the storage network for in-cluster data traffic.
```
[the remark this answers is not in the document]
c3y1huang: Updated.

enhancements/20220428-storage-network-through-grpc-proxy.md
```diff
@@ -0,0 +1,345 @@
...
+
+- Support backward compatibility of existing engine instance manager after upgrade.
```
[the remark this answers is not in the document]
c3y1huang: Updated

enhancements/20220428-storage-network-through-grpc-proxy.md
```diff
@@ -0,0 +1,345 @@
...
+- Setup and configure the Multus `NetworkAttachmentDefinition` CRs.
+
+- Monitor for `NetworkAttachmentDefintition` CRs. The user needs to ensure the traffic is reachable between pods and across different nodes. Without monitoring, Longhorn will not get notified of the update of the `NetworkAttachmentDefinition` CRs. Thus the user should create a new `NetworkAttachmentDefinition` CR and update the `storage-network` setting.
+
+- Out-cluster data traffic. For example, backing image upload and download.
+
+
+## Proposal
+
+### Proxy Manager request through Instance Manager gRPC proxy
```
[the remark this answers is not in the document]
c3y1huang: Updated.

enhancements/20220428-storage-network-through-grpc-proxy.md
```diff
@@ -0,0 +1,345 @@
...
+
+- Introduce a new gRPC server in Instance Manager.
+
+- Keep re-usable connections between Manager and Instance Managers.
+
+- Allow falling back to engine binary calls when gRPC connection to Instance Manager is not available.
```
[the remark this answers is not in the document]
c3y1huang: Updated.

enhancements/20220428-storage-network-through-grpc-proxy.md
c3y1huang: There will be one gRPC connection for each `controllerID`. I think we should keep the connections persistent to avoid the on/off intermediate state.

enhancements/20220428-storage-network-through-grpc-proxy.md
c3y1huang: Updated.

enhancements/20220428-storage-network-through-grpc-proxy.md
c3y1huang: Sounds good, updated.

enhancements/20220428-storage-network-through-grpc-proxy.md
c3y1huang: Reworded.

enhancements/20220428-storage-network-through-grpc-proxy.md
c3y1huang: We need to keep dual IP for the Manager to Backing Image Manager communication. Therefore, having the same storageIP in engine/replica status for consistency.

enhancements/20220428-storage-network-through-grpc-proxy.md
c3y1huang: https://github.com/longhorn/longhorn/pull/3277#discussion_r864709057

enhancements/20220428-storage-network-through-grpc-proxy.md
innobead: Ok, let's keep it as is first. It's actually nice to have.

enhancements/20220428-storage-network-through-grpc-proxy.md
```diff
@@ -0,0 +1,323 @@
...
+	    BackupGet(destURL string, credential map[string]string) (*Backup, error)
+	    BackupVolumeGet(destURL string, credential map[string]string) (volume *BackupVolume, err error)
+	    BackupNameList(destURL, volumeName string, credential map[string]string) (names []string, err error)
+	    BackupVolumeNameList(destURL string, credential map[string]string) (names []string, err error)
+	    BackupDelete(destURL string, credential map[string]string) (err error)
+	    BackupVolumeDelete(destURL, volumeName string, credential map[string]string) (err error)
+	    BackupConfigMetaGet(destURL string, credential map[string]string) (*ConfigMetadata, error)
+    }
+    ```
+  - The EngineClient interface is used for `Proxy`, existing `Engine` and existing `BackupTargetClient`. Update any inconsistent collection of methods.
```
innobead, line 193: This statement looks a little confusing, but I guess what you want to explain is:

`A new EngineClientProxy interface will be introduced for the Proxy, which includes proxy specific methods and implementation of the existing EnglineClient and BackupTargetClient interfaces. This will be adaptive when using EngineClient  for the proxy or non-proxy/fallback case.`

innobead (dismissed):
LGTM

enhancements/20220428-storage-network-through-grpc-proxy.md
c3y1huang: Exactly, updated.

enhancements/20220428-storage-network-through-grpc-proxy.md
```diff
@@ -0,0 +1,323 @@
...
+[Some old instance manager pods are still running after upgrade](https://longhorn.io/kb/troubleshooting-some-old-instance-manager-pods-are-still-running-after-upgrade/).
+Old engine instance managers do not have the gRPC proxy server for Manager to communicate.
+Hence, we need to support backward compatibility.
+
+Manager communication:
+- Bump instance manager API version.
+- Manager checks for incompatible version and fall back to requests through the engine binary.
+
+Engine live upgrade:
+- Do not support because of the instance manager API version bump.
```
shuo-wu, line 319: Will we disable the volume/engine live upgrade in v1.3.0?

shuo-wu (commented):
LGTM except for the live upgrade issue.

enhancements/20220428-storage-network-through-grpc-proxy.md
c3y1huang: Actually what I meant is the volume/engine live upgrade will be blocked by [CheckInstanceManagerCompatibilty](https://github.com/longhorn/longhorn-manager/blob/master/engineapi/instance_manager.go#L247). 
This means after the cluster upgrade, the volume needs to do a detach/attach first to restart the processes. Then the user can do the volume engine live upgrade.

c3y1huang: Reworded.

shuo-wu: For the old volumes, Longhorn should fall back to using the binary. And using binary means using the old version of IM client, Then can we do live upgrade in this case?

c3y1huang: We could probably work around the current IM compatibility check with some logic
```
if im.Status.APIVersion == InstanceManagerProxyMinAPIVersion {
}
```
But I feel if we enforce the volume detach/attach in this version we can ensure the engine binary is safe to deprecate in the next major release. What do you think?

shuo-wu: Personally, I prefer your idea. But we cannot determine this. 😂.
cc @innobead

c3y1huang: After talking to @innobead 
```
For better support perspective, we should do:
- keep live upgrade well, but we will claim the old way (engine-binary) being deprecated in 1.4

This will be a soft notice for users to know we will not enforce any change in 1.3, but it will happen in 1.4
```

c3y1huang: Updated.

innobead: - Should support live upgrade as usual
- For deprecate engine binary, we can do it in 1.4 instead, but users should be noticed this change in our doc/release note. 

Basically, for any enforced change, we should avoid doing it at the same time when it's introduced. WDYT

c3y1huang: Got it. Sounds fair to me.

innobead (approved):

PR #3323, opened 2021-11-25
patch longhorn backing image and share manager images

meldafrawi:
Signed-off-by: Mohamed Eldafrawi <mohamed.eldafrawi@suse.com>


meldafrawi:
Running regression core test
Ref: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/562/

innobead (approved):

Issue #3206, opened 2021-11-02
[BUG] Volume crash during migrating


shuo-wu:
**Describe the bug**

There are 4 cases related to the volume crash. Some of them will lead to migrating volume hanging there:
1. The old engine crashes only: The migration confirmation works but the migration rollback gets stuck (in state `Healthy`).
2. The new engine crashes only: Both the migration rollback and confirmation get stuck (in state `Healthy`).
3. All old replicas crash: The migration confirmation works but the migration rollback gets stuck (in state `Healthy`).
4. All new replicas crash: The migration rollback works but the migration confirmation gets stuck (in state `Healthy`).

If all nodes/instance manager pods get crashed, the migration  rollback and confirmation get stuck, which is similar to case 2.

**To Reproduce**
1. Deploy StorageClass https://raw.githubusercontent.com/longhorn/longhorn/master/examples/rwx/storageclass-migratable.yaml
2. Deploy a RWX PVC with the above StorageClass
3. Attach the volume to a node
4. Start migration. (Send an attach request with `hodeId` setting to another node for the volume)
5. Enter into the instance manager pod then crash the processes. e.g.,
```
$ kubectl -n longhorn-system exec -it instance-manager-r-537c53aa -- bash
root@instance-manager-r-537c53aa:/# longhorn-instance-manager process ls
{
	"pvc-99743237-a487-4a5d-8ea5-4cf5cf857b1a-r-50a0e42e": {
		"name": "pvc-99743237-a487-4a5d-8ea5-4cf5cf857b1a-r-50a0e42e",
		"binary": "/host/var/lib/longhorn/engine-binaries/longhornio-longhorn-engine-master-head/longhorn",
		"args": [
			"replica",
			"/host/var/lib/longhorn/replicas/pvc-99743237-a487-4a5d-8ea5-4cf5cf857b1a-73aee4b4",
			"--size",
			"1073741824",
			"--listen",
			"0.0.0.0:10015"
		],
		"portCount": 15,
		"portArgs": [
			"--listen,0.0.0.0:"
		],
		"processStatus": {
			"state": "running",
			"errorMsg": "",
			"portStart": 10015,
			"portEnd": 10029
		},
		"deleted": false
	},
	"pvc-99743237-a487-4a5d-8ea5-4cf5cf857b1a-r-6b887a5c": {
		"name": "pvc-99743237-a487-4a5d-8ea5-4cf5cf857b1a-r-6b887a5c",
[...]

PR #3332, opened 2021-11-26
Improve the way to validate the current engine

jenting:
Introduce a new field engine.Spec.Active to indicate if the engine is the current one
    
https://github.com/longhorn/longhorn/issues/3206


innobead (approved):

shuo-wu (approved):
LGTM. Thank you for helping update the Chart!

PR #3366, opened 2021-12-03
Update CRD with v1beta1 and v1beta2

derekbit:
Longhorn 2955

### DIFF
--- a/chart/templates/crds.yaml
+++ b/chart/templates/crds.yaml
@@ -98,6 +98,89 @@ spec:
             type: object
         type: object
     served: true
+    storage: false
+    subresources:
+      status: {}
+  - additionalPrinterColumns:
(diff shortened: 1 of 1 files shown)


chart/templates/crds.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.7.0
   creationTimestamp: null
-  labels: {{- include "longhorn.labels" . | nindent 4 }}
+  labels:
```
jenting, line 7: Please add this back `{{- include "longhorn.labels" . | nindent 4 }}`, similar below.

jenting (changes requested):

jenting (approved):

innobead (approved):

PR #3369, opened 2021-12-06
Configure Settings CR directly, removes longhorn-default-setting ConfigMap

jenting:
#### Proposal Change
- Install/Upgrade CRDs first (Make sure the CRDs be upgraded after helm upgrade).
- Install/Upgrade longhorn-manager (If the Setting CR does not exist, the longhorn-manager will create a new one with the default value inside the longhorn-manager [code](https://github.com/longhorn/longhorn-manager/blob/96a1fd9/datastore/longhorn.go#L47-L61)).
- Install/Upgrade customized setting by directly configuring the Setting CRs.

To detect the value is _nil_ or not, we use a solution `{{ kindIs "invalid" $value }}` to check it. Ref to https://github.com/Masterminds/sprig/issues/53#issuecomment-483414063.
If we check the value by `{{- if $value }}` only, then the `$value` with empty string would not be rendered.

#### Issue
Longhorn 2570


jenting:
This PR depends on #3366.

innobead (approved):
LGTM

Issue #3176, opened 2021-10-21
[FEATURE] Read side caching


yasker:
We can implement a read side caching in the engine, to increase the read performance and lower the latency if the data was cached.

innobead:
cc @keithalucas @derekbit

derekbit:
Rough design note
- a LRU cache
   - configurable cache size
   - key: offset, value: block

- IO path
  - read
      ```
       handleRead
        -> s.data.ReadAt
             -> cache.ReadAt(offset)
                 -> foreach block
                     -> hit, read data from cache
                     -> miss, read data from replica
                          -> cache.Update
                 -> merge and truncate block data to the request size
                 -> return data
      ```
  - write
      ```
       handleWrite
        -> s.data.WritedAt
             -> cache.Invalidate(offset, size)
             -> generic write path
      ```
- TBD
   - cache drop, need or not?

keithalucas:
> Rough design note
> 
> * a LRU cache

It seems like we want to have this cache in the longhorn-engine controller process not the replica, correct?   @yasker did some analysis of other storage solutions and found that they do caching.  It might be useful to look at several them to see if they use LRU caching or one of the other [caching approaches](https://en.wikipedia.org/wiki/Cache_replacement_policies).

We should investigate implementing the cache with lock free algorithm because the cache will be accessed by multiple goroutines constantly.  
>   
>   * configurable cache size

I think we will need to determine a default cache size eventually.  

>   * key: offset, value: block
> * IO path
>   
>   * read
>     ```
>      handleRead
>       -> s.data.ReadAt
>            -> foreach block
>                  -> cache.ReadAt(offset)
>                        -> hit, read data from cache
>                   -> miss, read data from replica (generic read path)
>                        -> cache.Update

The longhorn-engine controller doesn't have a fixed block size; it can handle `ReadAt` and `WriteAt`s with any offset and length.   The longhorn-engine controller currently requests the same offset and length from the longhorn-engine replica that it receives from tgtd.   One of the bottlenecks in the longhorn-engine is the communication between the longhorn-engine controller and the longhorn-engine replicas.  

[...]

PR #3448, opened 2021-12-21
Add LEP for read side cache

derekbit:
Implement a read side cache in the engine, to increase the read performance and lower the latency if the data was cached.

Longhorn-3176

### DIFF
--- a/enhancements/20211221-read-side-cache.md
+++ b/enhancements/20211221-read-side-cache.md
@@ -1,0 +1,242 @@
+# Add Read Side Cache
+
+## Summary
+This enhancement adds a read side cache in the engine to lower the read latenty if the data was cached and improve the read performance.
+
+## Related Issues
+- [[FEATURE] Read side caching #3176](https://github.com/longhorn/longhorn/issues/3176)
+
+## Motivation
+### Goals
+- The cache device is dedicated to read buffering, so there is no data loss concern if the cache device is lost.
+- User is able to create a volume with a cache device.
+- User is able to enable or disable the cache.
+- Cache percentage
+  - User is able to configure the global size of the cache device by setting the cache device to volume ratio, expressed as a percentage.
+  - User is able to override the ratio in StorageClass.
+- Cache block size
+  - User is able to configure the global block size of the cache device.
+  - User is able to override the block size in StorageClass.
+- Cache disk path
+  - User is able to configure the per-node cache disk path.
+  
+### Non-goals [optional]
+
+## Proposal
+### User Stories
+We would like to add a read side cache in the engine to reduce the read latency. In current read I/O path, the data from a replica to the application suffers from storage and network latencies, shown in Figure 1.
+
+```bash
+
+      +-------+
+      |  APP  |
+      +-------+
+          ▲  
+          | read data
++--------------------+
+|     filesystem     |
++--------------------+
+|   iSCSI initiator  |
++--------------------+
+|    iSCSI target    |
++--------------------+
+|     controller     |
++--------------------+
+|       replica      |
++--------------------+
+|        disk        |
++--------------------+ 
+
+       Figure 1
+```
+
+In the filesystem, the page cache caches frequently accessed data, but the limited memory capacity limits the size of the cache, and cache thrashing occurs when too many workloads compete for memory resources on a node. Those issues result in cache misses and extra data reads from the replicas. In addition, applications enabling direct I/O also suffer the long latencies to the disk on each read request.
+
+To mitigate the long latency of the read I/O path, a per-volume cache being dedicated to read buffering is introduced shown in Figure 2. A cache device creating from an preallocated file and `dm-cache` on the host storage is added on top of the iSCSI initiator. The size and the block size of the cache device is configurable according to workload requirements [1]. Caching frequently accessed data on the cache device improves the read performance and reduces I/O latency.
+
+```bash
+
+      +-------+
+      |  APP  |
+      +-------+
+          ▲  
+          | read data
++--------------------+
+|     filesystem     |
++--------------------+
+|    cache device    |
++--------------------+
+|   iSCSI initiator  |
++--------------------+
+|    iSCSI target    |
++--------------------+
+|     controller     |
++--------------------+
+|       replica      |
++--------------------+
+|        disk        |
++--------------------+ 
+
+       Figure 2
+```
+
+### Cache Management
+
+Cache allocation will be based on best effort. If there is no space, then the cache won't be allocated and the volume will function without cache.
+
+The cache is disabled by default. User can enable it by set `CacheEnabled` to `"true"` in the StorageClass.
+#### Cache Disk Path
+The cache data can be stored in `node.cacheDisks` or in a per-node dedicated cache disk [2]. The `node.cacheDisks` can be deleted if the scheduled storage size is `0`.
+
+#### Cache Size
+A global cahe size setting `SettingNameDefaultCachePercentage` in terms of cache device to volume ratio, expressed as a percentage, is introduced. The default value is `0` which disables the cache by default.
+
+In addition to the global setting, user is able to override the value in StorageClass. The cache size can be overridden by each volume.
+
+#### Cache Block Size
+A global cahe block size setting `SettingNameDefaultCacheBlockSize` is introduced. The default value is `128Ki` by default. User can modify the value for different applications.
+
+In addition to the global setting, user is able to override the value in StorageClass.
+
+### User Experience In Detail
+#### Configure Per-Node Cache Disk
+If the use would like to enable the read-side cache, the per-node cache disk needs to be specified.
+
+#### Create A Volume With Cache
+- Global Setting
+    - SettingNameDefaultCachePercentage:
+      - Type: int
+      - Range: 0-100
+      - Default: 0
+    - SettingNameDefaultCacheBlockSize
+      - Type: string
+      - Range: 4K - 128Ki
+      - Default: 32Ki
+- To enable the read side cache, create a storageClass and set `cacheEnabled` to be `"true"`. `cachePercentage` are `cacheBlockSize` are overridable.
+    - cacheEnabled: The read-side cache can be enabled or disabled by this flag. User can disable the read-side cache on the node who has a cache disk according to the application.
+    - cachePercentage: the capacity of the cache device in terms of the percentage of the volume size
+    - cacheBlockSize: the size of the cache device
+    ```
+    kind: StorageClass
+    apiVersion: storage.k8s.io/v1
+    metadata:
+    name: longhorn-crypto-global
+    provisioner: driver.longhorn.io
+    allowVolumeExpansion: true
+    parameters:
+      numberOfReplicas: "3"
+      staleReplicaTimeout: "2880" # 48 hours in minutes
+      fromBackup: ""
+      cacheEnabled: "true"
+      cachePercentage: "20"
+      cacheBlockSize: "64Ki"
+    ```
+
+- Create a pvc that references the created storage class.
+
+#### Failure
+If the cache disk is full or corrupted, Longhorn has the fallback to provide a volume in absence of the read-side cache.
+### API changes
+Add `cacheSize` string which is calculated from the `cachePercentage` and the volume size, `cacheBlockSize` string and `cachePath` string to the Volume struct utilized by the http client, this ends up with being stored in `Volume.Spec.cachePercentage`, `Volume.Spec.cacheBlockSize` and `Volume.Spec.cachePath`  of the volume CR.
+
+
+## Design
+### Implementation Overview
+### Volume with a Read Side Cache
+Host requires `dm-cache` kernel module as well as `dmsetup` and `losetup` userspace utilities.
+
+The logics are implemented in longhorn-manager CSI `NodeStageVolume`.
+
+1. Create a preallocated file under per-node cache disk.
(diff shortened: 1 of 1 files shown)


enhancements/20211221-read-side-cache.md
```diff
@@ -0,0 +1,186 @@
...
++--------------------+ 
+
+       Figure 2
+```
+
+### User Experience In Detail
+- To enable the read side cache, create a storage class and specify
+    - cacheSize: the capacity of the cache device
+    - cacheBlockSize: the size of the cache device
+    - cacheDir: the directory containing the data of the cache device
```
yasker, line 80: We can allow users to specific per volume cache size/block size (which overrides the global default values for those two options). But it's not a good experience to require users to choose the cacheDir. Users already specific certain disks/paths for Longhorn to store data. The cache directory needs to be managed by Longhorn, which includes allocation and clean-up properly.

One thing tricky about this is, the engine itself is stateless and it can be run on the nodes without disks for Longhorn. Sounds like we have to ask users to prepare the cache disks. Then we need to deal with the cache disk failure scenario differently from Longhorn disk failure. 

Can we use ramdisks? how big do we need the cache per volume to make it efficient?

yasker (changes requested):

enhancements/20211221-read-side-cache.md
derekbit: > We can allow users to specific per volume cache size/block size (which overrides the global default values for those two options). But it's not a good experience to require users to choose the cacheDir. Users already specific certain disks/paths for Longhorn to store data. The cache directory needs to be managed by Longhorn, which includes allocation and clean-up properly.
> 
> One thing tricky about this is, the engine itself is stateless and it can be run on the nodes without disks for Longhorn. Sounds like we have to ask users to prepare the cache disks. Then we need to deal with the cache disk failure scenario differently from Longhorn disk failure.

Agree. Specifying the `cacheDir` is not ideal.
The default replica path is `/var/lib/longhorn/replicas`, so how about put the cache file inside `/var/lib/longhorn/caches`? Therefore, we can manage the cache files similar to replica files.

> Can we use ramdisks? how big do we need the cache per volume to make it efficient? The space we require for cache likely needs to have a limit for all the volumes on that node.

I did a [benchmark](https://github.com/longhorn/longhorn/issues/3176#issuecomment-1004056334) comparing the 100% rand read/write performance using ramdisk/SSD as a cache device.
The randwrite results were different from previous tests (randread:randwrite = 90:10), which might have been affected by read operations. Based on the results, ramdisk or persistent disk both offer the same benefits.



As a cache device, I think using a persistent disk is preferred, because
1. Using ramdisks directly reduces available memories for applications.
2. Linux supports `rbd` (ramdisk block device) and `tmpfs`. `rbd` needs a pre-configuration and creates fixed-size and fixed number of ramdisks, so it's not flexible for our applications. Similar to using a persistent disk drive, the user has to mount the `tmpfs` before using it, and Longhorn has to be responsible for the allocation and free of the ramdisks.
3. The filesystem already has page cache, so the read side cache should not use a ramdisk which reduces the page cache size.

yasker: As I mentioned, since the engine is stateless, there is no guarantee that there will be a disk usable on the node that the engine is running on. We can add a directory to the data disk for the cache but that needs to be managed and probably taken into scheduling consideration. We cannot use `/var/lib/longhorn/caches` since it's likely the root partition. So the questions are:
1. What if the volume attached node doesn't have a disk?
2. What if there is not enough space for creating the cache file on the disk?
3. Engine does not associate with a disk. How to change the design of Longhorn to associate the engine process with a disk? Introducing a special disk (e.g. Data disk vs cache disk)?

I am a little curious as the ram cache doesn't yield better performance, but I agree that using a disk is likely better because memory is a more scarce resource.

Also, the most important thing of this implementation, is we cannot lose any data if we lose the whole volume attached node. All the data that are written into the disk (via the cache layer) must be persistent into the Longhorn data disk. We need to run more tests to ensure that.

derekbit: > As I mentioned, since the engine is stateless, there is no guarantee that there will be a disk usable on the node that the engine is running on. We can add a directory to the data disk for the cache but that needs to be managed and probably taken into scheduling consideration. We cannot use `/var/lib/longhorn/caches` since it's likely the root partition. So the questions are:
> 
> 1. What if the volume attached node doesn't have a disk?
> 2. What if there is not enough space for creating the cache file on the disk?

A way around the two problems is to provide an iSCSI blockdev without cache.

> 3. Engine does not associate with a disk. How to change the design of Longhorn to associate the engine process with a disk? Introducing a special disk (e.g. Data disk vs cache disk)?

Yes. I think we should introduce a special disk and need to design the management.

> 
> I am a little curious as the ram cache doesn't yield better performance, but I agree that using a disk is likely better because memory is a more scarce resource.

I think the ramdisk cannot yield better performance because the test file is 10G and cache block size is 32K, so for 4K fio test, cache miss and data promotion from SSD to ramdisk always happened. The promotion slowed down the performance. Using 32K cache block size and 32K fio test will see the benefit of the ramdisk.

> 
> Also, the most important thing of this implementation, is we cannot lose any data if we lose the whole volume attached node. All the data that are written into the disk (via the cache layer) must be persistent into the Longhorn data disk. We need to run more tests to ensure that.

Agree, it's the first priority.

derekbit: > As I mentioned, since the engine is stateless, there is no guarantee that there will be a disk usable on the node that the engine is running on.

This is a rough thought from my side.

If the node does not have a disk, the read cache is not allowed.

Two possible designs
1. Replicas and cache files are on the same disk.
   The user needs to disable disk scheduling if they do not want to store replicas on the disk, only cached files.

2. Replicas and cache files are on different disks.
    The design should be very similar to generic node.disk design.

I think the first design is much simpler and without lots of duplication implementations similar to node.disk.

@yasker Do you have any suggestion?

innobead (commented):
In general, LGTM. Few comments as below for your reference.

- Have a global setting for cache dir instead of each volume. (KISS first and see the further requests then to improve)
- Probably need to have a place to briefly describe cache lifecycle which should be together with volume cycle
- Cache eviction should be probably taken together when the node is unscheduled and all longhorn disks evicted 
- Cache GC ideally together with volume lifecycle, but in case some orphan caches happen, how to deal with that (at the moment, doc is enough, but this should be a general topic for other things GC like outdated replicas, maybe another ticket to follow up)

enhancements/20211221-read-side-cache.md
yasker: There can be multiple disks on a node. We can start with a simple way:
1. A cache disk is only a special property of the data disk. Need to introduce a new property here.
2. Only one disk per node will be designated as the cache disk. So if the user wants a dedicated cache disk, they can disable the scheduling on the disk.
3. Cache allocation will likely be a percentage of the allocated volume size. The option can be a global setting at first. `0` value will disable the cache. We can make it disabled by default for the first iteration.
4. The cache size can be overridden by each volume.
5. The cache percentage can be overridden by the storage class.
5. Cache allocation will be based on best effort. If there is no space (remember to take scheduled replicas into consideration), then the cache won't be allocated and the volume will function without cache. A condition will be introduced to prompt the user that the cache cannot be allocated due to the lack of space.

yasker: Also, once the LEP is updated, please make sure @joshimoo and @shuo-wu review it.

shuo-wu: > Only one disk per node will be designated as the cache disk. So if the user wants a dedicated cache disk, they can disable the scheduling on the disk.

Why can only one disk per node handle cache? I don't think we need to differentiate volume cache allocation from that of replicas or backing images. And it's better to have a separate field enabling/disabling cache for a disk.

yasker: Because then we don't need to add another scheduling method to choose the disk for the cache. 

We can add that ability later but we can keep it simple for the first iteration.

shuo-wu: Then which disk will Longhorn use for the cache if there are multiple disks on a node? I think we can skip the scheduling check for the cache but still have a separate field in DiskSpec that controls if the cache can be allocated to a disk.

derekbit: Only one disk per node can enable the `allowCacheStore` property, and the `allowCacheStore` should be enabled by user itself.
@shuo-wu Do you think it works?

derekbit: If we store the cache in a disk, we will encounter the management issues.

1. How to choose the disk for cache files?
2. If there are cache files for the live volumes but no replicas in the disk, the disk cannot be evicted before detaching the volume.
3. If user changes the disk for cache files to another, old and new volumes' cache files are on different disks. This will be confusing.

I think we can follow @innobead [comment 1](https://github.com/longhorn/longhorn/pull/3448#pullrequestreview-843361039) in the first iteration. There is a global cache dir setting, e.g. `/var/lib/longhorn-cache`, and a dedicated cache disk can be mounted to folder, then all volumes' cache files will be stored in this folder to separate the replicas and cache files. The introduction of cache files will not change the node.disks management.

derekbit: > Also, the most important thing of this implementation, is we cannot lose any data if we lose the whole volume attached node. All the data that are written into the disk (via the cache layer) must be persistent into the Longhorn data disk. We need to run more tests to ensure that.

The stress test steps are
- Write data using direct io to volume via the cache layer
- Stop the volume
- Attach the volume back
- Verify the file checksum

yasker: > If we store the cache in a disk, we will encounter the management issues.
> 
> 
> 
> 1. How to choose the disk for cache files?
> 
> 2. If there are cache files for the live volumes but no replicas in the disk, the disk cannot be evicted before detaching the volume.
> 
> 3. If user changes the disk for cache files to another, old and new volumes' cache files are on different disks. This will be confusing.
> 
> 
> 
> I think we can follow @innobead [comment 1](https://github.com/longhorn/longhorn/pull/3448#pullrequestreview-843361039) in the first iteration. Don't mix up the replica and client-side cache storage. We can have a global cache dir setting, e.g. `/var/lib/longhorn-cache` where a dedicated cache disk can be mounted to, and then all volumes' cache files are stored in this folder. Cache files and replicas are not mixed up. The introduction of cache files will not change the node.disks management. 

I agree that we should keep it simple now. The management concerns are valid. Though I really worried about mistakenly put the cache directory in the root disk, which will result in disaster. We have unmount protection for the data disk. If the disk was unmounted by mistake, the UUID detection will stop it from working. We will need that for this directory as well. 

Also, it will apply to all the nodes, which means the user won't have the ability to specify different path for each node. Well, you can do separate bind mount but the experience is pretty rough. 

Looks like in the future, we still need to introduce cache disk vs data disk to solve the eviction problem. 

I assume we would want the option to be enabled by default. To counter the risk of eating up the root disk, we should put some limitation parameter in place so we can stop adding more cache files before it's too late. (This become more and more similar to the data disk parameters now...)

Also, before starting the implementation, we need to verify there won't be data loss if we lost the cache.

yasker: > > Also, the most important thing of this implementation, is we cannot lose any data if we lose the whole volume attached node. All the data that are written into the disk (via the cache layer) must be persistent into the Longhorn data disk. We need to run more tests to ensure that.
> 
> 
> 
> The stress test steps are
> 
> - Write data using direct io to volume via the cache layer
> 
> - Stop the volume
> 
> - Attach the volume back
> 
> - Verify the file checksum

Hmm, this won't be enough since whatever in the cache will be flushed into the backend once you detached the volume.

enhancements/20211221-read-side-cache.md
```diff
@@ -0,0 +1,243 @@
...
+- User is able to create a volume with a cache device.
+- Cache size
+  - User is able to configure the global size of the cache device by setting the cache device to volume ratio, expressed as a percentage.
+  - User is able to override the ratio in StorageClass.
+- Cache block size
+  - User is able to configure the global block size of the cache device.
+  - User is able to override the block size in torageClass.
+- Cache directory path
+  - User is able to configure the global cache directory path.
+  - User is able to override the cache directory path in StorageClass.
```
yasker, line 21: No, the user won't be able to override that since cache directory is not per volume, it's per node. User won't know which node the volume will be scheduled to.

enhancements/20211221-read-side-cache.md
```diff
@@ -0,0 +1,243 @@
...
+
+In addition to the global setting, user is able to override the value in StorageClass. The cache size can be overridden by each volume.
+
+#### Cache Block Size
+A global cahe block size setting `SettingNameDefaultCacheBlockSize` is introduced. The default value is `32Ki` by default. User can modify the value for different applications.
+
+In addition to the global setting, user is able to override the value in StorageClass.
+
+### User Experience In Detail
+- Global Setting
```
yasker, line 106: Need to include the setup experience and failure scenarios as well.

enhancements/20211221-read-side-cache.md
derekbit: Right. Precisely, `Stop` includes `detach` and `crash`.

shuo-wu: > We can have a global cache dir setting, e.g. /var/lib/longhorn-cache where a dedicated cache disk can be mounted to, and then all volumes' cache files are stored in this folder.

Does this mean that there is no way to enable the cache for a volume if the node only contains one default data disk?

innobead:
Closing first, but will continue if needed.

PR #3485, opened 2022-01-03
feat(longhorn-UI): Allow specifying loadBalancer IP/SourceRanges

docent-net:
This change allows users to specify loadBalancerIP/SourceRanges while for loadBalancer service type in Longhorn-UI service.


innobead (approved):
LGTM

Issue #1473, opened 2020-06-09
[FEATURE] Implement Alerts and Notifications from Rancher for Longhorn


Sheldonwl:
**Is your feature request related to a problem? Please describe.**
This feature request relates to: https://github.com/longhorn/longhorn/issues/928https://github.com/longhorn/longhorn/issues/928
It would be nice to setup Alerts and Notifications in Rancher to get notified about certain things in Longhorn. 

**Describe the solution you'd like**
To be able to create Alerts from within Rancher and attached Notifiers to those Alerts. This would require an integration with Rancher. 

**Describe alternatives you've considered**
Being able to expose and export logs from Longhorn to a central logging tool. This would also be able to trigger alerts, if that tool support such a feature.

yasker:
@Sheldonwl  I think the user can already do this in v1.1.0, as a part of https://github.com/longhorn/longhorn/issues/1180 , e.g. https://longhorn.io/docs/1.1.0/monitoring/integrating-with-rancher-monitoring/ Can you confirm it's done?

amnk:
@yasker what do you think about having ServiceMonitor object support in Longhorn's chart? I can create a PR to enable that if you think this is a good idea.

yasker:
@amnk Unfortunately I am not super familiar with this part. @PhanLe1010 what do you think?

PhanLe1010:
@amnk 

Do you mean deploying this service monitoring CR with Longhorn by default? https://longhorn.io/docs/1.1.2/monitoring/integrating-with-rancher-monitoring/#add-longhorn-metrics-to-the-rancher-monitoring-system

It would mean that Longhorn also must be shipped with `ServiceMonitor` CRD depending on where user cluster already has it or not. IMO, it is best to just let user deploy the service monitor CR by themself if users want to enable monitoring. Similar to what we did with CSI snapshot.

amnk:
[...]

PR #3580, opened 2022-01-30
Add ServiceMonitor and PrometheusRule

reitermarkus:
Allow easily enabling monitoring and alerts for volumes.

### DIFF
--- a/chart/templates/prometheusrules.yaml
+++ b/chart/templates/prometheusrules.yaml
@@ -1,0 +1,91 @@
+{{- if and .Values.metrics.enabled .Values.metrics.prometheusRule.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels: {{- include "longhorn.labels" . | nindent 4 }}
+  name: longhorn
+  namespace: {{ include "release_namespace" . }}
+spec:
+  groups:
+    - name: longhorn
+      rules:
+        - alert: LonghornVolumeActualSpaceUsedWarning
+          annotations:
+            description: The actual space used by Longhorn volume {{ "{{" }} $labels.volume }} on {{ "{{" }} $labels.node }} is at {{ "{{" }} $value }}% capacity for
+              more than 5 minutes.
+            summary: The actual used space of Longhorn volume is over 90% of the capacity.
+          expr: (longhorn_volume_actual_size_bytes / longhorn_volume_capacity_bytes) * 100 > 90
+          for: 5m
+          labels:
+            issue: The actual used space of Longhorn volume {{ "{{" }} $labels.volume }} on {{ "{{" }} $labels.node }} is high.
+            severity: warning
+        - alert: LonghornVolumeStatusCritical
+          annotations:
+            description: Longhorn volume {{ "{{" }} $labels.volume }} on {{ "{{" }} $labels.node }} is Fault for
+              more than 2 minutes.
+            summary: Longhorn volume {{ "{{" }} $labels.volume }} is Fault
+          expr: longhorn_volume_robustness == 3
+          for: 5m
+          labels:
+            issue: Longhorn volume {{ "{{" }} $labels.volume }} is Fault.
+            severity: critical
+        - alert: LonghornVolumeStatusWarning
+          annotations:
+            description: Longhorn volume {{ "{{" }} $labels.volume }} on {{ "{{" }} $labels.node }} is Degraded for
+              more than 5 minutes.
+            summary: Longhorn volume {{ "{{" }} $labels.volume }} is Degraded
+          expr: longhorn_volume_robustness == 2
+          for: 5m
+          labels:
+            issue: Longhorn volume {{ "{{" }} $labels.volume }} is Degraded.
+            severity: warning
+        - alert: LonghornNodeStorageWarning
+          annotations:
+            description: The used storage of node {{ "{{" }} $labels.node }} is at {{ "{{" }} $value }}% capacity for
+              more than 5 minutes.
+            summary:  The used storage of node is over 70% of the capacity.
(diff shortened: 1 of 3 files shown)


reitermarkus:
I took the examples from https://longhorn.io/docs/1.2.3/monitoring/alert-rules-example/, i.e. https://github.com/longhorn/website/blob/450c526feaffce56ed2f8eafcc0ee76736bbe7f7/content/docs/1.2.3/monitoring/alert-rules-example.md.

It seems to me that the `LonghornVolumeActualSpaceUsedWarning` is pretty useless since almost all of my volumes have an actual size that is > 100% of the normal size. Am I right in this assumption, or should I be increasing all my volumes sizes to fit the actual size?

reitermarkus:
@jenting, any comment on the usefulness of the `LonghornVolumeActualSpaceUsedWarning`?

innobead:
> It seems to me that the `LonghornVolumeActualSpaceUsedWarning` is pretty useless since almost all of my volumes have an actual size that is > 100% of the normal size. Am I right in this assumption, or should I be increasing all my volumes sizes to fit the actual size?

@shuo-wu it seems related to a recent topic people are asking about `why actual size large than the volume size`. If that's the case, probably we need to revisit how to make this alert much more meaningful? 

cc @PhanLe1010

shuo-wu:
Yes. We need a separate doc that explains why and when the actual size can be greater than the spec size.

innobead (approved):
LGTM

reitermarkus:
> why and when the actual size can be greater than the spec size

I think I understand why it is larger: The actual size includes all snapshots.

The warning will be shown when the actual size is bigger than the allocated size, so my question was: Is a warning for this condition meaningful? I think this will happen with pretty much any volume once you have enough snapshot, won't it?

If the warning is meaningful, what is the correct path forward to fix it?

chart/values.yaml
```diff
@@ -176,6 +176,13 @@ resources: {}
   #  memory: 128Mi
   #
 
+metrics:
+  ## Enable the Prometheus ServiceMonitor.
+  enabled: false
```
PhanLe1010, line 181: The metrics are always exposed by default. Can we make it clear that this field enables the Prometheus service monitor?
```suggestion
  prometheusServiceMonitor:
    enabled: false
```

PhanLe1010 (changes requested):
> The warning will be shown when the actual size is bigger than the allocated size, so my question was: Is a warning for this condition meaningful? I think this will happen with pretty much any volume once you have enough snapshots, won't it?

I think this alert is meaning full to detect the case in which a volume has too many snapshots that the actual size is much bigger than the spec size (i.g. by 5 times). Users can change the threshold value to 2x/3x/4x/.. depending on their use case.

PhanLe1010:
Linking this PR to the corresponding issue https://github.com/longhorn/longhorn/issues/1473

chart/templates/prometheusrules.yaml
```diff
@@ -0,0 +1,91 @@
...
+spec:
+  groups:
+    - name: longhorn
+      rules:
+        - alert: LonghornVolumeActualSpaceUsedWarning
+          annotations:
+            description: The actual space used by Longhorn volume {{ "{{" }} $labels.volume }} on {{ "{{" }} $labels.node }} is at {{ "{{" }} $value }}% capacity for
+              more than 5 minutes.
+            summary: The actual used space of Longhorn volume is over 90% of the capacity.
+          expr: (longhorn_volume_actual_size_bytes / longhorn_volume_capacity_bytes) * 100 > 90
```
innobead, line 17: Just a revisit. I think it's also good to have the percentage here configurable w/ a default value, so users are able to adjust based on their need. 

The same suggestion is applying to other % rule expressions.

innobead (changes requested):

shuo-wu:
The doc/explanation related to the volume actual size is tracked in https://github.com/longhorn/longhorn/issues/3627

innobead:
ping @reitermarkus a few comments ☝️ waiting to resolve

chart/values.yaml
```diff
@@ -176,6 +176,19 @@ resources: {}
...
+  prometheusServiceMonitor:
+    enabled: false
+
+  ## Enable a default set of Prometheus alerting rules.
+  prometheusRule:
+    enabled: false
+
+    rules:
+      volumeActualSpaceUsedWarning:
+        threshold: 90 # warn when above this percentage
```
jenting, line 190: Any reason _why only_ make this threshold configurable?

reitermarkus: I didn't want to make too many more changes here since I migrated my cluster to Talos which doesn't support Longhorn at the moment so I have no way to test any more changes currently.

chart/values.yaml
```diff
@@ -176,6 +176,19 @@ resources: {}
   #  memory: 128Mi
   #
 
+metrics:
+  ## Enable the Prometheus ServiceMonitor.
+  prometheusServiceMonitor:
+    enabled: false
+
+  ## Enable a default set of Prometheus alerting rules.
```
jenting, line 184: Providing a default Prometheus service monitor is good.

I'm thinking that should we provide the default Prometheus rules in the helm chart? Since different users _probably_ would like to change different thresholds.

innobead: I would suggest exposing all adjustable thresholds w/ default values.

wirwolf:
Hello. I copy these rules to my helm chart
https://github.com/SomeBlackMagic/helm-charts/tree/master/charts/longhorn-monitoring
In this chart, i can disable some rules or change the severity level.

HujinoKun:
any ETAM?

innobead:
This contribution seems pending for a long while w/o an update. Let's follow up with https://github.com/longhorn/longhorn/issues/6585 instead.

@reitermarkus feel free to reopen and continue this contribution if you want. Thanks.

Issue #2955, opened 2021-08-31
[FEATURE] Match type of status.conditions of nodes.longhorn.io to other k8s objects


HubbeKing:
**Is your feature request related to a problem? Please describe.**
I currently have an ansible-playbook that I use in order to automate maintenance tasks.
As part of that playbook, I have a step that waits for longhorn to be ready before continuing.
However, because the nodes.longhorn.io object contains the node conditions under status.conditions as a map of condition name to condition information, rather than as a list of conditions, I cannot use `kubectl wait --for=condition=Ready=True` like I can for standard kubernetes objects.

If one tries to do so, i.e. `kubectl -n longhorn-system wait --for=condition=Ready nodes.longhorn.io atlantia`, kubectl returns the following error message:
`error: .status.conditions accessor error: map[MountPropagation:map[lastProbeTime: lastTransitionTime:2021-08-13T13:32:04Z message: reason: status:True type:MountPropagation] Ready:map[lastProbeTime: lastTransitionTime:2021-08-21T18:59:02Z message:Node atlantia is ready reason: status:True type:Ready] Schedulable:map[lastProbeTime: lastTransitionTime:2021-08-21T18:58:20Z message: reason: status:True type:Schedulable]] is of the type map[string]interface {}, expected []interface{}`

This means that waiting for a longhorn node to be ready requires a manual check and parse
(Currently I'm doing `kubectl -n longhorn-system get nodes.longhorn.io {{ host }} -o json|jq .status.conditions.Ready.status` and grepping for `True`)

**Describe the solution you'd like**
Ideally, I'd prefer for the nodes.longhorn.io object to support the same kubectl operations that the standard kubernetes nodes object does. However, as this would require drastically altering the CRD for the object, it may be unwise to do without proper warning ahead of time - I imagine more people than me have written scripts that parse through the nodes.longhorn.io object looking for things.

**Describe alternatives you've considered**
An alternative would be to look into extending kubectl with a plugin that has proper support for the Longhorn CRD objects, so one could more easily automate operations on them without having to restort to jq shenanigans.

**Additional context**
Standard kubernetes node status.conditions: (this is essentially how every native kubernetes object represents conditions - see also pods, jobs, etc)
```
"conditions": [
    {
        "lastHeartbeatTime": "2021-08-31T07:20:53Z",
        "lastTransitionTime": "2021-08-21T18:58:18Z",
        "message": "kubelet is posting ready status. AppArmor enabled",
[...]

PR #3586, opened 2022-02-01
Initial script for review

olegvorobiov:
My goal for this script is to be executed within the pipeline, while deploying Longhorn. 
There is a missing part, where a test load should be deployed - I will work on that soon.
Also, I realized that the listing nodes, the way I did - lists Kubernetes nodes, I will rework it with API calls, once I figure it out

Signed-off-by: Oleg Vorobiov <olegvorobyov90@gmail.com>

### DIFF
--- a/scripts/check_lh.sh
+++ b/scripts/check_lh.sh
@@ -119,7 +119,7 @@ do
         # Iterate through Longhorn nodes and determine if each node has Kubelet in Ready status
         for node in $(echo $LONGHORN_NODE_LIST)
         do
-            LONGHORN_NODE_STATUS=$(kubectl get nodes.longhorn.io/node1 -n longhorn-system -o json | jq -r '.status.conditions.Ready.status')
+            LONGHORN_NODE_STATUS=$(kubectl get nodes.longhorn.io/${node} -n longhorn-system -o json | jq -r '.status.conditions.Ready.status')
             add_desired "1"
             if [ -z "$LONGHORN_NODE_STATUS" ]
             then
--- a/scripts/check_lh.sh
+++ b/scripts/check_lh.sh
@@ -1,0 +1,174 @@
+#!/usr/bin/env sh
+# Run this script with:
+# env HOST=https://rancher-lab.home PASS=admin USER=admin CLUSTER=c-zlwr8 sh -x ./check_lh.sh
+# $CLUSTER - cluster id of a cluster, where you want to deploy Longhorn
+# List of required tools: kubectl, curl, jq
+CURL=${CURL:-curl}
+JQ=${JQ:-jq}
+
+CURL=$(command -v ${CURL})
+JQ=$(command -v ${JQ})
+# Helper functions to reduce the amount of args we pass in on each curl/jq call,
+# note we are calling whatever command -v curl/jq/etc.. found directly to avoid
+# recursion
+curl() {
+  ${CURL} --insecure --silent -H 'content-type: application/json' "$@"
+}
+
+jq() {
+  ${JQ} -r "$@"
+}
+
+# Login to rancher so we can have our bearer token to do work, the user needs to be a cluster admin
+LOGINJSON=$(curl ${HOST}/v3-public/localProviders/local?action=login --data-binary "$(cat <<FIN
+{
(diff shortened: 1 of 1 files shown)


scripts/check_lh.sh
```diff
@@ -0,0 +1,174 @@
...
+        add_desired "$DESIRED_LH_MANAGER_DS_NUMBER"
+        add_available "$READY_LH_MANAGER_DS_NUMBER"
+    else
+        echo "Longhorn Manager deamonsets are not fully deployed yet"
+    fi
+
+
+    ### CRDs
+    # Generate a list of nodes that Lonhorn is installed on
+    NODE_LIST=$(kubectl get nodes -n longhorn-system -o json | jq -r '.items[].metadata.name')
```
joshimoo, line 88: In regard to your comment you can get the longhorn nodes via `kubectl get nodes.longhorn.io -n longhorn-system`
In addition evaluating the kubernetes nodes might be nice as well as you are doing.

scripts/check_lh.sh
```diff
@@ -0,0 +1,174 @@
...
+
+    # Iterate through Longhorn nodes and determine if each node has Kubelet in Ready status
+    for node in $(echo $NODE_LIST)
+    do
+        NODE_STATUS=$(kubectl get nodes $node -n longhorn-system -o json | jq -r '.status.conditions[] | select(.reason == "KubeletReady") | .status')
```
joshimoo, line 93: Knowledge sharing: 
For longhorn node schema we are updating the conditions fields to be in line with the kubernetes node schema. As well as making some addiitonal CRD schema changes in v1.3
REF: https://github.com/longhorn/longhorn/issues/2955#issuecomment-968505098

joshimoo (commented):
LGTM, there is more than can be done, will add a list of nice to haves as a separate comment.
But since scripts don't need to be backwards compatible they can be built one change at a time, instead of needing to be complete and perfect right from the get go.

Adding @meldafrawi to the review since his bash fu is strong :)

scripts/check_lh.sh
oleg-vorobiov-suse: Kept Kubernetes nodes checks, added extra condition and added nodes.longhorn.io checks

scripts/check_lh.sh
oleg-vorobiov-suse: Understood, I will adjust as soon as it gets released

scripts/check_lh.sh
```diff
@@ -0,0 +1,213 @@
...
+  "username":"${USER}",
+  "password":"${PASS}"
+}
+FIN
+)")
+TOKEN=$(echo ${LOGINJSON} | jq -r '.token')
+
+# Generate Kubeconfig and forward it into a file
+curl --insecure --silent -H 'content-type: application/json' -X POST "${HOST}/v3/clusters/${CLUSTER}?action=generateKubeconfig" -H "Authorization: Bearer ${TOKEN}" -H 'Accept: application/json' | jq -r '.config' > cluster_config.yaml
+export KUBECONFIG=./cluster_config.yaml
```
meldafrawi, line 33: Since Longhorn is not always installed with Rancher managed clusters, you can add a check in the beginning of the script if the `KUBECONFIG` env variable is set, so there will be two ways to invoke the script


1. `env HOST=https://rancher-lab.home PASS=admin USER=admin CLUSTER=c-zlwr8 sh -x ./check_lh.sh`
2. `env KUBECONFIG="<PATH_TO_KUBECONFIG>"  ./check_lh.sh`

meldafrawi:
LGTM in general, left one comment.
Very good work @olegvorobiov, Thank you very much for your effort.

innobead:
@olegvorobiov Please make this ready for review instead of draft if this is still valid. Thanks.

scripts/check_lh.sh
LarsBingBong: I surely agree. This would be great. We're not having Longhorn on Rancher.

innobead:
This PR is pending for update. Let's follow up with https://github.com/longhorn/longhorn/issues/6586 and consider a more compete solution for healthy check purpose.

PR #3606, opened 2022-02-10
chart: add longhorn admission and conversion webhook servers

derekbit:


jenting (dismissed):

innobead:
@derekbit we should update chart files as well, right?

derekbit:
> @derekbit we should update chart files as well, right?

Oh! Yep, we need add a manifest for the webhooks.

jenting (approved):

PR #3662, opened 2022-02-24
adding cache-volume to longhorn-ui

mattmattox:
Signed-off-by: Matthew Mattox <Matt.Mattox@suse.com>

### DIFF
--- a/chart/templates/deployment-ui.yaml
+++ b/chart/templates/deployment-ui.yaml
@@ -20,13 +20,27 @@ spec:
         image: {{ template "registry_url" . }}{{ .Values.image.longhorn.ui.repository }}:{{ .Values.image.longhorn.ui.tag }}
         imagePullPolicy: IfNotPresent
         securityContext:
-          runAsUser: 0
+          runAsUser: 1000
+        volumeMounts:
+        - mountPath: /var/cache/nginx/
+          name: cache-volume
+        - mountPath: /var/config/nginx/
+          name: cache-volume
+        - mountPath: /var/run/
+          name: cache-volume
         ports:
         - containerPort: 8000
           name: http
         env:
           - name: LONGHORN_MANAGER_IP
             value: "http://longhorn-backend:9500"
+      securityContext:
+        fsGroup: 2000
+        runAsGroup: 3000
+        runAsUser: 1000
+      volumes:
+      - emptyDir: {}
+        name: cache-volume        
       {{- if .Values.privateRegistry.registrySecret }}
       imagePullSecrets:
       - name: {{ .Values.privateRegistry.registrySecret }}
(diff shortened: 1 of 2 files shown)


chart/templates/deployment-ui.yaml
```diff
@@ -20,13 +20,27 @@ spec:
         image: {{ template "registry_url" . }}{{ .Values.image.longhorn.ui.repository }}:{{ .Values.image.longhorn.ui.tag }}
         imagePullPolicy: IfNotPresent
         securityContext:
-          runAsUser: 0
+          runAsUser: 1000
```
innobead, line 23: Probably we can add n use a non root user in the longhorn ui image instead.

chart/templates/deployment-ui.yaml
```diff
@@ -20,13 +20,27 @@ spec:
...
+          name: cache-volume
+        - mountPath: /var/run/
+          name: cache-volume
         ports:
         - containerPort: 8000
           name: http
         env:
           - name: LONGHORN_MANAGER_IP
             value: "http://longhorn-backend:9500"
+      securityContext:
```
innobead, line 37: If having a non root user, probably we don't need this security context.

chart/templates/deployment-ui.yaml
```diff
@@ -20,13 +20,27 @@ spec:
         image: {{ template "registry_url" . }}{{ .Values.image.longhorn.ui.repository }}:{{ .Values.image.longhorn.ui.tag }}
         imagePullPolicy: IfNotPresent
         securityContext:
-          runAsUser: 0
+          runAsUser: 1000
+        volumeMounts:
+        - mountPath: /var/cache/nginx/
+          name: cache-volume
```
innobead, line 26: 3 duplicated names. If this is optional, suggest removing.

innobead (changes requested):

chart/templates/deployment-ui.yaml
mattmattox: No longer need because of https://github.com/longhorn/longhorn-ui/pull/495/commits/ced67e0649bcbc16e365f9ddf33e8360e2174331

chart/templates/deployment-ui.yaml
mattmattox: No longer need because of https://github.com/longhorn/longhorn-ui/pull/495/commits/ced67e0649bcbc16e365f9ddf33e8360e2174331

chart/templates/deployment-ui.yaml
mattmattox: Fixed, I broke it up into three volumes
- nginx-cache - tmpfs for NGINX
- nginx-config - config file location for nginx, Note this be RW because CMD writes to /var/config/nginx/nginx.conf during its start-up process
- var-run -  tmpfs for NGINX pid file and socket

PR #3681, opened 2022-03-01
introducing storage class locality

janesser:
Adopt feature from https://github.com/longhorn/longhorn/issues/1045 for helm charts.

Kind thanks to @jenting leading me here.

### DIFF
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -516,6 +516,15 @@ WARNING:
   min: 1
   max: 10
   default: 3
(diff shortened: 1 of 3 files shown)


innobead (changes requested):
Thanks for the contribution. 

Please also update questions.yml as other persistence settings.

janesser:
> Thanks for the contribution.
> 
> Please also update questions.yml as other persistence settings.

Let me know if that does the job. Cheers.

innobead (approved):

Issue #3636, opened 2022-02-17
[BUG] High CPU usage by instance manager


HonakerM:
## Describe the bug
This is a continuation of [this bug report](https://github.com/longhorn/longhorn/issues/1533#issuecomment-1038601469). I'm seeing 100% cpu usage by a `tgtd` process. 

## To Reproduce

1. Install longhorn via [helm](https://longhorn.io/docs/1.2.3/deploy/install/install-with-helm/)
2. Provision volumes with the `longhorn` storage class
3. Volumes will be successfully bound and created; however, a pod tried to mount the volume it will fail 
4. The node's instance manager will then expand to use 100% of the CPU

## Expected behavior
 Have a low CPU usage from pod instance-manager-e-xxxx (lower than consuming 4 CPU)

## Log or Support bundle

Here is a list of all of my longhorn pods and their resource usage (note instance-manager-e-167eff13 ):
```
NAME                                        CPU(cores)   MEMORY(bytes)
csi-attacher-6454556647-6xdl4               0m           15Mi
csi-attacher-6454556647-z52zh               0m           9Mi
csi-provisioner-869bdc4b79-l6t6w            0m           9Mi
csi-provisioner-869bdc4b79-nwrnm            0m           10Mi
csi-provisioner-869bdc4b79-znn4g            2m           14Mi
csi-resizer-6d8cf5f99f-gsx4p                0m           8Mi
csi-resizer-6d8cf5f99f-qm5h7                0m           15Mi
csi-resizer-6d8cf5f99f-zngh7                0m           12Mi
csi-snapshotter-588457fcdf-6pnbx            2m           11Mi
csi-snapshotter-588457fcdf-kd9mt            0m           13Mi
csi-snapshotter-588457fcdf-nr5zf            0m           9Mi
engine-image-ei-fa2dfbf0-dh46h              5m           9Mi
engine-image-ei-fa2dfbf0-f2c54              4m           12Mi
engine-image-ei-fa2dfbf0-w5db5              1m           6Mi
instance-manager-e-167eff13                 3433m        71Mi
instance-manager-e-50d209a9                 3m           17Mi
instance-manager-e-54542bca                 9m           22Mi
instance-manager-r-26157459                 3m           44Mi
instance-manager-r-b8b0cead                 89m          37Mi
instance-manager-r-e0818bc8                 32m          58Mi
longhorn-csi-plugin-cvm9s                   0m           15Mi
[...]

PR #3696, opened 2022-03-03
Update instance manager

keithalucas:
Use an instance manager version with the tgt 100% CPU usage fix.

https://github.com/longhorn/longhorn/issues/3636

Signed-off-by: Keith Lucas <keith.lucas@suse.com>

### DIFF
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -53,7 +53,7 @@ questions:
     label: Longhorn Instance Manager Image Repository
     group: "Longhorn Images Settings"
   - variable: image.longhorn.instanceManager.tag
-    default: v1_20211210
+    default: v1_20220303
     description: "Specify Longhorn Instance Manager Image Tag"
     type: string
     label: Longhorn Instance Manager Image Tag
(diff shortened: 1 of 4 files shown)


innobead (changes requested):
Also need to update chart files: questions.yml and values.yml

ref: https://github.com/longhorn/longhorn/commit/3301f9b9155ae4ee706efdcf625cbe4f01cb5554

P.S. do the same update for other backport PRs.

keithalucas:
> Also need to update chart files: questions.yml and values.yml
> 
> ref: [3301f9b](https://github.com/longhorn/longhorn/commit/3301f9b9155ae4ee706efdcf625cbe4f01cb5554)
> 
> P.S. do the same update for other backport PRs.

I updated the helm chart here and in the other PRs.

chart/questions.yaml
```diff
@@ -65,7 +65,7 @@ questions:
     label: Longhorn Share Manager Image Repository
     group: "Longhorn Images Settings"
   - variable: image.longhorn.shareManager.tag
-    default: v1_20211020
+    default: v1_20220303
```
innobead, line 68: Should instance manager not share manager

innobead (changes requested):

chart/questions.yaml
keithalucas: Fixed

innobead (approved):

Issue #3701, opened 2022-03-05
[BUG] Longhorn webhooks fail to deploy on hardened cluster


khushboo-rancher:
## Describe the bug

Longhorn admission webhook and Longhorn conversion webhook are unable to come up healthy on an RKE2 cluster with the cis-1.5 profile.

## To Reproduce

1. Create RKE2 cluster using the step listed at https://docs.rke2.io/security/hardening_guide/. eg - RKE2 version - v1.23.4+rke2r1 
2. Deploy Longhorn master-head in the cluster.
3. Longhorn deployment fails.

```
CreateContainerConfigError: container has runAsNonRoot and image will run as root (pod: "longhorn-admission-webhook-bcb579-xhffb_longhorn-system(475c3569-7dae-4427-8dca-d5eedde09c71)", container: wait-longhorn-conversion-webhook)

CreateContainerConfigError: container has runAsNonRoot and image will run as root (pod: "longhorn-conversion-webhook-55dd445b5-z57bg_longhorn-system(bae03ad5-b27a-4159-ab89-3770ce5cd7c4)", container: longhorn-conversion-webhook)
```

## Expected behavior

The webhooks should be deployed properly as other Longhorn workloads are getting deployed.

## Environment

 - Longhorn version: Longhorn-master `03/04/2022`
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl): kubectl
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: RKE2 v1.23.4+rke2r1 with cis-1.5 profile

kaxing:
Validation - **PASSED**

Follow the hardening guide and tested with latest manifests longhorn-manager master branch. Deploy without hitting any problem.

- Cluster spec: 1 control/etc node and 3 worker nodes
  - Node distro: CentOS Stream 8 4.18.0-305.3.1.el8.x86_64
  - K8s distro: v1.22.8+rke2r1

```
$ cat /etc/rancher/rke2/config.yaml
selinux: true
protect-kernel-defaults: true
profile: cis-1.6
```
 
Additionally, quick check with rancher-cis-benchmark:2.0.3 using rke2-cis-1.6-profile-hardened profile only had 1 failed item:
<img src=https://user-images.githubusercontent.com/602540/163323884-17c26245-bf9d-44ab-a58c-83ca8fb08a3e.png height=369px />

---
##### Note for testing and setup the security enforcing cluster:
[...]

PR #3702, opened 2022-03-05
chart: set securityContext of webhook

derekbit:
Longhorn #3701

### DIFF
--- a/chart/templates/deployment-webhook.yaml
+++ b/chart/templates/deployment-webhook.yaml
@@ -31,6 +31,8 @@ spec:
       - name: longhorn-conversion-webhook
         image: {{ template "registry_url" . }}{{ .Values.image.longhorn.manager.repository }}:{{ .Values.image.longhorn.manager.tag }}
         imagePullPolicy: IfNotPresent
+        securityContext:
+          runAsUser: 2000
         command:
         - longhorn-manager
         - conversion-webhook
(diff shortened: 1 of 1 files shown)


jenting (dismissed):

chart/templates/deployment-webhook.yaml
```diff
@@ -31,6 +31,8 @@ spec:
       - name: longhorn-conversion-webhook
         image: {{ template "registry_url" . }}{{ .Values.image.longhorn.manager.repository }}:{{ .Values.image.longhorn.manager.tag }}
         imagePullPolicy: IfNotPresent
+        securityContext:
+          privileged: true
```
innobead, line 35: This is like a workaround, because webhook containers don't really need to run in privileged mode. (I know we are reusing longhorn-manager image)

What about adding another non-default user in longhorn-manager image, then we specify `runAsUser` instead? Then, we can keep the least privilege principle, although this is still not a perfect way for me.

innobead (changes requested):

chart/templates/deployment-webhook.yaml
derekbit: > this is still not a perfect way for me

What's the better way to this issue except for the `runAsUser`?

innobead: The default user in the image is non-root then we don't need to update the spec actually.

derekbit: IIUC

We can add the uid and gid creation to the longhorn-manager Dockerfile
```
# Use an unprivileged user
RUN groupadd -g 10001 longhorn
RUN useradd -m -u 10001 longhorn -g longhorn
USER 10001:10001
```

By default, the webhook will run with user 10001.

But, the `runAsUser: 0` needs to be added in the remaining containers using longhorn-manager image those need to run with `root`.

Correct?

derekbit: IIUC

We can add the uid and gid creation to the longhorn-manager Dockerfile

# Use an unprivileged user
RUN groupadd -g 10001 longhorn
RUN useradd -m -u 10001 longhorn -g longhorn
USER 10001:10001
By default, the webhook will run with user 10001.

But, the runAsUser: 0 needs to be added in the remaining containers using longhorn-manager image those need to run with root.

Correct?

innobead: As discussed, just add `runAsUser` w/o updating the Dockerfile, because longhorn-manager is still used by other components which need root and privileged mode.

innobead (approved):

PR #3704, opened 2022-03-05
Update dev versions to components and chart for the master branch

innobead:

### DIFF
--- a/uninstall/uninstall.yaml
+++ b/uninstall/uninstall.yaml
@@ -102,7 +102,7 @@ spec:
       containers:
       - name: longhorn-uninstall
         image: longhornio/longhorn-manager:master-head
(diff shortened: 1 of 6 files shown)


jenting (approved):

uninstall/uninstall.yaml
```diff
@@ -101,7 +101,7 @@ spec:
     spec:
       containers:
       - name: longhorn-uninstall
-        image: longhornio/longhorn-manager:v1.2.3
+        image: longhornio/longhorn-manager:master-head
         imagePullPolicy: Always
```
shuo-wu, line 105: I just noticed that this value is still `Always`. Should we change it to `IfNotPresent`?

shuo-wu (approved):
LGTM

PR #3774, opened 2022-03-23
Update install.yaml

derekbit:
Longhorn 3772


jenting (approved):

jenting:
Leaving a note here:

It's better to handle in CI when PR create to check after running the script https://github.com/longhorn/longhorn/blob/master/scripts/generate-longhorm-yaml.sh, the existing directory is the same or not.

But for now, run the script by hand is the simple way.

innobead (approved):

PR #3785, opened 2022-03-24
Add LEP for orphaned data cleanup

derekbit:
The LEP describes the design of the orphaned data cleanup in longhorn ndoes.

Longhorn 685


enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,116 @@
+# Title
```
jenting, line 1:
```suggestion
# Orphaned Data Cleanup
```

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,116 @@
...
+The reasons for the orphaned replica directories might be because
+
+- The disks added contain replica directories that are not managed by the Longhorn cluster.
+- User rebalances the replicas of the volumes after one node is down. After the down node comes back, the replica directories on the node’s disks are not managed by Longhorn and become orphaned.
+
+### User Experience In Detail
+
+- Orphaned replica directories scan and identification process cannot be triggered by the user. The process is triggered by
+    - Node is added
+    - Disk is added
```
jenting, line 38: Should the process be triggered by the Node comes back?

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,116 @@
...
+     │                        queue.Add()               │                      │
+     └──────────────────────────────────────────────────┤  collect information │
+                                                        │                      │
+                                                        └──────────────────────┘
+    ```
+    
+
+- Introduce orphanedData CRD and controller
+    - orphanedData CRD: describe the orphaned data
+        - spe
```
jenting, line 73: spec

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,116 @@
...
+     │                                                  │   per-node monitor   │
+     │                        queue.Add()               │                      │
+     └──────────────────────────────────────────────────┤  collect information │
+                                                        │                      │
+                                                        └──────────────────────┘
+    ```
+    
+
+- Introduce orphanedData CRD and controller
+    - orphanedData CRD: describe the orphaned data
```
jenting, line 72: Do we need the status field to indicate that this orphanedData CR reconciles finished  (i.e., the orphan replica on the host be clean up)?

Do we need to delete the orphanedData CR once the orphan replica on the host is clean up?

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,116 @@
...
+        - Orphaned replica directories
+            - It is a time-consuming operation. Thus, the collection is triggered by adding a node and changing the disks.
+    - The reconciliation loop sync with the monitor and get the latest state. Then, create the `orphanedData` resources and update the state of the state of the node.
+    
+    - node eviction
+        - If the `node.Spec.EvictionRequested` is set to true, then delete the orphanedData resource associated with the node.
+
+- **orphaned data controller**
+    - A monitor is created for deleting the orphaned data if `DeletionTimestamp` is set.
+        - The deletion of the physical orphaned data is done in the monitor, because the operation is sometimes time-consuming.
```
jenting, line 100: Is it suitable that clean up logic implements inside the monitor routine?
Or should we have another routine to delete the orphaned data?

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,116 @@
...
+        - Once finish deletion, the `monitor.deletionState`  flag is updated to `completed`, and then the resource is enqueued.
+        - The reconciliation loop sync with the monitor and check the deletion state. If the state is `completed`, delete the resource.
```
jenting, line 102: Do you mean the orphaned data controller deletes the orphanedData CR once the monitor.deletionState is completed?

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,116 @@
...
+            - It is a time-consuming operation. Thus, the collection is triggered by adding a node and changing the disks.
+    - The reconciliation loop sync with the monitor and get the latest state. Then, create the `orphanedData` resources and update the state of the state of the node.
+    
+    - node eviction
+        - If the `node.Spec.EvictionRequested` is set to true, then delete the orphanedData resource associated with the node.
+
+- **orphaned data controller**
+    - A monitor is created for deleting the orphaned data if `DeletionTimestamp` is set.
+        - The deletion of the physical orphaned data is done in the monitor, because the operation is sometimes time-consuming.
+        - Once finish deletion, the `monitor.deletionState`  flag is updated to `completed`, and then the resource is enqueued.
```
jenting, line 101: Suggest describing the data struct of the monitor.

enhancements/20220324-orphaned-data-cleanup.md
derekbit: > Do we need the status field to indicate that this orphanedData CR reconciles finished (i.e., the orphan replica on the host be clean up)?

Do you mean the orphaned replica directory is cleaned up by a user rather than by the trigger from the UI?

I'm thinking the node resync period 30 seconds, but we can scan the orphaned data, for example, every 300 seconds in the node controller monitor. If the orphaned replicas does not exist, delete the resources. WDYT?

enhancements/20220324-orphaned-data-cleanup.md
derekbit: Yep!

enhancements/20220324-orphaned-data-cleanup.md
derekbit: > Is it suitable that clean up logic implements inside the monitor routine?

I'm not quite sure. Any concerns or potential issues, for example, time limit?

> Or should we have another routine to delete the orphaned data?

for example?

enhancements/20220324-orphaned-data-cleanup.md
derekbit: Yep. The full flow is
- user deletes the CR
 
- orphaned data controller is aware of the `deletionTimestamp` of the resources and start deleting the physical orphaned data in a monitor.
- The deletion is finished, and then the monitor set the  `monitor.deletionTimestamp`  and enqueue the resource.

- orphaned data controller's reconcile loop sync with monitor and check the flag.

- If the flag is true, delete the resource in the reconcile loop.

enhancements/20220324-orphaned-data-cleanup.md
jenting: The backup has 2 single sources of truth.
- If the snapshot backup is New/InProgress, the single source of truth is engine/replica
- If the snapshot backup is Completed or the backup is from another cluster, the single source of truth is the remote backup target

From my point of view, the monitor is dedicated to collecting the node/disk replica with a timer, and we could have another routine dedicated to cleaning up the orphan replica once it's triggered by the user. This makes the logic clearer.

derekbit: Yep. Got your point. Should not use the monitor for deletion.

derekbit:
Just discussed with @jenting,
we may add a `type` in the `replica` CRD, and recorded the orphaned replicas using the `replica` CRD with `type = orphaned`. The deletion is also handled in the replica controller.

enhancements/20220324-orphaned-data-cleanup.md
derekbit: Remove the monitor as the above discussion.

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,116 @@
...
+### User Stories
+
+The reasons for the orphaned replicas might be because
+
+- The disks added contain replicas that are not managed by the Longhorn cluster.
+- User rebalances the replicas of the volumes after one node is down. After the down node comes back, the replicas on the node’s disks are not managed by Longhorn and become orphaned.
+
+### User Experience In Detail
+
+- Orphaned replica scan and identification process cannot be triggered by a user. The process is triggered by
```
innobead, line 34: Is it possible to have replicas rebuilding to cause orphan replicas even the node and disks still alive? for example: disk space exhaustion

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,116 @@
...
+            
+    
+    - Information/state collection
+        
+        The monitor collects the following information/state, and then the data are stored in `monitor.nodeStatus`.
+        
+        - Disks status
+            - collected peridocally
+        - Orphaned replicas
+            - It is a time-consuming operation. Thus, the collection is triggered by
```
innobead, line 84: So the monitor is one-shot execution triggered by a specific even?

derekbit: No, the monitor is created in the first reconcile() and then keep collect the real world data, for example, disk status or orphaned replicas information.

For the disk status, it collects the disk status every 30 seconds as same as the resync time of the node controller.
But it dose not collect the orphaned replicas every 30 seconds, and only the events mentioned will trigger the collection.

enhancements/20220324-orphaned-data-cleanup.md
derekbit: After checking the replica rebuilding, the broken replica is deleted after failure.
I'm thinking that periodical scan is also necessary, because there might be orphaned replicas created unexpectedly.

enhancements/20220324-orphaned-data-cleanup.md
derekbit: Change the design, so close the conversation.

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,116 @@
...
+    
+
+- **longhorn-ui**
+    
+    Allow users to list the orphaned replicas on node page by sending `OrphanedReplicaDirectoryLiist` call to the backend.
+    
+    Allow users to select the orphaned replica directories to be deleted. The frontend needs to send `OrphanedReplicaDirectoryDelete` call to the backend.
+    
+
+### Test Plane
```
jenting, line 109: Test Plan

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,116 @@
...
+- Delete the orphaned replicas automatically
+
+## Proposal
+
+### User Stories
+
+The reasons for the orphaned replicas might be because
+
+- The disks added contain replicas that are not managed by the Longhorn cluster.
+- User rebalances the replicas of the volumes after one node is down. After the down node comes back, the replicas on the node’s disks are not managed by Longhorn and become orphaned.
```
jenting, line 30: The sentence behaviour should be double confirmed.
As we discussed, looks like the behaviour isn't like this.

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,116 @@
...
+            
+            The monitor is created in first reconciliations, and the subsequent reconciliations check and sync with the monitor.
+            
+    
+    - Information/state collection
+        
+        The monitor collects the following information/state, and then the data are stored in `monitor.nodeStatus`.
+        
+        - Disks status
+            - collected peridocally
```
jenting, line 82: periodically

innobead:
> Just discussed with @jenting,
> 
> we may add a `type` in the `replica` CRD, and recorded the orphaned replicas using the `replica` CRD with `type = orphaned`. The deletion is also handled in the replica controller.

Sounds good but is it possible to change lots of code in the current implementation to separate the handling between active replica and orphan replica? If that's a potential case, probably need to consider side effects.

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,116 @@
...
+        - If the node is down and detected, delete the replica resources with `type=orphaned`  in the node.
+
+- **replica controller**
+    
+    If the `deletionTimestamp` of the replica resource with `type=orphaned` is set,  `replica controller` is responsible for the deletion. Because the orphaned replicas does not have the instance manager, the deletion of the instance manager is not necessary.
+    
+
+- **longhorn-ui**
+    
+    Allow users to list the orphaned replicas on node page by sending `OrphanedReplicaDirectoryLiist` call to the backend.
```
jenting, line 104: `OrphanedReplicaDirectoryList`

The RESTful API could be designed to returns the node/disk replica but have a field to indicate which one(s) is/are orphan replica(s).

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,116 @@
...
+    
+    Allow users to select the orphaned replica directories to be deleted. The frontend needs to send `OrphanedReplicaDirectoryDelete` call to the backend.
```
jenting, line 106: If I understand correctly, the backend would delete the corresponding Replica CR?

enhancements/20220324-orphaned-data-cleanup.md
innobead: So how will you handle this case outside the trigger events you mentioned?

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,116 @@
...
+- The disks added contain replicas that are not managed by the Longhorn cluster.
+- User rebalances the replicas of the volumes after one node is down. After the down node comes back, the replicas on the node’s disks are not managed by Longhorn and become orphaned.
+
+### User Experience In Detail
+
+- Orphaned replica scan and identification process cannot be triggered by a user. The process is triggered by
+    - Node is added
+    - Down node comes back
+    - Disk is added
+- Longhorn UI node page shows the per-node orphaned replicas.
```
innobead, line 38: Suggest users have another batch operations by using custom resources to handle the query, select and deletion.

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,116 @@
...
+- User can select the orphaned replicas to be deleted on the Longhorn UI node page. After being selected, the backend will delete the replicas.
+
+## Design
+
+### **Implementation Overview**
+
+- **CRD**
+    
+    Add `spec.type` in `Replica` CRD. The `type` is either `active` or `orphaned`.
```
innobead, line 47: Do u think having an individual orphan resource definition with different types (here, it's replica)? So, we probably can extend this resource for other orphan cases in the future, and also there is no need to deal with the current code and upgrade migration because we will not change replica resource spec, but the cons is the replica spec can't be reused. WDYT?

derekbit:
> > Just discussed with @jenting,
> > we may add a `type` in the `replica` CRD, and recorded the orphaned replicas using the `replica` CRD with `type = orphaned`. The deletion is also handled in the replica controller.
> 
> Sounds good but is it possible to change lots of code in the current implementation to separate the handling between active replica and orphan replica? If that's a potential case, probably need to consider side effects.


Currently, we only focus on orphaned replicas.
If we would handle other types of orphaned data in the future. Using a dedicated crd and controller is also a good choice for extension.

enhancements/20220324-orphaned-data-cleanup.md
derekbit: Yep. Delete the physical data and CR.

enhancements/20220324-orphaned-data-cleanup.md
derekbit: Yep.

enhancements/20220324-orphaned-data-cleanup.md
derekbit: Yep. I'm still checking the ways for generating replicas.

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,116 @@
...
+        - Orphaned replicas
+            - It is a time-consuming operation. Thus, the collection is triggered by
+                - add a node,
+                - a down node back,
+                - changing the disks.
+    
+    - The reconciliation loop sync with the monitor and get the latest state. Then, create the `replica` resources with `type=orphaned` and update the node/disks status as well.
+    
+    - Node eviction
+        - If the `node.Spec.EvictionRequested` is set to `true`, delete the replica resources with `type=orphaned`  in the node.
```
innobead, line 92: We should not delete orphans until users decide to delete, right?

jenting:
For UX, using orphaned data CRD describes the orphaned data might not be confusing?

Yes, but using label selector on Replica CR is another way.

enhancements/20220324-orphaned-data-cleanup.md
derekbit: The monitor collect the data every 30 seconds. In every period, the disk status is collected.
I might add a counter in monitor struct, and every 10 period (i.e. 300 seconds), the orphaned data will be rescanned.

Any suggestion?

enhancements/20220324-orphaned-data-cleanup.md
derekbit: I see.
Every API should accept a list or return a list.

enhancements/20220324-orphaned-data-cleanup.md
derekbit: Yep. As you said, I prefer a dedicated crd and controller.

enhancements/20220324-orphaned-data-cleanup.md
derekbit: Yep. Only delete CRs and won't touch the physical data.

derekbit:
> For UX, using orphaned data CRD describes the orphaned data might not be confusing?
> 
> Yes, but using label selector on Replica CR is another way.

Yep. It's feasible.

enhancements/20220324-orphaned-data-cleanup.md
innobead: Actually I mean kubectl but for UI, we should have a consistent behavior of batch as other resources.

This can be a phased implementation, we can do CRD first in 1.3.0 first, UI later.

derekbit: Okay. I see. The consistency is a must.

enhancements/20220324-orphaned-data-cleanup.md
derekbit: One way to create the orphaned replicas
1. Create a volume with 3 replicas
2. Shutdown one node containing one of the three replicas
3. Scale down the replica number to 2
4. Delete the replica in error
5. Power the down node up, and then the orphaned replicas can be found in this node.

enhancements/20220324-orphaned-data-cleanup.md
jenting: Having a dedicated CRD is a good approach.

However, we need to take care that the users could create the Orphan Replica CR manually, would it cause the healthy CR deleted by the controller?
If yes, the real data would be deleted by accident.
If no, the controller needs to validate the replica on the disk is _really_ an orphan replica.

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,144 @@
+# ****Orphaned Replica Cleanup****
+
+## **Summary**
+
+Orphaned replica cleanup identifies unmanaged replica on the disks and provides a list of the orphaned replica on each node. Longhorn will not delete the replicas automatically, preventing deletions by mistake. Instead, it allows the user to select and trigger the deletion of the orphaned replica by manual.
```
jenting, line 5:
```suggestion
Orphaned replica cleanup identifies unmanaged replicas on the disks and provides a list of the orphaned replica on each node. Longhorn will not delete the replicas automatically, preventing deletions by mistake. Instead, it allows the user to select and trigger the deletion of the orphaned replica manual.
```

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,144 @@
...
+
+- Delete the orphaned replicas automatically
+
+## **Proposal**
+
+### **User Stories**
+
+The reasons for the orphaned replicas might be because
+
+- A disks added contains replicas that are not managed by the Longhorn cluster.
```
jenting, line 29:
```suggestion
- A disk added contains replicas that are not managed by the Longhorn cluster.
```

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,144 @@
...
+    
+    If the `deletionTimestamp` of the orphan resource is set, `orphan controller` is responsible for the deletion.
+    
+    The steps to delete an orphan are
+    
+    - Delete the physical data. If failed, the `status` and `reason` of `orphan.status.condition['deletable']` is set.
+    - If the physical data is deleted successfully, delete the `orphan` resource.
+- **longhorn-ui**
+    
+    Allow users to list the orphans on node page by sending `OrphanLiist` call to the backend.
```
jenting, line 132:
```suggestion
    Allow users to list the orphans on the node page by sending `OrphanList` call to the backend.
```

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,144 @@
...
+    The dedicated CRD and controller is designed for orphaned data. In the future, other types of orphaned data are described and processed by the CRD and controller.
+    
+    - CRD
+        
+        The CRD describes the type, properties and state of orphaned data
+        
+    
+    ```go
+    type OrphanSpec struct {
+    	// Can be "replica" or "unknown".
```
jenting, line 62: Looks like the `unknown` type would not be used

enhancements/20220324-orphaned-data-cleanup.md
derekbit: > However, we need to take care that the users could create the Orphan Replica CR manually, would it cause the healthy CR deleted by the controller?

Yep. Need to check the orphan resource with the active replica first before clean it up.
If it is illegal, set the orphan source's `condition['deletable']` to false and set the reason.

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,144 @@
...
+    	OwnerID string `json:"ownerID"`
+    
+    	// Directory or file name of the orphaned data
+    	DataName string `json:"dataName"`
+    
+    	DiskPath string `json:"diskPath"`
+    }
+    
+    type OrphanStatus struct {
+    	Conditions map[string]Condition `json:"conditions"`
```
jenting, line 74: Suggestion to follow Kubernetes conditions definition which is a slice, not a map.

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,144 @@
...
+    type OrphanSpec struct {
+    	// Can be "replica" or "unknown".
+    	Type OrphanType `json:"type"`
+    
+    	OwnerID string `json:"ownerID"`
+    
+    	// Directory or file name of the orphaned data
+    	DataName string `json:"dataName"`
+    
+    	DiskPath string `json:"diskPath"`
```
jenting, line 70: Looks like you want to design a _generic_ garbage collection CRD.
However, with these two fields, it won't be _generic_ anymore. WDYT?

enhancements/20220324-orphaned-data-cleanup.md
derekbit: Yep. Maybe need to use it in this phase.

enhancements/20220324-orphaned-data-cleanup.md
derekbit: Currently, we only handle the orphaned replicas.
If we handle the file or other directories, for example, there is a file called `thisisagarbage.txt' in `/var/lib/longhorn`,
the `DataName` and `DiskPath` of the orphan is `thisisagarbage.txt` and `/var/lib/longhorn`.

But yes, if the orphan is not a file or a directory, the two fields won't be generic.

enhancements/20220324-orphaned-data-cleanup.md
derekbit: Yep. It's a typo here.

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,144 @@
...
+        - If the `node.Spec.EvictionRequested` is set to `true`, delete the `orphan` resources in the node. The physical orphaned data should not be deleted.
+    - Node down
+        - If the node is down and detected, delete the `orphan` resources in the node.
+- **orphan controller**
+    
+    If the `deletionTimestamp` of the orphan resource is set, `orphan controller` is responsible for the deletion.
+    
+    The steps to delete an orphan are
+    
+    - Delete the physical data. If failed, the `status` and `reason` of `orphan.status.condition['deletable']` is set.
```
shuo-wu, line 128: Need to check `node.Spec.EvictionRequested` first.

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,144 @@
...
+            The monitor is created in first reconciliations, and the subsequent reconciliations check and sync with the monitor.
+            
+    - Information/state collection
+        
+        The monitor collects the following information/state, and then the data are stored in `monitor.nodeStatus`.
+        
+        - Disks status
+            - collected periodically
+        - Orphaned replicas
+            - It is a time-consuming operation. Thus, the collection is triggered by
```
shuo-wu, line 112: I don't think you need to differentiate the disk status detection from the orphaned replica identification:
  - With the separate monitor on each node, Longhorn can use a longer interval without affecting by the resync period.
  - The orphaned replica identification is a multi-step operation, which may be not more expensive than the disk status detection:
    - For each disk, Longhorn will need to list all replica directories. If there is nothing new, it does not need to do anything. Most of the time the orphaned replica identification should stop here, which involves in one read operation. The price is the same as that of the disk check.
    - Otherwise, it needs to check the files in the new directories and see if it match the structure of replica data directories, which is more expensive comparing with the previous step. (But if there is an unknown directory living in the disk path for a long time, Longhorn may repeatedly check it but won't create an orphaned. Then this may be a problem.)

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,144 @@
...
+### **Goals**
+
+- Identify the orphaned replicas
+- The scanning process should not stuck the reconciliation of the controller
+- Provide user a way to select and trigger the deletion of the orphaned replicas
+
+### **Non-goals**
+
+- Delete the orphaned replicas automatically
+
```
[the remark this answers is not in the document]
derekbit: So, it will be extended to unknown files/dirs, correct?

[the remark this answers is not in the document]
shuo-wu: This is a non-goal item. But if you want to handle the last case I mentioned in [this comment](https://github.com/longhorn/longhorn/pull/3785#discussion_r836086105), you may need to ignore this.

enhancements/20220324-orphaned-data-cleanup.md
derekbit: > For each disk, Longhorn will need to list all replica directories. If there is nothing new, it does not need to do anything. Most of the time the orphaned replica identification should stop here, which involves in one read operation. The price is the same as that of the disk check.

`Longhorn will need to list all replica directories` 

Do you mean listing on-disk replica directories, right?

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,144 @@
...
+The reasons for the orphaned replicas might be because
+
+- A disk added contains replicas that are not managed by the Longhorn cluster.
+- User rebalances the replicas of the volumes after one node is down. After the down node comes back, the replicas on the node’s disks are not managed by Longhorn and become orphaned. The steps to reproduce are:
+    1. Create a volume with 3 replicas
+    2. Shutdown one node containing one of the three replicas
+    3. Scale down the replica number to 2
+    4. Delete the replica in error
+    5. Power the down node up, and then the orphaned replicas can be found in this node.
```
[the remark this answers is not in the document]
derekbit: Okay!

enhancements/20220324-orphaned-data-cleanup.md
innobead: As I mentioned before, when having a generic, you always will have this question. Either keep extending the spec along the time for different info or use a generic data structure for that. (personally, I prefer the first, a strict typed, but if choosing this, we need to ensure if any compatibility we need to consider)

enhancements/20220324-orphaned-data-cleanup.md
innobead: Exactly, I mean `when`.

enhancements/20220324-orphaned-data-cleanup.md
derekbit: Yep. I put the `properties` of the orphan into `Parameters map[string][string]`.

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,162 @@
...
+     └──────────────────────────────────────────────────┤  collect information │
+                                                        │                      │
+                                                        └──────────────────────┘
+    
+    ```
+    
+    - The interaction between the reconciliation loop and the monitor
+        - Monitor creation
+            
+            The monitor is created in first reconciliations, and the subsequent reconciliations check and sync with the monitor.
```
shuo-wu, line 105: The monitor is started along with the node controller. I mean, at the end of `NewNodeController`.

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,162 @@
...
+            Conditions []Condition `json:"conditions"`
+        }
+        ```
+
+        - `Spec.Parameters["Name"]`: orphaned replica directory name
+        - `Spec.Parameters["DiskUUID"]`: The uuid of the disk the orphaned replica directory is in
+        - `Spec.Parameters["DiskPath"]`: The path of the disk the orphaned replica directory is in
+        
+        
+- **node controller**
```
shuo-wu, line 80: You can re-organize this part. The illustration is not necessary. e.g.,
```
Node controller is responsible for:
  - Start the monitor during init.
  - Sync with the monitor in each reconcile loop.
  - Update the node/disk status or Create Orphan CRs based on the monitor info.

The node monitor:
  - Struct:
    `type NodeMonitor struct {}`
  - Periodically detect/verify disk: Run `stat`; Check disk UUID in the metafile; balabala...
  - Periodically check and identify orphan directories: YOU SHOULD MENTION MORE DETAILS HERE, including how it can identify orphan replica directories and how it will handle unknown files/directories.
```

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,162 @@
...
+        - If the `node.Spec.EvictionRequested` is set to `true`, delete the `orphan` resources in the node. The physical orphaned data should not be deleted.
+    - Node down
+        - If the node is down and detected, delete the `orphan` resources in the node.
+- **orphan controller**
+    
+    If the `deletionTimestamp` of the orphan resource is set, `orphan controller` is responsible for the deletion.
+    
+    The steps to delete an orphan are
+
+    - Check if the node containing the orphan is evicted by checking `node.Spec.EvictionRequested`. If YES, the deletion is not allowed.
```
shuo-wu, line 144: Do you mean skipping actual data deletion?

shuo-wu (commented):
You can refer to other LEPs and see what each section is for.

innobead (approved):
In general, LGTM

enhancements/20220324-orphaned-data-cleanup.md
derekbit: Yep. If the node is evicted, Longhorn should not delete the physical data.

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,162 @@
...
+            Type OrphanType `json:"type"`
+            // The parameters of the orphaned data
+            // +optional
+            // +nullable
+            Parameters map[string]string `json:"parameters"`
+        }
+            
+        type OrphanStatus struct {
+            // +optional
+            OwnerID string `json:"OwnerID"`
```
jenting, line 68: Not sure if we need the `status.OwnerID` or not because if the orphan replica data is on Node A, then should be Node A has the ability to access the Node A host and perform the cleanup procedure.

derekbit: I don't think we need it, so it will be removed.

shuo-wu: If node A is down and we need to remove the CRs, there should be **only** **another** controller responsible for handling it.

jenting: > If a node is down and we need to remove the CRs on the node, there should be **only** **another** controller responsible for handling it.

I see. So if I understand correctly, the longhorn-manager should check the `spec.NodeID` and `status.OwnerID` to decide whether the reconcile loop should perform clean up orphan replica data on the disk _or_ should perform delete the orphan replica CR.

shuo-wu: YES. This could be a use case of this field.

shuo-wu (changes requested):

shuo-wu:
Based on today's meeting with the customers/users, they may want the auto-cleanup feature as well. Can we have a setting controlling it? cc @innobead

innobead:
> Based on today's meeting with the customers/users, they may want the auto-cleanup feature as well. Can we have a setting controlling it? cc @innobead

We can have that during the implementation, but by default, it should be disabled to ensure the users know the risk.

shuo-wu:
Yes, by default the setting should be disabled.

derekbit:
If enabl the auto-deletion, should Longhorn create the orphan CRs for the orphaned replica directories? Or, just delete them directory directories while scanning?

innobead:
> If enabl the auto-deletion, should Longhorn create the orphan CRs for the orphaned replica directories? Or, just delete them directory directories while scanning?

We should just do the same process, but only one more deletion action should be triggered by the monitor after the resource creation, then leave the controller to handle the deletion process. 
```
Create `orphan` CRs based on the information collected by the monitor.
```

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,156 @@
...
+    - Run `stat`
+    - Check disk FSID
+    - Check disk UUID in the metafile
+  - Periodically check and identify orphan directories
+
+    - List on-disk directories in `${disk_path}/replicas` and compare them with the last record stored in `monitor.onDiskDirectoriesInReplicas`.
+    - If the two lists are different, iterate all directories in `${disk_path}/replicas` and then get the list of the orphaned replica directories.
+    
+        A valid replica directory has the properties:
+        - The directory name format is `${disk_path}/replicas/pvc-${uuid}-${id}`
```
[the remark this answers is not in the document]
derekbit: yep. the 

> > If enabl the auto-deletion, should Longhorn create the orphan CRs for the orphaned replica directories? Or, just delete them directory directories while scanning?
> 
> We should just do the same process, but only one more deletion action should be triggered by the monitor after the resource creation, then leave the controller to handle the deletion process. (responsibility segregation)
> 
> ```
> Create `orphan` CRs based on the information collected by the monitor. Delete orphan CRs if auto-deletion is enabled. 
> ```

Sounds good. One more question, do we need a retention period for each CRs? If expired, delete the CRs.

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,199 @@
...
+      // The age in second of the orphaned data
+      Age metav1.Duration `json:"age"`
+      // +optional
+      // +nullable
+      Conditions []Condition `json:"conditions"`
+    }
+    ```
+    - If receive the deletion request, delete the on-disk orphaned replica directory and the `orphan` resource.
+
+    - If the auto-deletion is enabled, sync with the orphan monitor and update the `status.age` of the orphan resources periodically. If the orphans are expired, delete the orphans and its on-disk orphaed replica directories.
```
shuo-wu, line 162: Typically, the controller should not responsible for managing the lifecycle of its own CRD. I mean, sending CR creation or deletion requests.

shuo-wu (commented):
Do we need the TTL setting for the orphan CRs? If you want it, you can remove `orphan-auto-deletion` and use a negative TTL setting value to present the disabling.

IMO, TTL is unnecessary. And using a separate orphan monitor (with 2 extra fields) to handle it seems to be a little complicated (in term of the life cycle management). Can we directly issue deletion requests in the node controller (similar to the node eviction case) instead?

enhancements/20220324-orphaned-data-cleanup.md
derekbit: Got it!

derekbit:
> Do we need the TTL setting for the orphan CRs? If you want it, you can remove `orphan-auto-deletion` and use a negative TTL setting value to present the disabling.
> 
> IMO, TTL is unnecessary. And using a separate orphan monitor (with 2 extra fields) to handle it seems to be a little complicated (in term of the life cycle management). Can we directly issue deletion requests in the node controller (similar to the node eviction case) instead?

Maybe we can simplify it in this phase.
 auto-deletion is controlled by `node.Spec.orphanAutoDeletion` (per-node configuration)
- `node.Spec.orphanAutoDeletion=true`: node controller creates the orphan CRs and then issues the orphan deletion requests.
- `node.Spec.orphanAutoDeletion=false`: node controller creates the orphan CRs. Users delete the orphan manually.

How about it?

innobead:
I believe TTL is just another granular behavior in general but agree that we probably just KISS first to delete them directly, but it will not hurt to describe the idea in the proposal.

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,169 @@
...
+      // +optional
+      OwnerID string `json:"ownerID"`
+      // +optional
+      // +nullable
+      Conditions []Condition `json:"conditions"`
+    }
+    ```
+    - If receive the deletion request, delete the on-disk orphaned replica directory and the `orphan` resource.
+
+    - If the auto-deletion is enabled, sync with the orphan monitor and issues the orphans deletion requests.
```
innobead, line 153: Should the auto-deletion be triggered from the node monitor? Why we will have an orphan monitor?

derekbit: Oh, sorry. Won't have the orphan monitor. 
```
If the auto-deletion is enabled, node controller will issues the orphans deletion requests.
```

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,169 @@
...
+                                                        │                      │
+                                                        │   per-node monitor   │
+                                                        │                      |
+                                                        ┤  collect information │
+                                                        │                      │
+                                                        └──────────────────────┘
+    
+    ```
+
+3. The monitor on each orphan controller is created. If the auto-deletion is enabled, it periodically updates the `status.age` of the orphan resources cached in the `monitor` and then enqueues the orphans. Then, the reconciliation loop of of the orphan controller checks and deletes the expired orphaned replica directories and the corresponding resources.
```
shuo-wu, line 54: We won't launch this monitor, right?

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,169 @@
...
+### Implementation Overview
+**Settings**
+  - Add setting `orphan-auto-deletion`. Default value is `false`.
+
+**Node controller**
+  - Start the monitor during initialization.
+  - Sync with the monitor in each reconcile loop.
+  - Update the node/disk status.
+  - Create `orphan` CRs based on the information collected by the monitor.
+  - Delete the `orphan` CRs if the node/disk is requested to be evicted.
```
shuo-wu, line 86: Add 2 more cases: 
  1. Delete the `orphan` CRs if the corresponding directories disappear
  2. The auto-deletion setting is enabled.

shuo-wu (changes requested):
Please make sure the previous proposal (TTL) is removed

enhancements/20220324-orphaned-data-cleanup.md
derekbit: Right!

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -0,0 +1,169 @@
...
+    Allow users to list the orphans on the node page by sending `OrphanList` call to the backend.
+    
+    Allow users to select the orphans to be deleted. The frontend needs to send `OrphanDelete` call to the backend.
+    
+
+### Test Plan
+
+**Integration tests**
+
+- Create the orphaned replica directories based on the two scenarios described in User Stories and test orphaned replica directories identification.
```
shuo-wu, line 166: NIT: Need to test:
  1. Orphan CRs will be created correctly when there are multiple kinds of files/dirs in disk path. And it can be cleaned up with the dir.
  2. Orphan CRs will be removed when the directories disappear or the node/disk is evicted.
  3. The auto-deletion setting

shuo-wu (commented):
In general LGTM

innobead (dismissed):

shuo-wu (approved):

Issue #3787, opened 2022-03-25
[BUG] Data migration sample YAML moves data to the subfolder


c3y1huang:
## Describe the bug

Example [data_migration.yaml](https://github.com/longhorn/longhorn/blob/master/examples/data_migration.yaml#L22) moved data into subfolder /old.

## To Reproduce

1. Create [PVC and Pod](https://github.com/longhorn/longhorn/blob/master/examples/pod_with_pvc.yaml) and write some data to it.
2. Detach the volume (delete Pod).
3. Create a new PVC.
5. Update and deploy [data_migration.yaml](https://github.com/longhorn/longhorn/blob/master/examples/data_migration.yaml).
6. Wait for the job to complete.
7. Attach the volume to the new PVC.

Result
8. Data migrated to `/old` subfolder.

## Expected behavior

Data should be migrated to the correct path.

## Log or Support bundle

`n/a`

## Environment

 `n/a`
 
## Additional context

`n/a`

yangchiu:
Verified on master-head.

nested folders and their files can be successfully migrated from old path to new path.

PR #3788, opened 2022-03-25
Fix data migration YAML command args

c3y1huang:
https://github.com/longhorn/longhorn/issues/3787


innobead (approved):

innobead:
@c3y1huang Please help update to longhorn/longhorn-manager examples as well.

Issue #2203, opened 2021-01-21
[FEATURE] instance-manager compatibility with Cluster Autoscaler


drewhemm:
**Is your feature request related to a problem? Please describe.**
When using Cluster Autoscaler, nodes are not removed from the cluster if they have bare pods; an annotation can be added to pods to let CA know it is safe to evict them:

https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node

```
"cluster-autoscaler.kubernetes.io/safe-to-evict": "true"
```

A node that has no other workloads running should be a candidate for removal from the cluster, but alas:

```
I0121 16:44:11.071945       1 cluster.go:93] Fast evaluation: ip-10-xxx-xx-xxx.eu-west-1.compute.internal for removal
I0121 16:44:11.071966       1 cluster.go:107] Fast evaluation: node ip-10-xxx-xx-xxx.eu-west-1.compute.internal cannot be removed: longhorn-system/instance-manager-e-1c19a568 is not replicated
```

**Describe the solution you'd like**
The option to specify annotations to be added to instance-manager pods would solve this issue. When installing Longhorn via Helm, there is the option to provide annotations for the ingress in the `values.yaml` - similar functionality would be desirable. 

**Describe alternatives you've considered**
A mutating webhook might be an option, but that is a lot of work. 

**Additional context**
Without this functionality, it is not possible to use Longhorn in conjuction with Cluster Autoscaler as it CA will be blocked from removing any nodes from the cluster.

drewhemm:
I tried to find the source code for where the pod spec is defined, but could not find it.

joshimoo:
If you want to dynamically remove nodes from the cluster you want to use dedicated storage nodes, otherwise you risk removing the last replica of your data.

https://longhorn.io/docs/1.1.0/volumes-and-nodes/storage-tags/
https://longhorn.io/docs/1.1.0/advanced-resources/default-disk-and-node-config/

We implement a poddisruption budget to prevent eviction of the last healthy replica. To prevent a user from doing just that.
We cannot add the above since it's not safe to delete the nodes unless we know there are other nodes with a healthy replica of the data.

drewhemm:
[...]

PR #3831, opened 2022-04-08
Add LEP - Support K8s Cluster Autoscaler

c3y1huang:
https://github.com/longhorn/longhorn/issues/2203

### DIFF
--- a/enhancements/20220408-support-kubernetes-ca.md
+++ b/enhancements/20220408-support-kubernetes-ca.md
@@ -1,0 +1,177 @@
+# Support Kubernetes Cluster Autoscaler
+
+Longhorn should support Kubernetes Cluster Autoscaler.
+
+## Summary
+
+Currently, Longhorn pods are [blocking CA from removing a node](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node). This proposes to introduce a new global setting `kubernetes-cluster-autoscaler-enabled` that will annotate Longhorn components and also add logic for instance-manager PodDisruptionBudget management.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/2203
+
+## Motivation
+
+### Goals
+
+- Longhorn should block CA from scaling down if a node met ANY condition:
+    - Any volume attached
+    - Contains a backing image manager pod
+    - Contains a share manager pod
+- Longhorn should not block CA from scaling down if a node met ALL conditions:
+    - All volume detached and there is another schedulable node with volume replica and replica IM PDB.
+    - Not contain a backing image manager pod
+    - Not contain a share manager pod
+
+### Non-goals [optional]
+
+- CA setup.
+- CA blocked by kube-system components.
+- CA blocked by backing image manager pod. (TODO)
+- CA blocked by share manager pod. (TODO)
+
+## Proposal
+Set `kubernetes-cluster-autoscaler-enabled` adds `cluster-autoscaler.kubernetes.io/safe-to-evict` annotation to Longhorn pods that are not backed by a controller, or with local storage volume mounts. To avoid data loss, Longhorn does not annotate the backing image manager and share manager pods.
+
+Currently, Longhorn creates instance-manager PDBs for replica/engine regardless of the volume state.
+During scale down, CA tries to find a removable node but failed by those instance-manager PDBs.
+
+We can add IM PDB handling to create and retained when the PDB is required:
+
+- There are volumes/engines running on the node. We need to guarantee that the volumes won't crash.
+- The only available/valid replica of a volume is on the node. Here we need to prevent the volume data from being lost.
+
+### User Stories
+
+#### CA scaling
+Before the enhancement, CA will be blocked by
+- Pods that are not backed by a controller (engine/replica instance manager).
+- Pods with [local storage volume mounts](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/utils/drain/drain.go#L222) (longhorn-ui, longhorn-csi-plugin, csi-attacher, csi-provisioner, csi-resizer, csi-snapshotter).
+
+After enhancement, instance manager PDB will be actively managed by Longhorn:
+- Creates all engine/replica instance manager PDB when the volume is attached.
+- Delete engine instance manager PDB when the volume is detached.
+- Delete but keep 1 replica instance manager PDB when the volume is detached.
+
+the user can set a new global setting `kubernetes-cluster-autoscaler-enabled` to unblock CA scaling. This allows Longhorn to annotate Longhorn-managed deployments and engine/replica instance manager pods with `cluster-autoscaler.kubernetes.io/safe-to-evict`.
+
+
+### User Experience In Detail
+
+- Configure the setting via Longhorn UI or kubectl.
+- Ensure all volume replica count is set to more than 1.
+- CA is not blocked by Longhorn components when the node doesn't contain volume replica, backing image manager pod, and share manager pod.
+    - Engine/Replica instance-manager PDB will block the node if the volume is attached.
+    - Replica instance-manager PDB will block the node when CA tries to delete the last node with the volume replica.
+
+### API changes
+
+`None`
+
+## Design
+
+### Implementation Overview
+
+#### Global setting
+- Add new global setting `Kubernetes Cluster Autoscaler Enabled (Experimental)`.
+  - The setting is `boolean`.
+  - The default value is `false`.
+
+#### Annotations
+
+When setting `kubernetes-cluster-autoscaler-enabled` is `true`, Longhorn will add annotation `cluster-autoscaler.kubernetes.io/safe-to-evict` for the following pods:
+- The engine and replica instance-manager pods because those are not backed by a controller and use local storage mounts.
+- The deployment workloads are managed by the longhorn manager and using any local storage mount. The managed components are labeled with `longhorn.io/managed-by: longhorn-manager`.
+
+#### PodDisruptionBudget
+
+- No change to the logic to cleanup PDB if instance-manager doesn't exist.
+
+- Engine IM PDB:
+    - Delete PDB if volumes are detached;
+        - There is no instance process in IM (im.Status.Instance).
+        - The same logic applies when a node is un-schedulable. Node is un-schedulable when marked in spec or with CA tainted `ToBeDeletedByClusterAutoscaler`;
+    - Create PDB if volumes are attached; there are instance processes in IM (im.Status.Instance).
+
+- Replica IM PDB:
+    - Delete PDB if setting `allow-node-drain-with-last-healthy-replica` is enabled.
+    - Delete PDB if volumes are detached;
+        - There is no instance process in IM (im.Status.Instance)
+        - There are other schedulable nodes with healthy volume replica and have replica IM PDB.
+    - Delete PDB when a node is un-schedulable. Node is un-schedulable when marked in spec or with CA tainted `ToBeDeletedByClusterAutoscaler`;
+        - Check if the condition is met to delete PDB (same check as to when volumes are detached).
+        - Enqueue the replica instance-manager of another schedulable node with the volume replica.
+        - Delete PDB.
+    - Create PDB if volumes are attached:
+        - There are instance processes in IM (im.Status.Instance).
+    - Create PDB when volumes are detached;
+        - There is no instance process in IM (im.Status.Instance)
+        - The replica has been started. There are no other schedulable nodes with healthy volume replica and have replica IM PDB.
+
+### Test plan
+
+#### Scenario: test CA
+
+    Given Cluster with Kubernetes cluster-autoscaler.
+    And Longhorn installed.
+    And Set `kubernetes-cluster-autoscaler-enabled` to `true`.
+    And Create deployment with cpu request.
+    ```
+    resources:
+      limits:
+        cpu: 300m
+        memory: 30Mi
+      requests:
+        cpu: 150m
+        memory: 15Mi
+    ```
+
+    When Trigger CA to scale-up by increase deployment replicas.
+         (double the node number, not including host node)
+    ```
+    10 * math.ceil(allocatable_millicpu/cpu_request*node_number/10)
+    ```
+    Then Cluster should have double the node number.
+
+    When Trigger CA to scale-down by decrease deployment replicas.
+         (original node number)
+    Then Cluster should have original node number.
+
+#### Scenario: test CA scale down all nodes containing volume replicas
+
+    Given Cluster with Kubernetes cluster-autoscaler.
+    And Longhorn installed.
+    And Set `kubernetes-cluster-autoscaler-enabled` to `true`.
+    And Create volume.
+    And Attach the volume.
+    And Write some data to volume.
+    And Detach the volume.
+    And Create deployment with cpu request.
+
+    When Trigger CA to scale-up by increase deployment replicas.
+         (double the node number, not including host node)
+    Then Cluster should have double the node number.
+
+    When Annotate new nodes with `cluster-autoscaler.kubernetes.io/scale-down-disabled`.
+         (this ensures scale-down only the old nodes)
+    And Trigger CA to scale-down by decrease deployment replicas.
+        (original node number)
+    Then Cluster should have original node number + 1 blocked node.
+
+    When Attach the volume to a new node. This triggers replica rebuild.
+    And Volume data should be the same.
+    And Detach the volume.
+    Then Cluster should have original node number.
+    And Volume data should be the same.
+
+#### Scenario: test CA should block scale down of node running backing image manager pod
+
+Similar to `Scenario: test CA scale down all nodes containing volume replicas`.
+
+### Upgrade strategy
+
+`N/A`
+
+## Note [optional]
+
+`N/A`


enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,94 @@
...
+
+Currently, Longhorn creates instance-manager PDBs for replica/engine regardless of the volume state.
+During scale down, CA tries to find a removable node but failed by those instance-manager PDBs. We can remove the PDB when the volume is not attached and at least one replica is available.
+
+### User Stories
+
+#### CA scaling
+Before the enhancement, CA will be blocked by
+- Pods that are not backed by a controller (engine/replica instance manager).
+- Pods with [local storage volume mounts]( local storage volume mounts) (longhorn-ui, longhorn-csi-plugin, csi-attacher, csi-provisioner, csi-resizer, csi-snapshotter).
```
shuo-wu, line 36: How about share manager pods and backing image datasource/manager pods?

For backing image manager pods, if the disk/node contains the only file of one backing image, then you cannot add the annotation for the pod.

For share manager pods, the eviction means you need to stop the RWX volume service for a while and relaunch them on other nodes. Probably we cannot do anything for this until HA RWX support is done.

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### Annotations
+
+- Annotate engine and replica instance-manager pods when no instance process inside the instance manager.
+- Annotate longhorn-ui, longhorn-csi-plugin, csi-attacher, csi-provisioner, csi-resizer, and csi-snapshotter.
+
+#### PodDisruptionBudget
+
+- When the volume is attached, the logic of creating engine/replica IM PBD is the same.
+- Delete engine IM PBD when the volume is detached.
```
shuo-wu, line 64: Do you mean actively deleting the PDB when there is no process in the IM? It's better to add more details here. e.g., the new PDB sync-up function logic would be like:
1. Check if the PDB can be deleted (I am not sure if the following conditions are corrects or not, will leave them to you for further verifications)
    1. If the node is already gone/tainted/...
    2. If the IM or IM pod is already gone
    3. For engine IM PDB, if there is no engine process......
    4. For replica IM PDB, If there is no replica scheduled......
2. Check if the PDB can be created
    1. volume attachment requirement
    2. replicas are scheduled
    3. ......

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,95 @@
...
+- Keep/create one replica IM PBD when the volume is detached.
+- When the IM node is unschedulable (tainted by CA), delete the replica IM PBD when there are other schedulable nodes. Before deleting the replica IM PBD, enqueue the schedulable node IM.
+
+### Test plan
+
+Given Enable `kubernetes-autoscaler-support` setting.
```
shuo-wu, line 70: How about the cluster config?

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,95 @@
...
+And Create volume with 2 replicas.
+And Attach the volume to start the replicas.
+And Detach volume.
+
+When create deployment with cpu request.
+And trigger CA scale-up 2 nodes by increate deployment replicas.
+And annotate new nodes with `scale-down-disable` to test when CA tries to scale-down all volume replica nodes.
```
shuo-wu, line 77: You mean disabling the scaling down for the new node that does not contain any volume replica, then check if CA could kick out the 2 other nodes?

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,95 @@
...
+- Annotate longhorn-ui, longhorn-csi-plugin, csi-attacher, csi-provisioner, csi-resizer, and csi-snapshotter.
+
+#### PodDisruptionBudget
+
+- When the volume is attached, the logic of creating engine/replica IM PBD is the same.
+- Delete engine IM PBD when the volume is detached.
+- Keep/create one replica IM PBD when the volume is detached.
+- When the IM node is unschedulable (tainted by CA), delete the replica IM PBD when there are other schedulable nodes. Before deleting the replica IM PBD, enqueue the schedulable node IM.
+
+### Test plan
```
shuo-wu, line 68: As I mentioned above, need to have tests with backing images or share managers

shuo-wu (changes requested):

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: ATM I don't think there is anything we could do about backing image datasource/manager, and share manager pods. The nodes running with those pods will be blocked by CA local storage condition.

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,94 @@
+# Support Kubernetes CA
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: Updated.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: Reworded.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: Added.

enhancements/20220408-support-kubernetes-ca.md
shuo-wu: For the backing image manager, I think we can do something here. 
As I mentioned, if all backing image files on the node contain duplicates, it's fine to add annotation `cluster-autoscaler.kubernetes.io/safe-to-evict` for the related backing image manager.
The backing image managers won't remove the related files immediately when there is no replica using them. There is a setting named `Backing Image Cleanup Wait Interval` controlling it. Hence, this may unnecessarily block CA.
Of course, the handling of the backing image manager is nice to have. If it makes the logic too complicated, we can skip it now.

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,171 @@
...
+
+- Replica IM PBD:
+    - Delete PBD if setting `allow-node-drain-with-last-healthy-replica` is enabled.
+    - Delete PBD if volumes are detached;
+        - There is instance process in IM (im.Status.Instance)
+        - There are other schedulable nodes with healthy volume replica and have replica IM PDB.
+    - Create PBD if volumes are attached; there are instance processes in IM (im.Status.Instance).
+    - Create PBD when volumes are detached;
+        - There is no instance process in IM (im.Status.Instance)
+        - The replica has been started. There are no other schedulable nodes with healthy volume replica and have replica IM PDB.
```
shuo-wu, line 86: How about other corner cases I mentioned before? Maybe we can revisit this after the implementation

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,171 @@
...
+And Set `kubernetes-cluster-autoscaler-enabled` to `true`.
+
+When create deployment with cpu request.
+And trigger CA scale-up 2 nodes by increase deployment replicas.
+And trigger CA scale-down 2 nodes by decrease deployment replicas.
+
+Then CA should scale down 1 node.
+And cluster should have 3 nodes.
+And CA should scale down 1 node.
+And cluster should have 2 nodes.
```
shuo-wu, line 103: The statement here is a little bit confusing...

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: After the internal meeting discussion, this feature will be marked experiential and will skip handling of the CA blocked by backing image manager/share manager pod for now.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: Updated.

enhancements/20220408-support-kubernetes-ca.md
shuo-wu: I mean, the node is already gone/tainted, or the IM or IM pod is already gone. Some cases are already handled in the current implementation. You may need to reorganize them during this implementation.

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,171 @@
...
+And Detach volume.
+
+When create deployment with cpu request.
+And trigger CA scale-up 2 nodes by increase deployment replicas.
+And annotate new nodes with `scale-down-disable` to test when CA tries to scale-down all nodes containing volume replica.
+And trigger CA scale-down 2 nodes by decrease deployment replicas.
+
+Then CA should scale down 1 node.
+And cluster should have 3 nodes.
+And CA scale-down should be blocked by replica instance-manager pod PDB.
```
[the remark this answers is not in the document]
c3y1huang: I am thinking we can delete the engine IM PBD but keep at least 1 replica IM PBD when the volume is detached.

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,171 @@
...
+And Detach volume.
+
+When create deployment with cpu request.
+And trigger CA scale-up 2 nodes by increase deployment replicas.
+And annotate new nodes with `scale-down-disable` to test when CA tries to scale-down all nodes containing volume replica.
+And trigger CA scale-down 2 nodes by decrease deployment replicas.
+
+Then CA should scale down 1 node.
+And cluster should have 3 nodes.
+And CA scale-down should be blocked by share manager local storage mount.
```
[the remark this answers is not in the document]
c3y1huang: Removed.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: Added the implementation to cover node un-schedulable case.

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+- Engine IM PDB:
+    - Delete PDB if volumes are detached;
+        - There is no instance process in IM (im.Status.Instance).
+        - The same logic applies when a node is un-schedulable. Node is un-schedulable when marked in spec or with CA tainted `ToBeDeletedByClusterAutoscaler`;
+    - Create PDB if volumes are attached; there are instance processes in IM (im.Status.Instance).
+
+- Replica IM PDB:
+    - Delete PDB if setting `allow-node-drain-with-last-healthy-replica` is enabled.
+    - Delete PDB if volumes are detached;
+        - There is no engine instance process in IM (im.Status.Instance)
```
shuo-wu, line 94: NIT: When handling the Replica IM PDB, there is no need to consider engine processes.

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+        - There are other schedulable nodes with healthy volume replica and have replica IM PDB.
+    - Delete PDB when a node is un-schedulable. Node is un-schedulable when marked in spec or with CA tainted `ToBeDeletedByClusterAutoscaler`;
+        - Check if the condition is met to delete PDB (same check as to when volumes are detached).
+        - Enqueue the replica instance-manager of another schedulable node with the volume replica.
+        - Delete PBD.
+    - Create PDB if volumes are attached:
+        - There are instance processes in IM (im.Status.Instance).
```
shuo-wu, line 101: If the processes are redundant replica processes, will the replica IM PDB be removed then get recreated immediately?

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+### User Stories
+
+#### CA scaling
+Before the enhancement, CA will be blocked by
+- Pods that are not backed by a controller (engine/replica instance manager).
+- Pods with [local storage volume mounts](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/utils/drain/drain.go#L222) (longhorn-ui, longhorn-csi-plugin, csi-attacher, csi-provisioner, csi-resizer, csi-snapshotter).
+
+After enhancement, the user can set a new global setting `kubernetes-cluster-autoscaler-enabled` to unblock CA scaling.
+
+Longhorn considers a node can be removed when conditions are met, else it will block scale down to avoid potential data loss:
```
innobead, line 48: How about pods w/ local storage volume mounts mentioned above? the nodes they are running on will never be eligible for cluster autoscaler, right?

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/2203
+
+## Motivation
+
+### Goals
+
+- Longhorn should block CA from scaling down if a node met ANY condition:
+    - Volume attached
```
innobead, line 18:
```suggestion
    - Any volume attached
```

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+    - Contain backing image pod
+    - Contain share manager pod
+- Longhorn should not block CA from scaling down if a node met ALL conditions:
+    - Volume detached and can find another schedulable node with volume replica and replica IM PDB.
```
innobead, line 22:
```suggestion
    - All volume detached, replicas of each detailed volume should be replicated to other schedulable nodes protected by replica IM PDB.
```

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+    - Not contain a backing image pod
```
innobead, line 23:
```suggestion
    - Not contain a backing image manager pod
```

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+Before the enhancement, CA will be blocked by
+- Pods that are not backed by a controller (engine/replica instance manager).
+- Pods with [local storage volume mounts](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/utils/drain/drain.go#L222) (longhorn-ui, longhorn-csi-plugin, csi-attacher, csi-provisioner, csi-resizer, csi-snapshotter).
+
+After enhancement, the user can set a new global setting `kubernetes-cluster-autoscaler-enabled` to unblock CA scaling.
+
+Longhorn considers a node can be removed when conditions are met, else it will block scale down to avoid potential data loss:
+- all volumes on the node are detached and there is an available healthy replica on other nodes.
+- backing image manager not running on the node.
+- share-manager node not running on the node.
```
innobead, line 51: You can reuse the goals you mentioned above because these are the same.

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+#### Global setting
+- Add new global setting `Kubernetes Cluster Autoscaler Enabled (Experimental)`.
+  - The setting is `boolean`.
+  - The default value is `false`.
+
+#### Annotations
+
+When setting `kubernetes-cluster-autoscaler-enabled` to `true`:
+- Annotate engine and replica instance-manager pods when no IM not in state stop/error.
+- Annotate managed deployments with `cluster-autoscaler.kubernetes.io/safe-to-evict`. Managed deployments are labeled with `longhorn.io/managed-by: longhorn-manager` (longhorn-ui, longhorn-csi-plugin, csi-attacher, csi-provisioner, csi-resizer, and csi-snapshotter).
```
innobead, line 79:
```suggestion
- Annotate managed deployments with local storage volume mounts with `cluster-autoscaler.kubernetes.io/safe-to-evict`. Managed deployments are labeled with `longhorn.io/managed-by: longhorn-manager` (longhorn-ui, longhorn-csi-plugin, csi-attacher, csi-provisioner, csi-resizer, and csi-snapshotter).
```

enhancements/20220408-support-kubernetes-ca.md
innobead: This seems like a side-effect caused by the redundant replicas, so probably we should not tackle this here?

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+    And Attach the volume.
+    And Write some data to volume.
+    And Detach the volume.
+    And Create deployment with cpu request.
+
+    When Trigger CA to scale-up by increase deployment replicas.
+         (double the node number, not including host node)
+    Then Cluster should have double the node number.
+
+    When Annotate new nodes with `scale-down-disable`.
```
innobead, line 138: Do you mean `cluster-autoscaler.kubernetes.io/scale-down-disabled`? It's better to describe clearly.

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+        - The replica has been started. There are no other schedulable nodes with healthy volume replica and have replica IM PDB.
+
+### Test plan
+
+#### Scenario: test CA
+
+    Given Cluster with Kubernetes cluster-autoscaler.
+    And Longhorn installed.
+    And Set `kubernetes-cluster-autoscaler-enabled` to `true`.
+    And Create deployment with cpu request.
```
innobead, line 113: Is the deployment having a volume?

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+
+    When Trigger CA to scale-up by increase deployment replicas.
```
innobead, line 115: Do you mean the deployment replicas w/ node affinity to enforce a new node is needed?

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+        (original node number)
+    Then Cluster should have original node number + 1 blocked node.
+
+    When Attach the volume to a new node. This triggers replica rebuild.
+    And Volume data should be the same.
+    And Detach the volume.
+    Then Cluster should have original node number.
+    And Volume data should be the same.
+
+#### Scenario: test CA should block scale down of node running backing image manager pod
```
innobead, line 150:
```suggestion
#### Scenario: test CA should block scale down of node running backing image manager/share manager pod
```

innobead (commented):
In general, LGTM. However, some comments needed to be resolved/answered. 

Just ensure we will make this an experimental feature in 1.3.0.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: Fixed.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: Reworded.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: Fixed.

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+
+https://github.com/longhorn/longhorn/issues/2203
+
+## Motivation
+
+### Goals
+
+- Longhorn should block CA from scaling down if a node met ANY condition:
+    - Volume attached
+    - Contain backing image pod
```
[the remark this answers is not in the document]
c3y1huang: Fixed.

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+    - Contain share manager pod
```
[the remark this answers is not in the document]
c3y1huang: Fixed.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: Hmm..the suggested change looks like we will rebuild the replicas to another node. 

This is a condition statement to ensure the last replica of the detached volume is protected and we do not rebuild replicas to another node.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: Fixed.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: Maybe we don't need to duplicate this, I've removed this block.

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+- backing image manager not running on the node.
+- share-manager node not running on the node.
+
+
+### User Experience In Detail
+
+- Configure the setting via Longhorn UI or kubectl.
+- Ensure all volume replica count is set to more than 1.
+- CA is not blocked by Longhorn components when the node doesn't contain volume replica, backing image manager pod, and share manager pod.
+    - Engine/Replica instance-managerPDB will block the node if the volume is attached.
```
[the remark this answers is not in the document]
c3y1huang: Fixed.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: Fixed.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: Fixed.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: The deployment does not need volume, this is used to trigger CA scaling and will contain CPU requests.
```
        resources:
          limits:
            cpu: 300m
            memory: 30Mi
          requests:
            cpu: 150m
            memory: 15Mi
```

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: The deployment does not need node affinity. We will increase the deployment replica numbers and let it consume more CPU resources than the cluster nodes can provide. This triggers CA to scale up and vice versa.

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+         (this ensures scale-down only the old nodes)
+    And Trigger CA to scale-down by decrease deployment replicas.
+        (original node number)
+    Then Cluster should have original node number + 1 blocked node.
+
+    When Attach the volume to a new node. This triggers replica rebuild.
+    And Volume data should be the same.
+    And Detach the volume.
+    Then Cluster should have original node number.
+    And Volume data should be the same.
```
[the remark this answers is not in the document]
c3y1huang: Are you referring to `And Volume data should be the same.`?
This is followed by the `Then` statement above.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: https://github.com/longhorn/longhorn/pull/3831#discussion_r847911269

enhancements/20220408-support-kubernetes-ca.md
shuo-wu: I mean, the replica on the node is not the only healthy one of a volume. In the above deletion logic, the PDB will be removed since this kind of replica won't prevent the deletion. But the controller will recreate the PDB immediately based on this line.

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+
+#### Global setting
+- Add new global setting `Kubernetes Cluster Autoscaler Enabled (Experimental)`.
+  - The setting is `boolean`.
+  - The default value is `false`.
+
+#### Annotations
+
+When setting `kubernetes-cluster-autoscaler-enabled` to `true`:
+- Annotate engine and replica instance-manager pods when no IM not in state stop/error.
```
shuo-wu, line 78: I am confused about this. Why cannot we annotate the IM pods when the IM is state stop or error? As long as the PDBs are handled correctly, I don't think we need to consider the pods.

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+### Implementation Overview
+
+#### Global setting
+- Add new global setting `Kubernetes Cluster Autoscaler Enabled (Experimental)`.
+  - The setting is `boolean`.
+  - The default value is `false`.
+
+#### Annotations
+
+When setting `kubernetes-cluster-autoscaler-enabled` to `true`:
```
shuo-wu, line 77:
```suggestion
When setting `kubernetes-cluster-autoscaler-enabled` is `true`, Longhorn will add annotation `cluster-autoscaler.kubernetes.io/safe-to-evict` for the following pods:
```

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,160 @@
...
+- Annotate engine and replica instance-manager pods when no IM not in state stop/error.
+- Annotate managed deployments using local storage volume mounts with `cluster-autoscaler.kubernetes.io/safe-to-evict`. Managed deployments are labeled with `longhorn.io/managed-by: longhorn-manager` (longhorn-ui, longhorn-csi-plugin, csi-attacher, csi-provisioner, csi-resizer, and csi-snapshotter).
```
shuo-wu, line 79:
```suggestion
The workloads that are managed by longhorn-manager and contain local storage. The managed components are labeled with ......
```
BTW, do we need to handle longhorn-manager?

shuo-wu (commented):
BTW, you can add TODOs for backing image manager and share manager handling.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: We will only remove the PDB of the detached volume, and only if there is another node with the same volume replica IM PBD. In case all replicas are on the same node of the detached volume, then all replica IM PDB will get created and none will get removed.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: > I am confused about this. Why cannot we annotate the IM pods when the IM is state stop or error? 

We can probably disregard the IM state. Updated.

> As long as the PDBs are handled correctly, I don't think we need to consider the pods.

We need to consider the pods because they are bare pods and also using local storage mounts that will not pass the CA pre-check.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: Updated.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: Updated.

c3y1huang: > BTW, do we need to handle longhorn-manager?

No. CA pre-check bypasses DaemonSets.

enhancements/20220408-support-kubernetes-ca.md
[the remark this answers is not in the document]
shuo-wu: I think what happens here is:
1. Volume rebuilds new replicas on other nodes.
2. The CA finally evicts the blocking node which previously contains the only replica. This means the cluster node number should be the same as that at the beginning.
3. The volume data is still intact after the rebuilding & eviction.

You may need to explicitly mean the 2nd behavior & result.

enhancements/20220408-support-kubernetes-ca.md
shuo-wu: You can put this example here.

enhancements/20220408-support-kubernetes-ca.md
shuo-wu: the PDB of the detached volume?? another node with the same volume replica IM PBD?? Sorry I cannot follow...

shuo-wu: I think we can describe this part from another perspective: The IM PDB will be created **and retained** only when it's required. And once it's required, we cannot delete the PDB. Required means either of the following happen:
  1. There are volumes/engines running on the node. We need to guarantee that the volumes won't crash
  2. The only available/valid replica of a volume is on the node, no matter whether the volume itself is detached or not. Here we need to prevent the volume data from being lost.

c3y1huang: 1. The behavior of creating engine/replica IM PBD  is the same as now when there are volumes attached.
2. We keep only the replica IM PBD of 1 of the nodes when the volume is detached.

> I think we can describe this part from another perspective: The IM PDB will be created **and retained** only when it's required. And once it's required, we cannot delete the PDB. Required means either of the following happen:
> 
> 1. There are volumes/engines running on the node. We need to guarantee that the volumes won't crash
> 2. The only available/valid replica of a volume is on the node, no matter whether the volume itself is detached or not. Here we need to prevent the volume data from being lost.

The implementation overview is written based on scenarios, and I think we are in the same direction.

I've put your suggestion of the description under the proposal section to clarify what we want to achieve.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: Updated.

shuo-wu: Should the CPU request be large enough so that the current nodes cannot fit the requirement?

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,174 @@
...
+- The only available/valid replica of a volume is on the node. Here we need to prevent the volume data from being lost.
+
+### User Stories
+
+#### CA scaling
+Before the enhancement, CA will be blocked by
+- Pods that are not backed by a controller (engine/replica instance manager).
+- Pods with [local storage volume mounts](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/utils/drain/drain.go#L222) (longhorn-ui, longhorn-csi-plugin, csi-attacher, csi-provisioner, csi-resizer, csi-snapshotter).
+
+After enhancement, instance manager PBD will be actively managed by Longhorn:
```
shuo-wu, line 51: typo `PBD` -> `PDB`. Same below

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: We can calculate the number of deployment replicas with the allocatable CPU.
```
10 * math.ceil(allocatable_millicpu/cpu_request*node_number/10)
```

shuo-wu: NIT: The calculation result is for increasing 1-node scaling-up,  right? You can add this expectation there.

shuo-wu (commented):
In general LGTM except for the existing comments left by David.

innobead (approved):
LGTM

enhancements/20220408-support-kubernetes-ca.md
```diff
@@ -0,0 +1,177 @@
...
+    - Create PDB if volumes are attached; there are instance processes in IM (im.Status.Instance).
+
+- Replica IM PDB:
+    - Delete PDB if setting `allow-node-drain-with-last-healthy-replica` is enabled.
+    - Delete PDB if volumes are detached;
+        - There is no instance process in IM (im.Status.Instance)
+        - There are other schedulable nodes with healthy volume replica and have replica IM PDB.
+    - Delete PDB when a node is un-schedulable. Node is un-schedulable when marked in spec or with CA tainted `ToBeDeletedByClusterAutoscaler`;
+        - Check if the condition is met to delete PDB (same check as to when volumes are detached).
+        - Enqueue the replica instance-manager of another schedulable node with the volume replica.
```
innobead, line 103: Ensure not Enqueue.

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: The expectation is to double the cluster node number and I have it with the following `When Trigger CA to scale-up by increase deployment replicas.`

enhancements/20220408-support-kubernetes-ca.md
c3y1huang: We want to enqueue another schedulable replica IM to create the PDB when the current is un-schedulable.

innobead: Thanks, it's clear now, but it was not easy to understand a bit.

The point is to ensure at least one healthy replica is still available in another node when the volume is detached and the volume owner node is being eligible for CA by removing it's PDB.

Issue #3860, opened 2022-04-18
[BUG] PVC create from volumsanpshot stuck at pending status(type=snap)


cchien816:
## Describe the bug

After add volumesnapshotclass(type=snap) and create volumesnapshot, create PVC from volumesnapshot, the PVC will stuck at pending status

master-head 0418(1c261cab59443013b398c4a99889661cfe2e6fe5)

related to https://github.com/longhorn/longhorn/issues/2534

## To Reproduce

1. Enable [csi snapshot support](https://longhorn.io/docs/1.2.4/snapshots-and-backups/csi-snapshot-support/enable-csi-snapshot-support/)  
2. Create Longhorn volume test-vol of 5GB. Create PV/PVC for the Longhorn volume. Attach the volume
3. Deploy snapshotclass (type=snap)
```yaml
kind: VolumeSnapshotClass
apiVersion: snapshot.storage.k8s.io/v1beta1
metadata:
  name: longhorn-snapshot
driver: driver.longhorn.io
deletionPolicy: Delete
parameters:
  type: snap

```
4. Create volumesnapshot
```yaml
apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshot
metadata:
  name: test-snapshot-longhorn-snapshot
spec:
  volumeSnapshotClassName: longhorn-snapshot
  source:
    persistentVolumeClaimName: test-vol
```
5. Create PVC from volumesnapshot
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: test-restore-pvc
spec:
  storageClassName: longhorn
  dataSource:
    name: test-snapshot-longhorn-snapshot
    kind: VolumeSnapshot
    apiGroup: snapshot.storage.k8s.io
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 5Gi
```
6. Will see PVC `test-restore-pvc` stuck at pending status

## Expected behavior

PVC create from volumesnapshot should generated and bring up a PV

## Log or Support bundle

[longhorn-support-bundle_ab3e8554-0924-4c3f-a433-5e8774cceeb7_2022-04-18T01-29-04Z.zip](https://github.com/longhorn/longhorn/files/8502872/longhorn-support-bundle_ab3e8554-0924-4c3f-a433-5e8774cceeb7_2022-04-18T01-29-04Z.zip)

```
[...]

PR #3863, opened 2022-04-18
Fix volumeMounts missing in admission webhook

derekbit:
[Longhorn 3860](https://github.com/longhorn/longhorn/issues/3860)


innobead:
Closing, we don't need this anymore because moving privileged operations back to longhorn-manager instead of admission controllers.

Issue #3144, opened 2021-10-12
[FEATURE] CRD for snapshots


HubbeKing:
**Is your feature request related to a problem? Please describe.**
Snapshot RecurringJobs set to Retain X number of snapshots do not touch unrelated snapshots, so if one ever changes the name of the RecurringJob, the old snapshots will stick around forever. These then have to be manually deleted in the UI.
Having a CRD for snapshots would greatly simplify this, as one could prune snapshots using kubectl - much like how one can currently manage backups using kubectl due to the existance of the backups.longhorn.io CRD.

**Describe the solution you'd like**
A CRD for Snapshots. Something like snapshots.longhorn.io, and code in longhorn-manager to ensure that there exists a CRD object for each snapshot for each volume, and that they are in sync.

**Describe alternatives you've considered**
I suppose a browser automation framework might also work for pruning large numbers of snapshots, but this feels janky as all hell.

**Additional context**
Screenshot of v1.2.2 UI showing leftover snapshots from a deleted recurringjob, and a single snapshot from a new RecurringJob with retain set to 4:
https://i.imgur.com/SnlP23P.png

jenting:
I remember we discussed internally that Longhorn needs the snapshot CR. cc @joshimoo @shuo-wu

PhanLe1010:
## Test Plan:
### Innit
1. Create volume: `vol1`, `vol2`
1. Attach and write some data to it

### Create 
1. Create new snapshot CR, volume attached 
    1. Deloy the yaml
       ```
       apiVersion: longhorn.io/v1beta2
       kind: Snapshot
       metadata:
         name: snap001
         namespace: longhorn-system
       spec:
         volume: vol1
         createSnapshot: true    
       ```
    1. Verify that Longhorn provison a new snapshot `snap001` inside Longhorn UI
    1. Run `kubectl get snapshots -n longhorn-system snap001`. Verify the `status` field of the snapshot CR is good (all fields make sense)
    1. Run `kubectl describe snapshots -n longhorn-system snap001`. Verify that the related events sequence is good
1. Create new snapshot CR, volume detached
    1. Detach `vol1`
    1. Deloy the yaml
       ```
       apiVersion: longhorn.io/v1beta2
       kind: Snapshot
       metadata:
         name: snap002
         namespace: longhorn-system
       spec:
         volume: vol1
         createSnapshot: true    
       ```
[...]

PR #3876, opened 2022-04-20
Add LEP for Longhorn snapshot CRD feature

PhanLe1010:
longhorn/longhorn#3144


enhancements/20220420-longhorn-snapshot-crd.md
```diff
@@ -0,0 +1,160 @@
...
+## Design
+
+### Implementation Overview
+
+Introduce a new CRD, snapshot CRD and the snapshot controller.
+The snapshot CRD is:
+
+```yaml
+// SnapshotSpec defines the desired state of Longhorn Snapshot
+  type SnapshotSpec struct {
```
innobead, line 60: Will we allow users to update the snapshot resource after creation? If it's immutable after the creation, suggest describing it somewhere and it will be protected by the admission webhook.

enhancements/20220420-longhorn-snapshot-crd.md
```diff
@@ -0,0 +1,160 @@
...
+1. Controller workqueue
+    1. Disable resync period for snapshot informer
+    1. Enque snapshot only when:
+        1. There is a change in snapshot CR
+        1. There is a change in `engine.Status.CurrentState` (volume attach/detach event), `engine.Status.PurgeStatus` (for snapshot deletion event), `engine.Status.Snapshots` (for snapshot creation/update event)
+1. This enhancement proposal doesn't make additional call to engine process comparing to the existing design.
+
+## Todo
+
+I am still not sure how to handle the special snapshot `volume-head`. Should we create a snapshot CR for this special snapshot?
```
innobead, line 156: Probably we should because it's still a type of snapshot, this can make the consistent user experience from different interfaces (UI, CR, or even the current engine status). 

To avoid operations on the volume head, we should just leverage the admission webhook to protect.

enhancements/20220420-longhorn-snapshot-crd.md
```diff
@@ -0,0 +1,160 @@
...
+
+1. **Create**
+    1. When a snapshot CR is created, Longhorn mutation webhook will:
+        1. Add a volume label `longhornvolume: <VOLUME-NAME>` to the snapshot CR. This allow us to efficiently find snapshots corressponding to a volume without having listing potientially thoundsands of snapshots.
+        1. Add `longhornFinalizerKey` to snapshot CR to prevent it from being removed before Longhorn has change to clean up the corresponding snapshot
+        1. Populate the value for `Spec.VolumeRef` to uniquely indentify the volume of this snapshot. This field contains the volume UID to uniquely identify the volume in case  the old volume was deleted and a new volume was created with the same name.
+    1. For user createed snapshot CR, the field `Spec.CreateSnapshot` should be set to `true` indicating that Longhorn should provision a new snapshot for this CR.
+        1. Longhorn snapshot controller will pick up this CR, check to see if there already is a snapshot inside the `engine.Status.Snapshots`.
+        1. If there is there already a snapshot inside engine.Status.Snapshots, update the snapshot.Status with the snapshot info inside `engine.Status.Snapshots`
```
innobead, line 115: So we will not have an upgrade path to create those snapshot resources for existing snapshots, but rather rely on the controller to do the asynchronous creation?

enhancements/20220420-longhorn-snapshot-crd.md
```diff
@@ -0,0 +1,160 @@
...
+    1. For the snapshots that are already exist inside `engine.Status.Snapshots` but doesn't have corresponding snapshot CRs (i.g., system generated snapshots), the engine monitoring will generate snapshot CRs for them. The snapshot CR generated by engine monitoring with have `Spec.CreateSnapshot` set to `false`, Longhorn snapshot controller will not create a snapshot for those CRs. The snapshot controller only sync status for those snapshot CRs
+1. **Update**
+    1. Sync the snapshot infor from `engine.Status.Snapshots` to the `snapshot.Status`.
+    1. If there is any error or if the snapshot is marked as removed, set `snapshot.Status.ReadyToUse` to `false`
+    1. If there there is no snapshot info inside `engine.Status.Snapshots`, mark the `snapshot.Status.ReadyToUse` to `false`and populate the `snapshot.Status.Error` with the lost message. This snapshot will eventually be updated again when engine monitoring update `engine.Status.Snapshots` or it may be cleanup as the section below
+1. **Delete**
+    1. Engine monitor will responsible for removing all snapshot CR it auto generated, i.e., the snapshot CRs with `Spec.CreateSnapshot: false`.
```
innobead, line 126: `Spec.CreateSnapshot: false` means system-generated or existing user-created/system-generated snapshots before the upgrade? If existing user-created snapshots, we will still allow the engine monitor to be responsible for deletion? 

Should we differentiate `Spec.CreateSnapshot` for these cases?

enhancements/20220420-longhorn-snapshot-crd.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Upgrade strategy
+
+Anything that requires if user want to upgrade to this enhancement
+
+## Note [optional]
+
+How do we address scalability issue?
+1. Controller workqueue
+    1. Disable resync period for snapshot informer
```
innobead, line 148: YES, we should disable periodical resync for this case, otherwise, it will probably be a storm.

innobead (commented):
In general, LGTM. A few questions left. 

Need @shuo-wu check to see if anything missing or concerns.

enhancements/20220420-longhorn-snapshot-crd.md
```diff
@@ -0,0 +1,160 @@
+
+# Longhorn Snapshot CRD
+
+## Summary
+
+Supporting Longhorn snapshot CRD allows users to query/create/delete volume snapshots using kubectl. This is one step closer to making kubectl as Longhorn CLI. Also, this will be a bulding block for the future auto-attachment/auto-detachment refactoring for snapshot creation, deletion, volume cloning.
```
weihanglo, line 6:
```suggestion
Supporting Longhorn snapshot CRD allows users to query/create/delete volume snapshots using kubectl. This is one step closer to making kubectl as Longhorn CLI. Also, this will be a building block for the future auto-attachment/auto-detachment refactoring for snapshot creation, deletion, volume cloning.
```

enhancements/20220420-longhorn-snapshot-crd.md
```diff
@@ -0,0 +1,160 @@
...
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/3144
+
+## Motivation
+
+### Goals
+
+1. Support Longhorn snapshot CRD to allow users to query/create/delete volume snapshots using kubectl.
+2. A bulding block for the future auto-attachment/auto-detachment refactoring for snapshot creation, deletion, volume cloning.
```
weihanglo, line 17:
```suggestion
2. A building block for the future auto-attachment/auto-detachment refactoring for snapshot creation, deletion, volume cloning.
```

derekbit:
A bit of curious.
While calling the snapshot purge command, how does the purge work with the snapshot CR?

enhancements/20220420-longhorn-snapshot-crd.md
PhanLe1010: Thank you. Added the info

enhancements/20220420-longhorn-snapshot-crd.md
PhanLe1010: Yeah, you are correct

enhancements/20220420-longhorn-snapshot-crd.md
PhanLe1010: Sorry for the confusion. I have updated the LEP

`Spec.CreateSnapshot: false` means that the snapshot CR is created by the engine monitor (both user-created and system-generated snapshots). Snapshot controller doesn't need to provision a new snapshot for this Snapshot CR. The design is that when engine monitoring sees a snapshot that doesn't have the corresponding snapshot CR, it will create a snapshot CR. Then when the engine monitoring sees a snapshot CR that doesn't have a corresponding snapshot info, it will remove the snapshot CR

enhancements/20220420-longhorn-snapshot-crd.md
PhanLe1010: Thank you. Fixed

enhancements/20220420-longhorn-snapshot-crd.md
PhanLe1010: Thank you. Fixed

enhancements/20220420-longhorn-snapshot-crd.md
PhanLe1010: After discussion with Shuo, we think that there is not much benefit of populating volume-head snapshot CR in this release as a user cannot delete/update it. All snapshot info is still fetched from engine status in this version. We don't want to have disruptive changes in this initial version of snapshot CR (e.g., snapshot API create/delete shouldn't change. Snapshot status is still inside the engine status). 

We can wait for the snapshot CRD to be a bit mature (no issue with scalability) and make the disruptive changes in the next version of snapshot CR (e.g., snapshot API create/delete changes to create/delete snapshot CRs. Snapshot status is removed from inside the engine status). At this point, we will populate the volume-head snapshot. What do you think @innobead ?

PhanLe1010:
> While calling the snapshot purge command, how will the purge work with the snapshot CR?

Could you clarify a bit about `calling the snapshot purge command`? Do you mean user make the snapshotPurge API call?

derekbit:
> > While calling the snapshot purge command, how will the purge work with the snapshot CR?
> 
> Could you clarify a bit about `calling the snapshot purge command`? Do you mean user make the snapshotPurge API call?

Yes. IIUC, currently, the snapshotPurge API calls the engine to do purge the snapshot directly. Will the snapshot CR be updated or deleted after the call?

enhancements/20220420-longhorn-snapshot-crd.md
```diff
@@ -0,0 +1,161 @@
...
+}
+```
+The life cycle of a snapshot CR is as below:
+
+1. **Create**
+    1. When a snapshot CR is created, Longhorn mutation webhook will:
+        1. Add a volume label `longhornvolume: <VOLUME-NAME>` to the snapshot CR. This allow us to efficiently find snapshots corressponding to a volume without having listing potientially thoundsands of snapshots.
+        1. Add `longhornFinalizerKey` to snapshot CR to prevent it from being removed before Longhorn has change to clean up the corresponding snapshot
+        1. Populate the value for `snapshot.OwnerReferences` to uniquely indentify the volume of this snapshot. This field contains the volume UID to uniquely identify the volume in case  the old volume was deleted and a new volume was created with the same name.
+    2. For user createed snapshot CR, the field `Spec.CreateSnapshot` should be set to `true` indicating that Longhorn should provision a new snapshot for this CR.
```
derekbit, line 112:
```suggestion
    2. For user created snapshot CR, the field `Spec.CreateSnapshot` should be set to `true` indicating that Longhorn should provision a new snapshot for this CR.
```

PhanLe1010:
@derekbit 

I think I got your concern about the snapshot purge call. Are you thinking about how snapshot controller make the snapshot purge calls?

Right now, then users delete a snapshot CR, snapshot controller will:
* delete the snapshot by exec the engine binary with snapshot delete command
* Do snapshot purge by exec the engine binary with snapshot purge command (Are you worried about this part?)

Please see https://github.com/longhorn/longhorn-manager/pull/1304 for more detail

enhancements/20220420-longhorn-snapshot-crd.md
```diff
@@ -0,0 +1,161 @@
...
+           1. If there is there already a snapshot inside engine.Status.Snapshots, update the snapshot.Status with the snapshot info inside `engine.Status.Snapshots`
+           2. If there isn't a snapshot inside `engine.Status.Snapshots` then:
+               1.  making a call to engine process to check if there already a snapshot with the same name. This is to make sure we don't accidentally create 2 snapshots with the same name. This logic can be remove after [the issue](https://github.com/longhorn/longhorn/issues/3844) is resolved
+               1. If the snapshot doesn't inside the engine process, make another call to create the snapshot
+    3. For the snapshots that are already exist inside `engine.Status.Snapshots` but doesn't have corresponding snapshot CRs (i.g., system generated snapshots), the engine monitoring will generate snapshot CRs for them. The snapshot CR generated by engine monitoring with have `Spec.CreateSnapshot` set to `false`, Longhorn snapshot controller will not create a snapshot for those CRs. The snapshot controller only sync status for those snapshot CRs
```
derekbit, line 118: i.e. or e.g?
Does it mean `Spec.CreateSnapshot=false` is only for the system generated snapshot?

```suggestion
    3. For the snapshots that are already exist inside `engine.Status.Snapshots` but doesn't have corresponding snapshot CRs (i.e., system generated snapshots), the engine monitoring will generate snapshot CRs for them. The snapshot CR generated by engine monitoring with have `Spec.CreateSnapshot` set to `false`, Longhorn snapshot controller will not create a snapshot for those CRs. The snapshot controller only sync status for those snapshot CRs
```

derekbit:
> @derekbit
> 
> I think I got your concern about the snapshot purge call. Are you thinking about how snapshot controller make the snapshot purge calls?
> 
> Right now, then users delete a snapshot CR, snapshot controller will:
> 
> * delete the snapshot by exec the engine binary with snapshot delete command
> * Do snapshot purge by exec the engine binary with snapshot purge command (Are you worried about this part?)
> 
> Please see [longhorn/longhorn-manager#1304](https://github.com/longhorn/longhorn-manager/pull/1304) for more detail

Yep, that's my question. Thanks for your explanation. I will look into the details.

enhancements/20220420-longhorn-snapshot-crd.md
PhanLe1010: Thank you. Fixed

enhancements/20220420-longhorn-snapshot-crd.md
PhanLe1010: You are right. It should be `i.e.,` because `Spec.CreateSnapshot=false` is NOT only for the system generated snapshot

enhancements/20220420-longhorn-snapshot-crd.md
```diff
@@ -0,0 +1,161 @@
...
+    1. Enque snapshot only when:
+        1. There is a change in snapshot CR
+        1. There is a change in `engine.Status.CurrentState` (volume attach/detach event), `engine.Status.PurgeStatus` (for snapshot deletion event), `engine.Status.Snapshots` (for snapshot creation/update event)
+1. This enhancement proposal doesn't make additional call to engine process comparing to the existing design.
+
+## Todo
+
+I am still not sure how to handle the special snapshot `volume-head`. Should we create a snapshot CR for this special snapshot?
+1. From the usecase perspective, user cannot delete this snapshot anyway so there is no need to generate this snapshot
+1. The name `volume-head` is not globally uniquely, we might have to include volume name if we want to generate this snapshot CR
```
shuo-wu, line 159: Besides, the `VolumeHead` CR may suddenly point to another actual file during the snapshot creation. I am not sure if there will be any race conditions.

enhancements/20220420-longhorn-snapshot-crd.md
```diff
@@ -0,0 +1,161 @@
...
+    3. For the snapshots that are already exist inside `engine.Status.Snapshots` but doesn't have corresponding snapshot CRs (i.e., system generated snapshots), the engine monitoring will generate snapshot CRs for them. The snapshot CR generated by engine monitoring with have `Spec.CreateSnapshot` set to `false`, Longhorn snapshot controller will not create a snapshot for those CRs. The snapshot controller only sync status for those snapshot CRs
+2. **Update**
+    1. Snapshot CR spec and label are immutable after creation. It will be protected by the admission webhook
+    2. Sync the snapshot info from `engine.Status.Snapshots` to the `snapshot.Status`.
+    3. If there is any error or if the snapshot is marked as removed, set `snapshot.Status.ReadyToUse` to `false`
+    4. If there there is no snapshot info inside `engine.Status.Snapshots`, mark the `snapshot.Status.ReadyToUse` to `false`and populate the `snapshot.Status.Error` with the lost message. This snapshot will eventually be updated again when engine monitoring update `engine.Status.Snapshots` or it may be cleanup as the section below
```
shuo-wu, line 123: Can we use a `Condition` field instead?

enhancements/20220420-longhorn-snapshot-crd.md
```diff
@@ -0,0 +1,161 @@
...
+Anything that requires if user want to upgrade to this enhancement
+
+## Note [optional]
+
+How do we address scalability issue?
+1. Controller workqueue
+    1. Disable resync period for snapshot informer
+    1. Enque snapshot only when:
+        1. There is a change in snapshot CR
+        1. There is a change in `engine.Status.CurrentState` (volume attach/detach event), `engine.Status.PurgeStatus` (for snapshot deletion event), `engine.Status.Snapshots` (for snapshot creation/update event)
```
shuo-wu, line 152: Do you mean that you will check and compare the status before enqueuing?
`engine.Status.CurrentSize` will be modified frequently. If we don't ignore it, it may overwhelm the work queue of the SnapshotController.

shuo-wu:
@PhanLe1010 Remember to check this PR.

innobead:
This has been introduced in 1.3.0, but some comments from @shuo-wu have not been resolved yet.

@PhanLe1010 do you have anything you would like to update before merging?

PhanLe1010:
@innobead I forgot this PR. I will review it and push the update

enhancements/20220420-longhorn-snapshot-crd.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Goals
+
+1. Support Longhorn snapshot CRD to allow users to query/create/delete volume snapshots using kubectl.
+2. A building block for the future auto-attachment/auto-detachment refactoring for snapshot creation, deletion, volume cloning.
+3. Pay attention to scalability problem. A cluster with 1k volumes might have 30k snapshots. We should make sure not to overload the controller work-queue as well as making too many grpc calls to engine processes.
+
+## Proposal
+
+Introduce a new CRD, snapshot CRD and the snapshot controller. The life cycle of a snapshot CR is as below:
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. Fixed

enhancements/20220420-longhorn-snapshot-crd.md
```diff
@@ -0,0 +1,161 @@
...
+  // +optional
+  CreateSnapshot bool `json:"createSnapshot"`
+  // The labels of snapshot
+  // +optional
+  Labels map[string]string `json:"labels"`
+}
+
+  // SnapshotStatus defines the observed state of Longhorn Snapshot
+  type SnapshotStatus struct {
+  // The node ID on which the controller is responsible to reconcile this backup CR.
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. Fixed

enhancements/20220420-longhorn-snapshot-crd.md
```diff
@@ -0,0 +1,161 @@
...
+```yaml
+// SnapshotSpec defines the desired state of Longhorn Snapshot
+  type SnapshotSpec struct {
+  // the volume that this snapshot belongs to.
+  // This field is immutable after creation.
+  // Required
+  Volume string `json:"volume"`
+  // require creating a new snapshot
+  // +optional
+  CreateSnapshot bool `json:"createSnapshot"`
```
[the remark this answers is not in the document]
PhanLe1010: We already merged the code so keeping the LEP as it is

enhancements/20220420-longhorn-snapshot-crd.md
```diff
@@ -0,0 +1,161 @@
...
+        1. Add a volume label `longhornvolume: <VOLUME-NAME>` to the snapshot CR. This allow us to efficiently find snapshots corressponding to a volume without having listing potientially thoundsands of snapshots.
+        1. Add `longhornFinalizerKey` to snapshot CR to prevent it from being removed before Longhorn has change to clean up the corresponding snapshot
+        1. Populate the value for `snapshot.OwnerReferences` to uniquely indentify the volume of this snapshot. This field contains the volume UID to uniquely identify the volume in case  the old volume was deleted and a new volume was created with the same name.
+    2. For user created snapshot CR, the field `Spec.CreateSnapshot` should be set to `true` indicating that Longhorn should provision a new snapshot for this CR.
+        1. Longhorn snapshot controller will pick up this CR, check to see if there already is a snapshot inside the `engine.Status.Snapshots`.
+           1. If there is there already a snapshot inside engine.Status.Snapshots, update the snapshot.Status with the snapshot info inside `engine.Status.Snapshots`
+           2. If there isn't a snapshot inside `engine.Status.Snapshots` then:
```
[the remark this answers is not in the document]
PhanLe1010: Since we actually reach out to the engine process to get the info of the snapshot, we know for sure whether the actual snapshot file exists or not. Does it answer the concern?

enhancements/20220420-longhorn-snapshot-crd.md
PhanLe1010: We already merged the code so keeping the LEP as it is

enhancements/20220420-longhorn-snapshot-crd.md
PhanLe1010: We only enqueue when There is a change in `engine.Status.CurrentState` (volume attach/detach event), `engine.Status.PurgeStatus` (for snapshot deletion event), `engine.Status.Snapshots so there are not a lot of enqueuing

enhancements/20220420-longhorn-snapshot-crd.md
PhanLe1010: Thank you. Updated the LEP to not creating snapshot CR for volume-head

PhanLe1010:
Sorry, I forgot this PR. Since GitHub doesn't allow me to re-open this PR, creating a new one at https://github.com/longhorn/longhorn/pull/4743

PR #3879, opened 2022-04-21
Update deployment for cluster autoscaler support

c3y1huang:
https://github.com/longhorn/longhorn/issues/2203

### DIFF
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -499,10 +499,10 @@ WARNING:
     default: 12
 - variable: defaultSettings.kubernetesClusterAutoscalerEnabled
     label: Kubernetes Cluster Autoscaler Enabled (Experimental)
-    description: "Enable this setting will notify Longhorn that the cluster is using Kubernetes cluster-autoscaler.
-    Longhorn prevents data loss by only allowing CA to scale down a node that met all conditions:
-      - No volume attached
-      - Is not the last node containing the replica of a detached volume.
+    description: "Enabling this setting will notify Longhorn that the cluster is using Kubernetes Cluster Autoscaler.
+    Longhorn prevents data loss by only allowing the Cluster Autoscaler to scale down a node that met all conditions:
+      - No volume attached to the node.
+      - Is not the last node containing the replica of any volume.
       - Is not running backing image components pod.
       - Is not running share manager components pod."
     group: "Longhorn Default Settings"
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -497,6 +497,17 @@ WARNING:
     min: 0
     max: 40
     default: 12
+- variable: defaultSettings.kubernetesClusterAutoscalerEnabled
+    label: Kubernetes Cluster Autoscaler Enabled (Experimental)
+    description: "Enable this setting will notify Longhorn that the cluster is using Kubernetes cluster-autoscaler.
+    Longhorn prevents data loss by only allowing CA to scale down a node that met all conditions:
+      - No volume attached
+      - Is not the last node containing the replica of a detached volume.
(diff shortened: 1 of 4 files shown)


chart/questions.yaml
```diff
@@ -497,6 +497,17 @@ WARNING:
     min: 0
     max: 40
     default: 12
+- variable: defaultSettings.kubernetesClusterAutoscalerEnabled
+    label: Kubernetes Cluster Autoscaler Enabled (Experimental)
+    description: "Enable this setting will notify Longhorn that the cluster is using Kubernetes cluster-autoscaler.
```
innobead, line 502:
```suggestion
    description: "Enabling this setting will notify Longhorn that the cluster is using Kubernetes cluster-autoscaler.
```

chart/questions.yaml
```diff
@@ -497,6 +497,17 @@ WARNING:
...
+    Longhorn prevents data loss by only allowing CA to scale down a node that met all conditions:
```
innobead, line 503:
```suggestion
    Longhorn prevents data loss by only allowing the Cluster Autoscaler to scale down a node that met all conditions:
```

innobead (changes requested):
LGTM, just some rewords needed.

chart/questions.yaml
c3y1huang: Updated.

chart/questions.yaml
```diff
@@ -497,6 +497,17 @@ WARNING:
     min: 0
     max: 40
     default: 12
+- variable: defaultSettings.kubernetesClusterAutoscalerEnabled
+    label: Kubernetes Cluster Autoscaler Enabled (Experimental)
+    description: "Enabling this setting will notify Longhorn that the cluster is using Kubernetes Cluster Autoscaler.
+    Longhorn prevents data loss by only allowing the Cluster Autoscaler to scale down a node that met all conditions:
+      - No volume attached
```
shuo-wu, line 504: NIT: Just realized that this statement may be confusing. Should be `No volume attached to the node` instead.

chart/questions.yaml
```diff
@@ -497,6 +497,17 @@ WARNING:
...
+      - Is not the last node containing the replica of a detached volume.
```
shuo-wu, line 505: NIT: For a running volume on another node, we cannot remove the node either when it contains the last replica.

chart/templates/default-setting.yaml
```diff
@@ -42,3 +42,4 @@ data:
     backing-image-recovery-wait-interval: {{ .Values.defaultSettings.backingImageRecoveryWaitInterval }}
     guaranteed-engine-manager-cpu: {{ .Values.defaultSettings.guaranteedEngineManagerCPU }}
     guaranteed-replica-manager-cpu: {{ .Values.defaultSettings.guaranteedReplicaManagerCPU }}
+    kubernetes-cluster-autoscaler-enabled: {{ .Values.defaultSettings.kubernetesClusterAutoscalerEnabled }}
```
shuo-wu, line 45: Note: We will switch to using CRs for default settings in v1.3.0. But we can upgrade this whole YAML later.

chart/questions.yaml
c3y1huang: Fixed.

Also in https://github.com/longhorn/longhorn-manager/pull/1305

chart/templates/default-setting.yaml
c3y1huang: Fixed.

chart/questions.yaml
c3y1huang: Fixed.

innobead (approved):

Issue #3880, opened 2022-04-21
[IMPROVEMENT] Make storage parameter "migratable" available in the helm chart


ibrokethecloud:
## Is your improvement request related to a feature? Please describe
Currently harvester packaging for longhorn uses helm post-install and post-upgrade hooks to replace the longhorn-storageclass secret with a longhorn  storageclass override which introduces `migratable:true` in the storage class parameters.

Fleet is currently unable to handle post-install / post-ugprade hooks, and reports the longhorn-storageclass configmap as missing because this has been recreated by the post-install webhook.

This is possibly related to the issue: https://github.com/rancher/fleet/issues/257

## Describe the solution you'd like
We should expose the storage class parameter `migratable` in the helm chart for longhorn.

This will allow harvester to leverage this setting by just defining value overrides.

## Additional context

This should help us address:  https://github.com/harvester/harvester/issues/2088

joshimoo:
This can be done, but I am not sure why harvester isn't creating it's own storage class with parameters purposefully built for harvester which then gives them control/update ability independent from longhorn's default parameter set.
REF: https://github.com/harvester/harvester/issues/2088#issuecomment-1107551647
REF: https://github.com/longhorn/longhorn/issues/2594#issuecomment-843950488

ibrokethecloud:
@joshimoo would it be better in that case to make the storageclass definition configmap optional? with the default being true for backward compatibility? This we way we don't need to patch the default storage class created by the longhorn storage chart. I can create another PR for the same.

innobead:
> @joshimoo would it be better in that case to make the storageclass definition configmap optional? with the default being true for backward compatibility? This we way we don't need to patch the default storage class created by the longhorn storage chart. I can create another PR for the same.

No, `migratable` is a very implicit function actually for migratable volumes used in virtual workloads, but the change here is actually fine to expose in our helm template and users are able to configure it via values.yaml, because that's the public interface Longhorn provides to users. 

For `migratable` or any unconfigured parameters in the default storage class, we should not make them with the default value false, that's not intended.

For any downstream callers like Harvester, ideally, there should be a specific storage class created by itself as @joshimoo mentioned.

innobead:
@longhorn/qa For QA, we just need to verify if after the chart installation, the setting of migratable will take effect in the default storage class.

kaxing:
Will begin test when chart file for v1.3.0 preview or rc come out https://github.com/longhorn/charts

[...]

PR #3882, opened 2022-04-21
Retrofit of ability to expose migratable paraemter in the longhorn he…

ibrokethecloud:
This PR is related to: https://github.com/longhorn/longhorn/issues/3880 and exposes the storage class "migratable" in the helm chart.

This should help address: https://github.com/harvester/harvester/issues/2088

Same as PR# 3881 but for v1.2.x

### DIFF
--- a/chart/templates/storageclass.yaml
+++ b/chart/templates/storageclass.yaml
@@ -24,7 +24,7 @@ data:
       fsType: "{{.Values.persistence.defaultFsType}}"
       {{- end }}
       {{- if .Values.persistence.migratable }}
-      migratable: "{{.Values.persistence.migratable}}"
+      migratable: {{.Values.persistence.migratable}}
       {{- end }}  
       {{- if .Values.persistence.backingImage.enable }}
       backingImage: {{ .Values.persistence.backingImage.name }}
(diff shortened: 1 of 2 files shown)


innobead (approved):

chart/templates/storageclass.yaml
```diff
@@ -23,6 +23,9 @@ data:
       {{- if .Values.persistence.defaultFsType }}
       fsType: "{{.Values.persistence.defaultFsType}}"
       {{- end }}
+      {{- if .Values.persistence.migratable }}
+      migratable: "{{.Values.persistence.migratable}}"
```
innobead, line 27: It's bool value, so probably you don't need a double quote for this.

innobead (changes requested):

chart/templates/storageclass.yaml
ibrokethecloud: current override configmap quoted the value, so i was trying to follow convention:

https://github.com/harvester/harvester/blob/master/deploy/charts/harvester/templates/longhorn-storageclass-override.yaml#L28

ibrokethecloud: fixed up.

innobead (approved):

Issue #836, opened 2019-10-28
[FEATURE] Reclaim/Shrink space of volume


IAkumaI:
Hello, I khow volumes can grow only, but what if I need to reclaim space of deleted files?
Is there other choice but remove/create volume?

yasker:
@IAkumaI 

I don't think there is another way to reclaim the space of deleted files. Regarding the total size of the volume, you might reclaim some by removing old snapshots.

yasker:
Maybe we can try to implement `trim` on iscsi target.

SeMeKh:
@yasker I'm willing to spend some time on this. Any pointers on where and how the code should be changed?

yasker:
@SeMeKh Thanks!

After thinking more about it, I realized it's probably not a good idea to implement this.

My idea was by implementing `TRIM` command in iSCSI target (https://github.com/rancher/tgt/blob/longhorn_oct2019/usr/bs_longhorn.c#L80 ), we would know which area is no longer used by the filesystem. We will need a new command (https://github.com/longhorn/longhorn-engine/blob/master/pkg/engine/dataconn/types.go#L6) in data connection protocol to tell us about the TRIM. 

Longhorn's snapshot is immutable. That means we can only save the space in the `volume head` by punching holes in the `volume head` (similar to what we do in https://github.com/longhorn/sparse-tools/blob/50ffab4d432962b913d1497fd77f4878caf94077/sparse/rest/handlers.go#L119 ). So I am not sure if it's worth the effort.

The most helpful case would be the user write and delete a lot of data in the `volume head`, without making any snapshots. Once a snapshot was made, `TRIM` command will not be more helpful. It also depends on how OS uses `TRIM`. So it's a bit unclear how helpful this will be.

@SeMeKh If you want to take a look at other Longhorn issues instead, feel free to let me know.

SeMeKh:
@yasker Thanks for the explanation.

I experimented with running `fstrim` directly on the underlying file, and it did free up space. I'm afraid that could cause inconsistencies though. Do you think there is a feasible way to avoid losing the consistency, given that I am willing to avoid using snapshots if that's the cost?

yasker:
[...]

PR #3893, opened 2022-04-25
Create SPDK LEP

keithalucas:
Signed-off-by: Keith Lucas <keith.lucas@suse.com>

### DIFF
--- a/enhancements/20220325-spdk.md
+++ b/enhancements/20220325-spdk.md
@@ -1,0 +1,304 @@
+# Reimplement Longhorn Engine with SPDK
+
+## Summary
+
+[SPDK](https://spdk.io) is a framework for building high performance storage applications.  SPDK has several features that allow it perform tasks similar to what the longhorn-engine needs currently.  
+
+* [Block Devices](https://spdk.io/doc/bdev.html), or bdevs, are SPDK storage
+devices that support reading and writing data in fixed-size blocks.  Bdevs can
+be composed of other bdevs, and bdevs can also be stored in another bdev.   One
+built in example of a bdev being composed of another bdev is the RAID 0 bdev. 
+ 
+
+* [Logical volumes](https://spdk.io/doc/logical_volumes.html) is a feature in
+SPDK which allows one large bdev to be broken up into multiple smaller bdevs. 
+The large bdev has a _logical volume store_ which has information about each
+_logical volume_ contained within it.  The _logical volume store_ has a table
+of blocks for each _logical volume_; the data in a volume isn't necessarily
+continguous.  Also, the data in a volume can be thinly provisioned;  Like the
+current longhorn-engine, a thinly provision volume starts as a virtual device
+consisting of all zeroes.  When data is written to the volume, unused blocks
+for the data are used and the volume's table of blocks is updated. 
+
+Also, a logical volume supports snapshots like the longhorn-engine.  When data
+is read from a logical volume, for each block that is being read, that block
+number is looked up in the volume's table.  If the block is in the table, the
+block is read from the block specified.  If the block is not specified and the
+logical volume has a snapshot, it will look for the block in the snapshot.
+This will continue until either the data is found or a virtual zero logical
+volume is reached.
+
+A logical volume has the ability to perform snapshots which is a feature that
+is used as the basis for rebuilding, backups, and snapshots in general in the
+current longhorn engine.  Logical volumes are stored sparsely their parent's
+bdev similar to how the existing volume files are filesystem level sparse
+files.   
+
+* [NVMe over Fabrics](https://spdk.io/doc/nvmf.html)
+
+Any bdev can be made available as an NVMe over Fabrics target.  SPDK
+can internally make an NVMe over Fabrics target available as an bdev.
+For longhorn, a replica will make its logical volume available as an
+NVMe over Fabrics targets, and the longhorn bdev will _mount_ it as a
+bdev within SPDK.  This will allow a longhorn bdev to forward I/O
+operations to a remote logical volume.
+
+Also, a virtual longhorn volume can be made available as an NVMe over
+Fabrics target.  Linux can connect to these targets and make them
+available as a block device.  The `nvme` command can query and connect
+to NVMe over Fabrics targets, which makes them available as NVMe block
+devices to the Linux operating system (i.e.  `/dev/nvmeXn1`). 
+
+
+### Related Issues
+
+The URL For the related enhancement issues in Longhorn repo.
+
+## Motivation
+
+### Goals
+
+* Use SPDK to improve performance
+* Use SPDK to scale better
+
+### Non-goals [optional]
+
+* Providing the ability to continue to use the same
+`/var/lib/liblonghorn` data path.
+* Continuing to support qcow2 files.
+* Continuing to support multiple engine versions concurrently.
+
+## Proposal
+
+Create a new longhorn bdev in SPDK.  The longhorn bdev will be created 
+with a list of _replicas_ which are local and/or remote logical volumes.  
+The longhorn bdev is exposed over NVME over Fabrics.  The system will
+connect to the longhorn bdev over NVME over Fabrics, and it will be available
+as a device file.
+
+The longhorn bdev will handle I/O operations in a way such that all of the 
+replicas are in sync.  Every write operation is performed on every connected
+replica.  The longhorn bdev internally has a bdev for each replica.  A local
+replica is used directly; a remote replica is an NVME over Fabrics bdev.
+
+The longhorn bdev will write all write operations to all replicas; it will 
+choose a replica to read from in a round-robin manner.  As will the current
+longhorn-engine, a timeout can occur with each operation on replicas, 
+which will remove than replica from the list of replicas.
+
+
+### User Stories
+
+For the most part, the user experience between longhorn with SPDK should
+be the same as longhorn without SPDK.
+
+#### Storage Configuration
+
+Before SPDK, `/var/lib/longhorn` was used for storage and add additional disks
+as paths.  In all cases, the disk paths are formatted and mounted disks which
+can also be used for other things.
+
+With SPDK, a full disk can be used.  Each disk will be specified as a device
+file. 
+
+### User Experience In Detail
+
+### API changes
+
+The CRDs for longhorn will change slightly because of the new architecture.
+The new design eliminates the instance manager so the InstanceManager CRD isn't
+necessary.  Some of the specific fields in th Engine and Replica CRDs will need
+to be changed.
+
+
+## Design
+
+### Implementation Overview
+
+A new longhorn bdev will be added to SPDK.  This bdev will be a composite bdev
+of one or more replica bdevs.  Changing Longhorn's data plane will be a
+significant change.  One significant departure from the current design is that
+the logical volume store that will be used for longhorn volumes supports
+multiple volumes within one running instance of SPDK.  This will reduce the 
+number of processes running on the cluster.  Previously, every volume required
+a longhorn engine process.
+
+#### SPDK Programming Model
+
+SPDK's programming model rather different than the existing
+longhorn-engine.  The current longhorn-engine uses goroutines
+extensively.  For SPDK, an event driven model is used.  By default, 
+one actual thread runs on one core of the system's processor.  This 
+thread continuous polls for events, i.e. file descriptors becoming readable.
+When an event occurred, callbacks registered for that event are called.
+
+Callbacks should typically execute quickly.  In most cases, they will either
+finish handling the event or call another function cannot be performed 
+immediately.  This function will take another callback.  The original 
+callback will finish, relinquishing control the polling code, which should
+eventually receive the event that another callback is awaiting.
+
+Will the programming model, it's not possible to just do things
+step-by-step.  Almost everything requires a series of callbacks.
+
+By default, SPDK runs as just one thread.  However, it is possible to increase
+the number of threads.  In this case, multiple threads will be polling for
+events and executing the callbacks within different threads.  Longhorn 
+specific changes should be able to work under different threading conditions.
+
+Also, while there are different threading configuration for SPDK,
+SPDK internally assigns everything being executed to a virtual thread 
+which doesn't correspond to the actual thread that the process is executing
+in.    
+
+
+#### Longhorn Volume Bdev
+
+The longhorn volume bdev will be a virtual volume that relays I/O operations
+to a set of _replicas_.  
+
+#### Replica
+
+A replica will be a logical volume in a logical volume store. 
+
+#### Operations
+
+##### I/O
+
+* Read - The longhorn volume bdev will chose of the replicas to read from.
+* Write - The longhorn volume bdev will write to all of the replicas.
+
+##### Snapshot
+
+Snapshots need to be performed such that all replicas contain the same
+data when the snapshot occurs.
+
+1. Signal all the I/O threads to stop performing I/O.
+2. When all the writing threads are stopped, call snapshot API on 
+all the local replicas and issue a remote JSON-RPC call to perform a
+snapshot on the remote replicas.
+3. Resume the I/O threads.
+
+##### Adding new replica
+
+When a new replica is added to an existing volume, it 
+
+1. A snapshot operation is performed on the current set of replicas in
+the volume with an auto generated snapshot name.
+2. The new replica is added to the list of replicas as a _write only_ replica.
+3. The new replica queries one of the existing replicas to get the list of
+snapshots and the allocated blocks in each snapshot with JSON-RPC.
+4. The replica being queried assembles the snapshot info and makes each
+snapshot available over NVMe over Fabrics and sends this information to
+the new replica over JSON-RPC.
+5. The new replica recreates each snapshot by creating new logical volumes,
+registering the remote NVMe over Fabrics target can copying just the relative 
+blocks.
+6. The snapshot logical volumes on the new replica are linked together.
+7. The _write only_ replica is changed to _read/write_.
+
+##### Backups
+
+Using the S3 REST API from SPDK is somewhat impractical.  The S3 REST API
+is usually used with higher level programming language than C.  Also, 
+integrating an external network connection into the SPDK polling
+mechanism could affect other operations.  Because of this, SPDK will provide
+access to a volume so that a backup implementation in golang could 
+perform the backup outside of SPDK.
+
+Within SPDK, backup functionality consists of:
+1. Performing a snapshot
+2. Making that snapshot available over NVMe-oF.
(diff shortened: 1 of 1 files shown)


enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,299 @@
...
+perform the backup outside of SPDK.
+
+Within SPDK, backup functionality consists of:
+1. Performing a snapshot
+2. Making that snapshot available over NVMe-oF.
+
+The external backup program will mount the NVMe-oF target for the snapshot
+and perform the backup.
+
+For a backup, we need to create a new volume with the contents of the
```
joshimoo, line 212: This is a good approach moving backup logic into higher levels.
```suggestion
For a restores, we need to create a new volume with the contents of the
```

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,299 @@
...
+managers on each node to manage the engine and replicas processes is not 
+needed. 
+
+The longhorn-manager will be able to communication with SPDK directly.
+With this change, the following components will be removed:
+* Longhorn instance manager
+* Longhorn engine (longhorn-engine controller)
+* Longhorn replica
+
+Also longhorn will use NVMe over Fabrics to create the block device instead
```
joshimoo, line 234: Just thinking out loud here :)

This is okay, the manager can talk to spdk to get the volume up and running then the csi drive could be the one handling the connection to the nvme-of target + mount of the fs. 

This might or might not include some communication between the csi driver + manager as we have at the moment.
ControllerPublishVolume -> csi -> manager -> spdk 
NodeStageVolume -> csi -> mount the volume that has been connected to

**Depending on implementation complexity, it might also be easier to have the manager do the connection to `nvme-of` instead of the csi driver** (driver would still call on the manager backend), then just have the csi driver do the mount after the initial communication with the manager to get this setup.

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,299 @@
...
+Because of this, golang will be used to develop tests for this.  Python
+is currently used for tests but the global interpreter lock makes it 
+difficult to use to develop tests to exercise concurrent scenarios.
+
+
+### Upgrade strategy
+
+This will represent a significant change to how longhorn is deployed and
+how longhorn volumes are stored on disk.  Also, the longhorn has some custom
+resources which will no longer make sense, like the instance manager.
```
joshimoo, line 294: That was to be expected with a big change like this.

We can consider thinking about this some more down the line, since it might make sense to combine the im then have a go grpc control program/process per volume running in the im with which the manager can communicate via grpc like in the present.

Then the control program can handle the spdk interactions instead of the manager, this way there would still be a split between control plane / data plane. And the manager wouldn't need to know anything about spdk directly, instead it could treat the spdk interactions just like another engine type.

I considered combining the two im's before to potentially allow the engine/replica process to share memory / local resources (local socket) instead of having to go through the network layer when working with a single/primary replica etc.

joshimoo (commented):
LGTM so far
Just left some notes for ideas/knowledge sharing.

enhancements/20220325-spdk.md
keithalucas: I update the text.  I separated the logic because felt like it would be difficult to write the backup/restore to S3 in C. :)

enhancements/20220325-spdk.md
keithalucas: I don't know.  I thought about having some kind of gRPC to JSON-RPC proxy program.  We could investigate doing something like to ease the transition.   Also, it would be a way to achieve security via mTLS.

enhancements/20220325-spdk.md
keithalucas: I was just suggesting something similar to how it works now.   We can probably revisit exact what components do what in the process when it is being integrated.

enhancements/20220325-spdk.md
liyimeng: Is this still targeting 1.30?

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,299 @@
...
+	When data is read from a logical volume, for each block that is being
+	read, that block number is looked up in the volume's table.  If the
+	block is in the table, the block is read from the block specified.  If
+	the block is not specified and the logical volume has a snapshot, it
+	will look for the block in the snapshot.  This will continue until
+	either the data is found or a virtual zero logical volume is reached.
+
+
+	A logical volume has the the
+	ability to perform snapshots which is a feature that is used a basis for
```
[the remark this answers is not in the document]
keithalucas: Updated the text.

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,299 @@
...
+
+	Any bdev can be made available as an NVMe over Fabrics target.  SPDK
+	can internally make an NVMe over Fabrics target available as an bdev.
+	For longhorn, a replica will make its logical volume available an an
+	NVMe over Fabrics targets, and the longhorn bdev will _mount_ at as a
+	bdev within SPDK.  This will allow a longhorn bdev to forward I/O
+	operations to a remote logical volume.
+
+	Also, a virtual longhorn volume can be made available as an NVMe over
+	Fabrics target.  Linux can connect to these targets an make them
```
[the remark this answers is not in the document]
keithalucas: Updated the text.

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,299 @@
...
+With SPDK, a full disk can be used.  Each disk will be specified as a device
+file. 
+
+### User Experience In Detail
+
+Detail what user need to do to use this enhancement. Include as much detail as possible so that people can understand the "how" of the system. The goal here is to make this feel real for users without getting bogged down.
+
+### API changes
+
+The CRDs for longhorn will change slightly because there is an engine
```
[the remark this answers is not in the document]
keithalucas: Updated the text.

enhancements/20220325-spdk.md
keithalucas: The github issues were set to 1.3 because nothing higher was available.  I think it will be something like 2.0

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,299 @@
...
+	rebuilding, backups, and snapshots in general in the current longhorn
+	engine.  Logical volumes are stored sparsely their parent's bdev similar
+	to how the existing volume files are filesystem level sparse files.   
+
+* [NVMe over Fabrics](https://spdk.io/doc/nvmf.html)
+
+	Any bdev can be made available as an NVMe over Fabrics target.  SPDK
+	can internally make an NVMe over Fabrics target available as an bdev.
+	For longhorn, a replica will make its logical volume available an an
+	NVMe over Fabrics targets, and the longhorn bdev will _mount_ at as a
```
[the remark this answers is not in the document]
keithalucas: I proofread the text and updated it.

enhancements/20220325-spdk.md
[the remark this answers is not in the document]
keithalucas: I proofread the text and updated it.

enhancements/20220325-spdk.md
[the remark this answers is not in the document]
keithalucas: I proofread the text and updated it.

enhancements/20220325-spdk.md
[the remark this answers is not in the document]
keithalucas: I proofread the text and updated it.

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,299 @@
...
+
+
+## Design
+
+### Implementation Overview
+
+A new longhorn bdev will be added to SPDK.  This bdev will be a
+composite bdev of one or more replica bdevs.  The vo
+Changing Longhorn's data plan with be a significant change.  One
+significant departure from the current design is that the logical volume
```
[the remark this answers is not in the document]
keithalucas: I kind of don't like changing departure to difference.  https://www.freethesaurus.com/departure indicates difference is a synonym.

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,304 @@
...
+The longhorn bdev will write all write operations to all replicas; it will 
+choose a replica to read from in a round-robin manner.  As will the current
+longhorn-engine, a timeout can occur with each operation on replicas, 
+which will remove than replica from the list of replicas.
+
+
+### User Stories
+
+For the most part, the user experience between longhorn with SPDK should
+be the same as longhorn without SPDK.
```
derekbit, line 93: If the underlying storage does not fit the SPDK requirements, will it fallback to the data path without SPDK automatically or just return an error?

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,304 @@
...
+
+#### Storage Configuration
+
+Before SPDK, `/var/lib/longhorn` was used for storage and add additional disks
+as paths.  In all cases, the disk paths are formatted and mounted disks which
+can also be used for other things.
+
+With SPDK, a full disk can be used.  Each disk will be specified as a device
```
derekbit, line 101: > Each disk will ...

Just curious. Does the disk mean a physical disk or a logical disk?

liyimeng:
This seem dead again. Mayadata get acquired and commercialise its nvmeof solution https://blocksandfiles.com/2022/05/18/datacore-bolt-kubernetes/ 

Will longhorn  provide something similar? I mean a premium version with high performance.  Or comparable benchmark is unachievable under longhorn archtecture?

innobead:
This is still in the plan and continues.

enhancements/20220325-spdk.md
innobead: We will not do the failback, because when choosing which data engine to use, it will just use it successfully or error out.

enhancements/20220325-spdk.md
innobead: Theoretically, It should be any block device on the host, so it doesn't matter whether it's physical or virtual. 

There is one bdev driver we could leverage for local device. 

> Linux AIO bdev
> The SPDK AIO bdev driver provides SPDK block layer access to Linux kernel block devices or a file on a Linux filesystem via Linux AIO. Note that O_DIRECT is used and thus bypasses the Linux page cache. This mode is probably as close to a typical kernel based target as a user space target can get without using a user-space driver. To create AIO bdev RPC command bdev_aio_create should be used.

ref: https://spdk.io/doc/bdev.html

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,304 @@
...
+
+### Implementation Overview
+
+A new longhorn bdev will be added to SPDK.  This bdev will be a composite bdev
+of one or more replica bdevs.  Changing Longhorn's data plane will be a
+significant change.  One significant departure from the current design is that
+the logical volume store that will be used for longhorn volumes supports
+multiple volumes within one running instance of SPDK.  This will reduce the 
+number of processes running on the cluster.  Previously, every volume required
+a longhorn engine process.
```
innobead, line 124: YES. This is resource optimization based on logical volume stores, but we also need to understand more if we still can claim individual dependency between volumes/replicas.

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,304 @@
...
+Will the programming model, it's not possible to just do things
+step-by-step.  Almost everything requires a series of callbacks.
+
+By default, SPDK runs as just one thread.  However, it is possible to increase
+the number of threads.  In this case, multiple threads will be polling for
+events and executing the callbacks within different threads.  Longhorn 
+specific changes should be able to work under different threading conditions.
+
+Also, while there are different threading configuration for SPDK,
+SPDK internally assigns everything being executed to a virtual thread
```
innobead, line 150: `virtual thread` should be green thread or usually, we say coroutine.

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,304 @@
...
+
+
+#### Longhorn Volume Bdev
+
+The longhorn volume bdev will be a virtual volume that relays I/O operations
+to a set of _replicas_.  
+
+#### Replica
+
+A replica will be a logical volume in a logical volume store.
```
innobead, line 162: As said, need to think through the replica independency in the store.

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,304 @@
...
+#### Operations
+
+##### I/O
+
+* Read - The longhorn volume bdev will chose of the replicas to read from.
+* Write - The longhorn volume bdev will write to all of the replicas.
+
+##### Snapshot
+
+Snapshots need to be performed such that all replicas contain the same
```
innobead, line 173: SPDK logic volume natively supports snapshot, so no need to do what has been done in the current data plane.

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,304 @@
...
+3. The new replica queries one of the existing replicas to get the list of
+snapshots and the allocated blocks in each snapshot with JSON-RPC.
+4. The replica being queried assembles the snapshot info and makes each
+snapshot available over NVMe over Fabrics and sends this information to
+the new replica over JSON-RPC.
+5. The new replica recreates each snapshot by creating new logical volumes,
+registering the remote NVMe over Fabrics target can copying just the relative 
+blocks.
+6. The snapshot logical volumes on the new replica are linked together.
+7. The _write only_ replica is changed to _read/write_.
```
innobead, line 198: This part actually involves replica rebuilding. Operations between replicas or volumes need to tackle by our new data plane.

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,304 @@
...
+
+#### Deployment and Integration
+
+SPDK with longhorn support will have a JSON-RPC API for creating and managing
+longhorn volumes.   Because each SPDK instance will manage both multiple 
+volumes and multiple replicas, the current implementation with two instance
+managers on each node to manage the engine and replicas processes is not 
+needed. 
+
+The longhorn-manager will be able to communication with SPDK directly.
```
innobead, line 232: Still need to have another layer for data plane for better responsibility segregation, so it should be Ok to merge IM-E and IM-R.

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,304 @@
...
+* Longhorn replica
+
+Also longhorn will use NVMe over Fabrics to create the block device instead
+of iSCSI.  In order to use NVMe over Fabrics to connect to a remote device,
+the `nvme_tcp` kernel module needs to be installed on the node. There is
+a sysfs interface which can be used to connect to an NVMe over Fabrics
+target.  Using NVMe over Fabrics eliminates the following following
+components from longhorn:
+* iscsid/isciadm from the node.
+* tgt and liblonghorn
```
innobead, line 245: These are just for the current data plane.

innobead:
This proposal is based on replication and tries to achieve what we have done in the current data plane. 

However, we need to base on this and consider sharding in the design. 

cc @DamiaSan @derekbit @joshimoo @shuo-wu

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,304 @@
...
+The longhorn volume bdev will be a virtual volume that relays I/O operations
+to a set of _replicas_.  
+
+#### Replica
+
+A replica will be a logical volume in a logical volume store. 
+
+#### Operations
+
+##### I/O
```
shuo-wu, line 166: Need to support Unmap as well. See longhorn/longhorn#836

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,304 @@
...
+* Read - The longhorn volume bdev will chose of the replicas to read from.
+* Write - The longhorn volume bdev will write to all of the replicas.
+
+##### Snapshot
+
+Snapshots need to be performed such that all replicas contain the same
+data when the snapshot occurs.
+
+1. Signal all the I/O threads to stop performing I/O.
+2. When all the writing threads are stopped, call snapshot API on
```
shuo-wu, line 177: Do we need to flush data into the disk?

enhancements/20220325-spdk.md
innobead: ref: https://spdk.io/doc/group__bdev__io__submit__functions.html#ga680c9c302998f7b003e2476e35d9ae4b

enhancements/20220325-spdk.md
innobead: I think YES, we should do the same before calling the snapshot API as what we are doing now.

enhancements/20220325-spdk.md
```diff
@@ -0,0 +1,304 @@
...
+integrating an external network connection into the SPDK polling
+mechanism could affect other operations.  Because of this, SPDK will provide
+access to a volume so that a backup implementation in golang could 
+perform the backup outside of SPDK.
+
+Within SPDK, backup functionality consists of:
+1. Performing a snapshot
+2. Making that snapshot available over NVMe-oF.
+
+The external backup program will mount the NVMe-oF target for the snapshot
```
l-mb, line 213: The current backup exports the incremental diff efficiently and at the block layer; perhaps that can be retained if the SPDK snapshots were made more transparent to Longhorn?
Otherwise there'd be a net-loss of functionality?

enhancements/20220325-spdk.md
l-mb: If the higher layers call `unmap` (say, after deleting a large data set to free up space), the users will expect this to be passed down and eventually also release the allocated clusters for other volumes/bdevs in the SPDK blobstore. Not just when de-allocating the entire device.

enhancements/20220325-spdk.md
l-mb: SPDK's own freeze isn't enough since the data is replicated/possibly sharded across multiple devices (which obviously the one SPDK instance doesn't know about). The quiescence/freeze needs to be implemented from the top down.

enhancements/20220325-spdk.md
DamiaSan: Which allocated clusters should be released? In SPDK you can unmap an entire volume/bdev

l-mb: Depending on the shard size that Longhorn would use for multi-device distribution, we'd possibly be able to (eventually) release the whole shard/bdev (assuming no snaps exist, obviously).

Within one lvol, the current SPDK seems to only allow *allocating* clusters in thinly provisioned volumes, but has no support for releasing them at finer granularity than the entire bdev. (Unlike, say, ext4 or other file systems that can punch holes into files so they become sparse again.)

 That seems like something that'd need to be added to lvol interface to fully support thin provisioning. (Not for the initial revision, but it should be on the roadmap and discussed with SPDK upstream)

innobead:
Closed first. @DamiaSan will start another PR for the design based on @keithalucas contribution.

PR #3899, opened 2022-04-27
orphaned data cleanup: update deploy YAML

derekbit:
Longhorn 685

### DIFF
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -508,6 +508,12 @@ WARNING:
     group: "Longhorn Default Settings"
     type: boolean
     default: false
+- variable: defaultSettings.orphanAutoDeletion
+    label: Orphaned Data Cleanup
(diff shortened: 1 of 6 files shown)


chart/questions.yaml
```diff
@@ -508,6 +508,12 @@ WARNING:
     group: "Longhorn Default Settings"
     type: boolean
     default: false
+- variable: defaultSettings.orphanAutoDeletion
+    label: Orphaned Data Cleanup
+    description: "This setting allows Longhorn to delete the orphan resource and its corresponding orphaned data automatically. Orphan resources on down or unknown nodes will not be cleaned up automatically.
```
innobead, line 513:
```suggestion
    description: "This setting allows Longhorn to delete the orphan resource and its corresponding orphaned data automatically like stale replicas. Orphan resources on down or unknown nodes will not be cleaned up automatically.
```

innobead (approved):

Issue #2128, opened 2020-12-21
[FEATURE]Application consistent snapshot/backup


yasker:
**Is your feature request related to a problem? Please describe.**
Currently, Longhorn can do crash-consistent snapshot/backup. But for the applications like databases, quiescing is needed to create an application-consistent snapshot, so the application will make sure all the data in the memory has been written to the disk before creating the snapshot.

**Describe the solution you'd like**
Provide a way for the users to quiesce the workload before taking the snapshot/backup.

**Describe alternatives you've considered**
Users can also script the backup using CSI snapshotter and run quiescing/unquescing before taking the snapshot. But it won't work with the recurring snapshot/backup which is scheduled by Longhorn.

**Additional context**
N/A

innobead:
@jenting Please help with this. Thanks.

joshimoo:
linking the backup refactor issue https://github.com/longhorn/longhorn/issues/1761

janeczku:
Application state may be distributed across multiple persistent volumes (e.g. sharded DB). The application consistent backup must then ensure that snapshots are performed in a point-in-time consistent manner across all volumes.

Offering such feature Longhorn may allow users to create "volume groups" and define on-demand or scheduled backup plans for the whole group instead of individual volumes.

innobead:
> Application state may be distributed across multiple persistent volumes (e.g. sharded DB, distributed indexes). The application consistent backup must then ensure that snapshots are performed in a point-in-time consistent manner across all volumes.
> 
> 
> 
> Offering such feature Longhorn may allow users to create "volume groups" and define on-demand or scheduled backup plans for the whole group instead of individual volumes.

Thanks for the comment. The similar idea we had some discussion before. The goal is we need to introduce a mechanism to make the volumes of application atomic aware operations supported and like u said, not just individual volumes.

Please follow up the discussion and proposals in near future.

innobead:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Longhorn-Main-Board-59d8105b2fbdf4155c4589ef/issues/longhorn/longhorn/2128?planning-poker) with ZenHub @jenting @PhanLe1010 @shuo-wu @joshimoo

shuo-wu:
Leave a note here: 
Velero relies on annotation hooks to execute the cmd before/after backup. I can do something similar to archive application-consistent snapshot/backup. (run `sync` and `freezes` before the snapshot/backup)

The next:
Investigate Kanister

[...]

PR #3904, opened 2022-04-28
Create RecurringJobs with the helm chart

tyzbit:
In this PR, we add the `recurringJobs` object in `values.yaml`. This objects's purpose is to generate `RecurringJob`s (and associated `StorageClass`es) by simply specifying snapshot and backup intervals for one or more groups.

Some notes about this PR:
- I have not updated `questions.yml` or `app-readme.md` because I'm a little unfamiliar with these. Point me to some docs and I will give it a go.
- I have tried to comment as much as I can, but unfortunately it is still hard to follow in places, particularly when generating the `RecurringJob`s.


tyzbit:
I am still open to questions or improvements to this PR.

innobead:
cc @longhorn/maintainer

joshimoo:
Let's circle back on this after 1.3 we also need to check for upgrade compatibility.

tyzbit:
This is still in process.

tyzbit:
This has been updated to support any task type including the new `snapshot-cleanup` and `snapshot-delete` task types. Docs in `values.yaml` have been updated as well.

innobead:
cc @c3y1huang what do you think?

c3y1huang:
I am thinking if we can make this more intuitive by having a list for `recurringJobs`. Each item could have customizable values matching the RecurringJob spec.
```
recurringJobs:
 -  name: sample-snapshot
    concurrency: 1
    cron: '* * * * *'
    groups: []
    labels: {}
    retain: 0
    task: snapshot
 -  name: sample-backup
    concurrency: 1
    cron: '* * * * *'
    groups: []
    labels: {}
    retain: 0
    task: backup
```

tyzbit:
I arrived at this convention coming from the perspective that the best experience would be if users cared about, logically organized, and managed their storage tasks based primarily on group membership. When you configure a volume to take regular snapshots, backups, or perform the new cleanup tasks, you do so by assigning a group to the volume (manually or via the `StorageClass`), **not** by associating the volume with an individual `RecurringJob`.

In my opinion, the design of groups being implicitly created from `RecurringJob`s rather than being their own CRD or being otherwise explicitly managed makes it difficult to organize and maintain `RecurringJob`s and understand which tasks will be performed on which volumes at what times without studying the `RecurringJob`s closely. This PR is how I addressed that difficulty. I can glance at a volume and know how often my snapshots will be taken and I know how many snapshots will be retained simply by referencing the group (again, manually assigned or implicit by using one of the `StorageClass`es that can be generated automatically for you).

Having an array of `RecurringJob` config objects as you described definitely could make the Helm templating much simpler (and thus **much** easier to maintain. I cannot overstate how displeased I am that the Helm logic for this is so complex, but it is as simple as I could make it while meeting my goals, and I documented as much as possible). Having more than a couple items though would result in it being difficult to determine what jobs would be performed if a given group was assigned to a volume and would also require deciphering the cron specification for each `RecurringJob`. **Essentially, users would be required to memorize all of the details for all `RecurringJob`s to be certain about what the configuration is, or to make changes and be sure the new configuration is what they desire.** Although this PR doesn't allow for a wide range of cron specificiations (and I'd be open to adding an enhancement where it could be explicitly defined), it is very simple to find a group in the `values.yaml`, find the task type and then see how often the tasks will run (and how many snapshots/backups are retained).

I'd support changing from a dictionary of group/configuration objects (and by extension `StorageClass` names, if enabled) to a list of configuration objects, if that's preferred. In fact that would allow for a simple demonstrative example to be enabled by default in the `values.yaml` that users could override since specifying your own list would completely overwrite the default list (instead of how it is now that if the examples were uncommented, they would be merged with the groups the user specifies, creating both sets). I think I'll work on making that change regardless, unless there's a compelling reason not to.

c3y1huang:
I see. Creating RecurringJob and associating with a group sounds like a logical approach.

However, I'm not entirely clear on the need to create a separate StorageClass when the groups can simply be included in the default longhorn StorageClass by customizing the persistence.recurringJobSelector.

c3y1huang:
Got it. However, I think it would be better not to include the creation of the StorageClass in the `recurringJob`, as there are other fields and parameters to consider and not relevant to the RecurringJob.

tyzbit:
I agree with the point that `recurringJob` is not quite appropriate to control/create `StorageClass`es, but I do think the functionality of creating `StorageClass`es that correspond to a set of groups that also generate `RecurringJobs` according to the configured properties is worth keeping in.

To achieve this, what about reorganizing the values around groups?

To summarize:

- Add top-level key `groupRecurringJobGeneratorDefaults`. This has the defaults for the `RecurringJob` resources generated by the chart, such as `concurrency` and `minuteStep` (which is used to calculate the cron expression)
- Add top-level key `groupStorageClassGeneratorDefaults`. This has the defaults for the `StorageClass` resources generated by the chart, such as `parameters` and `allowVolumeExpansion`
- Add top-level key `volumeGroups` which is an array of group config objects that specify `task`s that generate `RecurringJob`s. Also, there is an optional `storageClass` property where users can enable/disable the creation of related `StorageClass`es and every group can accept the defaults or completely override the entire `StorageClass` in the group object.

Note: Some of the defaults may not be appropriate to specify as defaults, or there may be defaults set somewhere else in the chart that would need to be reconciled with this approach.

```yaml
groupRecurringJobGeneratorDefaults:
  concurrency: 4
  retain: 1
  spreadCronStartTimes: true
  minuteStep: 5
  hourStep: 2

groupStorageClassGeneratorDefaults:
  ## (This comment would be removed in the final values.yaml) Everything under this key will essentially be passed directly into StorageClass resources after the metadata section. This makes it future-proof for any possible settings for the StorageClasss, at the expense of being more verbose.
  allowVolumeExpansion: true
  parameters:
    numberOfReplicas: "3"
    staleReplicaTimeout: "2880" # 48 hours in minutes
    fromBackup: ""
    fsType: "ext4"

## Specify Longhorn volume groups to create. The group must have at least one task specified to exist.
## Specifying a task generates the appropriate `RecurringJob` for the volume group.
## Specifying storageClass.create: true will create a StorageClass with `parameters.recurringJobsSelector` selecting this group, so creating a volume with the StorageClass will assign the group automatically.
volumeGroups:

  ## Simple example
  - name: Database
    storageClass:
      create: true
    tasks:
      snapshot:
        - interval: 10m
          retain: 6
      backup:
        - interval: 12h
          retain: 2

  ## Fully customized example
  - name: Webapp
    storageClass:
      create: true
      ## (This comment would be removed in the final values.yaml) everything under `properties` overwrites the defaults set in groupStorageClassGeneratorDefaults and is essentially just appended onto the generated StorageClass after the metadata section. This makes it future-proof for any possible settings for the StorageClasss, at the expense of being more verbose.
      properties:
        allowVolumeExpansion: false
        parameters:
          numberOfReplicas: "2"
          staleReplicaTimeout: "60"
          fromBackup: ""
          fsType: "ext4"

    tasks:
      snapshot:
        ## Override everything that's possible to override
        - interval: 10m
          retain: 6
          concurrency: 2
          spreadCronStartTimes: false
          minuteStep: 30
          hourStep: 1
        ## Add another snapshot task but accept all defaults (interval is the only required field)
        - interval: 12h
      backup:
        - interval: 12h
          retain: 2
      snapshot-cleanup:
        - interval: 24h
      snapshot-purge:
        - interval: 7d
```

c3y1huang:
I like the volumeGroups approach. What do you think of re-arrange like below?
```yaml
volumeGroups:
  recurringJobDefaults:
    concurrency: 4
    retain: 1
    spreadCronStartTimes: true
    minuteStep: 5
    hourStep: 2
    
  storageClassDefaults:
    allowVolumeExpansion: true
    parameters:
      numberOfReplicas: "3"
      staleReplicaTimeout: "2880" # 48 hours in minutes
      fromBackup: ""
      fsType: "ext4"
      
  generators:
  ## Simple example
  - name: Database
    storageClass:
      create: true
    recurringJobs:
      snapshot:
        - interval: 10m
          retain: 6
      backup:
        - interval: 12h
          retain: 2

  ## Fully customized example
  - name: Webapp
    storageClass:
      create: true
      properties:
        allowVolumeExpansion: false
        parameters:
          numberOfReplicas: "2"
          staleReplicaTimeout: "60"
          fromBackup: ""
          fsType: "ext4"

    recurringJobs:
      snapshot:
        - interval: 10m
          retain: 6
          concurrency: 2
          spreadCronStartTimes: false
          minuteStep: 30
          hourStep: 1
        - interval: 12h
      backup:
        - interval: 12h
          retain: 2
      snapshot-cleanup:
        - interval: 24h
      snapshot-purge:
        - interval: 7d
```

tyzbit:
That's great! I like how it's easier to understand naturally based on simpler setting names (such as `recurringJobDefaults`) and where those settings are organized. I want to think a bit more about `generators`, maybe change that back to `groups` because then it's clear that every item in that list will result in a new group and the `storageClass` and `recurringJobs` settings on the group object make it clear that a `StorageClass` and the appropriate `RecurringJob`s will be generated.

In any case, I'll get to work reorganizing and we can circle back to tweak when the chart's updated.

c3y1huang:
Having this could simplify the management when the user is managing volumes with multiple StorageClasses. However, we should also take into account how this fits with the volume group concept mentioned in    https://github.com/longhorn/longhorn/issues/2128.

tyzbit:
I see your point, but I think they can both coexist nicely. This functionality creates groups, `StorageClass`es and `RecurringJobs` so that users can manage many volumes easily by grouping them according to the overall `RecurringJob` settings appropriate for similar volumes (according to data behavior).

#2128 looks like it is looking for solutions so the multiple volumes of a total workload can be grouped together and managed as well, which to me is as simple as creating another volume group and assigning it to every volume of a given workload.

This PR is more of a band-aid to deal with the difficulty that is using groups to manage volumes when groups are only either created implicitly by a `RecurringJob` or manually. To me this is kind of like creating groups on a Linux system according to what groups are specified on files throughout the filesystem. I think both items highlight the shortcomings of creating and managing volume groups when they are a critical tool for managing volumes. 

My suggestion, **far outside the scope of this PR**, is to create a `Group` CRD so users can see and manage their volume groups in the same way they manage all other Kubernetes resources and then create an annotation users can add to PV/PVCs to enroll them into the groups. Then, remove the group membership from `RecurringJobs` and add it to the `Group` CRD. Then the users can simply look at the `Group` CRD to see what `RecurringJob`s are enabled for that `Group`  after figuring out what group a volume is in by checking the annotation. Of course manual group membership would still be possible.

Something like:

```yaml
kind: Group
metadata:
  name: longhorn-database
  namespace: longhorn-system
spec:
  recurringJobs:
    - name: backup-12h-retain-2-concurrency-4
      namespace: longhorn-system
    - name: snapshot-1h-retain-6-concurrency-4
      namespace: longhorn-system
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: example-app-database
  namespace: example-app
  annotations:
    # note the new annotation
    longhorn.io/group: longhorn-database
spec:
  resources:
    requests:
      storage: 5Gi
  volumeMode: Filesystem
  storageClassName: longhorn
  accessModes:
    - ReadWriteOnce
---
kind: RecurringJob
metadata:
  name: backup-12h-retain-2-concurrency-4
  namespace: longhorn-system
spec:
  concurrency: 4
  cron: 15 6/12 * * *
  # groups would be removed from the CRD.
  labels:
    jobName: backup-12h-retain-2-concurrency-4
  retain: 2
  task: backup
---
kind: RecurringJob
metadata:
  name: snapshot-1h-retain-6-concurrency-4
  namespace: longhorn-system
spec:
  concurrency: 4
  cron: 35 * * * *
  labels:
    jobName: snapshot-1h-retain-6-concurrency-4
  name: snapshot-1h-retain-6-concurrency-4
  retain: 6
  task: snapshot
```

That would honestly take a lot of wind out of the sails of this PR because then the only complicated part is creating unique, appropriate RecurringJobs but that's so much easier than reviewing every `RecurringJob` to see what groups are enrolled where.

The `Group` CRD could also be extended to include other Longhorn volume settings, potentially all of them. It would be like a StorageClass except volumes could easily change groups whereas you can't change the `StorageClass` on an existing volume.

Until then, this PR can help bridge that usability gap, and if what I described above were to be implemented this functionality could be pared down substantially to be much simpler and just have the functionality of specifying intervals for different task types and then the template creates and deduplicates `RecurringJob`s that match the intervals specified for the task type. That would be much more maintainable for others, too.

karlkovaciny:
There are no docs anyone could give tyzbit about `questions.yml`? What is the purpose of this file?

tyzbit:
@karlkovaciny I looked again and found [this](https://ranchermanager.docs.rancher.com/v2.6/how-to-guides/new-user-guides/helm-charts-in-rancher/create-apps#questionsyml), I think this file is for Rancher's App Catalog. I don't see a way for the user to specify entirely new objects, only existing keys or select between predefined options so I don't think the App Catalog could support this natively (if there's a way to throw YAML in there that gets deep-merged, users of the App Catalog could still easily use this functionality)

tyzbit:
FYI `filesystem-trim` jobs introduced in https://github.com/longhorn/longhorn/issues/5186 are already supported per the recent changes to the chart that allow any job name except `options`.

tyzbit:
Is there any more feedback on this PR, or an idea of what else needs to happen to get it merged? I will close this soon unless it seems there is an appetite for this. A few people in the k8s@home Discord use this through [the standalone chart I created](https://github.com/tyzbit/helm-charts/tree/main/charts/longhorn-recurring-jobs) so maybe that's a better place for this functionality to live.

innobead:
@c3y1huang please help review this.

c3y1huang:
I am a bit concerned about managing the group via the helm chart and the overlap with the [volume group feature](https://github.com/longhorn/longhorn/issues/2128).

tyzbit:
I can't see a way to make `RecurringJob`s programatically without also implicitly creating groups with the same functionality unless you remove the functionality to specify groups on `RecurringJob`s. See https://github.com/longhorn/longhorn/pull/3904#issuecomment-1471243929

If group functionality is changing soon, I don't mind leaving this open until the changes are in and then updating the template to work with whatever the new convention is. If there are no plans to remove implicit group creation and membership functionality from `RecurringJob`s, then I believe this PR is in line with current and future architectural goals and the remaining discussion may only be about whether or not the official helm chart will include functionality to programatically generate `RecurringJob` CRDs.

tyzbit:
Abandoning this PR due to lack of feedback and interest.

PR #3927, opened 2022-05-05
Fix wrong type of parameters in the storage class

innobead:
[Longhorn #3880](https://github.com/longhorn/longhorn/issues/3880)


derekbit (approved):
LGTM.

PR #3928, opened 2022-05-05
Fix wrong type of parameters in the storage class

innobead:
Fix the wrong type of parameters in the storage class
    
Longhorn #3880
    
(cherry picked from commit f898dec142f1f603e95787dccca86c3e0481ff71)


derekbit (approved):
LGTM.

PR #3954, opened 2022-05-10
Fix incorrect indent and missing double-quote in questions.yaml

yangchiu:
Fix incorrect indent and missing double-quote in questions.yaml

Signed-off-by: Yang Chiu <yang.chiu@suse.com>


innobead (approved):

innobead:
Good catch @yangchiu

Issue #2570, opened 2021-05-05
[BUG] helm upgrade won't apply customize default setting


jenting:
**Is your feature request related to a problem? Please describe.**

The longhorn-default-setting ConfigMap sync to settings CR only when settings CR does not exist.
This means that during helm upgrade if the user changes the setting, then it won't applies to settings CR (but applies to longhorn-default-setting ConfigMap _only_)
```shell
helm upgrade longhorn longhorn/longhorn -n longhorn-system \
    --set defaultSettings.backupTarget=<new-backup-target> \
    --set defaultSettings.backupTargetCredentialSecret=<new-backup-target-credential-secret>
```

**Describe the solution you'd like**

I think we could configure the settings CR directly.
So, we don't have to write a setting controller to reconcile to longhorn-default-setting ConfigMap to settings CR.
However, to accomplish it, we need to have a structural schema on CRDs and also, have an admission webhook to validate the input of these settings.

**Describe alternatives you've considered**

N/A

**Additional context**

Related issues
- https://github.com/longhorn/longhorn/issues/2562#issuecomment-832413461
- https://github.com/longhorn/longhorn/issues/2539#issuecomment-827290662
- https://github.com/longhorn/longhorn/issues/2611
- https://github.com/longhorn/longhorn/issues/2744
- https://github.com/longhorn/longhorn/issues/2825
- https://github.com/longhorn/longhorn/issues/3398
- https://github.com/longhorn/longhorn/issues/3458

jenting:
Note: we had a [KB](https://longhorn.io/kb/troubleshooting-default-settings-do-not-persist/) to workaround this.

jenting:
We've lots of community issues related to this, we might need to consider raising the priority of this issue.

jenting:
We might need to consider the validating/mutating admission webhook #791 along with this issue to auto-correct the invalid user input, for example:
- if the user input <true|false>, the Longhorn manager help auto-correct to <"true"|"false"> and updates the Setting CR.
- if the user input value but not in our value range, the Longhorn manager falls back to use the default value and updates the Setting CR.

innobead:
> We might need to consider the validating/mutating admission webhook #791 along with this issue to auto-correct the invalid user input, for example:
> 
> * if the user input <true|false>, the Longhorn manager help auto-correct to <"true"|"false"> and updates the Setting CR.
> * if the user input value but not in our value range, the Longhorn manager falls back to use the default value and updates the Setting CR.

[...]

PR #3998, opened 2022-05-18
Skip rendering nil default settings

derekbit:
A default settings is nil when it is
- unset in values, or
- the value is "~"

[Longhorn 2570](https://github.com/longhorn/longhorn/issues/2570)

### DIFF
--- a/chart/templates/default-setting.yaml
+++ b/chart/templates/default-setting.yaml
@@ -6,42 +6,42 @@ metadata:
   labels: {{- include "longhorn.labels" . | nindent 4 }}
 data:
   default-setting.yaml: |-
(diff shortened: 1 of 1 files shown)


derekbit:
Before, all default settings are rendered. If the setting's value is not set or user unsets it, the value becomes empty after rendering. However, we cannot distinguish the two cases and determine if we need to update the value to empty.

Thus,  the commit is skipping rendering nil default settings to address the `helm upgrade` issue.

`longhorn.yaml` is generated by `./scripts/generate-longhorm-yaml.sh` and ConfigMap longhorn-default-setting is empty by default.  This will make the behaviors of helm and kubectl consistent.

innobead (approved):

PR #4041, opened 2022-05-27
Make imagePullPolicy is configurable

derekbit:


innobead:
@derekbit where is the change request from? any purpose to doing this change?

derekbit:
> @derekbit where is the change request from? any purpose to doing this change?

Ah, I've not opened an issue for this.
`pullPolicy` in `values.yaml` is actually not used, so I think it's designed for the imagePullPolicy originally.

PR #4043, opened 2022-05-27
Separate tls ingress option from secure backend

Morishiri:
Currently when setting `ingress.tls` to true it is trying to connect to the underlying service on port 443 which doesn't work.

Is it even supported to run on 443?

I separated the options, because theconfiguration with `secure-backends` was constantly giving me 503. When removing `secure-backends` - it works.

I'm terminating TLS on ingress level with cert-manager so I think I don't need to have secured service as well.


innobead:
> Currently when setting `ingress.tls` to true it is trying to connect to the underlying service on port 443 which doesn't work.
> 
I think you are right. @shuo-wu Please help review this. Thanks.

shuo-wu (approved):
LGTM

PR #4067, opened 2022-06-02
Move replicas from UI and driver to values

wirwolf:
Hello. I have some wrong behavior when i drain the node in my cluster. When I drain the node where run longhorn-ui, I see an error in UI because replicas of UI is only one and i am forced to wait when the container start on another node. I want to have possibility scale UI for high availability


chart/templates/deployment-driver.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   namespace: {{ include "release_namespace" . }}
   labels: {{- include "longhorn.labels" . | nindent 4 }}
 spec:
-  replicas: 1
+  replicas: {{ .Values.longhornDriver.replicas }}
```
innobead, line 8: For the driver, 1 should be intended so probably we don't need to export it. cc @shuo-wu

wirwolf: Fixed

innobead (approved):

Issue #4049, opened 2022-05-31
[BUG] Longhorn can't handle restore command error in DR volume incremental restore


cchien816:
## Describe the bug
From [e2e](https://ci.longhorn.io/job/public/job/master/job/sles/job/amd64/job/longhorn-tests-sles-amd64/139/testReport/junit/tests/test_ha/test_dr_volume_with_restore_command_error_s3_/)

Incremental restore will fail in DR volume when set restore command error

Tested on master-head 20220531
Not happen on V1.2.4

## To Reproduce

1. Set a random backupstore.
2. Create a volume, then create the corresponding PV, PVC and Pod.
3. Write data to the pod volume and get the md5sum
4. Create the 1st backup.
5. Create a DR volume from the backup.
6. Wait for the DR volume restore complete.
7. Create a non-empty directory `volume-delta-<last backup name>.img`
       in one replica directory of the DR volume. This will fail the
       restore command call later.
8. Write data to the original volume then create the 2nd backup.
9. Wait for incremental restore.

In step 9, the incremental restore won'y start or complete immediately without failed replica in DR volume

## Expected behavior

Longhorn can capture and handle the restore command error

- Incremental restore complete.
- DR volume is Degraded
- One failed replica.

## Log or Support bundle

N/A

## Environment

 - Longhorn version:
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl): Kubectl
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: K3s
   - Number of management node in the cluster: 1 
   - Number of worker node in the cluster: 3
 - Node config
 - Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal): VMWare
 - Number of Longhorn volumes in the cluster: 1

## Additional context

N/A

innobead:
cc @c3y1huang

derekbit:
```
time="2022-06-01T02:49:45Z" level=info msg="Prepare to restore backup, backup target: s3://backupbucket@us-east-1/backupstore, backup volume: longhorn-testvol-vigot8-std, requested restored backup name: backup-befb04effb564042" controller=longhorn-engine engine=longhorn-testvol-vigot8-dr-e-5b080655 node=rancher60-worker3
[...]

PR #4068, opened 2022-06-02
Update longhorn-instance-manager to v1_20220602

derekbit:
longhorn/longhorn#4049


innobead (changes requested):
Also questions.yaml?

derekbit:
> Also questions.yaml?

Addressed. Thank you.

innobead (approved):

Issue #4069, opened 2022-06-02
[IMPROVEMENT] less verbose log on longhorn-manager


mfbrj:
## Is your improvement request related to a feature? Please describe

longhorn-manager pods currently use level=debug on logs, such as:
```
time="2022-05-25T16:52:21Z" level=debug msg="Found 0 node has at least one schedulable disk" accessMode=rwo controller=longhorn-volume frontend=blockdev migratable=false node=k8s-worker-a1 owner=k8s-worker-a1 state=attached volume=pvc-11f3737c-0f4e-4bce-8378-8bfaf7d37d6b
time="2022-05-25T16:52:27Z" level=debug msg="Found 2 use zones [anexo1 anexo3]" accessMode=rwo controller=longhorn-volume frontend=blockdev migratable=false node=k8s-worker-a1 owner=k8s-worker-a1 replicaAutoBalanceType=zone state=attached volume=pvc-5ce9de33-3ed4-4cc2-b8dc-52269212b021
time="2022-05-25T16:52:27Z" level=debug msg="Found 1 use nodes [k8s-worker-b2]" accessMode=rwo controller=longhorn-volume frontend=blockdev migratable=false node=k8s-worker-a1 owner=k8s-worker-a1 replicaAutoBalanceType=zone state=attached volume=pvc-5ce9de33-3ed4-4cc2-b8dc-52269212b021
time="2022-05-25T16:52:27Z" level=debug msg="Balanced, 2 volume replicas are running on different zones" accessMode=rwo controller=longhorn-volume frontend=blockdev migratable=false node=k8s-worker-a1 owner=k8s-worker-a1 replicaAutoBalanceType=zone state=attached volume=pvc-5ce9de33-3ed4-4cc2-b8dc-52269212b021
time="2022-05-25T16:52:27Z" level=debug msg="Found 0 replica candidate for auto-balance" accessMode=rwo controller=longhorn-volume frontend=blockdev migratable=false node=k8s-worker-a1 owner=k8s-worker-a1 replicaAutoBalanceOption=least-effort state=attached volume=pvc-5ce9de33-3ed4-4cc2-b8dc-52269212b021
[...]

PR #4078, opened 2022-06-06
Add debug logs option to helm chart

corybolar:
Debug logs in helm deployments are enabled and hardcoded.  This results in large log files that are unnecessary. Add a setting in helm chart values to control this behavior defaulting to off.

Fixes: #4069

Signed-off-by: Cory Bolar <cory.bolar@gmail.com>


innobead (commented):
Ref: https://github.com/longhorn/longhorn/issues/4069#issuecomment-1165282395

Thanks for the contribution, but I would suggest having logs in the right level first instead of just enabling/disabling the current debug log during the deployment because it will probably cause another problem related to troubleshooting at runtime.

SPodjasek:
I came across your PR after posting mine in a similar topic #4177, and I wonder whether you would consider changing your value path to `longhornManager.logs.debug`. Or maybe I'll move mine to `longhornManager.jsonLogs` and make it `boolean`. Whatever suits this project best.

thunko:
hi, is there an option to achieve this with with the current way. I have json logs already set.

Issue #3714, opened 2022-03-08
[TASK] Create a migration path from deprecated Rancher catalog to Rancher marketplace


innobead:
## What's the task? Please describe

Longhorn supports 3 ways of installation including the Rancher app chart. Lots of community users installed Longhorn from the Rancher catalog which is getting deprecated and has been replaced by the new Rancher marketplace. 

There is no official way to migrate between two places, so we need to find out a way to help users do the migration. 

## Describe the items of the task (DoD, definition of done) you'd like
> Please use a task list for items on a separate line with a clickable checkbox https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists

- [x] Document or script the solution of migration

## Additional context

https://github.com/longhorn/longhorn/issues/3424#issuecomment-998308675

cc @meldafrawi @kaxing

timmy59100:
A migration path  from Rancher catalog to helm3 installation would be nice. Rancher marketplace installation is dependent on rancher version. Which results in delayed longhorn versions available.

innobead:
> A migration path from Rancher catalog to helm3 installation would be nice. Rancher marketplace installation is dependent on rancher version. Which results in delayed longhorn versions available.

Actually, this part has been improved in recent Rancher releases, also if we can't catch up with the upcoming Rancher release, we will try to do an out-of-band release to catch that.

innobead:
@jenting this probably we also need to check Rancher code base between catalog (old UI) and marketplace (new UI).

jenting:
After migrating from the Rancher catalog app to the Rancher marketplace app by adding the annotations and labels to Longhorn CRD and other resources
- CRD
  ```yaml
  spec:
    annotations:
      meta.helm.sh/release-name=longhorn-crd
      meta.helm.sh/release-namespace=longhorn-system
    labels:
      app.kubernetes.io/managed-by=Helm
  ```
- Other Resources
  ```yaml
    annotations:
      meta.helm.sh/release-name=longhorn-system
      meta.helm.sh/release-namespace=longhorn-system
    labels:
      app.kubernetes.io/managed-by=Helm
  ```

We are able to use the Rancher marketplace app to manage/upgrade the Longhorn app.
However, we are unable to use `helm3` to list the `longhorn-crd` and `longhorn-system` chart releases.

But, if we use the Rancher marketplace app to upgrade the Longhorn app to a newer version, the `helm3` is able to list the `longhorn-crd` and `longhorn-system` chart release.

[...]

PR #4095, opened 2022-06-09
Add rancher chart migration script

PhanLe1010:
The script help to migrate Longhorn installed in the old Rancher catalog UI to the new chart in Rancher app&marketplace UI. More details about how the process look like is at https://github.com/longhorn/longhorn/issues/3714#issuecomment-1150714463

longhorn/longhorn#3714


c3y1huang (commented):
Can we consolidate the scripts with another usage option `--delete`?

shuo-wu (commented):
1. Actually we can check if setting `concurrent-automatic-engine-upgrade-per-node-limit` is value 0 before at the beginning of the 1st script since we already know the kubeconfig.
2. Similarly, I am thinking if the 2nd script needs to verify the 1st script being already executed or the annotations being already applied at the beginning.
3. One more question, can we directly install the new chart then we don't need to ask users to do any manual steps? If NO, it's better to describe the functionality of each part in `--help` description.

PhanLe1010:
Thanks @c3y1huang and @shuo-wu . I updatec the PR

> can we directly install the new chart then we don't need to ask users to do any manual steps?

I couldn't find a way to do so

scripts/longhorn_rancher_chart_migration.sh
```diff
@@ -0,0 +1,200 @@
...
+  echo "ERROR: command (kubectl get-all) is not found. Please install it here: https://github.com/corneliusweig/ketall#installation"
+  exit 1
+fi
+
+set -e
+
+usage() {
+  echo ""
+  echo "The migration includes:"
+  echo "1. Running the part 1 of the migration script to migrate the labels and annotations for Longhorn resources"
```
shuo-wu, line 15: The users may be confused about "part 1" or "part 2". You can put the example here rather than the end of the description.

scripts/longhorn_rancher_chart_migration.sh
```diff
@@ -0,0 +1,200 @@
...
+  echo "2. Manually installing Longhorn chart in app&marketplace UI"
+  echo "3. Running the part 2 of the migration script to remove the old Longhorn chart from old catalog UI"
+  echo ""
+  echo "usage:"
+  echo "$0 [options]"
+  echo "  -u | --upstream-kubeconfig:      upstream rancher cluster kubeconfig path"
+  echo "  -d | --downstream-kubeconfig:    downstream cluster kubeconfig path"
+  echo "  -p | --part:                     specify the part you want to run (1 or 2)"
```
shuo-wu, line 23: NIT: It's better to rename this arg. e.g., `-t | --type`, and there are 2 options: `migrate`/`label` and `cleanup`.

scripts/longhorn_rancher_chart_migration.sh
```diff
@@ -0,0 +1,200 @@
...
+echo ""
+echo "Rancher Project App '${RELEASE_NAME}' found:"
+echo "  Project-Namespace: ${RANCHERAPPNS}"
+echo "  Catalog: ${RANCHERAPPCATALOG}"
+echo "  Template: ${RANCHERAPPTEMPLATE} (${RANCHERAPPTEMPLATEVERSION})"
+echo "  Answers:"
+printf '%s\n' "${RANCHERAPPVALUES:-$RANCHERAPPANSWERS}"
+echo ""
+
+if [ "$PART" -eq "2" ] ; then
```
c3y1huang, line 116: Nit: maybe we can also consider using pause and wait so the user doesn't have to run the script twice?
```
echo "Migration is completed"
echo "You need to install Longhorn chart in app&marketplace UI before continuing to remove the old Longhorn chart from old catalog UI."
read -p "Press any key..."

longhorn_rancher_chart_migration.sh -u <upstream-path> -d <downstream-path>" --part 2
```

scripts/longhorn_rancher_chart_migration.sh
PhanLe1010: Users are supposed to be reading the migration document when running this script. The migration doc will walk them through each step. Do you think that we still need to put all details into this script?

scripts/longhorn_rancher_chart_migration.sh
PhanLe1010: Good idea! Fixed. Thank you

scripts/longhorn_rancher_chart_migration.sh
PhanLe1010: Between part 1 and part 2, the users have to install a new chart and verify that everything is working fine. That process might take a bit long time. Users might cancel or forget about this running script in the meantime. So, I don't see much benefit. What do you think? 

Also, we will have the document for user to follow step by step, so everything will be clear

c3y1huang: You are right, doesn't sound like user would benefit by the pause. Thanks!

scripts/longhorn_rancher_chart_migration.sh
```diff
@@ -0,0 +1,200 @@
...
+  echo "--upstream-kubeconfig is mandatory"
+  usage
+  exit 1
+fi
+if [ -z "$DOWNSTREAM_KUBECONFIG" ]; then
+  echo "--downstream-kubeconfig is mandatory"
+  usage
+  exit 1
+fi
+if [ -z "$TYPE" ]; then
```
c3y1huang, line 70: Should we validate the type input?

scripts/longhorn_rancher_chart_migration.sh
shuo-wu: OK. Then please provide the doc link in the description.

PhanLe1010: The doc is coming after this script is merged. Is it enough to have the doc references the script?

> And now part1 & part2 are renamed. It's better to update the wording here as well.

Thank you. Fixed

shuo-wu (approved):

scripts/longhorn_rancher_chart_migration.sh
```diff
@@ -0,0 +1,200 @@
...
+  echo "--upstream-kubeconfig is mandatory"
+  usage
+  exit 1
+fi
+if [ -z "$DOWNSTREAM_KUBECONFIG" ]; then
+  echo "--downstream-kubeconfig is mandatory"
+  usage
+  exit 1
+fi
+if [ "$TYPE" != "migrate" ] &&  [ "$TYPE" != "cleanup" ] ; then
```
derekbit, line 70: nit: I prefer `[ x"$TYPE" != x"migrate" ]`
https://stackoverflow.com/questions/1805663/shell-script-purpose-of-x-in-xvariable

innobead:
Before this merge, we need @longhorn/qa to verify this.

derekbit (approved):
LGTM

innobead:
@cchien816 Please help with verifying this PR and update the result here.

scripts/longhorn_rancher_chart_migration.sh
```diff
@@ -0,0 +1,200 @@
...
+RESOURCES=$(kubectl get-all --kubeconfig ${DOWNSTREAM_KUBECONFIG} --exclude AppRevision -o name -l io.cattle.field/appId=${RELEASE_NAME} 2>/dev/null | sort)
+if [[ "$RESOURCES" == "No resources"* ]]; then
+  RESOURCES=""
+fi
+
+echo ""
+echo "Patching CRD Resources ..."
+for resource in $RESOURCES; do
+  if [[ $resource == "customresourcedefinition.apiextensions.k8s.io/"* ]]; then
+    kubectl --kubeconfig ${DOWNSTREAM_KUBECONFIG} -n ${RELEASE_NAMESPACE} ${KUBECTL_DRY_RUN} annotate --overwrite ${resource} "meta.helm.sh/release-name"="longhorn-crd" "meta.helm.sh/release-namespace"="${RELEASE_NAMESPACE}" "helm.sh/resource-policy"="keep"
```
innobead, line 172: `"helm.sh/resource-policy"="keep"` I know you want to keep resources safe, but does it mean these resources will be out of helm management later? I think this will not impact chart upgrade, but how about uninstall longhorn?

cchien816 (dismissed):

innobead:
@PhanLe1010 Please help update https://github.com/longhorn/longhorn/issues/3714#issuecomment-1150714463 to reflect to the current change, then @longhorn/qa can help with the following testing.

innobead:
cc @PennyScissors

yangchiu:
> Thanks @c3y1huang and @shuo-wu . I updated the PR
> 
> > can we directly install the new chart then we don't need to ask users to do any manual steps?
> 
> I couldn't find a way to do so

Can https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/app_v2 be used to automate the manual steps?

PhanLe1010:
Thanks @cchien816 for reporting. Fixed the mistake. Could you give it a try again

PennyScissors:
I don't have permission to set myself as a reviewer and approve but LGTM.

scripts/longhorn_rancher_chart_migration.sh
PhanLe1010: Yeah, users will need to manually remove the Longhorn CRDs after uninstalltion. I don't think that this is a big concern over the safety benefit.

cchien816 (approved):
**Verified on `935092`
Result Pass**

cchien816:
Hi @PhanLe1010 

I verified with below combination, and migrate script all work well. 
- Migrate v1.2.0 from old UI to v1.2.2 new UI
- Migrate v1.2.1 from old UI to v1.2.2 new UI
- Migrate v1.2.2 from old UI to v1.2.2/v1.2.3/v1.2.4 new UI
- Migrate v1.2.3 from old UI to v1.2.3/v1.2.4 new UI
- Migrate v1.2.4 from old UI to v1.2.4 new UI

But I encounter a coroner case

PhanLe1010:
Thanks @cchien816 for the good catch with multiple cluster scenarios!
I fixed it. Could you give it a try again?

scripts/longhorn_rancher_chart_migration.sh
shuo-wu: One question, can this script help remove this annotation after the migration or at the end of the cleanup part?

scripts/longhorn_rancher_chart_migration.sh
```diff
@@ -0,0 +1,201 @@
...
+
+echo ""
+echo "Patching Other Resources ..."
+for resource in $RESOURCES; do
+  if [[ $resource == "customresourcedefinition.apiextensions.k8s.io/"* ]]; then
+    continue
+  fi
+  kubectl --kubeconfig ${DOWNSTREAM_KUBECONFIG} -n ${RELEASE_NAMESPACE} ${KUBECTL_DRY_RUN} annotate --overwrite ${resource} "meta.helm.sh/release-name"="longhorn" "meta.helm.sh/release-namespace"="${RELEASE_NAMESPACE}"
+  kubectl --kubeconfig ${DOWNSTREAM_KUBECONFIG} -n ${RELEASE_NAMESPACE} ${KUBECTL_DRY_RUN} label --overwrite ${resource} "app.kubernetes.io/managed-by"="Helm"
+done
```
shuo-wu, line 188: Will this update all longhorn CRs? If yes, will be there a scalability issue?

PhanLe1010: The script only consider the resource managed by the helm 2 (which has the label `-l io.cattle.field/appId=${RELEASE_NAME} `) . So I think the answer is no, the script will not update CRs of Longhorn CRDs. 

```
RESOURCES=$(kubectl get-all --kubeconfig ${DOWNSTREAM_KUBECONFIG} --exclude AppRevision -o name -l io.cattle.field/appId=${RELEASE_NAME} 2>/dev/null | sort)
```

@cchien816 When you retest the issue could you check:
1. Create some Longhorn volumes 
1. running the script with the flag `--type=migrate`
2. Check if the the volume CR has the label `app.kubernetes.io/managed-by"="Helm"` (could you paste it here)

Thank you very much!

scripts/longhorn_rancher_chart_migration.sh
PhanLe1010: It can but, I don't think that it is necessary since at this point user no longer need Longhorn. How do you think?

scripts/longhorn_rancher_chart_migration.sh
shuo-wu: BTW, please check the backup CRs as well.

scripts/longhorn_rancher_chart_migration.sh
shuo-wu: After removing the annotations, I think users would uninstall Longhorn via the new Rancher APP UI, right? But it's nice-to-have rather than must-have anyway and we don't need to spend too much time on it.

scripts/longhorn_rancher_chart_migration.sh
PhanLe1010: cc @cchien816 Could we prioritize this ticket as it is needed to decide to stop the 1.3.0 chart release for old rancher UI

cchien816: Hi @PhanLe1010 , @shuo-wu ,

After command `-t migrate` run,  I can see label appear `app.kubernetes.io/managed-by"="Helm"` in CRD `backups.longhorn.io `and `volumes.longhorn.io` as below

```
"labels": {
            "app.kubernetes.io/instance": "longhorn",
            "app.kubernetes.io/managed-by": "Helm",
            "app.kubernetes.io/name": "longhorn",
            "app.kubernetes.io/version": "v1.2.0",
            "helm.sh/chart": "longhorn-1.2.0",
            "io.cattle.field/appId": "longhorn-system",
            "longhorn-manager": "Volume"
        },
```
```
"labels": {
            "app.kubernetes.io/instance": "longhorn",
            "app.kubernetes.io/managed-by": "Helm",
            "app.kubernetes.io/name": "longhorn",
            "app.kubernetes.io/version": "v1.2.0",
            "helm.sh/chart": "longhorn-1.2.0",
            "io.cattle.field/appId": "longhorn-system",
            "longhorn-manager": "Backup"
        },

```

cchien816:
Hi @PhanLe1010 , I verified again on `8e3664` with below combination, and migrate script all work well.
Result passed

Rancher server imported single cluster and imported multiple clusters were both verified
- Migrate v1.2.0 from old UI to v1.2.2 new UI
- Migrate v1.2.1 from old UI to v1.2.2 new UI
- Migrate v1.2.2 from old UI to v1.2.2/v1.2.3/v1.2.4 new UI
- Migrate v1.2.3 from old UI to v1.2.3/v1.2.4 new UI
- Migrate v1.2.4 from old UI to v1.2.4 new UI

scripts/longhorn_rancher_chart_migration.sh
shuo-wu: @cchien816 I mean the backup CRs `kubectl -n longhorn-system get lhb` rather than CRD `kubectl get crd backups.longhorn.io`. Can you check if the labels or annotations in the CRs need to be updated?

PhanLe1010: I am checking them

scripts/longhorn_rancher_chart_migration.sh
PhanLe1010: Yeah, after the migration, the user will use the new Rancher UI to upgrade/uninstall Longhorn. 
Then the user has to delete the Longhorn CRD manually since the "helm.sh/resource-policy"="keep" annotation instructs Helm to keep Longhorn CRDs for safety. It is true that we can modify the script to help user to remove the annotation and thus remove the CRDs but I think it does not have much benefit

PhanLe1010:
@innobead All concerns are addressed. Could you please help merging this PR?

PR #4177, opened 2022-06-24
Add value to output manager logs to json format

SPodjasek:
As stated above I miss a flag to set logs output format to json which was implemented in longhorn/longhorn-manager#605.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -125,6 +125,9 @@ privateRegistry:
   registrySecret: ~
 
 longhornManager:
+  log:
+    ## Allowed values are `plain` or `json`.
+    format: plain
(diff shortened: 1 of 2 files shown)


chart/values.yaml
```diff
@@ -125,6 +125,9 @@ privateRegistry:
   registrySecret: ~
 
 longhornManager:
+  logs:
```
innobead, line 128: I would suggest using `logging` or `log`.

SPodjasek: Changed

SPodjasek:
@innobead Rebased to current `master`

innobead (approved):

PR #4257, opened 2022-07-20
Annotation for service LonghornManager will be configurable.

aamkye:
I found that there is no possibility to add those annotations to the Longhorn Manager service via helm-chart:
```
  prometheus.io/scrape: 'true'
  prometheus.io/port: '9500'
  prometheus.io/path: '/metrics'
```
So, here it is.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -142,7 +142,7 @@ longhornManager:
   #  label-key1: "label-value1"
   #  label-key2: "label-value2"
   serviceAnnotations: {}
-  ## If you want to set annotations for the Longhorn Manager delete the `{}` in the line above
+  ## If you want to set annotations for the Longhorn Manager service, delete the `{}` in the line above
   ## and uncomment this example block
   #  annotation-key1: "annotation-value1"
   #  annotation-key2: "annotation-value2"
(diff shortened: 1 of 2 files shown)


innobead (commented):
LGTM, but we do support https://longhorn.io/docs/1.3.0/monitoring/prometheus-and-grafana-setup/#install-longhorn-servicemonitor. Have you trited this instead?

aamkye:
Yes, I've checked it, but in general I wanted to try setup in more “standard” way without “ServiceMonitor” object.

aamkye:
Except that, cert-manager for example, allows adding annotations for its service, so Longhorn also should IMHO.

chart/values.yaml
```diff
@@ -141,6 +141,11 @@ longhornManager:
   ## and uncomment this example block
   #  label-key1: "label-value1"
   #  label-key2: "label-value2"
+  serviceAnnotations: {}
+  ## If you want to set annotations for the Longhorn Manager delete the `{}` in the line above
```
innobead, line 145:
```suggestion
  ## If you want to set annotations for the Longhorn Manager service, delete the `{}` in the line above
```

innobead (changes requested):

innobead (approved):

innobead (approved):

Issue #4246, opened 2022-07-18
[IMPROVEMENT] Recognize default toleration and node selector to allow Longhorn run on the RKE mixed cluster


innobead:
## Is your improvement request related to a feature? Please describe

When users install Longhorn on an RKE mixed cluster with Win nodes, users need to manually update the tolerations and node selector for user deployed components and the default settings for system-managed ones. 

However, RKE mixed clusters provisioned by Rancher will have some taint tolerations and node labels for Linux nodes, so it will be good to have an auto-recognized mechanism to allow Longhorn to install successfully w/o any manual intervention to provide a better user experience. 

## Describe the solution you'd like

https://github.com/rancher/rancher/issues/37030#issuecomment-1186876148

## Describe alternatives you've considered

As now, users just update those settings by themselves. 

## Additional context

https://github.com/rancher/rancher/issues/37030#issuecomment-1185787616

innobead:
cc @longhorn/qa

innobead:
@c3y1huang if having any questions, check with @PhanLe1010 . He has the whole context. Thanks.

cchien816:
**Verified on longhorn `81a094`
Result Pass**

Steps
1. Deploy Rancher mixed cluster (1 Windows node, 3 Linux node)
2. Download kubeconfig from Rancher in exported to local
3. In local, clone https://github.com/longhorn/longhorn.git, create namespace `longhorn-system`
4. Change `windowsCluster` enable to `true` in chart/values.yaml
```
  windowsCluster:
      # Enable this to allow Longhorn to run on the Rancher deployed Windows cluster
      enabled: true
```
5. Run `helm install longhorn ./longhorn/chart -n longhorn-system`
6. Observed Longhorn deploy well

c3y1huang:
@derekbit instead of reopening this, can you create another issue for the fix on the generated longhorn.yaml? 

cc @innobead

innobead:
> @derekbit instead of reopening this, can you create another issue for the fix on the generated longhorn.yaml?
> 
> cc @innobead

Yeah, I think this is fair because this is not about function not working but rather a side-effect. Let's create an issue to track the fix and re-connect to this issue. @derekbit Could you help with that? Thanks!

PR #4262, opened 2022-07-21
chart: support Rancher deployed Windows Cluster

c3y1huang:
https://github.com/longhorn/longhorn/issues/4246

### DIFF
--- a/chart/templates/default-setting.yaml
+++ b/chart/templates/default-setting.yaml
@@ -20,8 +20,26 @@ data:
     {{ if not (kindIs "invalid" .Values.defaultSettings.defaultDataLocality) }}default-data-locality: {{ .Values.defaultSettings.defaultDataLocality }}{{ end }}
     {{ if not (kindIs "invalid" .Values.defaultSettings.defaultLonghornStaticStorageClass) }}default-longhorn-static-storage-class: {{ .Values.defaultSettings.defaultLonghornStaticStorageClass }}{{ end }}
     {{ if not (kindIs "invalid" .Values.defaultSettings.backupstorePollInterval) }}backupstore-poll-interval: {{ .Values.defaultSettings.backupstorePollInterval }}{{ end }}
-    {{ if not (kindIs "invalid" .Values.defaultSettings.taintToleration) }}taint-toleration: {{ .Values.defaultSettings.taintToleration }}{{ end }}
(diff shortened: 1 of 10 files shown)


chart/templates/default-setting.yaml
```diff
@@ -20,8 +20,16 @@ data:
...
     {{ if not (kindIs "invalid" .Values.defaultSettings.backupstorePollInterval) }}backupstore-poll-interval: {{ .Values.defaultSettings.backupstorePollInterval }}{{ end }}
-    {{ if not (kindIs "invalid" .Values.defaultSettings.taintToleration) }}taint-toleration: {{ .Values.defaultSettings.taintToleration }}{{ end }}
-    {{ if not (kindIs "invalid" .Values.defaultSettings.systemManagedComponentsNodeSelector) }}system-managed-components-node-selector: {{ .Values.defaultSettings.systemManagedComponentsNodeSelector }}{{ end }}
+    {{- if or (not (kindIs "invalid" .Values.defaultSettings.taintToleration)) (not (kindIs "invalid" .Values.global.cattle.windowsCluster.enabled)) }}
+    taint-toleration: {{$windowsDefaultSettingTaintToleration := list -}}
+      {{- if .Values.global.cattle.windowsCluster.enabled }}{{ $windowsDefaultSettingTaintToleration = .Values.global.cattle.windowsCluster.defaultSetting.taintToleration }}{{ end -}}
+      {{- $taintToleration := list $windowsDefaultSettingTaintToleration .Values.defaultSettings.taintToleration }}{{ join ";" (compact $taintToleration) -}}
```
shuo-wu, line 26: Should we check `(not (kindIs "invalid" .Values.defaultSettings.taintToleration))` before applying this?

c3y1huang: Updated, thanks!

c3y1huang: Updated, thanks!

shuo-wu (approved):

chart/values.yaml
```diff
@@ -4,6 +4,22 @@
 global:
   cattle:
     systemDefaultRegistry: ""
+    windowsCluster:
+      # Enable this to allow Longhorn to run on the Rancher deployed Windows cluster
+      enabled: false
```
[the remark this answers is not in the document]
innobead: After discussing with @c3y1huang , we agree that adding the vendor-specific annotations automatically is not suitable here, so we will just keep the current way as is first, then see the feedback.

innobead (approved):
LGTM

innobead:
@c3y1huang Please backport to 1.3.x and 1.2.x. Thanks.

Issue #3898, opened 2022-04-27
[IMPROVEMENT] Support failed/obsolete orphaned backup cleanup


innobead:
## Is your improvement request related to a feature? Please describe

In some cases, there will be some failed backups still staying in the system, so we should provide an option for users to clean them up automatically. After failed backups cleanup, users should be still able to check the previous audit events of failed backups as usual. 

cc @joshimoo

innobead:
cc @PhanLe1010 @joshimoo @derekbit

innobead:
@mantissahz A new setting was introduced, so you need to update manifests in longhorn/longhorn as well w/ question.yaml updated as well.

chrisc1981:
**Verified on longhorn-manager master-head `76f0fb`
Result Pass**

Test steps or (https://github.com/longhorn/longhorn/issues/3898#issuecomment-1214773993):
1. Deploy Longhron and backupstore
2. Set `Failed Backup Time to Live ` to 5 (by default it's 14400 minutes)
3. Create a volume then attach to worker node and write some data
4. Create a backup to backup target then crash the instance-manager-r such as this command `kubectl delete pods -l longhorn.io/instance-manager-type=replica -n longhorn-system --wait`.
5. Then backup state should become Error in a few seconds.
6. Wait after `Failed Backup Time to Live ` passed
7. Verify failed backup had been deleted from UI and command `kubectl get backups.longhorn.io -n longhorn-system`
8. Test on both S3 and nfs type of backupstore

PR #4264, opened 2022-07-21
Extend LEP for orphaned data cleanup

mantissahz:
The LEP extends the design of the orphaned data cleanup for orphaned type `backup`.

longhorn/longhorn#3898

Signed-off-by: James Lu <james.lu@suse.com>

### DIFF
--- a/enhancements/20220324-orphaned-data-cleanup.md
+++ b/enhancements/20220324-orphaned-data-cleanup.md
@@ -1,38 +1,42 @@
-# Orphaned Replica Directory Cleanup
+# Orphaned Data Cleanup
 
 ## Summary
 
-Orphaned replica directory cleanup identifies unmanaged replicas on the disks and provides a list of the orphaned replica directory on each node. Longhorn will not delete the replicas automatically, preventing deletions by mistake. Instead, it allows the user to select and trigger the deletion of the orphaned replica directory manually or deletes the orphaned replica directories automatically.
+1. Orphaned replica directory cleanup identifies unmanaged replicas on the disks and provides a list of the orphaned replica directory on each node. Longhorn will not delete the replicas automatically, preventing deletions by mistake. Instead, it allows the user to select and trigger the deletion of the orphaned replica directory manually or deletes the orphaned replica directories automatically.
+
+2. Orphaned backup identifies a failed backup on the remote backup target or not when making a backup to remote backup target failed. Longhorn will not delete the backups automatically either. Also it allows the user to select and trigger the deletion of the orphaned backups manually or automatically.
 
 ### Related Issues
 
-[https://github.com/longhorn/longhorn/issues/685](https://github.com/longhorn/longhorn/issues/685)
+[[FEATURE] Clean up orphaned unused volume replicas](https://github.com/longhorn/longhorn/issues/685)
+
+[[IMPROVEMENT] Support failed/obsolete orphaned backup cleanup](https://github.com/longhorn/longhorn/issues/3898)
 
 ## Motivation
 
 ### Goals
 
-- Identify the orphaned replica directories
-- The scanning process should not stuck the reconciliation of the controller
-- Provide user a way to select and trigger the deletion of the orphaned replica directories
-- Support the global auto-deletion of orphaned replica directories
+- Identify the orphaned replica directories and orphaned backups
+- The scanning process should not stuck the reconciliation of the controllers
+- Provide user a way to select and trigger the deletion of the orphaned replica directories and backups
+- Support the global auto-deletion of orphaned replica directories and backups
+
 ### Non-goals
 
-- Clean up unknown files or directories in disk paths
+- Clean up unknown files or directories in disk paths or on the remote backup target.
 - Support the per-node auto-deletion of orphaned replica directories
 - Support the auto-deletion of orphaned replica directories exceeded the TTL
 
 ## Proposal
 
-1. Introduce a new CRD `orphan` and controller that represents and tracks the orphaned replica directories. The controller deletes the physical data and the resource if receive a deletion request.
-
+1. Introduce a new CRD `orphan` and controller that represents and tracks the orphaned replica directories and backups. The controller deletes the physical data and the resource if receive a deletion request.
 
 2. The monitor on each node controller is created to periodically collects the on-disk replica directories, compares them with the scheduled replica, and then finds the orphaned replica directories.
 
-    The reconciliation loop of the node controller gets the latest disk status and orphaned replica directories from the monitor and update the state of the node. Additionally, the `orphan` resources associated with the orphaned replica directories are created.
+    The reconciliation loop of the node controller gets the latest disk status and orphaned replica directories from the monitor and update the state of the node. Additionally, `orphan` resources with type `OrphanTypeReplica` associated with the orphaned replica directories are created.
 
     ```
-    
+
           queue           ┌───────────────┐             ┌──────────────────────┐
          ┌┐ ┌┐ ┌┐         │               │             │                      │
      ... ││ ││ ││ ──────► │   syncNode()  ├────────────►│     reconcile()      │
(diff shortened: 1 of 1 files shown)


enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -148,25 +191,26 @@ After the enhancement, Longhorn automatically finds out the orphaned replica dir
...
-- `orphan` CRs will be removed when the replica directories disappear.
-- `orphan` CRs will be removed when the node/disk is evicted or down. The associated orphaned replica directories should not be cleaned up.
+- `orphan` CRs with `OrphanTypeReplica` type will be created correctly in the disk path. And they can be cleaned up with the directories.
+- `orphan` CRs with `OrphanTypeReplica` type will be created correctly when there are multiple kinds of files/directories in the disk path. And they can be cleaned up with the directories.
+- `orphan` CRs with `OrphanTypeReplica` type will be removed when the replica directories disappear.
+- `orphan` CRs with `OrphanTypeReplica` type will be removed when the node/disk is evicted or down. The associated orphaned replica directories should not be cleaned up.
+- `orphan` CRs with `OrphanTypeBackup` type will be created correctly when there are failed/unknown backups. And they can be cleaned up with the failed/unknown bakcups.
+- `orphan` CRs with `OrphanTypeBackup` type will be removed when the a backup volume deleted.
```
c3y1huang, line 213:
```suggestion
- `orphan` CRs with `OrphanTypeBackup` type will be removed when the backup volume is deleted.
```

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -48,42 +52,74 @@ Orphaned replica directory cleanup identifies unmanaged replicas on the disks an
...
+  - Users can list the orphaned replica directories by `kubectl -n longhorn-system get orphans`.
+  - Users can delete the orphaned replica directories by `kubectl -n longhorn-system delete orphan <name>`.
+  - Users can enable the global auto-deletion by `kubectl -n longhorn-system edit settings orphan-auto-deletion`
 
 ## Design
 
 ### Implementation Overview
+
 **Settings**
   - Add setting `orphan-auto-deletion`. Default value is `false`.
```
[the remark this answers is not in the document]
mantissahz: Yes, that's good. Maybe we would have another one orphaned type.

enhancements/20220324-orphaned-data-cleanup.md
mantissahz: Modified, thanks.

enhancements/20220324-orphaned-data-cleanup.md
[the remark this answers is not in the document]
c3y1huang: Or instead of `disable` and `all` user can select `replica`, `backup` from the UI and the setting can take a comma-separated string. When none or both is selected it's equivalent to `disable` and `all`. WDYT?

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -48,42 +52,78 @@ Orphaned replica directory cleanup identifies unmanaged replicas on the disks an
...
 
-After the enhancement, Longhorn automatically finds out the orphaned replica directories on Longhorn nodes. Users can visualize and manage the orphaned replica directories via Longhorn GUI or command line tools. Additionally, Longhorn can deletes the orphaned replica directories automatically if users enable the global auto-deletion option.
+#### Orphaned backup
+
+When a user or recurring job tries to make a backup and store it in the remote backup target, many situations will cause the backup procedure failed. In some cases, there will be some failed backups still staying in the Longhorn system and this kind of backups are no longer tracked by the Longhorn system. These backups becomes orphaned as well.
+
+After the enhancement, Longhorn automatically finds out the orphaned replica directories on Longhorn nodes and the orphaned backups of backup volumes. Users can visualize and manage the orphaned replica directories and backups via Longhorn GUI or command line tools. Additionally, Longhorn can deletes the orphaned replica directories and backups automatically if users enable the global auto-deletion option.
```
derekbit, line 94: This paragraph focuses on `orphaned backups`, so the `orphaned replica directories` can be removed here.

enhancements/20220324-orphaned-data-cleanup.md
```diff
@@ -113,23 +153,30 @@ After the enhancement, Longhorn automatically finds out the orphaned replica dir
...
-        - `<disk path>/replicas/<replica name>-<random string>/volume.meta` is parsible and follows the `volume.meta`'s format.
+
+      A valid replica directory has the properties:
+      - The directory name format is `<disk path>/replicas/<replica name>-<random string>`
+      - `<disk path>/replicas/<replica name>-<random string>/volume.meta` is parsible and follows the `volume.meta`'s format.
 
     - Compare the list of the orphaned replica directories with the `node.status.diskStatus.scheduledReplica` and find out the list of the orphaned replica directories. Store the list in `monitor.node.status.diskStatus.orphanedReplicaDirectoryNames`
 
+**Backup Controller**
+  - Start the monitor and sync the backup status with the monitor in each reconcile loop.
```
derekbit, line 164: Do we really need a monitor here?

enhancements/20220324-orphaned-data-cleanup.md
mantissahz: This might be like a conclusion of two paragraphs, but it could be separated.

enhancements/20220324-orphaned-data-cleanup.md
mantissahz: Backup monitor would be involved but not changed, do you think we still need to mention it about what it is doing?

enhancements/20220324-orphaned-data-cleanup.md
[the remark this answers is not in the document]
mantissahz: Agree

enhancements/20220324-orphaned-data-cleanup.md
derekbit: Got it. Thanks.

enhancements/20220324-orphaned-data-cleanup.md
derekbit: What's the purpose of the backup monitor here? I think whether or not to have a monitor depends on how you identify a failed backup.

enhancements/20220324-orphaned-data-cleanup.md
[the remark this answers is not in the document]
mantissahz: Modified, thanks

shuo-wu (commented):
Thank you for extending the LEP!

mantissahz:
>  Does error/unknown backup cleanup involve in the orphan mechanism?

@shuo-wu Yes, error/unknown backups cleanup involves in the orphan mechanism.

> If a Backup CR needs to be removed, the caller can be BackupVolumeController 

Thanks, I would modify this part. It would be BackupVolumeController to remove a Backup CR.

mantissahz:
After a discussion with @shuo-wu, I would like to close this LEP and have a new LEP for failed backups cleanup mechanism without a new orphaned type, there are some reasons why I close this one:
1. We already have the Backup CR to handle the backup resources and it is not like `orphaned replica` is not owned by any volume at the beginning. And both in this LEP or the new LEP we would delete the failed backup by BackupVolume controller.
2. Cascading deletion of `Orphaned` CR and `Backup` CR would be more complicated than we just handle the failed backups immediately when backup procedure failed.
3. Listing orphaned backups and failed backups on both two UI pages `Orphaned Data` and `Backup` might be a bit confusing for users. Deleting items manually on anyone of two pages would be involved in what it mentioned at part 2.

WDYT? @innobead, @shuo-wu, @c3y1huang, @derekbit.

derekbit:
> After a discussion with @shuo-wu, I would like to close this LEP and have a new LEP for failed backups cleanup mechanism without a new orphaned type, there are some reasons why I close this one:

Sounds good to me. Let orphaned data cleanup focus on the data without owners.


> Cascading deletion of Orphaned CR and Backup CR would be more complicated than we just handle the failed backups immediately when backup procedure failed.

Does it mean the failed backup will be deleted once the failure happens if user enables the auto-deletion?
Or, should we always delete failed backups?

mantissahz:
> Does it mean the failed backup will be deleted once the failure happens if user enables the auto-deletion?
Or, should we always delete failed backups?

@derekbit Yes, failed backups will be deleted once auto-deletion is enable.
Although we have event logs to record the backup failed, 
I am afraid of that users do not know what happened if backup failed and there is no anything left on the UI or CLI then users would not go to check the event log either.

c3y1huang:
> @derekbit Yes, failed backups will be deleted once auto-deletion is enable. Although we have event logs to record the backup failed, I am afraid of that users do not know what happened if backup failed and there is no anything left on the UI or CLI then users would not go to check the event log either.

Maybe it is ok if we mention this in the setting definition and the doc.

mantissahz:
Closed because https://github.com/longhorn/longhorn/pull/4334

Issue #4136, opened 2022-06-20
[IMPROVEMENT] Replace the GCC allocator in liblonghorn with a more efficient memory allocator


derekbit:
## Is your improvement request related to a feature? Please describe

Replace the gcc allocator in liblonghorn with a more efficient memory allocator

## Describe the solution you'd like

The memory buffers are frequently allocated and freed in liblonghorn during sending and receiving requests. GCC's memory allocator is less efficient than other memory allocators such as [tcmalloc](https://github.com/google/tcmalloc), [jemalloc](https://github.com/jemalloc/jemalloc), [mimalloc](https://github.com/microsoft/mimalloc) and etc.

shuo-wu:
Do you have a rough benchmark/estimation over the improvement after switching to `tcmalloc`?

innobead:
We can just follow up on the e2e result to see if any outstanding impact for this introduction. 

@yangchiu Please verify this by running the benchmark pipeline to compare the result before and after introducing this memory allocator, tcmalloc. Close it afterward. Thanks.

liyimeng:
> I did the test on Equinix m3.small.x86 machin. The read iops increases by 3%, and the read latency decreases by 10%.
> The value may differ in different platform.


What about write, any improvement?

yangchiu:
**Test environment:**

machine: AWS c5d.2xlarge
test size: 80G
master-head longhorn-manager version: 70203d6

**Test Result:**

Bandwidth: KiB/s
Latency: ns

_(1) Replica count = 1_

  |   | IOPS |   | Bandwidth |   | Latency |  
-- | -- | -- | -- | -- | -- | -- | --
  |   | read | write | read | write | read | write
v1.3.0 | rand | 22,606 | 16,282 | 362,226 | 161,778 | 353,714 | 303,726
  | seq | 37,057 | 28,135 | 372,140 | 164,233 | 285,136 | 303,390
master-head | rand | 23,040 | 16,105 | 362,268 | 161,873 | 354,084 | 298,671
  | seq | 36,323 | 28,391 | 372,157 | 164,191 | 279,423 | 300,060
Improvement | rand | 1.92% | -1.09% | 0.01% | 0.06% | -0.1% | 1.66%
  | seq | -1.98% | 0.91% | 0% | -0.03% | 2% | 1.1%

_(2) Replica count = 2_

  |   | IOPS |   | Bandwidth |   | Latency |  
-- | -- | -- | -- | -- | -- | -- | --
[...]

PR #4269, opened 2022-07-22
Update longhorn-instance-manager to v1_20220722

derekbit:
[Longhorn 4136](https://github.com/longhorn/longhorn/issues/4136)


innobead (approved):

innobead:
Need to backport to 1.3 and 1.2.

PR #4316, opened 2022-07-29
Add PSP support to environment_check.sh

spijet:
This will allow hardened cluster users (e.g. RKE2 with `cis-1.6` profile enabled) to run the environment check script painlessly.

I'm not sure if this behaviour should be on by default, so I'm willing to add a knob to turn the PSP support on or off as needed. Or maybe try to auto-detect it by asking cluster. :)

### DIFF
--- a/scripts/environment_check.sh
+++ b/scripts/environment_check.sh
@@ -131,6 +131,56 @@ check_dependencies() {
 
 create_ds() {
 cat <<EOF > $TEMP_DIR/environment_check.yaml
+---
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  name: longhorn-environment-check-sa
+  namespace: default
+  labels:
+    app: longhorn-environment-check
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: longhorn-environment-check-role
+  labels:
+    app: longhorn-environment-check
+rules:
+- apiGroups: ['policy']
+  resources: ['podsecuritypolicies']
+  verbs:     ['use']
+  resourceNames:
+  - longhorn-environment-check-psp
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: longhorn-environment-check-binding
+  labels:
+    app: longhorn-environment-check
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: longhorn-environment-check-role
+subjects:
+  - kind: ServiceAccount
+    name: longhorn-environment-check-sa
+    namespace: default
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: longhorn-environment-check-psp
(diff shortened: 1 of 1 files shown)


scripts/environment_check.sh
```diff
@@ -127,6 +127,56 @@ check_dependencies() {
...
+  labels:
+    app: longhorn-environment-check
+spec:
+  privileged: true
+  allowPrivilegeEscalation: true
+  allowedCapabilities: ['*']
+  volumes: ['*']
+  hostNetwork: true
+  hostIPC: true
 apiVersion: apps/v1
```
shuo-wu, line 180: It seems that `---` is missed

scripts/environment_check.sh
```diff
@@ -127,6 +127,56 @@ check_dependencies() {
 
 create_ds() {
 cat <<EOF > $TEMP_DIR/environment_check.yaml
+---
```
shuo-wu, line 130: It seems that this `---` is redundant.

shuo-wu (changes requested):
Or maybe we can add an option/arg to enable the PSP support. By default, this option can be disabled.

spijet:
Hello everyone!
I'm currently out of town, will get back to this issue when I'm back home.

innobead:
FYI. From 1.4, the PSP enablement will be removed by default, so basically we don't need this anymore. Any concerns? cc @PhanLe1010

spijet:
I fixed the issue with wrongly-placed `---` in `environment_check` using the GitHub UI and forgot to add the sign-off. D:

spijet:
OK, I figured it out. :)

Now I'll check if I can add a knob to make it configurable.

spijet:
@shuo-wu I'm not sure if I fully understand how exactly the `environment_check.sh` script gets called when deploying Longhorn, so I'll need some help. :)

I can wrap this whole section into an `if [ "${LONGHORN_ENABLE_PSP}" = 'true' ]` or something like that and then we can pass this variable via environment, if that's OK?

scripts/environment_check.sh
```diff
@@ -131,6 +131,56 @@ check_dependencies() {
...
+kind: PodSecurityPolicy
+metadata:
+  name: longhorn-environment-check-psp
+  labels:
+    app: longhorn-environment-check
+spec:
+  privileged: true
+  allowPrivilegeEscalation: true
+  allowedCapabilities: ['*']
+  volumes: ['*']
```
shuo-wu, line 180: I am thinking if allowing `hostPath` only is enough

scripts/environment_check.sh
```diff
@@ -131,6 +131,56 @@ check_dependencies() {
...
+  hostNetwork: true
```
shuo-wu, line 181: Does this script touch the host network?

shuo-wu (changes requested):
Can we add an option for the script to disable/enable PSP? Since PSP is deprecated in Kubernetes v1.25.

innobead:
It seems pending for a while . Closing first but feel free to continue contributing. Thank you.

Issue #4165, opened 2022-06-22
[FEATURE] Restore BackingImage for BackupVolume in a new cluster


FrankYang0529:
## Is your feature request related to a problem? Please describe

A normal `Backup` for `Volume` can be restored in another cluster. However, a `Volume` based on a `BackingImage` **_can't_** be restored because we don't help users backup `BackingImage` in the backup target.

## Describe the solution you'd like

Store `BackingImage` to the backup target if a `BackupVolume` is based on it. Restore `BackingImage` without manual uploading backing images in a new cluster.

## Additional context
https://github.com/harvester/harvester/issues/2237

yasker:
Based on the user feedback, we might want to enhance this to be able to do delta backup/restore, so it would be easier to distribute golden images for volumes.

FrankYang0529:
I did some research about delta backup. There are two use cases, and I would like to know whether we want to implement the feature.

**Case 1: Users upload an image twice with different names**
In volume delta backup, we store volumes in different folders, so each volume only compares to itself. However, in BackingImage, we don't have this kind of relationship. To save storage space, we need to compare the new image with all existing images.

**Case 2: Users upload an image, use it to create a volume, and export the volume as a new image**
In this case, we can check the volume StorageClass to find the relationship between old and new images. However, in my experiment, it looks like we can't save much space. I uploaded a 609M BackingImage, created a 3G Volume, and exported the volume as an image. I set a block as 256 bytes, and we can only save 198,912 bytes.

```bash
> go run main.go origin.img first-exported.img
File 1 size:  637665280
File 2 size:  3221225472
Same bytes:  198912
```

Added a file with content `123` to the volume and exported an image again. We can save 1,455,144,192 bytes between two exported images.

```bash
> go run main.go first-exported.img second-exported.img
File 1 size:  3221225472
File 2 size:  3221225472
Same bytes:  1455144192
```

Based on the experiment result, we can only save space if we export a volume multiple times. However, if we store different BackingImages in different folders, it's hard to get this benefit, because exported BackingImages only link to the uploaded BackingImage. Exported BackingImages from the same volume don't link to each other.
```mermaid
graph TD;
    UploadedImage-->Volume;
    Volume-->FirstExportedImage;
    Volume-->SecondExportedImage;
```

Experiment code: https://go.dev/play/p/qVJrEY0ra6_x

FrankYang0529:
[...]

PR #4385, opened 2022-08-09
feat(lep): add backing-image-backup-restore

FrankYang0529:
Issue: https://github.com/longhorn/longhorn/issues/4165

PR: https://github.com/longhorn/backing-image-manager/pull/65, https://github.com/longhorn/longhorn-manager/pull/1465


enhancements/20220809-backing-image-backup-restore.md
```diff
@@ -0,0 +1,141 @@
...
+    string uuid = 2;
+	string backup_target = 3;
+	map<string, string> credential = 4;
+}
+```
+    
+2. Implement the `Backup` rpc in `manager` service:
+    1. Get `BackingImage` file path.
+    2. Setup `backupstore` driver.
+    3. Check whether the `BackingImage` is already existent in backup-target and the size is same:
```
c3y1huang, line 79: Would there be a case for the backing image when the size is the same but have different content?

enhancements/20220809-backing-image-backup-restore.md
```diff
@@ -0,0 +1,141 @@
...
+        - If yes, skip the operation.
+        - If not, Upload the `BackingImage` to `backupstore/backingimages/<backing-image-name>/backingimage`.
```
c3y1huang, line 81: What happens when the backing image is used for more than one volume?

enhancements/20220809-backing-image-backup-restore.md
FrankYang0529: Yes, it may happen, but I am unsure how to check the file's checksum without downloading it, so I choose the easiest way to determine it.

enhancements/20220809-backing-image-backup-restore.md
FrankYang0529: No, it doesn't change.

enhancements/20220809-backing-image-backup-restore.md
```diff
@@ -0,0 +1,141 @@
...
+
+message BackupRequest {
+	string name = 1;
+    string uuid = 2;
+	string backup_target = 3;
+	map<string, string> credential = 4;
+}
+```
+    
+2. Implement the `Backup` rpc in `manager` service:
```
shuo-wu, line 76: Notice that the restore function is implemented in `sync` service rather than `datasource` service, do we need to put this backup function in `sync` as well?

enhancements/20220809-backing-image-backup-restore.md
```diff
@@ -0,0 +1,141 @@
...
+    * If not, send a `Backup` request to `backing-image-manager`.
+
+#### Restore flow
+
+**longhorn-manager**
+1. Add a new soruce type and new parameters to `backingimagedatasource.longhorn.io`.
+```go
+const (
+    // ...
+	DataSourceTypeDownloadFromBackupTargetParameterBackupTarget     = "backup-target"
```
shuo-wu, line 100: NIT: Do we need to tell the backup target type? it can be identified by checking the path prefix.

enhancements/20220809-backing-image-backup-restore.md
```diff
@@ -0,0 +1,141 @@
...
+    - If not, check whether related `BackingImage` CR is in the cluster:
+        - If yes, skip download operation.
+        - If not, create a new `BackingImage` CR with `download-from-backup-target` source type.
+3. In `BackingImageDataSource` controller:
+    1. Add `securityContext` when creating the `backing-image-data-source` pod, so `backing-image-data-source` has enough permission operate NFS driver.
+    1. If `BackupTarget` is s3, mount `backup-target-credential` secret as a volume under `/backup-target-credential`.
+
+ **backing-image-manager**
+1. In `SyncingFile` struct, add a `DownloadFromBackupTarget` function:
+    1. Get s3 credential from `/backup-target-credential` folder.
```
shuo-wu, line 122: What if the credential changes? For example, the Longhorn suddenly switches to use another s3.

enhancements/20220809-backing-image-backup-restore.md
```diff
@@ -0,0 +1,141 @@
...
+
+[#4165](https://github.com/longhorn/longhorn/issues/4165)
+
+## Motivation
+
+### Goals
+
+1. A `BackupVolume` with non-empty `backingImageName` can be restored correctly in the new cluster without manual creating `BackingImage` again.
+2. Longhorn only upload a `BackingImage` when there is a `BackupVolume` based on it.
+3. Different `BackupVolume`s use a same `BackingImage`. The `BackingImage` is uploaded once.
```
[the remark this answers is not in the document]
FrankYang0529: I think we can consider this question first: If there is a `BackupVolume` based on a `BackingImage`, do we want to allow users to delete the `BackingImage`? Currently, we don't check this. However, if our answer is no, there is no `BackupVolume` depending on the previous `BackingVolume`, then we can safely overwrite it on the backup target.

enhancements/20220809-backing-image-backup-restore.md
```diff
@@ -0,0 +1,141 @@
...
+### Related Issues
+
+[#4165](https://github.com/longhorn/longhorn/issues/4165)
+
+## Motivation
+
+### Goals
+
+1. A `BackupVolume` with non-empty `backingImageName` can be restored correctly in the new cluster without manual creating `BackingImage` again.
+2. Longhorn only upload a `BackingImage` when there is a `BackupVolume` based on it.
```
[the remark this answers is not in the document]
FrankYang0529: Fixed it. Thank you.

enhancements/20220809-backing-image-backup-restore.md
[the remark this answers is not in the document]
shuo-wu: The answer is probably YES... Since I prefer not to add more restrictions after setting the backup target.

enhancements/20220809-backing-image-backup-restore.md
FrankYang0529: Yeah, it's good to handle all file uploads/downloads in the `sync` service. I will update it. Thank you.

enhancements/20220809-backing-image-backup-restore.md
FrankYang0529: I think we don't need it because we use one variable to set up the backupstore driver.

https://github.com/longhorn/backupstore/blob/56ddc538b85950b02c37432e4854e74f2647ca61/driver.go#L62-L74

enhancements/20220809-backing-image-backup-restore.md
FrankYang0529: This is a good question. The `BackingImageDataSource` pod is created when we want to download `BackingImage` from the backup target. If the secret is changed before the pod is created, it will use the wrong credential, so the `BackingImage` will not be ready. If the secret is changed after the pod is created, it depends on whether the original credential is still valid for the S3 backup target.

I suggest we add a webhook to check whether there is a backup/restore in progress when users want to change the backup target so that we can avoid this complex situation.

shuo-wu: OK. I was worried about the backing image manager pods... Do the BIM pods require the credential? If NO, then this should be fine. 
As for the backing image data source pods, the change of the credential may lead to the restore failure then Longhorn can retry it with the updated credential.

FrankYang0529: No, BIM pods don't need the credential. Thanks for clarifying the question.

innobead:
This will be postponed to 1.5.0. cc @FrankYang0529 @shuo-wu

FrankYang0529:
Close the PR since we will use VolumeSnapshot in the future.
ref: https://github.com/longhorn/longhorn/issues/4165#issuecomment-1338960402

Issue #4003, opened 2022-05-19
[IMPROVEMENT] Support K8s 1.25 by updating removed deprecated resource versions like PodSecurityPolicy


innobead:
## Is your improvement request related to a feature? Please describe

PodSecurityPolicy has been deprecated and will be removed from K8s 1.25, so we need to find an alternative way to resolve the need for PSP in Longhorn to support 1.25.

Also, some deprecated resource versions are also removed from 1.25. Need to resolve this via https://github.com/longhorn/longhorn/issues/4239 or even consider to backport this to 1.3 & 1.2 via an adaptive way to determine the K8s version of the cluster to use which API resource version if possible (except PSP, because it's totally removed instead of version bump). 

- Cronjob v1beta1 -> v1
- EndpointSlice v1beta1 -> v1
- Event v1beta1 -> v1
- HorizontalPodAutoscaler v2beta1 -> v2
- PodDisruptionBudget v1beta1 -> removed

Note: client-go is backward compatible with K8s any version.

> Compatibility: client-go <-> Kubernetes clusters
> Since Kubernetes is backwards compatible with clients, older client-go versions will work with many different Kubernetes cluster versions.

## Describe the solution you'd like

Deprecate PSP if it's not needed. Otherwise, we need an alternative solution like https://kubernetes.io/docs/tasks/configure-pod-container/migrate-from-psp/.

## Describe alternatives you've considered

N/A

## Additional context

https://www.kubernetes.dev/resources/release/#timeline
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-25
https://github.com/longhorn/longhorn/issues/4239

innobead:
This needs to consider to backport.

mattmeye:
This was released in [Longhorn v1.2.5-rc1 release](https://github.com/longhorn/longhorn/releases/tag/v1.2.5-rc1) but not in [Longhorn v1.2.5 release](https://github.com/longhorn/longhorn/releases/tag/v1.2.5). Is there an ETA for an official release?

innobead:
> This was released in [Longhorn v1.2.5-rc1 release](https://github.com/longhorn/longhorn/releases/tag/v1.2.5-rc1) but not in [Longhorn v1.2.5 release](https://github.com/longhorn/longhorn/releases/tag/v1.2.5). Is there an ETA for an official release?

It was wrongly mentioned in the rc release note. It is planned for 1.4.0 instead, the end of this year.

[...]

PR #4388, opened 2022-08-09
feat(system-backup/restore): add LEP

c3y1huang:
https://github.com/longhorn/longhorn/issues/1455

[View file](https://github.com/longhorn/longhorn/blob/7e9de910c369b114060d2531c701b2b3b3b08cd3/enhancements/20220913-longhorn-system-backup-restore.md)

### DIFF
--- a/enhancements/20220913-longhorn-system-backup-restore.md
+++ b/enhancements/20220913-longhorn-system-backup-restore.md
@@ -1,0 +1,723 @@
+# Longhorn System Backup/Restore
+
+## Summary
+
+This feature is to support the Longhorn system backup and restore. And also allows the user to rollback the Longhorn
+system to the previous healthy state
+after a failed upgrade.
+
+Currently, we have documents to guide users on how to restore Longhorn:
+- [Restore to a new cluster using Velero](https://longhorn.io/docs/1.3.0/advanced-resources/cluster-restore/restore-to-a-new-cluster-using-velero/)
+- [Restore to a cluster contains data using Rancher snapshot](https://longhorn.io/docs/1.3.0/advanced-resources/cluster-restore/restore-to-a-cluster-contains-data-using-rancher-snapshot/)
+
+However, the solution relies on third-party tools, not out-of-the-box, and involves tedious human intervention.
+
+With this new feature, Longhorn's custom resources will be backed up and bundled into a single system backup file, then
+saved to the remote backup target.
+
+Later, users can choose a system backup to restore to a new cluster or restore to an existing cluster; nevertheless,
+this allows for cluster rollback to fix the corrupted cluster state just after a failed upgrade.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/1455
+
+## Motivation
+### Goals
+
+- Support Longhorn system backup by backing up Longhorn custom resources, bundling them to a single file, and uploading
+  it to the backup target.
+- Support Longhorn system restore to a new/existing cluster from a system backup.
+- Support Longhorn system restoration to the previous healthy state when encountering the failed upgrade.
+- Support restoring volume from the `lastBackup` when the volume doesn't exist in the cluster during the system restore.
+- Support not restoring volume from the `lastBackup` when the volume exists in the cluster during the system restore.
+
+### Non-goals [optional]
+
+- This feature does not deploy the Longhorn cluster. Users need to have a running Longhorn cluster to run the system
+  restore.
+- Do backup/restore the downstream workloads attached to Longhorn volumes. For example, if a cluster has a Pod with a
+  PersistentVolumeClaim. Longhorn will restore the PersistentVolumeClaim, PersistentVolume, and Volume only. After a
+  successful system restoration, the user can re-deploy the pod with the same manifest.
+- Restore while there are volumes still attached.
+- Automatically create a system backup before an upgrade and run the system restore when the Longhorn upgrade fails.
+  Instead, the user needs to run a system backup and restore it manually.
+- Delete resources none-existing in the system backup. We can probably enhance it later and make this an option for the
+  users.
+- Support restore from `BackingImage`. **Dependent on https://github.com/longhorn/longhorn/issues/4165**
+
+## Proposal
+### System Backup
+
+```
+
+        |------------ [[ longhorn-backup-target ]] ----------|
+        |                     controller                     |
+        |                                                    |
+        v                                                    v
+[ SystemBackup ] ---> [[ longhorn-system-backup ]] ---> [ object store ]
+       CR                        controller                  |
+                                                             |
+                                    backupstore/system-backups/<longhorn-version>/<name>/
+                                                             |
+                                   [ system-backups.zip ] <--|--> [ system-backup.cfg ]
+```
+
+1. Introduce new [v1/systembackups](#longhorn-manager-http-api) HTTP APIs.
+1. Introduce a new [SystemBackup](#manager-systembackup-custom-resource) custom resource definition.
+    - A new custom resource triggers the creation of the system [resource](#system-backup-resources) backup.
+    - Deleting the custom resource triggers deletion of the system backup in the backup target.
+      This behavior is similar to the current backup resource handling.
+    - The system backups stored in the backup target will get synced to the custom resources list.
+1. Introduce new responsibility to `longhorn-backup-target` controller.
+    - Responsible for syncing system backups in the backup target to the `SystemBackup` list.
+1. Introduce a new `longhorn-system-backup` controller.
+    - Responsible for generating the system backup file, bundling them to a single file, and uploading it to the backup
+      target.
+      1. Generates the Longhorn resources YAML files.
+      1. Compress resources YAML files to a zip file.
+      1. Upload to the backup target `backupstore/system-backups/<longhorn-version>/<system-backup-name>`.
+      >**Note:** Do not create/upload the system backup when the `SystemBackup` is created by the backup target controller.
+
+    - Responsible for deleting system backup in the backup target.
+    - Responsible for updating `SystemBackup` status.
+    - Responsible for updating the error message in the `SystemBackup` status condition.
+      Reference [SystemRestore](#manager-systemrestore-custom-resource) condition as the example.
+1. Introduce `SystemBackup` webhook validator for [condition validation](#validator-system-backup).
+
+### System Restore
+
+```
+
+                                            [ system-backup.zip ]   [ backups ]]]
+                                                            ^          ^
+                                                            |          |
+            [ system-backup.cfg ] <-- backupstore/system-backups/...   |
+                                                            |          |
+                                                            |       backupstore/volumes/
+                                                            |          |
+                                                            |          |
+[ SystemRestore ] ---> [[ longhorn-system-restore ]] ---> [ object-store ]
+        CR                      controller                  |          |
+                                     |                      |          |
+                                     |                      |          |
+                                  [ Job ]                   |    [ system-backup.cfg: engine ]
+                                     |                      |          |               
+                                     v                      |          |               
+                [ system-backup.cfg: manager ] ---> [[ longhorn-system-rollout ]]  
+                                                             controller
+                                                            |          |
+                                                            |          v               
+                                                            | <--- [ Volume: from backup ]
+                                                            |
+                                                           V
+                                                    [ Resources ]]]]
+```
+
+1. Introduce new [v1/systemrestores](#longhorn-manager-http-api) HTTP APIs.
+1. Introduce a new [SystemRestore](#manager-systemrestore-custom-resource) custom resource definition.
+    - A new custom resource triggers the creation of a system restore job.
+    - Deleting the custom resource triggers the deletion of the system restore job.
+1. Introduce a new `longhorn-system-restore` controller.
+    - Responsible for creating a new system restore job that runs a `longhorn-system-rollout` controller.
+    - Responsible for deleting the system restore job.
+1. Introduce a new `longhorn-system-rollout` controller. This controller is similar to the `uninstall controller`.
+    - Run inside the pod created by the system restore job.
+    - Responsible for downloading the system-backup from the backup target.
+    - Responsible for restoring [resources](#system-backup-resources) from the system-backup file.
+    - Responsible for updating `SystemRestore` status during system restore.
+    - Responsible for adding `longhorn.io/last-system-restore`, and `longhorn.io/last-system-restore-at` annotation to the
+    restored resources.
+    - Responsible for adding `longhorn.io/last-system-restore-backup`annotation to the restored volume resources.
+    - Responsible for updating the error message in the [SystemRestore](#manager-systemrestore-custom-resource) status
+    condition.
+
+    > **Note:** There are 2 areas covered for cross version restoration:
+    > 1. The system restore will use the manager and engine image in the system backup config to run the `longhorn-system-rollout` so the controller is compatible with the restoring resources.
+    > 1. When the CustomResourceDefinition is missing the version for the restoring resource, the system restore doesn't replace or remove the existing CustomResourceDefinitions. Instead, the controller adds to its versions. So system restoration doesn't break existing resources.
+    >
+    > See [Specify multiple versions](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#specify-multiple-versions) for details.
+
+1. Introduce `SystemRestore` webhook validator for [condition validation](#validator-system-restore).
+
+
+### User Stories
+
+#### Longhorn system restore
+
+Before the enhancement, the user can follow the [documents](#summary) to leverage external solutions to backup and
+restore the Longhorn system.
+
+After the enhancement, the user can backup/restore Longhorn system to/from the backup target using this Longhorn native
+solution.
+
+#### Upgrade rollback (downgrade)
+
+Before the enhancement, the user cannot downgrade Longhorn.
+
+After the enhancement, the user can downgrade Longhorn when there is a pre-upgrade system backup in the backup target.
+
+> **Note:** For the Longhorn cluster version before v1.4.0, users still need to follow the Longhorn [documents](#summary)
+            for backup and restore the Longhorn system.
+
+### User Experience In Detail
+
+### Longhorn UI
+
+1. Go to `System Backup` from the Setting drop-down menu.
+    ```
+    | Dashboard | Node | Volume | Recurring Job | Backup | Setting v |
+                                                           + ======================= +
+                                                           | General                 |
+                                                           | Engine Image            |
+                                                           | Orphaned Data           |
+                                                           | Backing Image           |
+                                                           | Instance Manager Image  |
+                                                           | System Backup           |
+                                                           + ======================= +
+    ```
+1. View `System Backups` and `System Restores` on the same page.
+    ```
+    System Backups                                                                                      [Custom Column]
+    ====================================================================================================================
+    [Create] [Delete] [Restore]                                                       [Search Box   v ][__________][Go]
+                                                                                          + ======= +
+    ======================================================================================| Name    |===================
+    [] | Version  | Name   | State   | Error                                              | State   |
+    ---+----------+--------+--------+-----------------------------------------------------| Version |-------------------
+    [] | 1.4.0    | demo-1 | Error   | error uploading system backup: failed to execute:  + ======= + /engine-binaries/c3y1huang-research-000-lh-ei/longhorn [system-backup
+       :          :        :         : upload --source /tmp/demo-2.zip --dest s3://backupbucket@us-east-1/ --name demo-2 --manager-image c3y1huang/research:000-lh-manager
+       :          :        :         : --engine-image c3y1huang/research:000-lh-ei], output , stderr, time=\"2022-08-16T03:52:09Z\" level=fatal msg=\"Failed to run upload
+       :          :        :         : system-backup command\" error=\"missing required parameter --longhorn-version\"\n, error exit status 1
+    [] | 1.4.0    | demo-2 | Ready   |
+    ====================================================================================================================
+                                                     [<] [1] [>]
+
+    System Restores                                                                                     [Custom Column]
+    ====================================================================================================================
+    [Delete]                                                                           [Search Box   v ][_________][Go]
+                                                                                           + ======= +
+    ====================================================================================== | Name    |==================
+    [] | Name           | Version   | State        | Age   | Error                         | State   |
+    ---+----------------+-----------+--------------+-------+------------------------------ | Version |------------------
+    [] | demo-1-restore | v1.4.0    | Completed    | 2m26s |                               + ======= +
+    [] | demo-2-foobar  | v1.4.0    | Error        | 64s   | Download: sample error message
+    [] | demo-2-restore | v1.4.0    | Initializing | 1s    |
+    ===================================================================================================================
+                                                     [<] [1] [>]
+    ```
+
+### System Backup
+***Longhorn GUI***
+  - The user can create system backups to the backup target.
+  - The system backup will be uploaded to backup target `backupstore/system-backups/<longhorn-version>/<system-backup-name>`.
+  - The user can view the system backup status.
+
+***Command kubectl***
+  - The user can create `SystemBackup` to backup Longhorn system to the backup target.
+    ```yaml
+    apiVersion: longhorn.io/v1beta2
+    kind: SystemBackup
+    metadata:
+      name: demo-2
+      namespace: longhorn-system
+    ```
+  - The user can view the system backups.
+    ```
+    > kubectl -n longhorn-system get lhsb
+    NAMESPACE         NAME     VERSION   STATE   CREATED
+    longhorn-system   demo-1   v1.4.0    Error   2022-08-23T00:25:29Z
+    longhorn-system   demo-2   v1.4.0    Ready   2022-08-24T02:34:57Z
+    ```
+
+### System Restore
+***Longhorn GUI***
+  - The users can restore a system backup in the backup target.
+  - The users can view the system restore status.
+  - The users can restore from a different Longhorn version.
+
+***Command kubectl***
+  - The user can create `SystemRestore` to restore system backup in the backup target.
+    ```yaml
+    apiVersion: longhorn.io/v1beta2
+    kind: SystemRestore
+    metadata:
+      name: demo-2-restore
+      namespace: longhorn-system
+    spec:
+      systemBackup: demo-2
+    ```
+  - The users can view the system restores.
+    ```
+    > kubectl -n longhorn-system get lhsr
+    NAME             STATE         AGE
+    demo-1-restore   Completed     2m26s
+    demo-2-foobar    Error         64s
+    demo-2-restore   Initializing  1s
+    ```
+
+### API changes
+
+#### Longhorn manager HTTP API
+
+| Method     | Path                             | Description                                                            |
+| ---------- | -------------------------------- | ---------------------------------------------------------------------- |
+| **POST**   | `/v1/systembackups`              | Generates system backup file and upload to the backup target           |
+| **GET**    | `/v1/systembackups`              | Get all system backups. Including ones already exist in the backup target and ones that are initialized but do not exist in the backup target |
+| **DELETE** | `/v1/systembackups/{name}`       | Delete system backup saved in the backup target                        |
+| **POST**   | `/v1/systemrestores`             | Download a system backup from the backup target and restore it         |
+| **GET**    | `/v1/systemrestores`             | Get all system restores                                                |
+| **DELETE** | `/v1/systemrestores/{name}`      | Delete a system restore                                                |
+|            | `/v1/ws/{period}/systembackups`  | Websocket stream for system backups                                    |
+|            | `/v1/ws/{period}/systemrestores` | Websocket stream for system restores                                   |
+
+## Design
+### Implementation Overview
+
+#### Manager: SystemBackup custom resource
+```yaml
+apiVersion: longhorn.io/v1beta2
+kind: SystemBackup
+metadata:
+  creationTimestamp: "2022-08-25T02:50:06Z"
+  finalizers:
+  - longhorn.io
+  generation: 1
+  labels:
+    longhorn.io/version: v1.4.0
+  name: demo-2
+  namespace: longhorn-system
+  resourceVersion: "420138"
+  uid: 41aac4e1-4367-4e17-b4b6-cb7c19151442
+spec: {}
+status:
+  conditions: null
+  createdAt: "2022-08-24T04:44:32Z"
+  gitCommit: 95292c60bb17b77591d6dde5c8636fe6bb4de60d-dirty
+  lastSyncedAt: "2022-08-25T02:50:19Z"
+  managerImage: "longhornio/longhorn-manager:v1.4.0"
+  ownerID: ip-10-0-1-105
+  state: Ready
+  version: v1.4.0
+```
+
+#### Manager: sync system backups in the backup target
+
+***longhorn-backup-target-controller***
+1. Execute engine binary [system-backup list](#engine-commands).
+1. Check for system backups in the backup target that are not in the `SystemBackup` list.
+1. Create new `SystemBackup`s and label with `longhorn.io/version: <version>` for the non-existing system backups.
+1. Check for system backups in the `SystemBackup` list that are not in the backup target.
+1. Delete `SystemBackup` for the non-existing system backups.
+1. Delete `SystemBackup` custom resources when the backup target is empty.
+
+***longhorn-system-backup-controller***
+1. For `SystemBackup` with the `longhorn.io/version: <version>` label, execute the engine binary
+   [system-backup get-config](#engine-commands) using the `<version>`.
+1. Update `SystemBackup` status from the [system backup config](#system-backup-cfg).
+
+#### Manager: system backup to the backup target
+
+***POST /v1/systembackups***
+```golang
+type SystemBackupInput struct {
+	Name string `json:"name"`
+}
+```
+1. Create `SystemBackup`.
+   ```yaml
+   apiVersion: longhorn.io/v1beta2
+   kind: SystemBackup
+   metadata:
+     name: <name>
+     namespace: longhorn-system
+   ```
+1. Return system backup resource.
+   <a name="system-backup-resource"></a>
+   ```golang
+    type SystemBackup struct {
+    	client.Resource
+    	Name         string                     `json:"name"`
+    	Version      string                     `json:"version,omitempty"`
+    	ManagerImage string                     `json:"managerImage,omitempty"`
+    	State        longhorn.SystemBackupState `json:"state,omitempty"`
+    	CreatedAt    string                     `json:"createdAt,omitempty"`
+    	Error        string                     `json:"error,omitempty"`
+    }
+   ```
+
+***webhook validator*** <a name="validator-system-backup"></a>
+1. Skip validation for `SystemBackup` created by the backup target controller.
+1. Allow `SystemBackup` to create if met conditions.
+   - The backup target is set.
+   - System backup does not exist in the backup target.
+
+***longhorn-system-backup-controller***
+```none
+system-backup.zip
++ metadata.yaml
+|
++ yamls/
+  + apiextensions/
+  | + customresourcedefinitions.yaml
+  |  
+  + kubernetes/
+  | + clusterrolebindings.yaml
+  | + clusterroles.yaml
+  | + configmaps.yaml
+  | + daemonsets.yaml
+  | + deployments.yaml
+  | + persistentvolumeclaims.yaml
+  | + persistentvolumes.yaml
+  | + podsecuritypolicies.yaml
+  | + rolebindings.yaml
+  | + roles.yaml
+  | + serviceaccounts.yaml
+  | + services.yaml
+  |
+  + longhorn/
+    + engineimages.yaml
+    + recurringjobs.yaml
+    + settings.yaml
+    + volumes.yaml
+```
+1. Create metadata file.
+    ```golang
+    type SystemBackupMeta struct {
+    	LonghornVersion        string      `json:"longhornVersion"`
+    	LonghornGitCommit      string      `json:"longhornGitCommit"`
+    	KubernetesVersion      string      `json:"kubernetesVersion"`
+    	LonghornNamespaceUUID  string      `json:"longhornNamspaceUUID"`
+    	SystemBackupCreatedAt  metav1.Time `json:"systemBackupCreatedAt"`
+    	ManagerImage           string      `json:"managerImage"`
+    }
+    ```
+1. Generated the resource YAML files:
+   <a name="system-backup-resources"></a>
+   1. Generate the API extension resource YAML file.
+      - CustomResoureDefinitions with API group `longhorn.io`.
+   1. Generate the Kubernetes resources YAML files.
+      - ServiceAccounts in the Longhorn namespace.
+      - ClusterRoleBinding with any Longhorn ServiceAccounts in the `subjects`.
+      - ClusterRoles with any Longhorn ClusterRoleBindings in the `roleRef`.
+      - Roles in Longhorn namespace.
+      - PodSecurityPolicies with Longhorn Role in the `rules`.
+      - RoleBindings in the Longhorn namespace.
+      - DaemonSets in Longhorn namespace.
+      - Deployments in Longhorn namespace.
+      - `longhorn-storageclass` ConfigMap.
+      - Services in Longhorn namespace. The ClusterIP and ClusterIPs will get removed before converting to the YAML.
+      - StorageClasses with provisioner `driver.longhorn.io`.
+      - PersistentVolumes with Longhorn StorageClass in spec.
+      - PersistentVolumeClaims with Longhorn StorageClass in spec.
+   1. Generate the Longhorn resources YAML files.
+      - Longhorn Settings.
+      - Longhorn EngineImages.
+      - Longhorn Volumes.
+      - Longhorn RecurringJobs.
+1. Archive the files to a zip file.
+1. Execute engine binary [system-backup upload](#engine-commands) to upload to the backup target
+   `backupstore/system-backups/<longhorn-version>/<system-backup-name>`.
+
+#### Manager: list system backups (GET /v1/systembackups)
+1. List `SystemBackup`.
+1. Return collection of [SystemBackup](#system-backup-resource) resource.
+
+#### Manager: delete system backup in the backup target
+
+***DELETE /v1/systembackups/{name}***
+1. Deletes `SystemBackup`.
+
+***longhorn-system-backup-controller***
+1. Execute engine binary [system-backup delete](#engine-commands) to remove the system backup
+   in the backup target.
+1. Cleanup local generated system backup files and directory that has not been uploaded to the backup target.
+
+#### Manager: SystemRestore custom resource
+```yaml
+apiVersion: longhorn.io/v1beta2
+kind: SystemRestore
+metadata:
+  annotations:
+    kubectl.kubernetes.io/last-applied-configuration: |
+      {"apiVersion":"longhorn.io/v1beta2","kind":"SystemRestore","metadata":{"annotations":{},"name":"demo-2-restore","namespace":"longhorn-system"},"spec":{"systemBackup":"demo-2"}}
+  creationTimestamp: "2022-08-24T04:44:51Z"
+  finalizers:
+  - longhorn.io
+  generation: 1
+  name: demo-2-restore
+  namespace: longhorn-system
+  resourceVersion: "283819"
+  uid: ef93355d-3b73-4fdd-bed3-ec5016a6784d
+spec:
+  systemBackup: demo-2
+status:
+  conditions:
+  - lastProbeTime: ""
+    lastTransitionTime: "2022-08-24T04:44:59Z"
+    message: sample error message
+    reason: Download
+    status: "True"
+    type: Error
+  ownerID: ip-10-0-1-113
+  sourceURL: s3://backupbucket@us-east-1/backupstore/system-backups/v1.4.0/demo-2
+  state: Error
+```
+
+#### Manager: restore system backup from the backup target 
+
+***POST /v1/systemrestores***
+```golang
+type SystemRestoreInput struct {
+	Name         string `json:"name"`
+	Version      string `json:"version"`
+	SystemBackup string `json:"systemBackup"`
+}
+```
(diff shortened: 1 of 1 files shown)


c3y1huang:
test

c3y1huang:
test

c3y1huang:
@c3y1huang

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/1455
+
+## Motivation
+### Goals
+
+- Support to create Longhorn-relate resources dump and upload to the backup target.
```
innobead, line 23: This seems duplicated as the second goal, suggest removing.

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+- Support restore from `lastBackup` when the volume doesn't exist in the cluster.
+- Support restore from `BackingImage`. **Dependent on https://github.com/longhorn/longhorn/issues/4165**
+
+### Non-goals [optional]
+
+- This feature does not deploy the Longhorn cluster. Users need to have a running Longhorn cluster to rollout.
+- Do backup/restore the downstream workloads attached to Longhorn volumes. For example, if a cluster has a Pod with a
+  PersistentVolumeClaim. Longhorn will rollout the PersistentVolumeClaim, PersistentVolume and Volume only. After a
+  successful rollout, the user can re-deploy the pod with the same manifest.
+- Rollout while there are volumes still attached.
```
innobead, line 36: This is a criteria, so needs to mention in the design like rollout will be only worked during maintenance.

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+- Automatically save pre-upgrade rollout and load when Longhorn upgrade fails.
+- Delete resources none-existing to the rollout.
```
innobead, line 38: Do you mean Longhorn resources?

So is it possible for users to understand what resources are created from a rollout? I feel rollout is a system snapshot, any risk to managing longhorn resources cleanly?

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+### Save Rollout
+
+```
+
+        |----------- [[ longhorn-backup-target controller ]] ----------|
+        |                                                              |
+        v                                                              v
+[ RolloutSave ] ---> [[ longhorn-rollout-save controller]] ---> [ object store ]
+                                                                       |
+                                                               backupstore/rollouts/
```
innobead, line 50: Just a note: right now, the folder usage and design on the backup target usage is only for a single cluster.

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+                                                                       |
+                                                   [  rollout.zip ] <--|--> [ rollout.cfg ]
```
innobead, line 52: Suggest having a checksum when having any compressed artifact.

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+```
+
+1. Introduce new [v1/rolloutsaves](#longhorn-manager-http-api) HTTP APIs.
+1. Introduce a new [RolloutSave](#manager-rolloutsave-custom-resource) custom resource definition.
+    - A new custom resource triggers the resources dump and uploads the generated rollout file to the backup target.
+    - Delete the custom resource triggers deletion of the rollout saved in the backup target.
```
innobead, line 58: You also can mention that rolloutsave resource management is similar to the current backup resource.

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+    - The saved rollouts in the backup target will sync to the custom resources list.
+1. Introduce new responsibility to `longhorn-backup-target` controller.
+    - Responsible to sync rollout in the backup target to the `RolloutSave` list.
+    - Each of the `RolloutSave` created by this controller will have the `longhorn.io/version: <version>` label.
```
innobead, line 62: Not sure if it's possible to add the cluster name as well? users will be easy to recognize where the rollback comes from.

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+1. Introduce a new `longhorn-rollout-save` controller.
+    - Responsible for generating the rollout file and uploading it to the backup target.
```
innobead, line 64: If a rolloutsave is from the backup target, the corresponding rollback generation should not be created. This should be the same as backup resource.

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+                                                          |       backupstore/volumes/
+                                                          |          |
+                                                          |          |
+[ Rollout ] ---> [[ longhorn-rollout controller ]] ---> [ object-store ]
+                          |                               |          |
+                          |                               |          |
+                       [ Job ]                            |       [ rollout.cfg: engine ]
+                          |                               |          |               
+                          v                               |          |               
+              [ rollout.cfg: manager ] ----> [[ longhorn-rollout-load controller ]]
```
innobead, line 91: Why need another `longhorn-rollout-load controller`? `longhorn-rollout controller` not enough?

enhancements/20220902-longhorn-rollout.md
innobead, line 58:
```suggestion
    - Deleting the custom resource triggers the deletion of the rollout saved in the backup target.
```

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+                                                          | <--- [ Volume: from backup ]
+                                                          |
+                                                          V
+                                                   [ Resources ]]]]
+```
+
+1. Introduce new [v1/rollouts](#longhorn-manager-http-api) HTTP APIs.
+1. Introduce a new [Rollout](#manager-rollout-custom-resource) custom resource definition.
+    - A new custom resource triggers the creation of a rollout job.
+    - Delete the custom resource triggers the deletion of the rollout job.
```
innobead, line 103:
```suggestion
    - Deleting the custom resource triggers the deletion of the rollout job.
```

innobead: How about if the rollout is in the middle? How to make the system back to normal? Do we support retry or just replay another new rollout?

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+1. Introduce a new `longhorn-rollout` controller.
+    - Responsible for creating a new rollout job that runs a `longhorn-rollout-load` controller.
+    - Responsible for deleting the rollout job.
+1. Introduce a new `longhorn-rollout-load` controller.
```
innobead, line 107: Then, this is not a controller but a just task instead so it will be another app in longhorn-manager/

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+      - Deployments
+      - DaemonSets
+      - EngineImages
+      - Settings
+      - Volumes
+      - PersistentVolumes
+      - PersistentVolumeClaims
+      - RecurringJobs
+  - Responsible for updating `Rollout` status during rollout.
+  - Responsible for adding `longhorn.io/last-rollout`, and `longhorn.io/last-rollout-at` annotation to rolled-out resources.
```
innobead, line 128: These annotations are good.

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+restore the Longhorn system.
+
+After the enhancement, the user can save and load Longhorn-related resources to/from the backup target with Longhorn
+native solution.
+
+#### Upgrade rollback, downgrade
+
+Before the enhancement, the user cannot downgrade Longhorn.
+
+After the enhancement, the user can downgrade Longhorn when there is a pre-upgrade rollout saved in the backup target.
```
innobead, line 145: Does this mean users can downgrade to any previous version? As I mentioned above, should clarify what version are allowed to rollback like the just previous version or any compatible previous version.

And, what validation we should provide is for avoiding an unexpected downgrade.

innobead (changes requested):

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+                                                          V
+                                                   [ Resources ]]]]
+```
+
+1. Introduce new [v1/rollouts](#longhorn-manager-http-api) HTTP APIs.
+1. Introduce a new [Rollout](#manager-rollout-custom-resource) custom resource definition.
+    - A new custom resource triggers the creation of a rollout job.
+    - Delete the custom resource triggers the deletion of the rollout job.
+1. Introduce a new `longhorn-rollout` controller.
+    - Responsible for creating a new rollout job that runs a `longhorn-rollout-load` controller.
```
innobead, line 105: - Can we have the rollout job not immediately triggered but activatable instead? 
- Another question: more than one rollouts, how will they be handled? execute one by one by the creation time?

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+                                                           | Backing Image           |
+                                                           | Instance Manager Image  |
+                                                           | Rollout                 |
+                                                           + ======================= +
+    ```
+1. View `Saved Rollouts` and `Rollouts` on the same page.
+    ```
+    Saved Rollouts                                                                                      [Custom Column]
+    ====================================================================================================================
+    [Save] [Load] [Delete]                                                            [Search Box   v ][__________][Go]
```
innobead, line 170: What does `Load` mean here? any different from `Save`?

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+                                                                                          + ======= +
+    ======================================================================================| Name    |===================
+    [] | Version  | Name   | State   | Error                                              | State   |
+    ---+----------+--------+--------+-----------------------------------------------------| Version |-------------------
+    [] | 1.4.0    | demo-1 | Error   | error uploading rollout: failed to execute: /var/l + ======= + ngine-binaries/c3y1huang-research-000-lh-ei/longhorn [rollout upload
```
innobead, line 175: If it becomes `Error`, what does it mean for users? Do users need to create another rolloutsave to trigger again?

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+    NAMESPACE         NAME     VERSION   STATE   CREATED
+    longhorn-system   demo-1   v1.4.0    Error   2022-08-23T00:25:29Z
+    longhorn-system   demo-2   v1.4.0    Ready   2022-08-24T02:34:57Z
+    ```
+
+### Rollout
+***Longhorn GUI***
+  - The users can load a rollout in the backup target.
+  - The users can view the rollout status.
+  - The users can load the rollout from a different Longhorn version.
```
innobead, line 225: Should be a compatible version

innobead (approved):

innobead (changes requested):
Not approved yet, wrong click

enhancements/20220902-longhorn-rollout.md
c3y1huang: I couldn't find a way to do so. Ref https://github.com/kubernetes/apimachinery/blob/7fb03423f864da053f3849d76d4c58d0a3442fe6/pkg/apis/meta/v1/types.go#L258-L260

enhancements/20220902-longhorn-rollout.md
c3y1huang: Having an extra rollout-load controller is for compatibility purposes.
- The `rollout controller` is running on the current cluster manager and is responsible for managing rollout CRs and creating rollout jobs.
- The `longhorn-rollout controller` runs on an independent manager image (the image will be stored in the rollout.cfg). The controller will also be using the correct engine image(also stated in the rollout.cfg) for the backup target interactions.

enhancements/20220902-longhorn-rollout.md
c3y1huang: Reworded. Let me know if these still feels duplicated.

enhancements/20220902-longhorn-rollout.md
c3y1huang: Updated.

c3y1huang: Updated.

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+https://github.com/longhorn/longhorn/issues/1455
+
+## Motivation
+### Goals
+
+- Support to create Longhorn-relate resources dump and upload to the backup target.
+- Support rollout Longhorn resources saved in the backup target.
+- Support rollout Longhorn at a different version (upgrade rollback).
+- Support operate rollout with Longhorn UI or kubectl.
+- Support restore from `lastBackup` when the volume doesn't exist in the cluster.
```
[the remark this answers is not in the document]
c3y1huang: After discussing with @innobead , the rollout-load controller will not handle concurrent volume restoring from the backup.
https://github.com/longhorn/longhorn/issues/4558

enhancements/20220902-longhorn-rollout.md
c3y1huang: Don't think we should delete existing volumes even if the volume doesn't exist in the rollout.

However, as discussed with @innobead , we can probably enhance it later and make this an option for the users.

enhancements/20220902-longhorn-rollout.md
c3y1huang: The rollout is a one-shot operation, if the rollout status Errors, the user can create a new rollout.

enhancements/20220902-longhorn-rollout.md
c3y1huang: https://github.com/longhorn/longhorn/pull/4388#discussion_r968262470

enhancements/20220902-longhorn-rollout.md
c3y1huang: > * Can we have the rollout job not immediately triggered but activatable instead?

As discussed with @innobead , the use case is unclear. We can enhance it later when the use case is clearer.

> * Another question: more than one rollouts, how will they be handled? execute one by one by the creation time?

We are allowing one rollout at a time. It will be handled in the validator.

enhancements/20220902-longhorn-rollout.md
c3y1huang: https://github.com/longhorn/longhorn/pull/4388#discussion_r970290793

enhancements/20220902-longhorn-rollout.md
c3y1huang: `Save` is to save the current Longhorn system to the backup store.
`Load` is to rollout the Longhorn system from the backup store.

enhancements/20220902-longhorn-rollout.md
c3y1huang: Added `webhook validator` sections.

enhancements/20220902-longhorn-rollout.md
c3y1huang: Updated.

enhancements/20220902-longhorn-rollout.md
c3y1huang: Added note below.

enhancements/20220902-longhorn-rollout.md
c3y1huang: Like discussed. I think it's fine to roll back or roll forward since the rollout will also handle custom resource definitions. It will create the CRD if it is missing, or add to the existing one if the version is different from the current cluster.

innobead: Good to know, then test cases should be covered.

c3y1huang: Added note to the design.

enhancements/20220902-longhorn-rollout.md
c3y1huang: https://github.com/longhorn/longhorn/pull/4388#discussion_r970483912

enhancements/20220902-longhorn-rollout.md
```diff
@@ -0,0 +1,690 @@
...
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/1455
+
+## Motivation
+### Goals
+
+- Support to create Longhorn-relate resources dump and upload to the backup target.
+- Support rollout Longhorn resources saved in the backup target.
+- Support rollout Longhorn at a different version (upgrade rollback).
```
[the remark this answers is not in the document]
c3y1huang: https://github.com/longhorn/longhorn/pull/4388#discussion_r970483912

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,713 @@
...
+    	engineImageList  *longhorn.EngineImageList
+    	recurringJobList *longhorn.RecurringJobList
+    	settingList      *longhorn.SettingList
+    	volumeList       *longhorn.VolumeList
+    }
+    ```
+    - Kubernetes resources from files in the `kubernetes` directory.
+    - API extension resources from files in the `apiextensions` directory.
+    - Longhorn resources from files in the `longhorn` directory.
+1. Restore resources asynchronously and annotate with `longhorn.io/last-rollout`, and `longhorn.io/last-rollout-at`.
```
shuo-wu, line 558: But the below resources will be restored in sequence, right?

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,713 @@
...
+    + engineimages.yaml
+    + recurringjobs.yaml
+    + settings.yaml
+    + volumes.yaml
+```
+1. Create metadata file.
+    ```golang
+    type RolloutMeta struct {
+    	LonghornVersion       string      `json:"longhornVersion"`
+    	LonghornGitCommit     string      `json:"longhornGitCommit"`
```
shuo-wu, line 381: This git commit typically means the manager git commit ID. Do we need to store the commit id for other images? NO?

shuo-wu (commented):
The naming in the LEP is confusing.  The term` rollout` can be:
1. CRD `Rollout`. The indicator of starting to restore resources from the backup target.
2. Or the file stored in the backup target

I recommended renaming the file in the backup target to `RolloutSave`. This

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,713 @@
...
+  | + deployments.yaml
+  | + persistentvolumeclaims.yaml
+  | + persistentvolumes.yaml
+  | + podsecuritypolicies.yaml
+  | + rolebindings.yaml
+  | + roles.yaml
+  | + serviceaccounts.yaml
+  | + services.yaml
+  |
+  + longhorn/
```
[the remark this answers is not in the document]
c3y1huang: The ShareManager is ignored because rollout volume goes through the normal volume handing flow,
- The share manager controller will create ShareManager for none-existing `rwx` volumes.
- The volume controller will sync ShareManager's image for existing detached volumes.

Any reason we need to include the ShareManager in the rollout file?

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,713 @@
...
+  creationTimestamp: "2022-08-24T04:44:51Z"
+  finalizers:
+  - longhorn.io
+  generation: 1
+  name: demo-2-rollout
+  namespace: longhorn-system
+  resourceVersion: "283819"
+  uid: ef93355d-3b73-4fdd-bed3-ec5016a6784d
+spec:
+  rollout: demo-2
```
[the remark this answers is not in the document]
c3y1huang: The intention to separate the CR name and `Spec.rollout` is to allow the user to keep records of the past success/failed Rollouts instead of having to replace the CR for every rollout.

enhancements/20220913-longhorn-rollout.md
c3y1huang: There will be a go routine running for each KIND of resource. The go routine will also check for resource dependencies before rolling out, ex: persistent volume must exist before rolling out the persistent volume claim.

enhancements/20220913-longhorn-rollout.md
c3y1huang: Actually, I'm not clear on the use case of having the commit ID for all images. Can you elaborate? Or we can add them as enhancement later.

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,713 @@
...
+  version: v1.4.0
```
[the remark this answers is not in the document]
c3y1huang: Yes, the intention is to group the backup store rollouts base on the Longhorn versions.

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,713 @@
...
+### Goals
+
+- Can generate Longhorn-relate resources dump and upload to the backup target.
+- Can rollout Longhorn resources from the backup target.
+- Support rollout Longhorn at a different version (ex: upgrade rollback).
+- Support rollout operation using the Longhorn UI or kubectl.
+- Support restoring from `lastBackup` when the volume doesn't exist in the cluster.
+- Support restore from `BackingImage`. **Dependent on https://github.com/longhorn/longhorn/issues/4165**
+
+### Non-goals [optional]
```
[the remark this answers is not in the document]
c3y1huang: Updated.

enhancements/20220913-longhorn-rollout.md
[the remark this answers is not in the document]
shuo-wu: No. Then ignoring share managers is fine. Thank you for the explanation!

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,713 @@
...
+    }
+    ```
+1. Generate the API extension resource YAML file.
+   - CustomResoureDefinitions with API group `longhorn.io`.
+1. Generate the Kubernetes resources YAML files.
+   - ServiceAccounts in the Longhorn namespace.
+   - ClusterRoleBinding with any Longhorn ServiceAccounts in the `subjects`.
+   - ClusterRoles with any Longhorn ClusterRoleBindings in the `roleRef`.
+   - Roles in Longhorn namespace.
+   - PodSecurityPolicies with Longhorn Role in the `rules`.
```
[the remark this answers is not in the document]
c3y1huang: We can remove PSP from the rollout if we've decided to deprecate PSP in https://github.com/longhorn/longhorn/issues/4003. Since the issue hasn't been concluded, we will still handle it and I've added a note.

enhancements/20220913-longhorn-rollout.md
shuo-wu: The engine controller will ignore the live upgrade requests when the engine images before and after the upgrade have the same commit id:
https://github.com/longhorn/longhorn-manager/blob/ce5b0c0aeefe545e9bab8e69d3f261e6c5db8bdf/controller/engine_controller.go#L1636-L1644

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,713 @@
...
+   apiVersion: longhorn.io/v1beta2
+   kind: RolloutSave
+   metadata:
+     name: <name>
+     namespace: longhorn-system
+   ```
+1. Return rollout save resource.
+   <a name="rollout-save-resource"></a>
+   ```golang
+   type RolloutSave struct {
```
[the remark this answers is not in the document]
c3y1huang: Yes, for the UI. Updated.

enhancements/20220913-longhorn-rollout.md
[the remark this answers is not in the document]
shuo-wu: Got it. This is the specified RolloutSave name. The naming is a little bit confusing..

enhancements/20220913-longhorn-rollout.md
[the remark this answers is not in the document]
shuo-wu: Then why does the rollout record it? This is already recorded in `RolloutSave.Status`

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,716 @@
...
+1. Return rollout save resource.
+   <a name="rollout-save-resource"></a>
+   ```golang
+    type RolloutSave struct {
+    	client.Resource
+    	Name         string                    `json:"name"`
+    	Version      string                    `json:"version,omitempty"`
+    	ManagerImage string                    `json:"managerImage,omitempty"`
+    	State        longhorn.RolloutSaveState `json:"state,omitempty"`
+    	Created      string                    `json:"created,omitempty"`
```
shuo-wu, line 337: `CreatedAt`

enhancements/20220913-longhorn-rollout.md
shuo-wu: As for the manager commit ID, IIRC, it will be shown in the bottom left of the UI page when the non-official version is used

enhancements/20220913-longhorn-rollout.md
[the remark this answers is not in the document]
c3y1huang: You are right! We don't need it.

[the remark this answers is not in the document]
c3y1huang: Removed.

enhancements/20220913-longhorn-rollout.md
c3y1huang: Updated.

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,713 @@
...
+https://github.com/longhorn/longhorn/issues/1455
+
+## Motivation
+### Goals
+
+- Can generate Longhorn-relate resources dump and upload to the backup target.
+- Can rollout Longhorn resources from the backup target.
+- Support rollout Longhorn at a different version (ex: upgrade rollback).
+- Support rollout operation using the Longhorn UI or kubectl.
+- Support restoring from `lastBackup` when the volume doesn't exist in the cluster.
```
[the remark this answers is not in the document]
c3y1huang: Mainly for new clusters, but also applies to an existing cluster if volume doesn't already exist.

enhancements/20220913-longhorn-rollout.md
c3y1huang: Not sure if we will hit the same case. ATM We don't support rolling out to a cluster while volumes are attached.

shuo-wu: OK. I just explain the general user case of the commit ID in Longhorn. It's not related to this feature.
I am fine with either retaining this or removing it.

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,715 @@
...
+
+Currently, Longhorn has documents to help to restore the Longhorn cluster:
+- [Restore to a new cluster using Velero](https://longhorn.io/docs/1.3.0/advanced-resources/cluster-restore/restore-to-a-new-cluster-using-velero/)
+- [Restore to a cluster contains data using Rancher snapshot](https://longhorn.io/docs/1.3.0/advanced-resources/cluster-restore/restore-to-a-cluster-contains-data-using-rancher-snapshot/)
+
+However, the solution is external and not out-of-the-box.
+
+This document proposes to improve user experience by introducing a new `rollout` feature that will dump Longhorn-related
+resources and save them as a single file to the backup target. Later, users can load the rollout to the same cluster to
+revert the Longhorn resources or restore it to another Longhorn cluster.
```
innobead, line 13:
```suggestion

This feature is to support Longhorn cluster backup & restore, and also helps with rollback the cluster to the previous healthy state after a failed upgrade.

Currently, Longhorn has documents to guide users on how to restore the Longhorn cluster as below; however, the solution relies on third-party tools, not out-of-the-box, and also involves tedious human intervention.
- [Restore to a new cluster using Velero](https://longhorn.io/docs/1.3.0/advanced-resources/cluster-restore/restore-to-a-new-cluster-using-velero/)
- [Restore to a cluster contains data using Rancher snapshot](https://longhorn.io/docs/1.3.0/advanced-resources/cluster-restore/restore-to-a-cluster-contains-data-using-rancher-snapshot/)

With this new feature, Longhorn's custom resources will be backed up and bundled as a system backup file, then saved to the remote backup target. 

Later, users can choose a system backup to restore to a new cluster or an existing cluster; nevertheless, it can help with the cluster rollback to fix the corrupted cluster state just after a failed upgrade.

```

innobead: Don't suggest making the new term in the summary, but rather keep the summary w/ clear descriptions of what problems you would like to resolve instead. 

The new terminologies can be introduced later to avoid some confusion w/ the correct backup or context.

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,715 @@
...
+https://github.com/longhorn/longhorn/issues/1455
+
+## Motivation
+### Goals
+
+- Can generate Longhorn-relate resources dump and upload to the backup target.
+- Can rollout Longhorn resources from the backup target.
+- Support rollout Longhorn at a different version (ex: upgrade rollback).
+- Support rollout operation using the Longhorn UI or kubectl.
+- Support restoring from `lastBackup` when the volume doesn't exist in the cluster.
```
innobead, line 27: For some items, you can aggregate to make goals clear and complete.

```suggestion
- Support system/cluster backup via backing up Longhorn custom resources, bundle them, and upgrade to the backup target.  (Original: `RolloutSave`, but how about `SystemBackup`?)
- Support system/cluster restore to a new/existing cluster from a system backup (Original: `Rollout`, but how about `SystemRestore`?)
- Support restoring volume from `lastBackup` when the volume doesn't exist in the cluster during the system restore.
```

innobead (changes requested):

enhancements/20220913-longhorn-rollout.md
c3y1huang: I am ok with different namings.
So we also need to have different names for the 3 controllers:
- rollout-save controller > system-backup controller
- rollout controller > system-restore controller
- rollout-load controller > system-rollout controller

Also the file name:
- rollout.zip > system-backup.zip

WDYT @innobead @shuo-wu ?

shuo-wu: I am fine with the new naming. Or as I mentioned, modifying the remote file naming as rollout-save.zip

innobead: `system-rollout controller`, probably `system-restore-rollout controller`?

@c3y1huang I know you already spent efforts on it, so if you are confident about these changes, and no one has different preferences, then let's go with the new naming to make it much meaningful.

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,715 @@
...
+
+### Non-goals [optional]
```
innobead, line 29: Update each item in this section accordingly if needed due to naming change.

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,715 @@
...
+- Automatically save pre-upgrade rollout and load when Longhorn upgrade fails.
+- Delete resources none-existing to the rollout. We can probably enhance it later and make this an option for the users.
+- Support restore from `BackingImage`. **Dependent on https://github.com/longhorn/longhorn/issues/4165**
+
+## Proposal
+### Save Rollout
+
+```
+
+        |----------- [[ longhorn-backup-target controller ]] ----------|
```
innobead, line 45: This part is very similar to what we do for Backup? I will be great for design if something we can abstract or consolidate here, but it can be a later work. 

The flow is general, but just the resource different: one is backup, and the other is system-backup.

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,715 @@
...
+1. Introduce new [v1/rolloutsaves](#longhorn-manager-http-api) HTTP APIs.
+1. Introduce a new [RolloutSave](#manager-rolloutsave-custom-resource) custom resource definition.
+    - A new custom resource triggers the resources dump and uploads the generated rollout file to the backup target.
+    - Deleting the custom resource triggers deletion of the rollout saved in the backup target.
+      This behavior is similar to the current backup resource handling.
+    - The saved rollouts in the backup target will sync to the custom resources list.
+1. Introduce new responsibility to `longhorn-backup-target` controller.
+    - Responsible to sync rollout in the backup target to the `RolloutSave` list.
+    - Each of the `RolloutSave` created by this controller will have the `longhorn.io/version: <version>` label.
+1. Introduce a new `longhorn-rollout-save` controller.
```
innobead, line 64: As I mentioned, this is similar to `backup-controller` today.

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,715 @@
...
+[ RolloutSave ] ---> [[ longhorn-rollout-save controller]] ---> [ object store ]
+                                                                       |
+                                                               backupstore/rollouts/
+                                                                       |
+                                                   [  rollout.zip ] <--|--> [ rollout.cfg ]
+```
+
+1. Introduce new [v1/rolloutsaves](#longhorn-manager-http-api) HTTP APIs.
+1. Introduce a new [RolloutSave](#manager-rolloutsave-custom-resource) custom resource definition.
+    - A new custom resource triggers the resources dump and uploads the generated rollout file to the backup target.
```
innobead, line 57: Can you mention what CRDs and their resources will be backed up? as you explained in Rollout/system restore section.

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,715 @@
...
+      1. Compress resources to a zip file.
+      1. Upload to the backup target `backupstore/rollouts/<longhorn-version>/<rollout-name>`.
+      >**Note:** Do not generate the rollout archive for the backup target controller created `RolloutSave`.
+
+    - Responsible for deleting rollout saved in the backup target.
+    - Responsible for updating `RolloutSave` status.
+    - Responsible for updating the error message in the `RolloutSave` status condition.
+      Reference [Rollout](#manager-rollout-custom-resource) condition as the example.
+
+### Rollout
```
innobead, line 76: When doing the system restore, does the system need to be maintenance mode like all volumes needed to detached? I feel it's not possible to do something similar to engine upgrade during the system upgrade.


If there is some limitation or pre-conditions, need to emphasize or add to the implementation.

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,715 @@
...
+- Do backup/restore the downstream workloads attached to Longhorn volumes. For example, if a cluster has a Pod with a
+  PersistentVolumeClaim. Longhorn will rollout the PersistentVolumeClaim, PersistentVolume and Volume only. After a
+  successful rollout, the user can re-deploy the pod with the same manifest.
+- Rollout while there are volumes still attached.
+- Automatically save pre-upgrade rollout and load when Longhorn upgrade fails.
+- Delete resources none-existing to the rollout. We can probably enhance it later and make this an option for the users.
+- Support restore from `BackingImage`. **Dependent on https://github.com/longhorn/longhorn/issues/4165**
+
+## Proposal
+### Save Rollout
```
innobead, line 41: When doing system backup, anything users need to pay attention or prerequisites like I asked in the system restore section.

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,715 @@
...
+1. Introduce a new [Rollout](#manager-rollout-custom-resource) custom resource definition.
+    - A new custom resource triggers the creation of a rollout job.
+    - Deleting the custom resource triggers the deletion of the rollout job.
+1. Introduce a new `longhorn-rollout` controller.
+    - Responsible for creating a new rollout job that runs a `longhorn-rollout-load` controller.
+    - Responsible for deleting the rollout job.
+1. Introduce a new `longhorn-rollout-load` controller. This controller is similar to the `uninstall controller`.
+    - Run inside the pod created by the rollout job.
+    - Responsible for downloading the rollout from the backup target.
+    - Responsible for restoring resources from the rollout file.
```
innobead, line 113: As discussed before, few requestions:
- Can explain how it will work for restoring the system backup to an existing cluster or how is the difference than rollback the failed cluster to a previous healthy state?
- For restoring to the existing cluster, how to handle the resource version different (for example: v1beta1 -> v1, v1 -> v1beta1). (We discussed it should work, but it is worth explaining here) 
- If cross-version system restore will be a concern, what version policy will be to limit the restore path

enhancements/20220913-longhorn-rollout.md
```diff
@@ -0,0 +1,715 @@
...
+1. Output the [rollout config](#rollout-cfg) from the object store.
+
+### Test plan
+
+#### Rollout Save
+- Test rollout save to the backup target.
+- Test rollout save should fail when the backup target is empty.
+- Test rollout save should fail when the backup target is unreachable.
+
+#### Rollout
```
innobead, line 694: Can we add a test case when a rollout failed in the middle? What's the state and how to recover by retry/create another rollout.

innobead (changes requested):
In general, LGTM. 

Some feedbacks to make the proposal much clear to understand. We also need to pay attention to if the system-restore failed, how to handle for day2.

enhancements/20220913-longhorn-rollout.md
c3y1huang: Updated.

enhancements/20220913-longhorn-rollout.md
c3y1huang: Updated.

enhancements/20220913-longhorn-rollout.md
c3y1huang: Updated.

enhancements/20220913-longhorn-rollout.md
c3y1huang: Yes, it's similar.

enhancements/20220913-longhorn-rollout.md
c3y1huang: Added anchor and link.

enhancements/20220913-longhorn-rollout.md
c3y1huang: They are covered in the `webhook validator` sections.

c3y1huang: Also noted to the proposal sections for visibility.

enhancements/20220913-longhorn-rollout.md
c3y1huang: https://github.com/longhorn/longhorn/pull/4388#discussion_r997806385

enhancements/20220913-longhorn-rollout.md
c3y1huang: Added notes.

enhancements/20220913-longhorn-rollout.md
c3y1huang: Updated.

enhancements/20220913-longhorn-system-backup-restore.md
```diff
@@ -0,0 +1,727 @@
...
+    - Responsible for updating `SystemRestore` status during system restore.
+    - Responsible for adding `longhorn.io/last-system-restore`, and `longhorn.io/last-system-restore-at` annotation to the
+    restored resources.
+    - Responsible for adding `longhorn.io/last-system-restore-backup`annotation to the restored volume resources.
+    - Responsible for updating the error message in the [SystemRestore](#manager-systemrestore-custom-resource) status
+    condition.
+
+    > **Note:** There are 2 areas covered for cross version restoration:
+    > 1. The system restore will use the manager and engine image in the system backup config to run the `longhorn-system-rollout` so the controller is compatible with the restoring resources.
+    > 1. When the CustomResourceDefinition is missing the version for the restoring resource, the system restore doesn't replace or remove the existing CustomResourceDefinitions. Instead, the controller adds to its versions. So system restoration doesn't break existing resources.
```
innobead, line 138: Also, need to pay attention to if any file/meta saved on disks will impact the compatibility as well like metadata for volume head, snapshots, backing image, etc.

innobead (approved):
In general, LGTM>

enhancements/20220913-longhorn-system-backup-restore.md
innobead: cc @shuo-wu @derekbit

shuo-wu: Here, I think the system restore just needs to try its best effort with a warning message. 
It's hard to guarantee this compatibility... But typically we will add some fields with backward compatibility for the resources rather than directly modifying/deleting fields. Hence this should be fine.

enhancements/20220913-longhorn-system-backup-restore.md
```diff
@@ -0,0 +1,727 @@
...
+- Test system restore sync from the backup target.
+- Test system restore each resource when exist in the cluster.
+- Test system restore each resource when not exist in the cluster.
+- Test system restore failed to unzip.
+- Test system restore failed to restore resources.
+
+***Version Jump***
+- Test system restore to lower Longhorn version of each Longhorn installation method (kubectl/helm/Rancher).
+- Test system restore to higher Longhorn version of each Longhorn installation method (kubectl/helm/Rancher).
+- Test system restore to cluster with multiple engine images.
```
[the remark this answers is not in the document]
shuo-wu: Yes it's better to add some clarifications. For example, if we plan to deprecate and remove some fields in v1.4.0, we should announce the incompatibility between v1.4.0 and the old versions.

enhancements/20220913-longhorn-system-backup-restore.md
```diff
@@ -0,0 +1,727 @@
...
+    - DaemonSets.
+    - EngineImages.
+    - Volumes. Annotate with `longhorn.io/last-system-restore-backup` if the volume is restored from the backup.
+    - PersistentVolumes.
+    - PersistentVolumeClaims.
+    - RecurringJobs.
+1. Update [SystemRestore](#manager-systemrestore-custom-resource) status and error.
+1. Shutdown longhorn-system-rollout controller.
+
+#### Engine: commands
```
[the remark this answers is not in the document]
c3y1huang: Currently, the communication to the backup target is using engine binary. So I think we should do the same for the system backup/restore.

enhancements/20220913-longhorn-system-backup-restore.md
```diff
@@ -0,0 +1,722 @@
...
+
+1. Introduce new [v1/systembackups](#longhorn-manager-http-api) HTTP APIs.
+1. Introduce a new [SystemBackup](#manager-systembackup-custom-resource) custom resource definition.
+    - A new custom resource triggers the creation of the system [resource](#system-backup-resources) backup.
+    - Deleting the custom resource triggers deletion of the system backup in the backup target.
+      This behavior is similar to the current backup resource handling.
+    - The system backups stored in the backup target will get synced to the custom resources list.
+1. Introduce new responsibility to `longhorn-backup-target` controller.
+    - Responsible for syncing system backups in the backup target to the `SystemBackup` list.
+    - Each of the `SystemBackup` created by this controller will have the `longhorn.io/version: <version>` label.
```
shuo-wu, line 74: NIT: We don't need this kind of detail in the proposal section. And this will be mentioned in the below implementation part.

enhancements/20220913-longhorn-system-backup-restore.md
```diff
@@ -0,0 +1,722 @@
...
+      - ClusterRoles with any Longhorn ClusterRoleBindings in the `roleRef`.
+      - Roles in Longhorn namespace.
+      - PodSecurityPolicies with Longhorn Role in the `rules`.
+      - RoleBindings in the Longhorn namespace.
+      - DaemonSets in Longhorn namespace.
+      - Deployments in Longhorn namespace.
+      - `longhorn-storageclass` ConfigMap.
+      - Services in Longhorn namespace. The ClusterIP and ClusterIPs will get removed before converting to the YAML.
+      - PersistentVolumes with Longhorn StorageClass in spec.
+      - PersistentVolumeClaims with Longhorn StorageClass in spec.
```
shuo-wu, line 412: How about the StorageClasses with `provisioner: driver.longhorn.io`

enhancements/20220913-longhorn-system-backup-restore.md
```diff
@@ -0,0 +1,722 @@
...
+
+#### BackupStore: commands
+
+<a name="cmd-system-backup-upload"></a>
+
+***Command: system-backup upload***
+| Argument         | Usage                             |
+| -----------------| --------------------------------- |
+| 0                | the source local file path        |
+| 1                | the destination system backup URL |
```
shuo-wu, line 607: NIT: Do we need to use 2 flags `--src-path` and `--dest-url` instead to make this cmd more easy-to-use?

shuo-wu (commented):
In general LGTM

enhancements/20220913-longhorn-system-backup-restore.md
c3y1huang: Removed.

enhancements/20220913-longhorn-system-backup-restore.md
c3y1huang: I've missed this. Updated, thanks!

enhancements/20220913-longhorn-system-backup-restore.md
c3y1huang: Since those are mandatory fields, I prefer to use arguments instead. This is the same as `cp` and `rsync`.

enhancements/20220913-longhorn-system-backup-restore.md
```diff
@@ -0,0 +1,723 @@
...
+                                                           + ======================= +
+                                                           | General                 |
+                                                           | Engine Image            |
+                                                           | Orphaned Data           |
+                                                           | Backing Image           |
+                                                           | Instance Manager Image  |
+                                                           | System Backup           |
+                                                           + ======================= +
+    ```
+1. View `System Backups` and `System Restores` on the same page.
```
innobead, line 179: Remember adding events when any operations happening or status changed as backup.

Also, for failed system backup, need to have a consistent behavior for the clean up. https://github.com/longhorn/longhorn/blob/master/enhancements/20220801-failed-backups-cleanup.md

innobead (approved):

enhancements/20220913-longhorn-system-backup-restore.md
c3y1huang: Got it.

Issue #2293, opened 2021-03-03
[FEATURE] NFS support (RWX) GA


yasker:
We want to improve our NFS solution to be highly available, so one node down (especially the node running the NFS server) will result in minimal downtime of the NFS server and near-zero interruption for the NFS client.

One note: since the NFS server uses `sync` mode by default, all the data are written by the client will be written on the underlying block device before the request was acked.

Ref: https://www.avidandrew.com/understanding-nfs-caching.html (Server-side caching)

### Tasks: 
- [x] https://github.com/longhorn/longhorn/pull/4299
- [x] https://github.com/longhorn/longhorn/issues/4351
- [x] dedicated recovery backend: [longhorn-manager side](https://github.com/longhorn/longhorn-manager/pull/1464)
- [x] dedicated recovery backend: [nfs-ganesha side](https://github.com/longhorn/nfs-ganesha/pull/2)
- [x] nfs server grace priod, lease time and hostname issues (https://github.com/longhorn/longhorn-share-manager/pull/36)
- [x] longhorn-manager client mount options (https://github.com/longhorn/longhorn-manager/pull/1464)

derekbit:
**Investigate two scenarios**

**1. node, which has the volume and the share-manager pod, is not crashed or shut down, but the share-manager pod is stopped unexpectedly**
    The  share-manager pod  is restart on the same node, and then everything works well again.

**2. node, which has the volume and the share-manager pod, is crashed or suddenly shut down**
    The volume and share-manager are re-created on another node.
    The NFS4 client try to resend SEQUENCE to the new server but encounters timeout error for 5-6 mins. [client-rpc.log](https://github.com/longhorn/longhorn/files/8887142/client-rpc.log)
    The connect is stuck somewhere, and the SEQUENCE is not received at the server side. 
    After 5-6 mins, the old connection is cleaned up successfully, and the NFS4 connection becomes normal.

derekbit:
**OS: ubuntu 22.04 (kernel: 5.15.0)**

The incomplete IO operation is retried in the second case in the above comment after the volume and share-manager pod are re-created on another node. The NFS-HA is achieved.

The iptable rules contain more sophisticated traffic rules between the app pod share-manager pod in this environment. These rules are not seen in the above one. We need to investigate iptables more thoroughly.

voarsh2:
Could making the share-manager run thru a headless service or ClusterIP, with replica's? K8 design is to have more than one instance running and share-manager is not using replica's.....

[...]

PR #4437, opened 2022-08-17
chart: add and update manifests for recovery-backend

derekbit:
https://github.com/longhorn/longhorn/issues/2293

Signed-off-by: Derek Su [derek.su@suse.com](mailto:derek.su@suse.com)

### DIFF
--- a/chart/templates/services.yaml
+++ b/chart/templates/services.yaml
@@ -36,6 +36,23 @@ apiVersion: v1
 kind: Service
 metadata:
   labels: {{- include "longhorn.labels" . | nindent 4 }}
+    app: longhorn-recovery-backend
+  name: longhorn-recovery-backend
+  namespace: {{ include "release_namespace" . }}
+spec:
+  type: ClusterIP
+  sessionAffinity: ClientIP
+  selector:
+    app: longhorn-recovery-backend
+  ports:
+  - name: recovery-backend
+    port: 9600
+    targetPort: recov-backend
+---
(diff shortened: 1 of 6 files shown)


chart/templates/services.yaml
```diff
@@ -34,6 +34,23 @@ spec:
...
+  namespace: {{ include "release_namespace" . }}
+spec:
+  type: ClusterIP
+  sessionAffinity: ClientIP
+  selector:
+    app: longhorn-recovery-backend
+  ports:
+  - name: recovery-backend
+    port: 9600
+    targetPort: recov-backend
```
innobead, line 50: Probably we can just the same name `recovery-backend`as the service name.

innobead (commented):
LGTM. Just one comment for the consistency.

chart/templates/services.yaml
derekbit: targetPort must be no more than 15 characters. how about using `recov-backend`?

innobead: I see. This is why the name is used on purpose. Let's just keep it as is.

innobead (dismissed):

derekbit:
Waiting for https://github.com/longhorn/longhorn-share-manager/pull/36

derekbit:
e2e
test_rwx.py
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1934/

innobead (approved):

PR #4447, opened 2022-08-19
Release 1.2.5-rc1

innobead:


innobead:
sharing-manager, backing-image-manager, instance-manager image will be updated in 1.2.5-rc2. (just host package refresh)

PR #4482, opened 2022-08-25
Add require/automation-e2e to enhance/improvement ticket

innobead:


innobead:
@longhorn/dev add `require/automation-e2e` by default for any enhancement or improvement ticket. Assignees can check if their tickets are suitable for that. If not, just replace it with `require/manual-test-plan`.

cc @longhorn/qa

Issue #4423, opened 2022-08-16
[IMPROVEMENT] Create a pipeline to automatically update Longhorn upgrade responder when we release a new version


PhanLe1010:
The upgrade responder server is at: https://github.com/longhorn/upgrade-responder/tree/longhorn

Pipeline could be drone.io or jenkins

innobead:
cc @yangchiu

PhanLe1010:
Possible helm plugin for drone.io https://github.com/ipedrazas/drone-helm

innobead:
Moving into the QA board instead. @yangchiu should work with @PhanLe1010 to get this done in the future. Not a priority for now.

PR #4518, opened 2022-09-02
Add upgrade responder server helm values

PhanLe1010:
This is the first step for the pipeline to automatically update the upgrade responder server

longhorn/longhorn#4423


innobead (approved):

innobead:
cc @yangchiu @longhorn/qa

Issue #2380, opened 2021-03-19
[IMPROVEMENT] Dump NFS ganesha logs to pod stdout


c3y1huang:
**Is your feature request related to a problem? Please describe.**
Currently, ganesha dumps logs to /tmp/ganesha.log file inside the share-manager container. It could be difficult to retrieve the log file if the pod goes into a crash loop.

**Describe the solution you'd like**
Able to see ganesha log included when using `kubectl logs`

**Describe alternatives you've considered**
`none`

**Additional context**
https://github.com/longhorn/longhorn/issues/2284

innobead:
This will help us to improve the support bundle enrichment w/ more helpful logs.

weizhe0422:
If I add output log with stdout in manager shutdown function to implement gracefully shutdown, am I right?

https://github.com/longhorn/longhorn-share-manager/pkg/server/share_manager.go

innobead:
@weizhe0422 Remember updating the Zenhub pipeline status when working on the issue, so right now, it's time to move to ready-for-testing and provide the testing steps in the check-list-comment which will be created when the issue is moved to `review` pipeline.

<img width="227" alt="image" src="https://user-images.githubusercontent.com/899492/186138926-956082ae-f466-4897-bb51-dc9a98801711.png">

Lichiehyao:
**Verified on master-head 20220913**

- longhorn master-head ([8c87681](https://github.com/longhorn/longhorn/commit/8c876810bcf36a1c7222b59cf8fce49e16a44373)) 
- longhorn-share-manager [v1_20220825](https://github.com/longhorn/longhorn-share-manager/releases/tag/v1_20220825) ( [2cba526](https://github.com/longhorn/longhorn-share-manager/commit/2cba526a55368a7f5f263b204d87b39836855dff))

**The test steps**

1. Install Longhorn Master Ref.([install-with-kubectl](https://longhorn.io/docs/1.4.0/deploy/install/install-with-kubectl/))
2. Create rwx volume and attach to pod. Deploy pod “mypod1”
3. `kubectl apply -f pod_with_pvc.yaml`
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: longhorn-volv-pvc
  namespace: default
spec:
  accessModes:
    - ReadWriteMany
  storageClassName: longhorn
  resources:
    requests:
      storage: 0.5Gi
---
apiVersion: v1
kind: Pod
metadata:
  name: volume-test
  namespace: default
[...]

PR #4562, opened 2022-09-14
Update longhorn-share-manager to v1_20220914

weizhe0422:
[Longhorn 2380](https://github.com/longhorn/longhorn/issues/2380)


innobead (approved):

innobead:
@weizhe0422 Also longhorn-manager, thanks.

Issue #4574, opened 2022-09-16
[IMPROVEMENT] add possibility to use nodeSelector on the storageClass


msoGTA:
## Is your improvement request related to a feature? Please describe

In order to mount the volume only on the necesary nodes. It is useful to add the possibility to use nodeSelector with the storageClass template like in the documentation.

https://longhorn.io/docs/1.3.1/volumes-and-nodes/storage-tags/

## Describe the solution you'd like

Editing the storageClass.yaml file and the values.yaml file in order to offer this possibility.

Once done, the users could enable this feature and on the UI or within the K8s cluster directly tag the nodes like in the documentation below.

https://longhorn.io/docs/1.4.0/advanced-resources/default-disk-and-node-config/#adding-node-tags-to-new-nodes

Lichiehyao:
Hi @weizhe0422 :
     https://github.com/longhorn/website/pull/591 is not merged yet.

weizhe0422:
Hi @roger-ryao, 

Since longhorn/website#591 has been merged, I changed the issue status to Ready for Testing. 

Thanks a lot!

roger-ryao:
**Verified on master-head 20221117**

- longhorn master-head ([dab0603](https://github.com/longhorn/longhorn/commit/dab06038476c337266936b0fe47a5f0ae0cdaed7)) 
- longhorn-manager master-head ([ed1f742](https://github.com/longhorn/longhorn-manager/commit/ed1f74273fb2d86d85d910ec7a08e2654f477536)) 


**The test steps**

[Ref #4574 (To Reproduce)](https://github.com/longhorn/longhorn/issues/4574#issuecomment-1263185329)

- 1. Add annotation with kubectl `kubectl annotate node <node-name> node.longhorn.io/default-node-tags='["medium"]'`
- 2. Download the latest [helm chart](https://github.com/longhorn/longhorn/tree/master/chart). And 
   edit the `values.yaml` in the PATH `/longhorn/chart`
   and adjust the `persistence.defaultNodeSelector` section like below
    ```yaml
    defaultNodeSelector:
        enable: true # disable by default
        selector: "medium"
    ```
- 3. Install Longhorn via Helm:  `helm install longhorn . --namespace longhorn-system --create-namespace`
- 4. Type `kubectl get sc longhorn -n longhorn-system -o yaml` to check the storageclass has `nodeSelector: "medium"` text
[...]

PR #4576, opened 2022-09-16
Add nodeSelector to storageClass

msoGTA:
This PR is link to the following issue [https://github.com/longhorn/longhorn/issues/4574]
 add possibility to use nodeSelector on the storageClass #4574

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -85,6 +85,9 @@ persistence:
     dataSourceType: ~
     dataSourceParameters: ~
     expectedChecksum: ~
+  defaultNodeSelector:
+    enable: false # disable by default
+    selector: []
 
 csi:
   kubeletRootDir: ~
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -583,6 +583,20 @@ WARNING:
     group: "Longhorn Storage Class Settings"
     type: string
     default:
+- variable: defaultSettings.defaultNodeSelector.enable
+  description: "Enable recurring Node selector for Longhorn StorageClass"
+  group: "Longhorn Storage Class Settings"
+  label: Enable Storage Class Node Selector
+  type: boolean
+  default: false
+  show_subquestion_if: true
+  subquestions:
+  - variable: defaultSettings.defaultNodeSelector.selector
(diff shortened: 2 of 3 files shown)


mantissahz:
@msoGTA appreciate your contribution,
Check the DCO failed, `There are 2 commits incorrectly signed off`.
Please use `git commit -s` to sign-off in your commits. Thanks.

chart/values.yaml
```diff
@@ -85,6 +85,10 @@ persistence:
     dataSourceType: ~
     dataSourceParameters: ~
     expectedChecksum: ~
+   defaultNodeSelector:
+    enable: false # disable by default
+    name: longhorn # longhorn is the value of the label and can be named differnetly
```
innobead, line 90: Suggest renaming `name` to `selector` and the default value should be `[]`, because it supports type list instead.

innobead (changes requested):
Thanks for the contribution!

Please also update questions.yaml which is used for Rancher chart interactive installation.

chart/values.yaml
innobead: cc @c3y1huang we should make the same pattern for `recurringJobSelector`. You can create a ticket for future refactoring. 

```yaml
  recurringJobSelector:
    enable: false
    jobList: []
```

msoGTA: Hi innobead, thanks for your review. I just pushed a new commit with your suggestions. Regards.

innobead: Thanks, please consolidate your commits to one, and update questions.yaml (you can reference previous commit history to know how to update that from other settings) as I mentioned above. Also, remember to sign off the commit.

msoGTA: Hi, I made the requested changes. 
Regards

innobead (approved):
LGTM

But just notice some naming of settings in persistence seems are inconsistent, so we probably need to refactor them later. (minor)

innobead:
@weizhe0422 Please help handle this contribution later at https://github.com/longhorn/longhorn/issues/4574. Thanks. 

@msoGTA thanks for the contribution!

msoGTA:
You are welcome :) @innobead

Issue #4210, opened 2022-07-06
[IMPROVEMENT] Introduce checksum for snapshots


shuo-wu:
## Is your improvement request related to a feature? Please describe

After creating a snapshot, Longhorn volume can calculate the checksum then store it in the metafile along with the modification time. Later on when the volume needs to rebuild/reuse a failed replica or somehow need to validate a replica, checking the checksum & modification time matching would help skip unnecessary snapshot syncing, which actually speeds up the whole progress.

## Describe the solution you'd like
- [ ] Store the modification time during snapshot creation
- [ ] Calculate the checksum with an async goroutine after snapshot creation.
- [ ] Check and match the checksum & modification time before syncing a snapshot file
- [ ] (Optional, maybe we can open another ticket for this if required) Verify the snapshot file consistency by checking the modification time & checksum during attachment.

## Additional context
This one is similar to what I have done for the backing image management.

Martin-Weiss:
In case this helps for „delta-sync“ of data - this would be a really important feature! Especially in the area of Harvester it is really important to get a journal / delta-sync added! In the case of Harvester we have volumes with several hundreds of GB and we have node capacity of 4, 8 or more TB of data - and full re-replication of such an amount of data during node failures / maintenance / data-center failures etc. Is really a difficult challenge..

shuo-wu:
cc @joshimoo

derekbit:
> After creating a snapshot, Longhorn volume can calculate the checksum then store it in the metafile along with the modification time. 

Can the checksume be stored in the `snapshot.spec`?

shuo-wu:
Probably it can be updated and stored in `snapshot.Status` but with an async way.

derekbit:
### Benchmark

#### x86_64

|  --  | --  |
|  ----  | ----  |
| Kubernetes | v1.24.4+rke2r1 |
| Longhorn |  master-head with snapshot checksum feature |
| Host | AWS EC2 c5d.2xlarge |
| CPU | Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz |
| RAM | 16 GiB |
| Disk model | Amazon EC2 NVMe Instance Storage |

- Volume size: 100 GiB with random data
- Checksumming time: 325 seconds
- Checksumming throughput: 315 MiB/s
- CPU usage (`kubectl -n longhorn-system top instance-manager-r-xxx`):  900m (2m at idle state) 

cc @shuo-wu @joshimoo @innobead

[...]

Issue #3198, opened 2021-10-31
[FEATURE] Automatic identifying of corrupted replica (bit rot detection)


chickeaterbanana:
see also: https://longhorn.io/docs/1.2.2/advanced-resources/data-recovery/corrupted-replica/

**Describe the solution you'd like**
You already described a way of identifying corrupted replicas. This could also work on blocks. If you add some meta data to the snapshot in form of hashes of the changed blocks. It would be possible to identify corrupted blocks automatically. This would help to identify common storage problems. These hashes don't have to be cryptography correctly(sha). You could use [CRC](https://en.wikipedia.org/wiki/Cyclic_redundancy_check) for this. This wouldn't lower the performance drastically but increase the snapshot, backup and replica integrity significantly.

**Describe alternatives you've considered**
A cron job which automatically calculates a hash of the mounted devices and compare these. If a corrupt volume is detected with the python interface remove the replica. To work like this, the disk have to be without changes for some time. This would lead to a drastically performance decrease if we set them read only for the hash time. We also considered if we could use the snapshot data for this...

**Additional context**
Consumer SSDs are dying in a very bad way. We already saw disks which start to add 0-Bytes somewhere on the disks. We also saw different other situation where the data is completely corrupted at the end of live of the ssd. We know Ceph is using some automatic healing and detection of disk failures, but ceph is quite more complicated than longhorn. We like the simplicity of longhorn and would like to use consumer hardware for the storage. But without automatic detection/healing it would end in a couple of problems which would cost a lot of manual work, individual development on our side or cost in form of buying enterprise ssds. So we have to consider if these costs would compensate the costs of the introduction of ceph.

jenting:
Probably we could choose [hash](https://cyan4973.github.io/xxHash/) since the calculation time is faster.

Adding a hash to the replica is a good idea, to automatically detect the corrupted replica.

The hash calculation could based on per-replica or per-sector. I think per-sector is more suitable because we could calculate the hash of each write operation to the sector, instead of after each writes operation is finished, calculate the full replica hash.

The current sector size is 512, so the overhead after each write operation would be affected a lot. But we need to do the performance test on write IOPS after doing the code change.

jenting:
I think we need this one to automatic identifying which replica is corrupted.
Moving to the planning.

cc @innobead

voarsh2:
> 2\. Predict the disk failure: monitor the per-disk SMART

[...]

PR #4606, opened 2022-09-22
feat(snapshot): LEP for Snapshot Checksum Calculation and Bit Rot Detection

derekbit:
[Longhorn 4210](https://github.com/longhorn/longhorn/issues/4210)
[Longhorn 3198](https://github.com/longhorn/longhorn/issues/3198)


derekbit:
@PhanLe1010 
The feature willl resync snapshots periodically. Do you have any concern for the resync and the period?

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,90 @@
...
+                             │                                     │
+                             └─────────────────────────────────────┘
+    ```
+    
+    1. Query the snapshot checksum progress and status from each replica, and then store the checksum value in `snapshot.status.checksum`
+
+- **Snapshot Checksum Verification**
+    1. snapshot controller periodically reconciles the snapshot resources
+    2. If  `snapshot.status.checksum` is set, create a monitor and issue a snapshot checksum verification request. Then, the checksums of snapshot files in each replica will be calculated individually.
+    3. Then, compare the checksums with  `snapshot.status.checksum`. If the snapshot’s checksum is mismatched with  `snapshot.status.checksum`, set the replica to `ERR` and start rebuilding.
```
shuo-wu, line 77: As we discussed, we can:
  1. Record the file modification time before and after the calculation. Then we know whether the calculation result is valid or not.
  2. If the file modification time since the last calculation, we don't need the repeat calculation.

@joshimoo mentioned that we can record the versioning info rather than the file modification time in the metafile. But I don't know how that can avoid the timestamp mess-up issue.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,90 @@
...
+## Design
+
+### Implementation Overview
+
+- **CRD**
+    - Add `snapshot.status.checksum` for recording the snapshot `SHA256` checksum.
+    - Add `snapshot.status.lastVerifyAt` for recording the timestamp of last checksum verification.
+- **Automatic Snapshot Checksum Calculation**
+    1. snapshot controller periodically reconciles the snapshot resources
+    2. If  `snapshot.status.checksum` is empty, create a monitor and issue a snapshot checksum calculation request. Then, the checksums of snapshot files in each replica will be calculated individually.
```
shuo-wu, line 42: As we discussed in the meeting, the per-snapshot monitor implementation may lead to a scalability issue. You can use per-node monitoring or rely on the engines.

shuo-wu (changes requested):

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: Can you elaborate more on the `rely on the engines`?
cc @PhanLe1010

derekbit: Use per-node monitor instead.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,97 @@
...
+
+- **CRDs**
+  - Snapshot
+    - Add `snapshot.status.checksum` for recording the snapshot `SHA256` checksum.
+    - Add `snapshot.status.lastVerifyAt` for recording the finished timestamp of the last checksum verification.
+
+  - Node
+    - Add `node.status.lastSnapshotVerifyAt` for recording the start timestamp of the last checksum verification.
+  
+- **Automatic Snapshot Checksum Calculation and Verification**
```
shuo-wu, line 54: How about the engine or replica part?
As I mentioned before, we can rely on the file modification time to skip some unnecessary calculations.

derekbit: Added a section `Snapshot Disk File Hashing` for the replica part.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,129 @@
...
+    1. Node controller
+      Create a snapshot monitor for hashing snapshot disk files as well as checking their data integrity. The monitor is consist of
+       - N task workers
+          Responsible for executing hashing/checking tasks
+
+       - 1 goroutine `processSnapshotChangeEvent()`
+          Send a snapshot hashing/checking task to  `snapshotVerifyTaskChan` channel after receiving one snapshot `UPDATE` event.
+
+       - 1 goroutine `processPeriodicSnapshotVerify()`
+          Periodically create snapshot hashing/checking tasks. The period is determined by the global setting `snapshot-data-integrity-interval`. If current time exceeds the sum of `node.status.lastVerifyAt` and `snapshot-data-integrity-interval`, then it iterates the engines' snapshots and send snapshot hashing/checking tasks to the channel `snapshotVerifyTaskChan`.
```
innobead, line 81: This is trying to use a setting to relax system loading due to polling way.

See https://github.com/fsnotify/fsnotify can help with this case to use an asynchronous way by being notified by system if there are watched snapshots get changed.

derekbit: The walkthrough of all snapshots and hash calculation/verification is necessary. The filesystem cannot be aware of the data corruption caused by the underlying block device's bit rot. fsnotify seems helpless in the bit rot detection feature.

derekbit: But fsnotify might be helpful to detect any snapshot creation/purge/merge. Once per-node monitor gets the change notification, and then it send a hash calculation request to longhorn-engine.

innobead: You should be right, especially for silent data corruption.

derekbit: Just note here: a typical way to detect the silent corruption. 
https://github.com/trapexit/scorch/tree/1.0.1

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,129 @@
...
+    - Value:
+      - Minima: 1
+      - Maxima: 3
+
+- **CRDs**
+  - Snapshot
+    - Add `snapshot.status.checksum` for recording the snapshot `SHA256` checksum.
+
+  - Node
+    - Add `node.status.lastSnapshotVerifyAt` for recording the start timestamp of the last checksum verification.
```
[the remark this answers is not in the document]
derekbit: Updated, thank you.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,141 @@
...
+            - Set the replica to ERR if a snapshot file's checksum is not equal to `snapshot.status.checksum`
+
+- **Snapshot Disk File Hashing**
+  - Prerequisite
+    - `snapshot-data-integrity` is set to `hash-only`
+
+  - While the replica process is receiving the snapshot disk file hashing request
+
+    1. Get `ctime` information of the disk file.
+    2. Get the value of the extended attribute `user.longhorn-system.checksum` recording the checksum and `ctime` of the disk file in last calculation.
```
shuo-wu, line 117: Using the extended attribute is really a good idea!

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,141 @@
...
+    5. Return checksum or error to engine.
+
+- **Snapshot Disk File Verification**
+  - Prerequisite
+    - `snapshot-data-integrity` is set to `enabled`
+
+  - The silent data corruption in snapshot disk files can be caused by the host's storage device such as bit rot or somewhere within the storage stack. Filesystem cannot be not aware of the corruption. In order to detect the corruption, Longhorn system calculates the checksum of the disk files and compares the results with the values recorded in the snapshot resources.
+    1. Calculate the checksum of the snapshot disk file
+    2. Update the extended attribute `user.longhorn-system.checksum` with the checksum and `ctime` of the file
+    3. Return checksum or error to engine.
```
shuo-wu, line 135: NIT: This part looks a little bit redundant...

derekbit: I will reorganize `Snapshot Disk File Hashing`` and `Snapshot Disk File Verification` and make them more clear.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,129 @@
...
+       - 1 goroutine `processSnapshotChangeEvent()`
+          Send a snapshot hashing/checking task to  `snapshotVerifyTaskChan` channel after receiving one snapshot `UPDATE` event.
+
+       - 1 goroutine `processPeriodicSnapshotVerify()`
+          Periodically create snapshot hashing/checking tasks. The period is determined by the global setting `snapshot-data-integrity-interval`. If current time exceeds the sum of `node.status.lastVerifyAt` and `snapshot-data-integrity-interval`, then it iterates the engines' snapshots and send snapshot hashing/checking tasks to the channel `snapshotVerifyTaskChan`. 
+   
+
+    2. The task workers fetch tasks from `snapshotVerifyTaskChan` channel. Then, issue the snapshot hashing request to engine. Then, the checksum of each replica's snapshot file is calculated individually.
+   
+    3. The worker waits until each replica's snapshot file's checksum calculation has been completed. It periodically polls the engine and checks the status during waiting periods.
```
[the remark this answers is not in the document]
derekbit: Yeah, both backup and snapshot calculation need same improvement later.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,141 @@
...
+
+  - While the replica process is receiving the snapshot disk file hashing request
+
+    1. Get `ctime` information of the disk file.
+    2. Get the value of the extended attribute `user.longhorn-system.checksum` recording the checksum and `ctime` of the disk file in last calculation.
+    3. Compare the `ctime` from 1. and 2. Recalculate the checksum if one of the conditions is met
+         - The two `ctime` are mismatched.
+         - 2's `ctime` is not existing.
+
+    4. After the checksum calculation, get `ctime` information of the disk file again. Then, compare 1's and 4's `ctime`.
```
innobead, line 122: Do we need the step 4? I thought after each recalculation, the new checksum will always be updated to the extended attribute.

innobead (commented):
In general, LGTM, but some comments needed to resolve first.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,141 @@
...
+            "pvc-abc-r-003": ce7c12a4d568...fddf2f706952,
+        }
+        ```
+
+    5. The expected checksum is determined by a majority vote between the checksums
+        - When all checksums differ and the expected checksum cannot be determined
+          - If  `snapshot.status.checksum` is empty
+            - Set all replicas are set to ERR
+          - If  `snapshot.status.checksum` is not empty
+            - Use the  `snapshot.status.checksum` as the ground truth, set the replicas with different checksums to ERR
```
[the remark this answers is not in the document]
derekbit: Yeah, the replicas cannot be fixed (rebuilding) in some extreme cases.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,141 @@
...
+
+    ![Snapshot Checksum Calculation Flow](image/snapshot_checksum_calculation_flow.png)
+
+    1. Node controller
+      Create a snapshot monitor for hashing snapshot disk files as well as checking their data integrity. The monitor is consist of
+       - N task workers
+          Responsible for executing hashing/checking tasks
+
+       - 1 goroutine `processSnapshotChangeEvent()`
+          Send a snapshot hashing/checking task to  `snapshotVerifyTaskChan` channel after receiving one snapshot `UPDATE` event.
```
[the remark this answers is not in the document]
derekbit: Yeah, I also have the same concern. Can I introduce sqlite or other persistent and lightweight database.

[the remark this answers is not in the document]
innobead: Do we really need to have that queue persistent? I thought you already introduced the concept of periodical snapshot-check, so probably volatility is good enough, because in the end it will be requested to calculate the checksum again even if it was missed last time, right?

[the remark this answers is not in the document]
derekbit: The channel receives the snapshot change events ([codes](https://github.com/longhorn/longhorn-manager/pull/1519/files#)). If we use a bounded channel,  it might result in a stuck node controller if the snapshot verification workers consumes the tasks slowly. So, putting the change events into a persistent db can avoid the memory pressure and the potention stuck issue. WDYT?

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
innobead: I see, then we need to emphasize this reason in the doc because it is worth understanding the purpose.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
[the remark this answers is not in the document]
innobead: After discussing with @derekbit , propose to adopt the below items.
- Use a buffered channel for snapshot update events, then process them w/ best efforts.
- Use non-blocking to send events to the channel, so if the channel is full, just simply drop, because we would expect the events will be required to be processed periodically in the following iterations. It will be eventual consistency for snapshot checksum calculation and verification. 
- However, it's unable to guarantee that the unprocessed events just dropped will be processed soon or later, because there is no queue for them. So, suggest the remaining unprocessed snapshots of the volume, the partial snapshots of which have been processed, should be prioritized earlier to ensure the volume can be processed and completed in a reasonable time which could be expected.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,141 @@
...
+    - Add `node.status.lastSnapshotVerifiedAt` for recording the start timestamp of the last checksum verification.
+  
+- **Automatic Snapshot Checksum Hashing and Checking**
+
+    ![Snapshot Checksum Calculation Flow](image/snapshot_checksum_calculation_flow.png)
+
+    1. Node controller
+      Create a snapshot monitor for hashing snapshot disk files as well as checking their data integrity. The monitor is consist of
+       - N task workers
+          Responsible for executing hashing/checking tasks
```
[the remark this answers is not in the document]
shuo-wu: This means that each node monitor should track the in-process replica calculation on all other nodes, which seems pretty complicated and it's hard to avoid race conditions...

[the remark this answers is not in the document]
shuo-wu: Or maybe we can add something like labels for the replicas before starting the checksum calculation. This should simplify the issue but we cannot strictly guarantee the concurrent count.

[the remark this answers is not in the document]
derekbit: Yeah, just do best effort.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,141 @@
...
+            - Assign the expected checksum to `snapshot.status.checksum`
+            - Set the replica to ERR if a snapshot file's checksum is not equal to `snapshot.status.checksum`
+          - If  `snapshot.status.checksum` is not empty
+            - Set the replica to ERR if a snapshot file's checksum is not equal to `snapshot.status.checksum`
+
+- **Snapshot Disk File Hashing**
+  - Prerequisite
+    - `snapshot-data-integrity` is set to `hash-only`
+
+  - While the replica process is receiving the snapshot disk file hashing request
```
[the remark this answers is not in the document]
derekbit: Reorganized, thank you.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: Updated. Thank you.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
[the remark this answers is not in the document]
derekbit: Add the worker pool in the replica's sync-agent for controlling number of concurrent calculations.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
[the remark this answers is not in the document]
derekbit: Use a non-blocking channel instead.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,157 @@
+# Snapshot Checksum Calculation and Bit Rot Detection
+
+## Summary
+
+Longhorn system supports volume snapshotting and stores the snapshot disk files on the local disk. However, it is impossible to verify the data integrity of snapshots due to the lack of the checksums of the snapshots in current implementation. As a result, if the underlying storage bit rots, no way is available to detect the data corruption and repair the replicas. In the enhancement, the snapshot checksum is calculated after the snapshot is taken and the is verified periodically. When a corrupted snapshot is detected, a replica rebuild is triggered to repair the snapshot.
```
c3y1huang, line 5:
```suggestion
Longhorn system supports volume snapshotting and stores the snapshot disk files on the local disk. However, it is impossible to verify the data integrity of snapshots due to the lack of the checksums of the snapshots in current implementation. As a result, if the underlying storage bit rots, no way is available to detect the data corruption and repair the replicas. In the enhancement, the snapshot checksum is calculated after the snapshot is taken and is verified periodically. When a corrupted snapshot is detected, a replica rebuild is triggered to repair the snapshot.
```

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,157 @@
...
+## Proposal
+
+### User Stories
+
+Bit rot in storage is rare but real, and it can corrupt the data silently. Longhorn supports volume snapshotting and restoring a volume to a previous version. However, due to the lack of the checksums of the snapshots in current implementation, it is impossible to ensure the data integrity of the replicas/snapshots. Although, we provide a method ([ref](https://longhorn.io/docs/1.3.1/advanced-resources/data-recovery/corrupted-replica/)) to identify the corrupted snapshots/replicas, the process is tedious and time-consuming for users.
+
+### User Experience In Detail
+
+1. Users' operations will not be affected by snapshot hashing and checking.
+2. The system will consume computing and disk IO resources while hashing snapshot disk files. In the meantime, the performance will be impacted.
```
c3y1huang, line 32: Do you have an estimate on much impact will be on the performance?

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,157 @@
...
+#### CLI
+Add `snapshot hash` and `snapshot hash-status` commands
+- `snaphost hash` issues a snapshot hashing request to engin.
+  - Usage: `longhorn --url ${engine-ip}:{$engine-port} snapshot hash tcp://${replica-sync-agent-ip}:${replica-sync-agent-port} --snapshot-name ${name}`
+- `snapshot hash-status` requests the snapshot hashing status from engine.
+  - Usage: `longhorn --url ${engine-ip}:{$engine-port} snapshot hash-status tcp://${replica-sync-agent-ip}:${replica-sync-agent-port}`
+
+#### Sync-Agent gRPC API
+Add `SnapshotHash` and `SnapshotHashStatus` methods and their request and response messages.
+- `SnapshotHash` issues a snapshot hashing request to engin.
```
c3y1huang, line 43:
```suggestion
- `SnapshotHash` issues a snapshot hashing request to engine.
```

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,157 @@
...
+### Implementation Overview
+
+#### Global Settings
+- **snapshot-data-integrity**
+  - Description: A global setting for enabling or disabling snapshot hashing and data integrity checks
+  - Type: string
+  - Value:
+    - disabled: Disable snapshot disk file hashing and data integrity checking.
+    - enabled: Enables periodic snapshot disk file hashing and data integrity checking. To detect the filesystem-unaware corruption caused by bit rot or other issues in snapshot disk files, Longhorn system periodically hashes files and finds corrupted ones. Hence, the system performance will be impacted during the periodical checking.
+    - fast-check: Enable snapshot disk file hashing and fast data integrity checking. Longhorn system only hashes snapshot disk files if their are not hashed or the modification time are changed. In this mode, filesystem-unaware corruption cannot be detected, but the impact on system performance can be minimized.
```
c3y1huang, line 57:
```suggestion
    - fast-check: Enable snapshot disk file hashing and fast data integrity checking. Longhorn system only hashes snapshot disk files if they are not hashed or if the modification time changed. In this mode, filesystem-unaware corruption cannot be detected, but the impact on system performance can be minimized.
```

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,157 @@
...
+  
+#### Automatic Snapshot Checksum Hashing and Checking
+
+![Snapshot Checksum Hashing and Checking Flow](image/snapshot_checksum_calculation_flow.png)
+
+1. Node controller creates a snapshot monitor for hashing snapshot disk files as well as checking their data integrity. The monitor is consist of
+    - **1 goroutine `processSnapshotChangeEvent()`**: Send a snapshot hashing/checking task to the bounded and non-blocking `snapshotVerifyTaskChan` buffered channel after receiving one snapshot `UPDATE` event. If the channel is full, the task is simply dropped and ignored. The dropped snapshot disk files will be processed in the following hashing and checking iterations.
+
+    - **1 goroutine `processPeriodicSnapshotVerify()`**: Periodically create snapshot hashing/checking tasks. The period is determined by the global setting `snapshot-data-integrity-interval`. If current time exceeds the sum of `node.status.lastVerifiedAt` and `snapshot-data-integrity-interval`, then it populates engines' snapshots and sends snapshot hashing/checking tasks to the `snapshotVerifyTaskChan` channel.
```
c3y1huang, line 85:
```suggestion
    - **1 goroutine `processPeriodicSnapshotVerify()`**: Periodically create snapshot hashing/checking tasks. The period is determined by the global setting `snapshot-data-integrity-interval`. If the current time exceeds the sum of `node.status.lastVerifiedAt` and `snapshot-data-integrity-interval`, then it populates engines' snapshots and sends snapshot hashing/checking tasks to the `snapshotVerifyTaskChan` channel. 
```

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,157 @@
...
+      - Volume-head disk file, i.e. `Volume Head` in the following figure
+      - System-generated snapshot disk file, e.g. `ccb017f6` and `9a8d5c9c`.
+    - Hash a snapshot disk file if its parent is not system-generated. For example, `2965ed10`, `b55d30a2` and `1bf5e639`
+    - If the parent of a snapshot disk file is system-generated, then merge the disk file with its parent and calculate the checksum. For example, merge of `8d5e5c5d` and `9a8d5c9c`
+
+
+      ![Snapshot Hash](image/snapshot_hash.png)
+
+
+3. Issue snapshot hashing requests to the their associated engines. Then, the checksum of the snapshot disk file is calculated individually in the replica process.
```
c3y1huang, line 100:
```suggestion
3. Issue snapshot hashing requests to their associated engines. Then, the checksum of the snapshot disk file is calculated individually in the replica process.
```

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,157 @@
...
+      "pvc-abc-r-003": ce7c12a4d568fddf,
+    }
+    ```
+
+6. The **final checksum** is determined by the majority of the checksums from replicas. For instance, the **final checksum** of the result in 4. is `0470c08fbc4dc702`.
+   - When all checksums differ, the **final checksum** is unable to be determined
+      - If  `snapshot.status.checksum` is empty
+        - Set all replicas to `ERR`
+      - If  `snapshot.status.checksum` is already set
+        - Use the  `snapshot.status.checksum` as the **final checksum**, set the replicas that have mismatching checksums to `ERR`
```
c3y1huang, line 118:
```suggestion
        - Use the `snapshot.status.checksum` as the **final checksum**, and set the replicas that have mismatching checksums to `ERR`
```

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,157 @@
...
+
+- **If the global setting `snapshot-data-integrity` is set to `enabled`**
+  - Because the silent data corruption in snapshot disk files can be caused by the host's storage device such as bit rot or somewhere within the storage stack. Filesystem cannot be aware of the corruption. To detect the corruption, the checksums of the disk files are always be recalculated and return back to engine.
+
+### Test Plan
+**Integration tests**
+- Test snapshot disk files hashing and checking when the some of snapshot disk files are corrupted.
+
+## Note[optional]
+### CRC64 (ISO)
```
c3y1huang, line 154: Can we have more info on what this section is for?

c3y1huang (changes requested):

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: I only measure the CPU usage during checksumming.
Do you want to check the io performance? If yes, I can benchmark it.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: No prob. Updated it as well as the test plan. Thank you.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,157 @@
...
+      - If  `snapshot.status.checksum` is already set
+        - Use the  `snapshot.status.checksum` as the **final checksum**, set the replicas that have mismatching checksums to `ERR`
+   - When the **final checksum** is successfully determined
+      - Assign the **final checksum** to `snapshot.status.checksum`
+      - Set the replica to `ERR` if its snapshot disk file's checksum is not equal to `snapshot.status.checksum`
+  
+7. Then, the replicas in `ERR` mode will be rebuilt and fixed.
+
+#### Concurrent Snapshot Disk File Hashing in Replica Process
+A snapshot hashing worker pool is created while starting the sync-agent server. The sync-agent receives the snapshot hashing requests and push them into the boundded and non-blocking `taskQueue` channel. The tasks are then consumed by the workers.
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: Updated. Thank you.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,157 @@
...
+    - Hash a snapshot disk file if its parent is not system-generated. For example, `2965ed10`, `b55d30a2` and `1bf5e639`
+    - If the parent of a snapshot disk file is system-generated, then merge the disk file with its parent and calculate the checksum. For example, merge of `8d5e5c5d` and `9a8d5c9c`
+
+
+      ![Snapshot Hash](image/snapshot_hash.png)
+
+
+3. Issue snapshot hashing requests to the their associated engines. Then, the checksum of the snapshot disk file is calculated individually in the replica process.
+   
+4. The worker waits until each snapshot disk file's checksum calculation has been completed. It periodically polls the engine and checks the status during waiting period.
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: Updated. Thank you.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: Updated. Thank you.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: Updated. Thank you.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: Updated. Thank you.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: Updated. Thank you.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
c3y1huang: > I only measure the [CPU usage](https://github.com/longhorn/longhorn/issues/4210#issuecomment-1286484094) during checksumming. 

It would be nice if we could put it here, so there is a better picture of the impact mentioned.

> Do you want to check the impact on io performance? If yes, I can benchmark it.

Do you think there will be an impact on io performance?

derekbit: Yes, I think io perfmance should be impacted.
No prob. I will add the results here. Thank you.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,160 @@
...
+
+### User Experience In Detail
+
+1. Users' operations will not be affected by snapshot hashing and checking.
+2. The system will consume computing and disk IO resources while hashing snapshot disk files. In the meantime, the performance will be impacted.
+
+#### CLI
+Add `snapshot hash` and `snapshot hash-status` commands
+- `snaphost hash` issues a snapshot hashing request to engine.
+  - Usage: `longhorn --url ${engine-ip}:{$engine-port} snapshot hash tcp://${replica-sync-agent-ip}:${replica-sync-agent-port} --snapshot-name ${name}`
```
shuo-wu, line 37: NIT: I am thinking if we need to use the first port (replica gPRC server port) for simplicity. Then cmd function should be able to get the sync agent port based on the first port.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,160 @@
...
+A snapshot hashing worker pool is created while starting the sync-agent server. The sync-agent receives the snapshot hashing requests and pushes them into the bounded and non-blocking `taskQueue` channel. The tasks are then consumed by the workers.
+
+A per-node concurrent snapshot hashing limit at the same time is configurable by editing the global setting `snapshot-data-integrity-per-node-concurrent-limit` to prevent exhaustion of computing and disk IO resources. In the case of snapshot monitor issues a new snapshot hashing request via `SnapshotHash()`, the value is sent to the sync-agent and then the active workers will be resized according to the value.
+
+#### Snapshot Disk File Hashing in Replica Process
+When the replica process received the request of snapshot disk file hashing
+
+- **If the global setting `snapshot-data-integrity` is set to `fast-check`**
+  1. Get `ctime` information of the snapshot disk file.
+  2. Get the value of the extended attribute `user.longhorn-system.checksum` recording the checksum and `ctime` of the file in the last calculation.
```
shuo-wu, line 135: NIT: `user.longhorn-system.checksum.crc64`

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,160 @@
...
+
+- Node
+  - Add `node.status.lastSnapshotVerifiedAt` for recording the start timestamp of the last checksum checking.
+  
+#### Automatic Snapshot Checksum Hashing and Checking
+
+![Snapshot Checksum Hashing and Checking Flow](image/snapshot_checksum_calculation_flow.png)
+
+1. Node controller creates a snapshot monitor for hashing snapshot disk files as well as checking their data integrity. The monitor is consist of
+    - **1 goroutine `processSnapshotChangeEvent()`**: Send a snapshot hashing/checking task to the bounded and non-blocking `snapshotVerifyTaskChan` buffered channel after receiving one snapshot `UPDATE` event. If the channel is full, the task is simply dropped and ignored. The dropped snapshot disk files will be processed in the following hashing and checking iterations.
```
shuo-wu, line 83: NIT: Can we use a workqueue (like the controllers) instead? IIRC, the workqueue can help eliminate duplications.

shuo-wu (commented):
In general LGTM

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: Provided the perfomance comparison table.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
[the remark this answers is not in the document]
derekbit: Will be improved in https://github.com/longhorn/longhorn/issues/4725

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: Using a workqueue is a good method. The workqueue's capacity is unlimited, so the system's memory might be run out if there are lots of verification tasks. So, I use the bounded channel here.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: I'm thinking if `user.longhorn-system.metadata` is better. `user.longhorn-system.checksum` contains
hashing method, hash value and ctime. WDYT?

shuo-wu: `user.longhorn-system.checksum` with hashing method, hash value and time looks good.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
shuo-wu: The queue depth is determined by the snapshot count in the cluster. Hence I think the unlimited capacity should be fine. Besides, the monitor can forget some keys when the keys reach the retry limit.

derekbit: Got it. Replacing the channel with workqueue is good to me.

derekbit:
I think `SnapshotDataIntegrityInterval` might not be user-friendly.
If we use `SnapshotDataIntegrityCron` instead, users can specify the off-peak time for hashing/verifying snapshots. WDYT? @shuo-wu @innobead

shuo-wu:
> I think `SnapshotDataIntegrityInterval` might not be user-friendly. If we use `SnapshotDataIntegrityCron` instead, users can specify the off-peak time for hashing/verifying snapshots. WDYT? @shuo-wu @innobead

Good idea! It's nice to have this feature!

innobead:
> I think `SnapshotDataIntegrityInterval` might not be user-friendly. If we use `SnapshotDataIntegrityCron` instead, users can specify the off-peak time for hashing/verifying snapshots. WDYT? @shuo-wu @innobead

Do you mean users could have more control to decide how often or frequently to run those operations? If YES, would we have any validation to avoid too aggressive execution? I assume we will do the same for interval if still using `SnapshotDataIntegrityInterval`.

derekbit:
> > I think `SnapshotDataIntegrityInterval` might not be user-friendly. If we use `SnapshotDataIntegrityCron` instead, users can specify the off-peak time for hashing/verifying snapshots. WDYT? @shuo-wu @innobead
> 
> Do you mean users could have more control to decide how often or frequently to run those operations?

Yes, for exmaple, set `0 0 * * SUN` for one execution at 00:00 on Sunday.

> If YES, would we have any validation to avoid too aggressive execution? 

If the previous verification is not finished yet, the new job won't start. Then, we can avoid the aggressive execution.

>  I assume we will do the same for interval if still using SnapshotDataIntegrityInterval.

Yes,. But users have to calculate the interval and set the value at a specific moment, then the next job can run at the next desired moment. I think it is not friendly and intuitive enough.

innobead:
@derekbit then, LGTM

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: Updated. Thank you.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,168 @@
...
+![Snapshot Checksum Hashing and Checking Flow](image/snapshot_checksum_calculation_flow.png)
+
+1. Node controller creates a snapshot monitor for hashing snapshot disk files as well as checking their data integrity. The monitor is consist of
+    - **1 goroutine `processSnapshotChangeEvent()`**: Send a snapshot hashing/checking task to the `snapshotVerifyTaskQueue` workqueue after receiving one snapshot `UPDATE` event.
+
+    - **1 goroutine `processPeriodicSnapshotVerify()`**: Periodically create snapshot hashing/checking tasks. The period is determined by the global setting `snapshot-data-integrity-cron-job`. When the job is started, it populates engines' snapshots and sends snapshot hashing/checking tasks to the `snapshotVerifyTaskQueue` channel.
+  
+     - **N task workers**: Issue hashing requests to engines and detect corrupted replicas according to the results.
+
+2. Task workers fetch tasks from `snapshotVerifyTaskQueue` and check if the snapshot disk file needs to be hashed. The rules are
```
shuo-wu, line 97: NIT: mention the concurrent limit issue

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,168 @@
...
+   - When the **final checksum** is successfully determined
+      - Assign the **final checksum** to `snapshot.status.checksum`
+      - Set the replica to `ERR` if its snapshot disk file's checksum is not equal to `snapshot.status.checksum`
+  
+7. Then, the replicas in `ERR` mode will be rebuilt and fixed.
+
+#### Concurrent Snapshot Disk File Hashing in Replica Process
+A snapshot hashing worker pool is created while starting the sync-agent server. The sync-agent receives the snapshot hashing requests and pushes them into the bounded and non-blocking `taskQueue` channel. The tasks are then consumed by the workers.
+
+A per-node concurrent snapshot hashing limit at the same time is configurable by editing the global setting `snapshot-data-integrity-per-node-concurrent-limit` to prevent exhaustion of computing and disk IO resources. In the case of snapshot monitor issues a new snapshot hashing request via `SnapshotHash()`, the value is sent to the sync-agent and then the active workers will be resized according to the value.
```
shuo-wu, line 136: The sync agent server is a per-replica server. Does sending the value to it work?

derekbit: No. The design is changed and will be based on our discussion. I'll update this part.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: Update the steps.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: Updated the concurrent limit part.

derekbit:
Please help review again. Thank you.
cc @shuo-wu @c3y1huang @innobead @joshimoo

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,161 @@
...
+     - **N task workers**: Issue hashing requests to engines and detect corrupted replicas according to the results.
+
+2. Task workers fetch tasks from `snapshotCheckTaskQueue` and check if the snapshot disk file needs to be hashed. The rules are
+    - If one of the following conditions are met, do not hash the file
+      - Volume-head disk file, i.e. `Volume Head` in the following figure
+      - System-generated snapshot disk file, e.g. `ccb017f6` and `9a8d5c9c`.
+
+      ![Snapshot Hash](image/snapshot_hash.png)
+
+3. Issue snapshot hashing requests to their associated engines. Then, the checksum of the snapshot disk file is calculated individually in the replica process. To ensure only one in-progress calculation, the worker holds the per-node file lock (`/host/var/lib/longhorn/.lock/hash`) when calculating the checksum to avoid significant storage performance drop caused by the concurrent calculations.
```
shuo-wu, line 106: NIT: need to talk a little bit more about how the workers rely on the file lock to achieve the concurrent limit... 
`To ensure only one in-progress calculation, the worker holds the per-node file lock`. This is confusing. It's more like we are hashing the snapshot one by one.

innobead (commented):
MacOS's .DS_Store in this PR. Suggest adding to .gitignore, because it seems quite commont to hit this wrong commit when using Mac.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,161 @@
...
+- **snapshot-data-integrity-cron-job**
+  - Description: The setting is a set of five fields in a line, indicating when Longhorn checks the data integrity of snapshot disk files.
+  - Type: string (Cron job format)
+
+#### CRDs
+- Snapshot
+  - Add `snapshot.status.checksum` for recording the snapshot `crc64(iso)` checksum.
+
+- Node
+  - Add `node.status.snapshotPeriodicCheckStatus.state` for indicating current periodic check state. The value can be `idle` or `in-progress`.
```
innobead, line 85: Probably can check what state we use in other status to make the pattern consistent like using suffix `ed` or `ing`.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,161 @@
...
+1. Node controller creates a snapshot monitor for hashing snapshot disk files as well as checking their data integrity. The monitor is consist of
+    - **1 goroutine `processSnapshotChangeEvent()`**: Send a snapshot hashing/checking task to the `snapshotCheckTaskQueue` workqueue after receiving one snapshot `UPDATE` event.
+
+    - **1 goroutine `processPeriodicSnapshotCheck()`**: Periodically create snapshot hashing/checking tasks. The period is determined by the global setting `snapshot-data-integrity-cron-job`. When the job is started, it populates engines' snapshots and sends snapshot hashing/checking tasks to the `snapshotCheckTaskQueue` channel.
+  
+     - **N task workers**: Issue hashing requests to engines and detect corrupted replicas according to the results.
+
+2. Task workers fetch tasks from `snapshotCheckTaskQueue` and check if the snapshot disk file needs to be hashed. The rules are
+    - If one of the following conditions are met, do not hash the file
+      - Volume-head disk file, i.e. `Volume Head` in the following figure
```
innobead, line 101: Local volume or 1-replica volume can skip hash and check.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,161 @@
...
+      - System-generated snapshot disk file, e.g. `ccb017f6` and `9a8d5c9c`.
+
+      ![Snapshot Hash](image/snapshot_hash.png)
+
+3. Issue snapshot hashing requests to their associated engines. Then, the checksum of the snapshot disk file is calculated individually in the replica process. To ensure only one in-progress calculation, the worker holds the per-node file lock (`/host/var/lib/longhorn/.lock/hash`) when calculating the checksum to avoid significant storage performance drop caused by the concurrent calculations.
+   
+4. The worker waits until each snapshot disk file's checksum calculation has been completed. It periodically polls the engine and checks the status during the waiting period.
```
innobead, line 108: Because we don't have global even queue/broker, it's possible to make replica call back to the engine instead of polling? (probably can do it in later version)

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
innobead: I would rather suggest adding concurrent check to non-goals w/ the reason we found at the moment.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,161 @@
...
+  - Value:
+    - disabled: Disable snapshot disk file hashing and data integrity checking.
+    - enabled: Enables periodic snapshot disk file hashing and data integrity checking. To detect the filesystem-unaware corruption caused by bit rot or other issues in snapshot disk files, Longhorn system periodically hashes files and finds corrupted ones. Hence, the system performance will be impacted during the periodical checking.
+    - fast-check: Enable snapshot disk file hashing and fast data integrity checking. Longhorn system only hashes snapshot disk files if they are not hashed or if the modification time changed. In this mode, filesystem-unaware corruption cannot be detected, but the impact on system performance can be minimized.
+
+- **snapshot-data-integrity-immediate-checking-after-snapshot-creation**
+  - Description: Hashing snapshot disk files impacts the performance of the system. The immediate snapshot hashing and checking can be disabled to minimize the impact after creating a snapshot.
+  - Type: bool
+
+- **snapshot-data-integrity-cron-job**
```
innobead, line 76: We will have a default value when enabling `snapshot-data-integrity`?

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,161 @@
...
+   - When the **final checksum** is successfully determined
+      - Assign the **final checksum** to `snapshot.status.checksum`
+      - Set the replica to `ERR` if its snapshot disk file's checksum is not equal to `snapshot.status.checksum`
+  
+7. Then, the replicas in `ERR` mode will be rebuilt and fixed.
+
+#### Snapshot Disk File Hashing in Replica Process
+When the replica process received the request of snapshot disk file hashing
+
+- **If the global setting `snapshot-data-integrity` is set to `fast-check`**
```
innobead, line 134: If a replica is found corrupted, could send an event to make it outstanding to you when monitoring K8s events.

innobead (changes requested):
In general, LGTM. Just some minor items needed to resolve, then we can merge this.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: The default value is `disabled`.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: Yeah. The event is emitted in the implementation.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: I thought the single replica case.
Because we are still able to detect the silent corruption in single replica's snapshot disk files, I don't skip the hash and check.

innobead: But how do you recognize there is an error when only one replica? just compare the new value with the previously saved info in the snapshot checksum field?

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
innobead: Asked about the cron job instead of the global setting for `snapshot-data-integrity`

The answer from @derekbit , `default value: “0 0 */7 * *“. One week might be reasonable`

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,161 @@
...
+Add `SnapshotHash`, `SnapshotHashStatus` and `SnapshotHashCancel` methods and their request and response messages.
+- `SnapshotHash` issues a snapshot hashing request to engine.
+- `SnapshotHashStatus` requests the snapshot hashing status from engine.
+- `SnapshotHashCancel` cancels the snapshot hashing task.
+
+## Design
+
+### Implementation Overview
+
+#### Global Settings
```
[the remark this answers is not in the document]
derekbit: We can add `snapshot-data-integrity` for volume level setting. The other two settings can only be configured globally.

How about it?

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,161 @@
...
+          - CPU: Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz
+          - Memory: 16 GB
+          - Network: Up to 10Gbps
+        - Kubernetes: v1.24.4+rke2r1
+     - Result
+       - Disk: 200 GiB NVMe SSD as the instance store
+        ![Snapshot Hash Performance Impact (SSD)](image/snapshot_hash_ssd_perf.png)
+       - Disk: 500 GiB throughput optimized HDD (st1)
+        ![Snapshot Hash Performance Impact (HDD)](image/snapshot_hash_hdd_perf.png)
+#### CLI
```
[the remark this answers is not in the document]
derekbit: Only for troubleshooting and consistency.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: Yes, I think it is possible, but not sure if there is any potential issue. We can improve in the later version.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: > But how do you recognize there is an error when only one replica?

Compare the ctime and checksum of the disk file and the values recorded in the xattr.
If ctime are identical, but checksums are mismatched, it indicats the silent corruption happen. In this case, we just emit an error event and won't do any handling.

derekbit: BTW, the silent corruption detection algo is described in `If the global setting snapshot-data-integrity is set to enabled`

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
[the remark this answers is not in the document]
innobead: Sounds good to me. It seems the other two are sub settings.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
innobead: Then we are good.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
innobead: Ok, so the same as n replicas. Let's keep it as is first, then see any feedback later.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
innobead: Sounds good. Please create a following issue then. Thanks.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: BTW, we will per volume setting. User can disable it manually if replica=1.

innobead: It sounds good, so they can do the customization if they want.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: Updated. Thank you.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
```diff
@@ -0,0 +1,161 @@
...
+Add `snapshot hash` and `snapshot hash-status` commands
+- `snaphost hash` issues a snapshot hashing request to engine.
+  - Usage: `longhorn --url ${engine-ip}:${engine-port} snapshot hash tcp://${replica-sync-agent-ip}:${replica-sync-agent-port} --snapshot-name ${name}`
+- `snapshot hash-status` requests the snapshot hashing status from engine.
+  - Usage: `longhorn --url ${engine-ip}:${engine-port} snapshot hash-status tcp://${replica-sync-agent-ip}:${replica-sync-agent-port}`
+- `snapshot hash-cancel` cancels the snapshot hashing task.
+  - Usage: `longhorn --url ${engine-ip}:${engine-port} snapshot hash-cancel tcp://${replica-sync-agent-ip}:${replica-sync-agent-port}`
+#### Sync-Agent gRPC API
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: > I would rather suggest adding concurrent check to non-goals w/ the reason we found at the moment.

Added. Thank you.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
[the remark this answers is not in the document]
derekbit: Added the `volume.spec.dataIntegrityCheck`.

derekbit:
Add `volume.spec.snapshotDataIntegrity`. The value can be `global`, `disabled`, `enabled` and `fast-check`.
`global` means the setting will following the value in the global setting `snapshot-data-integrity`.
@innobead

innobead:
The enum of value should be ignored, enabled, disabled, and fast-check.

enhancements/20220922-snapshot-checksum-and-bit-rot-detection.md
derekbit: > But it’s better to have a brief test over the calculation time. For example, when calculating a 100G or 1T snapshot (empty, full of data, intermittent data blocks), how long it will take? Then we can provide some suggestions about the hashing check interval or if the fast-check should be used when there are lots of snapshots in the cluster.

@shuo-wu I think this can be added in official doc for users' reference. WDYT?

shuo-wu: Sure!

Issue #4601, opened 2022-09-21
[BUG] the values.yaml in the longhorn helm chart contains values not used.


SweBarre:
## Describe the bug

The [values.yaml](https://github.com/longhorn/charts/blob/master/charts/longhorn/values.yaml#L172-L183) file contains the `resource: {}` block and description on how to set recourse limits and requests, however this field is not used or referenced in the chart templates.

## Expected behavior

To minimize confusion the values.yaml shouldn't specifies variables not used.

Lichiehyao:
**Verified on master-head 20221003**

- longhorn master-head ([502dcb7](https://github.com/longhorn/longhorn/commit/502dcb72ed676cc19941fd5106f7ab1ddcd79cb4)) 

**The test steps**

1.[Test steps](https://github.com/longhorn/longhorn/issues/4601#issuecomment-1254718552)


**Result Passed**

PR #4610, opened 2022-09-22
Remove resources block in values.yaml

weizhe0422:
[Longhorn 4601](https://github.com/longhorn/longhorn/issues/4601)


c3y1huang (approved):

innobead (approved):

innobead:
> LGTM
> 
> But suggest mentioning why we don't support/recommend the customized resource (cpu and memory) settings in the official doc.

@weizhe0422 Please create an issue for this, so we can have that explanation in the doc. Thanks.

PR #4659, opened 2022-10-03
Update system managed componente images for v1.3

innobead:


chart/questions.yaml
```diff
@@ -65,7 +65,7 @@ questions:
     label: Longhorn Share Manager Image Repository
     group: "Longhorn Images Settings"
   - variable: image.longhorn.shareManager.tag
-    default: v1_20220808
+    default: v1_20221003
```
derekbit, line 68: The image is not created, right?

innobead: Just published :smile:

derekbit (approved):

Issue #4525, opened 2022-09-06
[TASK] Add K8s version limitation to < 1.4 chart


innobead:
## What's the task? Please describe

Before 1.4, Longhorn does not support K8s 1.25 which will be tackled at #4490 . However, we don't add K8s upper version limitation when installing/upgrading Longhorn via Helm. This should be improved in upcoming patch releases including 1.2.6 and 1.3.2. 

## Describe the items of the task (DoD, definition of done) you'd like
> Please use a task list for items on a separate line with a clickable checkbox https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists

- [x] As titled, add K8s version range for the upstream chart

## Additional context

Add any other context or screenshots about the task request here.

cc @PhanLe1010

chriscchien:
**Verified on Longhorn v1.3.x head `47012e` and v1.2.x head `80a7ea`
Result Pass**

In k3s v1.25.2+k3s1 environment, using helm to install Longhorn v1.2.x and v1.3.x, the installation won't get started 
```
Error: INSTALLATION FAILED: chart requires kubeVersion: >=1.18.0-0 <1.25.0-0 which is incompatible with Kubernetes v1.25.2+k3s1
```

PR #4662, opened 2022-10-03
Add k8s version limitation to < 1.4 chart

weizhe0422:
[Longhorn 4525](https://github.com/longhorn/longhorn/issues/4525)


weizhe0422:
Close PR since the limitation only before v1.4.0, I will create PR based on v1.3.x and v1.2.x branch.

Issue #4675, opened 2022-10-04
[TASK] Add a network policy example for rwx volumes' recovery backend service


derekbit:
## What's the task? Please describe

We've implemented recovery backend service for rwx volumes. Then, we can provide a network policy example for addressng the connectivity between kube-apiserver and the service like  
https://github.com/longhorn/longhorn/blob/master/examples/network-policy/webhook-network-policy.yaml

## Describe the items of the task (DoD, definition of done) you'd like
> Please use a task list for items on a separate line with a clickable checkbox https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists

- [ ] `item 1`

## Additional context

Add any other context or screenshots about the task request here.

PR #4678, opened 2022-10-04
example: add network policy for recovery backend

derekbit:
longhorn/longhorn#4675


innobead (approved):
LGTM

Issue #4239, opened 2022-07-15
[FEATURE] Align with upstream K8s version support from Longhorn 1.4 and update component/pkg/build dependencies


innobead:
## What's the task? Please describe

For each feature release, we will revisit if need to bump the minimum support version of K8s and K8s dependent libraries based on different reasons like K8s support version cycle, the minimum versions of K8s used by CSI sidecar components planned to deliver/support, etc.  

https://kubernetes.io/releases/patch-releases/
At the moment, 1.25, 1.24, and 1.23 are actively maintained branches, and others will still have patch support except for [versions less and equal to 1.21, which are EOL](https://kubernetes.io/releases/patch-releases/#non-active-branch-history), so the base min version now is 1.22. 

## Describe the items of the task (DoD, definition of done) you'd like

- [x] Bump the recent versions of K8s-dependent libraries to the recently newer version given no backward compatibility concerns
- [x] Fix deprecated resources being used/removed from unsupported K8s versions https://github.com/longhorn/longhorn/issues/4003 (ex: https://github.com/longhorn/longhorn/issues/4490#issuecomment-1240400365)
- [x] Update the min version support of K8s
- [x] Update CSI sidecar versions
https://github.com/longhorn/longhorn-manager/pull/1518#discussion_r992824473 (release note required because snapshot API v1beta1 will be deprecated since 4.1. 
- [x] Update Golang version
- [x] Update the K8s version in e2e testing to the updated nearly K8s version @longhorn/qa

innobead:
ref: #4003

joshimoo:
Issues to solve for kubernetes 1.25 support:
PSP https://github.com/longhorn/longhorn/issues/4003
PDB https://github.com/longhorn/longhorn/issues/4689
Cronjob https://github.com/longhorn/longhorn/issues/3160#issuecomment-947007800

innobead:
About removed deprecated versions, it will be tackled at https://github.com/longhorn/longhorn/issues/4003 together with PSP.

PhanLe1010:
Test plan
(in-progress)

* Test generate support bundler generation 
* Test upgrade 
* Test zone and topology 
* Test uninstallation

PhanLe1010:
Todo:
1. Upding docs:
    * github.com/container-storage-interface/spec v1.3.0
    * Min Kubernetes 1.21
[...]

PR #4695, opened 2022-10-11
Kubernetes 1.2.5 support

PhanLe1010:
1. Set enablePSP to false by default
1. Bump K8s min version support and CSI sidecar versions 
    * Min Kubernetes version from 1.18 to 1.21 
    * longhornio/csi-resizer v1.2.0 -> v1.3.0 
    * longhornio/csi-snapshotter: v3.0.3 -> v5.0.1
1. Update CSI snapshot examples from v1beta1 to v1
1. Updating images file
1. Gnerating new longhorn.yaml from the chart for kubectl
1. Sync uninstallation manifest from longhorn/longhorn-manager repo

longhorn/longhorn#4003
longhorn/longhorn#4239

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 name: longhorn
 version: 1.4.0-dev
 appVersion: v1.4.0-dev
-kubeVersion: ">=1.18.0-0"
+kubeVersion: ">=1.21.0-0"
 description: Longhorn is a distributed block storage system for Kubernetes.
 keywords:
 - longhorn
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -125,7 +125,7 @@ questions:
     label: Longhorn CSI Driver Resizer Image Repository
     group: "Longhorn CSI Driver Images"
   - variable: image.csi.resizer.tag
-    default: v1.2.0
+    default: v1.3.0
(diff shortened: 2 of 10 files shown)


innobead (commented):
LGTM, but the same question for uninstall at https://github.com/longhorn/longhorn-manager/pull/1518#issuecomment-1275002235

joshimoo (commented):
LGTM need to check on the uninstall job namespace change. https://github.com/longhorn/longhorn-manager/pull/1518#discussion_r992826517

As well as make sure we communicate the potential issue with older invalid csi snapshots, which were created before the upstream snapshot controller implemented a webhook for these.
https://github.com/longhorn/longhorn-manager/pull/1518#discussion_r992824473

innobead (changes requested):
I didn't see `longhorn/templates/psp.yaml` removed from this PR? anything missing here?

PhanLe1010:
> I didn't see longhorn/templates/psp.yaml removed from this PR? anything missing here?

The default value for `enablePSP` is switched to false which means we don't shi it by default. Some users on Kubernetes < 1.25 may still need it so we keep the template there

innobead:
> > I didn't see longhorn/templates/psp.yaml removed from this PR? anything missing here?
> 
> The default value for `enablePSP` is switched to false which means we don't shi it by default. Some users on Kubernetes < 1.25 may still need it so we keep the template there

100% correct!

Issue #4401, opened 2022-08-11
[TASK] Refresh distro version support for 1.4


innobead:
## What's the task? Please describe

Longhorn supports FS volume which actually relies on some kernel modules to achieve that like ext4 and xfs. In different distros w/ different Kernel versions, probably there will some compatibility problems like unsupported mount options to cause the volume mount to fail. The doc needs to update to ensure people understand the criteria for the host kernel.

## Describe the items of the task (DoD, definition of done) you'd like

- [x] Update the supported distro in the master (1.4), 1.3, and 1.2 documentation (https://longhorn.io/docs/1.3.0/best-practices/#software) @yangchiu has updated versions for 1.4, https://longhorn.io/docs/1.4.0/best-practices/#software and for older releases, just keep it as is. 
- [ ] ~Update the environment check script to provide early notice to users~

## Additional context

https://github.com/longhorn/longhorn/issues/2640 
https://github.com/ceph/ceph-csi/issues/966#issuecomment-620655796

cc  @longhorn/qa @PhanLe1010 @derekbit

joshimoo:
We previously had a similiar issuse to support an older suse distro.
https://github.com/longhorn/longhorn/issues/1208#issuecomment-624353654

We also previously had the slow extents retrieval on EXT4 linked above already.
https://github.com/longhorn/longhorn/issues/2640
https://github.com/longhorn/longhorn/issues/2507#issuecomment-867800662


If we want to allow usage on a older kernel we can add a `fsParam` to the storage class to allow the user to specify custom params. Since a storage class always only uses a single fsType that would make sense. Since I already wrote the paramOverrider for EXT4 we can deprecate the setting and instead allow the users to use the fsParam on the storage class.
We can use a similiar approach to allow the user to specify args for the fsCHK  when provided.

Just for future reference one can create a xfs fs on the older kernels by passing:
`-m reflink=0` to the `mkfs.xfs` call.

weizhe0422:
According to the conclusion of the last meeting, since the fs type will continue to be backported in the Linux kernel.

So the task will add a check function to the `environment_check.sh` to check the supported fs types, currently ext4 and xfs.

innobead:
ref: https://github.com/longhorn/longhorn/issues/4642

innobead:
Don't really need to figure out the min Kernel support but rather provide a flexible way to let users customize fs parameters, so let's focus on https://github.com/longhorn/longhorn/issues/4642. 

[...]

PR #4708, opened 2022-10-12
feat(script): add filesystem check function in environment check script

weizhe0422:
[Longhorn 4401](https://github.com/longhorn/longhorn/issues/4401)

image

### DIFF
--- a/scripts/environment_check.sh
+++ b/scripts/environment_check.sh
@@ -6,6 +6,8 @@ export CYAN='\x1b[36m'
 export YELLOW='\x1b[33m'
 export NO_COLOR='\x1b[0m'
 
+export need_support_filesystems=(ext4 xfs)
+
 if [ -z "${LOG_TITLE}" ]; then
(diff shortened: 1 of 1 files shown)


scripts/environment_check.sh
```diff
@@ -285,12 +287,34 @@ check_iscsid() {
   fi
 }
 
+check_filesystem(){
+  supported_filesystems="Required filesystems are supported:"
+  local allFound=true
+
+  for filesystem in "${need_support_filesystems[@]}"; do
+    if [ "$(cat /proc/filesystems | grep -E $filesystem | xargs)" != "$filesystem" ]; then
```
innobead, line 295: This is not right, what you need to check if each longhorn disk should be xfs|ext4 instead, but frankly, it's not possible to check beforehand.

innobead (commented):
As I mentioned, the check should be against each longhorn disk, so it's not easy to check beforehand. 

Suggest not adding this check.

Issue #4711, opened 2022-10-13
[BUG] Cannot re-run helm uninstallation if the first one failed and cannot fetch logs of failed uninstallation pod


PhanLe1010:
## Describe the bug

Cannot re-run helm uninstallation if the first one failed and cannot fetch logs of failed uninstallation pod

## To Reproduce

1. Clone the chart repo and remove the `--force` flag [here](https://github.com/longhorn/longhorn/blob/97cb124f247e736a77c4e73e319cc4b5c2f1bb46/chart/templates/uninstall-job.yaml#L27) as a way to artificially fail the uninstalltion job
2. Install the Helm chart
3. Create some volumes and attach them
4. Do helm uninstall
5. The uninstallation job will fail because there are attached volumes and `--force` flag is not set
6. Now detach the volumes
7. Re-run helm uninstalltion again. This command always fail since the previous uninstallation job already exist 
8. Also we cannot check the log of the uninstallation pod because it is already cleanup so we are kind of lost not knowing what was the failed reason.

## Expected behavior

Users can re-run helm uninstallation and can see logs of failed uninstallation pod

## Environment

 - Longhorn version: <= 1.3.2
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl): Helm/Rancher UI

PhanLe1010:
**Test steps:**

1. Clone the chart repo and remove the `--force` flag [here](https://github.com/longhorn/longhorn/blob/97cb124f247e736a77c4e73e319cc4b5c2f1bb46/chart/templates/uninstall-job.yaml#L27) as a way to artificially fail the uninstallation job
1. Install the Helm chart
1. Create some volumes and attach them
1. Do helm uninstall
1. The uninstallation job will fail because there are attached volumes and `--force` flag is not set
1. Verify that we can check the log of the failed uninstallation pod inside `longhorn-system` namespace
1. Now detach the volumes
1. Re-run `helm uninstall` again. Very that the uninstallation succeeded

Lichiehyao:
**Verified on master-head 20221027**

- longhorn master-head ([777bc5b](https://github.com/longhorn/longhorn/commit/777bc5b0c025ea2e8bf497bf22fc9d701db50bf9)) 
- longhorn-manager master-head ([76468fa](https://github.com/longhorn/longhorn-manager/commit/76468fa5013b70c5715977e6ba2dc35c3f024974)) 


**The test steps**

Test steps:

[...]

PR #4712, opened 2022-10-13
Fix bug: cannot re-run helm uninstall if the first one failed and cannot fetch logs of failed uninstalltion job

PhanLe1010:
1. Adding `before-hook-creation` annotation value instruct Helm to remove the previous uninstallation job and before try to avoid the resource conflict https://helm.sh/docs/topics/charts_hooks/#:~:text=Description-,before%2Dhook%2Dcreation,-Delete%20the%20previous
2. Set `restartPolicy: Never` to keep the log of failed uninstallation pod so users can identify the reason that it is failing, see more https://kubernetes.io/docs/concepts/workloads/controllers/job/#:~:text=Note%3A%20If,not%20lost%20inadvertently.

longhorn/longhorn#4711

### DIFF
--- a/chart/templates/uninstall-job.yaml
+++ b/chart/templates/uninstall-job.yaml
@@ -3,7 +3,7 @@ kind: Job
 metadata:
   annotations:
     "helm.sh/hook": pre-delete
-    "helm.sh/hook-delete-policy": hook-succeeded
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
   name: longhorn-uninstall
   namespace: {{ include "release_namespace" . }}
   labels: {{- include "longhorn.labels" . | nindent 4 }}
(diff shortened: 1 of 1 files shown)


chart/templates/uninstall-job.yaml
```diff
@@ -30,7 +30,7 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
-      restartPolicy: OnFailure
+      restartPolicy: Never
```
innobead, line 33: `backoffLimit` has been configured, so is it possible `OnFailure` is intended somehow?

PhanLe1010: backoffLimit is set to 1 so I don't see a difference between `restartPolicy: Never` and `restartPolicy: restartPolicy: Never` in terms of restarting the failed container (a.k.a if container failed only 1 time, stop the job)

PhanLe1010: Also, user can always trigger `helm uninstall` again to retry so I am not too worry about `restartPolicy`

innobead (approved):

chart/templates/uninstall-job.yaml
innobead: LGTM!

PR #4735, opened 2022-10-18
feat: Snapshot checksum calculation and verification

derekbit:
https://github.com/longhorn/longhorn/issues/4210
https://github.com/longhorn/longhorn/issues/3198


innobead (approved):
LGTM

PR #4743, opened 2022-10-20
Add LEP for Longhorn snapshot CRD feature

PhanLe1010:
longhorn/longhorn#3144


innobead (approved):
LGTM. This feature has been introduced in 1.3.0.

Issue #3353, opened 2021-12-01
[FEATURE] Allow customization of the cipher used by cryptsetup in volume encryption


neilgoodman:
## Is your feature request related to a problem? Please describe

I am using a Raspberry Pi 4 and would like to use the `xchacha12,aes-adiantum-plain64` cipher for encrypted volumes, as I am following [this guidance](https://source.android.com/security/encryption/adiantum) from Android for enabling encryption on ARM devices that don't have encryption acceleration.

Outside of Longhorn, I can do this today by running `cryptsetup` with the following command:

```
cryptsetup luksFormat \
  --cipher xchacha12,aes-adiantum-plain64 \
  --hash sha256 \
  --type luks2 \
  --key-size 256 \
  --sector-size 4096 \
  --use-urandom <DEVICE>
```

I did some research and I found this [issue](https://github.com/longhorn/longhorn/issues/3285). In that issue, a link to the code that runs the `cryptsetup` command was shared [here](https://github.com/longhorn/longhorn-manager/blob/42f27d4013d8624c3f3b5c9e5f020550709fe120/csi/crypto/luks.go#L27). It looks like the defaults for the `luksFormat` action are being used, which results in the `aes-xts-plain64` cipher being applied.

## Describe the solution you'd like

I would like a configuration option to at least set the cipher being used on the `luksFormat` action when creating an encrypted volume. It would be nice to also be able to configure:

- hash algorithm
- key size

## Describe alternatives you've considered

Right now I can just encrypt my devices out-of-band from Longhorn, create an ext4 filesystem on the encrypted device, and then provide a mount to that filesystem in the node settings in the Longhorn UI. I have done this and it works, but it has the following drawbacks:

- Backups aren't encrypted
- Data is not encrypted in-flight between nodes ~~(this is an assumption I'm making based on reading the docs, I haven't had time to test and make sure)~~
  - I did some testing and confirmed that with a default install of k3s + Longhorn that traffic between nodes for replicas is not encrypted in-flight, I will have to enable IPSec or WireGuard in Flannel to get encryption in-flight

derekbit:
May can be handled with #3026

neilgoodman:
#3026 seems concerned with storing the key for an encrypted volume as an encrypted value instead of a plain text value in a Secret resource. I'm not sure it's helpful here, but please let me know if I'm not following.

[...]

PR #4777, opened 2022-10-24
Extend the LEP to allow customization of the cipher

mantissahz:
Extend the old LEP `pv-encryption.md` to allow users customize the options of cipher used by `cryptsetup`.

longhorn/longhorn#3353

Signed-off-by: James Lu <james.lu@suse.com>


enhancements/20221024-pv-encryption.md
```diff
@@ -29,20 +29,20 @@ this in return means that backups of that volume end up also being encrypted.
...
 ### User Stories
-All regular longhorn operations should also be supported for encrypted volumes, 
+All regular longhorn operations should also be supported for encrypted volumes,
 therefore the only user story that is mentioned is
 how to create and use an encrypted volume.
 
 #### Create and use an encrypted volume
 - create a storage class with (encrypted=true) and either a global secret or a per volume secret
-- create the secret for that volume in the configured namespace
+- create the secret for that volume in the configured namespace with customization options of the cipher for instance `cipher`, `key-size` and `hash`
```
c3y1huang, line 38: Do we need to add those to the test-plan?

enhancements/20221024-pv-encryption.md
```diff
@@ -112,18 +115,26 @@ parameters:
...
+- `CRYPTO_KEY_CIPHER` allow users to choose the cipher algorithm when creating an encrypted volume by `cryptsetup`,
+- `CRYPTO_KEY_HASH` specifies the hash used in the LUKS key setup scheme and volume key digest
+- `CRYPTO_KEY_SIZE` sets key size in bits. The argument has to be a multiple of 8 and maximum interactive passphrase length is 512 (characters.)
+
 ```yaml
   CRYPTO_KEY_VALUE: "Simple passphrase"
   CRYPTO_KEY_PROVIDER: "secret" # this is optional we currently only support direct keys via secrets
+  CRYPTO_KEY_CIPHER: "aes-xts-plain64" # this is optional
+  CRYPTO_KEY_HASH: "sha256" # this is optional
+  CRYPTO_KEY_SIZE: "256" # this is optional
```
derekbit, line 137: What are the default values? Any tradeoff about different settings?

mantissahz: The default values are listed as the example.
Yes, cipher and key-size have the dependency, some ciphers could not use 512 key-size.
And There are different speeds of encryption and decryption for different cipher choices.

enhancements/20221024-pv-encryption.md
```diff
@@ -112,18 +115,26 @@ parameters:
...
+Storing the `Encrypted` value is necessary to support encryption for RWX volumes.
 
 ## Design
 
 ### Implementation Overview
 Host requires `dm_crypt` kernel module as well as `cryptsetup` installed.
-We utilize the below parameters from a secret, `CRYPTO_KEY_PROVIDER` allows us in the future to add other key management systems.
+We utilize the below parameters from a secret,
+- `CRYPTO_KEY_PROVIDER` allows us in the future to add other key management systems,
+- `CRYPTO_KEY_CIPHER` allow users to choose the cipher algorithm when creating an encrypted volume by `cryptsetup`,
```
c3y1huang, line 128:
```suggestion
- `CRYPTO_KEY_CIPHER` allow users to choose the cipher algorithm when creating an encrypted volume by `cryptsetup`
```

enhancements/20221024-pv-encryption.md
```diff
@@ -112,18 +115,26 @@ parameters:
...
+- `CRYPTO_KEY_HASH` specifies the hash used in the LUKS key setup scheme and volume key digest
+- `CRYPTO_KEY_SIZE` sets key size in bits. The argument has to be a multiple of 8 and maximum interactive passphrase length is 512 (characters.)
```
c3y1huang, line 130:
```suggestion
- `CRYPTO_KEY_SIZE` sets the key size in bits. The argument has to be a multiple of 8 and the maximum interactive passphrase length is 512 (characters)
```

enhancements/20221024-pv-encryption.md
```diff
@@ -146,6 +157,14 @@ We utilize the below parameters from a secret, `CRYPTO_KEY_PROVIDER` allows us i
 - create a pod that uses that pvc for a volume mount
 - wait for pod up and healthy
 
+#### Successful Creation of an encrypted volume with customization of the cipher
+- create a storage class with (encrypted=true) and either a global secret or a per volume secret
+- create the secret with customized options of the cipher for that volume in the configured namespace
+- create a pvc that references the created storage class
+- create a pod that uses that pvc for a volume mount
+- wait for pod up and healthy
+- check if customized options of the cipher are correct
```
c3y1huang, line 166:
```suggestion
- check if the customized options of the cipher are correct
```

enhancements/20221024-pv-encryption.md
```diff
@@ -16,7 +16,7 @@ this in return means that backups of that volume end up also being encrypted.
 ## Motivation
 
 ### Goals
-- user is able to create & use an encrypted volume
+- user is able to create & use an encrypted volume with customization options of the cipher
```
c3y1huang, line 19:
```suggestion
- user is able to create & use an encrypted volume with cipher customization options
```

c3y1huang (changes requested):

enhancements/20221024-pv-encryption.md
mantissahz: Modified. Thanks.

enhancements/20221024-pv-encryption.md
mantissahz: Modified. Thanks.

enhancements/20221024-pv-encryption.md
mantissahz: Modified. Thanks.

enhancements/20221024-pv-encryption.md
mantissahz: Modified. Thanks.

enhancements/20221024-pv-encryption.md
derekbit: Is CRYPTO_KEY_SIZE  necessary If `CRYPTO_KEY_HASH` is known?

mantissahz: No, it seems that they are not dependent on each other.

joshimoo (approved):
LGTM

c3y1huang (approved):

innobead (approved):
LGTM.

But think further, probably we need to have specific settings for each provider in the future, even though right now all default to dm_crypt except CRYPTO_KEY_PROVIDER.

Issue #3715, opened 2022-03-08
[IMPROVEMENT] Consolidate volume attach/detach implementation


innobead:
## Is your improvement request related to a feature? Please describe

There are several cases related to (auto) volume attach/detach, right now we leverage the volume attributes to achieve that, but it's better to introduce a new resource (longhorn volumeattachement) to have complete context for each scenario. 

## Describe the solution you'd like

This is an idea from https://github.com/longhorn/longhorn/issues/2527#issuecomment-951901533

## Describe alternatives you've considered

N/A

## Additional context

https://github.com/longhorn/longhorn-manager/pull/1037#issuecomment-1061642104
https://github.com/longhorn/longhorn/issues/2527

cc @shuo-wu @c3y1huang @PhanLe1010 @joshimoo 

Sub-issues:
- [x] https://github.com/longhorn/longhorn/issues/3692
- [x] Refactor auto attachment for recurring job
- [x] https://github.com/longhorn/longhorn/issues/3726
- [x] Refactoring auto attachment when volume is detach unexpectedly. This is a part of the ticket https://github.com/longhorn/longhorn/issues/2527#issuecomment-966597537

joshimoo:
Here is an old PR where I explored the VA refactoring, this wasn't using a dedicated VA crd instead there was dedicated controllers per operation that would be managing desired attachment via setting `v.spec.NodeID` 

while the controller was utilizing `v.status.NodeID` to inform about attachment, there is some refactoring in there that should be applied to the volume controller indepedent of using a VA object. 

-------------

Have a look at the commit messages to see where I was going with this. Some of these have already been applied over time.
Like the csi driver having less and less logic instead having the attach call on the api making the decisions and checking all conditions.

- https://github.com/longhorn/longhorn-manager/pull/857/commits/22eff5d593d49bdc32e43a5852563114da9e95d9
- https://github.com/longhorn/longhorn-manager/pull/857/commits/8926c040e8dd8381ccfb10a05b7d87464d4268d0

-------------

This is the main part of the internal controller change:
- https://github.com/longhorn/longhorn-manager/pull/857/commits/2ac48fd936ee65fd4d9b37d445f76e698319e3a4

-------------

Using a dedicated VA resource to determine attachment seems reasonable, but we would need to think about it whether we want to have it per Node/User and we also need to consider reference counting for the VA, so that a system initiated action can also use the VA and not get cleaned up / clean up a user initiated action.

PhanLe1010:
[...]

PR #4805, opened 2022-11-01
Add LEP for AD refactoring

PhanLe1010:
longhorn/longhorn#3715


innobead:
Is this still valid? or we just base on the version 2?

joshimoo:
@innobead we will most likely base on version 2 design, @PhanLe1010 is currently working on implementation of some of the AD refactoring before we will proceed with pair programming on the implementation.

Issue #4491, opened 2022-08-26
[IMPROVEMENT]  Adjust the iSCSI timeout and the engine-to-replica timeout settings


derekbit:
## Is your improvement request related to a feature? Please describe

The engine to replica timeout is 8 seconds ([code](https://github.com/longhorn/longhorn-engine/blob/master/pkg/dataconn/client.go#L20)) in current implementaion. In an unstable network environment, the network outage triggers frequent replica rebuildings due to the short timeout setting and moreover potentially results in a worse networking status. To mitigate the issue, using a larger timeout is a possible solution. However, when adjusting the engine to replica timeout, the open-iscsi timeout should be also adjusted to avoid that the volume is remounted as read-only. The ticket aims to investigate a better iSCSI and engine-to-replica timeout settings and adjust to better values.

## Additional context

https://github.com/open-iscsi/open-iscsi/blob/master/etc/iscsid.conf
https://github.com/longhorn/longhorn-engine/commit/773b0254ba3ab02c036f1cc9d785a517d44a8f1c

mstrent:
I believe this might help some things we're seeing. We are overdue for a network equipment refresh and experience some congestion from time to time. We see these errors sometimes after a congestion "blip":
`5m          Warning   UnknownNodeConditionTrue   node/oly-k3node-53d                                      Unknown condition true of kubernetes node oly-k3node-53d: condition type is ReadonlyFilesystem, reason is FilesystemIsReadOnly, message is EXT4-fs (sdd): Remounting filesystem read-only`

Attached are dmesg errors from that node during the congestion blip:
[Node dmesg errors congestion blip Longhorn volume.txt](https://github.com/longhorn/longhorn/files/9448126/Node.dmesg.errors.congestion.blip.Longhorn.volume.txt)

My feeling is if timeouts were a bit longer, Longhorn would weather these much better.

innobead:
As discussed, not just the timeout between engine and replica, but also other timeouts like iscsi target or initiator should be considered in this fine grain tunning. 

Also, consider making these settings reasonably configurable with a best-practice range to ensure they can be adaptively adjusted as per different dynamic user environments or platform architectures.

joshimoo:
Increasing timeouts needs to consider time of failure detection, stability and performance.

[...]

PR #4880, opened 2022-11-15
chart: add engineReplicaTimeout

derekbit:
[Longhorn 4491](https://github.com/longhorn/longhorn/issues/4491)


innobead (approved):
LGTM

PR #4882, opened 2022-11-15
Update longhorn-instance-manager to v2_20221115

derekbit:
[Longhorn 4491](https://github.com/longhorn/longhorn/issues/4491)


innobead:
@derekbit Also need to update longhorn manager manifest.

derekbit:
> @derekbit Also need to update longhorn manager manifest.

Yes, updated.

Issue #4642, opened 2022-09-30
[FEATURE] Make FS format options configurable in storage class


innobead:
## Is your feature request related to a problem? Please describe

As per https://github.com/longhorn/longhorn-manager/pull/1495#issuecomment-1262972486, after a potential data corruption issue is resolved in 1.4/1.3.2, we would like to generalize the file system format options able to configure in each storage class to accompany with fstype instead of only having a global setting for a specific filesystem, ext4.

The global setting for ext4 format options will be deprecated and removed in the next feature release, so need to have a release & deprecation note.

## Describe the solution you'd like

As above

## Additional context

https://github.com/longhorn/longhorn-manager/pull/1495#issuecomment-1262972486

chriscchien:
**Verified in longhorn master `a0a641`[steps](https://github.com/longhorn/longhorn/issues/4642#issuecomment-1322883145)
Result Pass**

Setting in storage class patameter `mkfsParams` will reflect in volume which using the storageclass as below:

![Screenshot from 2022-11-23 10-56-54](https://user-images.githubusercontent.com/73337386/203461461-d80870aa-d3db-4817-8e00-7a7cf3aef818.png)
![Screenshot from 2022-11-23 10-57-49](https://user-images.githubusercontent.com/73337386/203461477-6816d679-75a6-4c27-b246-3137b23653db.png)

innobead:
@longhorn/qa I think we need an auto e2e for this feature, just added the automation label.

innobead:
@weizhe0422 I think we missed the doc here? reopened and moved to `implementation`.

chriscchien:
Verified again on longhorn master `207e74` with test [steps](https://github.com/longhorn/longhorn/issues/4642#issuecomment-1322883145)
Result **Pass**

Use customized setting `mkfsParams: "-I 512 -b 2048 -O ^metadata_csum,^64bit"` or `mkfsParams: "-I 256 -b 4096 -O ^metadata_csum,^64bit"` to deploy [storagessclass](https://raw.githubusercontent.com/longhorn/longhorn/master/examples/storageclass.yaml), `Inode Size` and `Block Size` in attaced node were all the same as mkfsParams value in storageclass.yaml.

khushboo-rancher:
@chriscchien Could you please also check a scenario of when you create a volume with the `longhorn-test` storage class and take a backup and restore it?

roger-ryao:
Hi @khushboo-rancher :
[...]

PR #4909, opened 2022-11-21
feat: Added mkfsParams to storage class example

weizhe0422:
[Longhorn 4642](https://github.com/longhorn/longhorn/issues/4642)


innobead (changes requested):
Please also update storageclass.yaml in the chart template to make it opt-in like other persistence settings in values.yaml

BTW, in 06-storageclass.yaml of longhorn-manager repo, you still need to add this new parameter but commentted out because we would like to have all parameter info aware for users, even it's not the default value.

weizhe0422:
I've add the parameter `mkfsParams` in helm chart files includes `storageclass.yaml` and `values.yaml`.

innobead (approved):

Issue #2759, opened 2021-07-01
[FEATURE] Support bundle enhancement


yasker:
We will replace the Longhorn support bundle mechanism with a general-purpose support bundle kit (WIP https://github.com/rancher/support-bundle-kit) which will support:
1. Gather YAML and logs (including previous pod log) from specified namespaces in the Kubernetes cluster
    * including automatically collect all known CRs in the namespace
2. Generate the support bundle either using binary or as a service to integrate with existing applications.
3. Gather system log on each node in the support bundle.
4. Make it easier for the user to analyze the support bundle by start a mocked Kubernetes API server with imported YAMLs and logs.

innobead:
@keithalucas will not work on this item for now.

yasker:
Move to v1.2.1 since the code change is too big for v1.2.0 after RC.

yasker:
Update on this: https://github.com/rancher/support-bundle-kit now features `simulator` command to mock a Kubernetes cluster, making it extremely easy to look into the k8s environment for troubleshooting.

innobead:
Assigned to @joshimoo who is working on the support bundle related tasks.

innobead:
@chriscchien because support-bundle-kit didn't be verified on arm64 before, so needs to ensure the tests run well on arm64 as well because arm64 is GA in 1.4.0. 

P.S. @c3y1huang has implemented e2e test cases as well, so you can verify them too. 

cc @longhorn/qa

chriscchien:
Hi @c3y1huang ,

I tried to download support bundle from ARM64 ubuntu20.04 machine, but the support bundle manager faced CrashLoopBackOff state, below are information for your reference, thank you.

```
root@ip-172-31-29-108:/home/ubuntu# cat /proc/version
Linux version 5.15.0-1019-aws (buildd@bos02-arm64-025) (gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0, GNU ld (GNU Binutils for Ubuntu) 2.34) #23~20.04.1-Ubuntu SMP Thu Aug 18 03:23:03 UTC 2022
root@ip-172-31-29-108:/home/ubuntu# 
root@ip-172-31-29-108:/home/ubuntu# k get pods | grep support
longhorn-support-bundle-manager-support-bundle-2022-11-30tnn79g   0/1     CrashLoopBackOff   6 (2m46s ago)   8m34s
```
```
[...]

PR #4915, opened 2022-11-22
feat(support bundle): chart, YAML update

c3y1huang:
https://github.com/longhorn/longhorn/issues/2759

### DIFF
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -354,6 +354,15 @@ Disabling **Backupstore Poll Interval** also means to disable failed backup auto
     type: int
     min: 0
     default: 1
+  - variable: defaultSettings.supportBundleFailedHistoryLimit
+    label: SupportBundle Failed History Limit
+    description: This setting specifies how many failed support bundles can exist in the cluster.
+The retained failed support bundle is for analysis purposes and needs to clean up manually.
+Set this value to **0** to have Longhorn automatically purge all failed support bundles.
+    group: "Longhorn Default Settings"
+    type: int
+    min: 0
+    default: 1
   - variable: defaultSettings.autoSalvage
     label: Automatic salvage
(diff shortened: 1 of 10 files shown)


chart/templates/clusterrolebinding.yaml
```diff
@@ -11,3 +11,17 @@ subjects:
...
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: longhorn-support-bundle
+  labels: {{- include "longhorn.labels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
```
innobead, line 23: Need to mention this in the doc.

chart/values.yaml
```diff
@@ -41,6 +41,9 @@ image:
     backingImageManager:
       repository: longhornio/backing-image-manager
       tag: v3_20220808
+    supportBundleKit:
+      repository: rancher/support-bundle-kit
+      tag: v0.0.13
```
innobead, line 46: `longhornio/support-bundle-kit` instead

chart/questions.yaml
```diff
@@ -354,6 +354,12 @@ Disabling **Backupstore Poll Interval** also means to disable failed backup auto
     type: int
     min: 0
     default: 1
+  - variable: defaultSettings.supportBundleFailedHistoryLimit
```
innobead, line 357: Need the description.

innobead (changes requested):

innobead (approved):

chart/templates/clusterrolebinding.yaml
c3y1huang: Update to https://github.com/longhorn/website/pull/620.

chart/values.yaml
innobead: Will need another new release, then update here before merging

PR #4916, opened 2022-11-22
feat(system-backup/restore): chart, YAML update

c3y1huang:
https://github.com/longhorn/longhorn/issues/1455

### DIFF
--- a/chart/templates/clusterrole.yaml
+++ b/chart/templates/clusterrole.yaml
@@ -11,7 +11,7 @@ rules:
   verbs:
   - "*"
 - apiGroups: [""]
-  resources: ["pods", "events", "persistentvolumes", "persistentvolumeclaims","persistentvolumeclaims/status", "nodes", "proxy/nodes", "pods/log", "secrets", "services", "endpoints", "configmaps"]
+  resources: ["pods", "events", "persistentvolumes", "persistentvolumeclaims","persistentvolumeclaims/status", "nodes", "proxy/nodes", "pods/log", "secrets", "services", "endpoints", "configmaps", "serviceaccounts"]
   verbs: ["*"]
 - apiGroups: [""]
   resources: ["namespaces"]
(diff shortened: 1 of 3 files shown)


chart/templates/deployment-webhook.yaml
```diff
@@ -164,3 +167,7 @@ spec:
         {{- end }}
       {{- end }}
       serviceAccountName: longhorn-service-account
+      volumes:
+      - name: longhorn
+        hostPath:
+          path: /var/lib/longhorn/
```
innobead, line 173: Why need this?

innobead (changes requested):

chart/templates/deployment-webhook.yaml
innobead: I remembered we have a privileged issue related to webhooks, so we don't want to do some validation but rely on the host. @derekbit do you remember that?

c3y1huang: We had no validation against the backup store.

derekbit: > The validator will execute the engine binary to check if the backup already exists in the backup store.

It's a newly added feature, right?

> I remembered we have a privileged issue related to webhooks, so we don't want to do some validation but rely on the host. @derekbit do you remember that?

Yes, we limit the permission
```
        securityContext:
          runAsUser: 2000
```

c3y1huang: https://github.com/longhorn/longhorn/issues/3701

c3y1huang: Reverted the privilege change.

c3y1huang: https://github.com/longhorn/longhorn-manager/pull/1593

innobead (approved):

PR #4919, opened 2022-11-22
Update YAMLs and Chart for Filesystem Trim

shuo-wu:
#836

### DIFF
--- a/examples/storageclass.yaml
+++ b/examples/storageclass.yaml
@@ -16,6 +16,7 @@ parameters:
 #  backingImageDataSourceType: "download"
 #  backingImageDataSourceParameters: '{"url": "https://backing-image-example.s3-region.amazonaws.com/test-backing-image"}'
 #  backingImageChecksum: "SHA512 checksum of the backing image"
+#  unmapMarkSnapChainRemoved: "ignored"
 #  diskSelector: "ssd,fast"
 #  nodeSelector: "storage,fast"
 #  recurringJobSelector: '[{"name":"snap-group", "isGroup":true},
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -595,6 +595,15 @@ WARNING:
   group: "Longhorn Default Settings"
   type: string
   default: "0 0 */7 * *"
+- variable: defaultSettings.removeSnapshotsDuringFilesystemTrim
+  label: Remove Snapshots During Filesystem Trim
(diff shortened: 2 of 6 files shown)


shuo-wu:
One minor question: The YAML in this repo does not contain the ConfigMap `default-settings`. Should we add it?

examples/storageclass.yaml
```diff
@@ -16,6 +16,7 @@ parameters:
 #  backingImageDataSourceType: "download"
 #  backingImageDataSourceParameters: '{"url": "https://backing-image-example.s3-region.amazonaws.com/test-backing-image"}'
 #  backingImageChecksum: "SHA512 checksum of the backing image"
+#  unmapMarkSnapChainRemoved:
```
innobead, line 19: Give the possible values of the value.

innobead:
> One minor question: The YAML in this repo does not contain the ConfigMap `default-settings`. Should we add it?

@shuo-wu We have that in default-setting.yaml and the generated deploy/longhorn.yaml, or what you mean is different?

innobead (approved):
LGTM

shuo-wu:
> We have that in default-setting.yaml and the generated deploy/longhorn.yaml, or what you mean is different?
@innobead there is no content in ConfigMap `deploy/longhorn.yaml/longhorn-default-setting`:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: longhorn-default-setting
  namespace: longhorn-system
  labels:
    app.kubernetes.io/name: longhorn
    app.kubernetes.io/instance: longhorn
    app.kubernetes.io/version: v1.4.0-dev
data:
  default-setting.yaml: |-
```

In the old manager file, we will list all options and comment them.

innobead:
> > We have that in default-setting.yaml and the generated deploy/longhorn.yaml, or what you mean is different?
> > @innobead there is no content in ConfigMap `deploy/longhorn.yaml/longhorn-default-setting`:
> 
> ```
> apiVersion: v1
> kind: ConfigMap
> metadata:
>   name: longhorn-default-setting
>   namespace: longhorn-system
>   labels:
>     app.kubernetes.io/name: longhorn
>     app.kubernetes.io/instance: longhorn
>     app.kubernetes.io/version: v1.4.0-dev
> data:
>   default-setting.yaml: |-
> ```
> 
> In the old manager file, we will list all options and comment them.

I see, then we don't need this anymore, because users can check the values.yaml to know what settings are provided.

innobead (approved):

Issue #3957, opened 2022-05-10
[FEATURE] Local volume for distributed data workloads


innobead:
## Is your feature request related to a problem? Please describe

Longhorn is a highly available replica-based storage system, and it's good for fault tolerance, read performance, data protection, etc, but on the other side, it also needs some extra costs like requiring more disk paces for replication. 

In some cases, especially for distributed data workloads (SS) like databases (ex: Cassandra, Kafka, etc), they already have their own data replication, sharding, etc, so we should provide a better volume type for these use cases but also still support existing specific volume functionalities like snapshotting, backup/restore, etc. 

## Describe the solution you'd like

- Extend Data Locality with `strict` or `enforced` mode to require one replica should be local and next to the workload
- Connecting the local replica via a local socket file instead of a TCP connection

## Describe alternatives you've considered

N/A

## Additional context

https://github.com/longhorn/longhorn/issues/1965

derekbit:
As the benchmarking result, if the volume and the single replica are on the same node, the latency and iops can improved significantly.

![image](https://user-images.githubusercontent.com/12527233/173726307-022c758d-ff75-41f9-af4b-987b0cbe1e5e.png)

In addition, the tcp can be replaced with unix domain socket to gain more performance in this case.

derekbit:
Replace the tcp connection between the engine and the single replica with unix-domain-socket.
- The IOPS and Latency are improved.
- The bandwidth number is saturated, so the improvement is not seen here.

![image](https://user-images.githubusercontent.com/12527233/173792789-3a189f49-30f1-412a-8733-fde0bcc39510.png)

joshimoo:
@derekbit good job on the evaluation :)

innobead:
@derekbit as we discussed, let's add this to 1.4.0.

Bessonov:
This is a great news! May I ask what brings IO down in comparison to the baseline?

innobead:
> This is a great news! May I ask what brings IO down in comparison to the baseline?

It's more about the improvement of latency. Basically, there are 3 primary things:
- Data locality
- Data transfer over Unix socket instead of TCP stack
- No remote replication

Bessonov:
> It's more about the improvement of latency. Basically, there are 3 primary things:
> 
> * Data locality
> * Data transfer over Unix socket instead of TCP stack
> * No remote replication

[...]

PR #4928, opened 2022-11-23
lep: add local-volume

derekbit:
[longhorn #3957](longhorn/longhorn#3957)

### DIFF
--- a/enhancements/20221123-local-volume.md
+++ b/enhancements/20221123-local-volume.md
@@ -1,0 +1,68 @@
+# Local Volume
+
+## Summary
+Longhorn can support local volume to provide better IO latencies and IOPS.
+
+### Related Issues
+https://github.com/longhorn/longhorn/issues/3957
+
+## Motivation
+### Goals
+- Longhorn can support local volume (data locality=strict-local) for providing better IO latencies and IOPS.
+- A local volume can only have one replica.
+- A local volume supports the operations such as snapshot, backup and etc.
+
+### Non-goals
+- A local volume's data locality cannot be converted to other modes when volume is not detached.
+- A local volume does not support multiple replicas in the first version. The local replication could be an improvement in the future.
+
+## Proposal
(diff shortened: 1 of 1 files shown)


enhancements/20221123-local-volume.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Motivation
+### Goals
+- Longhorn can support local volume (data locality=strict-local) for providing better IO latencies and IOPS.
+- A local volume can only have one replica.
+- A local volume supports the operations such as snapshot, backup and etc.
+
+### Non-goals
+- A local volume's data locality cannot be converted to other modes when volume is not detached.
+- A local volume does not support multiple replicas.
```
innobead, line 17: note: local replication could be an improvement in the future.

enhancements/20221123-local-volume.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Validating Webhook
+- If a volume with `strict-local` data locality, the `numberOfReplicas` should be 1.
+- If a local volume is attached, the conversion between `strict-local` and other data localities is not allowable.
+- If a local volume is attached, the update of the replica count is not allowable.
+
+### Test Plan
+#### Integration tests
+1. Successfully create a local volume with `numberOfReplicas=1` and `dataLocality=strict-local`.
+2. Check the validating webhook can reject the following cases
```
innobead, line 65:
```suggestion
2. Check the validating webhook can reject the following cases when the volume is created or attached
```

innobead (approved):
LGTM

enhancements/20221123-local-volume.md
derekbit: Updated. Thank you.

enhancements/20221123-local-volume.md
derekbit: Updated. Thank you.

innobead (approved):

PR #4931, opened 2022-11-24
chore: missing mkfsParams

c3y1huang:
https://github.com/longhorn/longhorn/pull/4909
https://github.com/longhorn/longhorn/issues/4642

### DIFF
--- a/chart/templates/storageclass.yaml
+++ b/chart/templates/storageclass.yaml
@@ -23,9 +23,7 @@ data:
       {{- if .Values.persistence.defaultFsType }}
       fsType: "{{ .Values.persistence.defaultFsType }}"
       {{- end }}
-      {{- if .Values.persistence.defaultMkfsParams }}
       mkfsParams: "{{ .Values.persistence.defaultMkfsParams }}"
-      {{- end }}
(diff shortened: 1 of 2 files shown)


c3y1huang:
Already fixed. close.

c3y1huang:
Actually the removal of `mkfsParams: ""` might not be intended. Reopen.

weizhe0422:
I think the parameter 'mkfsParam' in 'deploy\longhorn.yaml' should be exists.

Also, I would like to ask why the if condition in 'template\storageclass.yaml' should be removed.

c3y1huang:
> In addition, I would like to ask why the if condition in 'template\storageclass.yaml' should be removed.

When running [generate-longhorn-yaml.sh](https://github.com/longhorn/longhorn/blob/master/scripts/generate-longhorn-yaml.sh) `'mkfsParam` will not render `if .Values.persistence.defaultMkfsParams` is an empty value.

c3y1huang:
> I think the parameter 'mkfsParam' in 'deploy\longhorn.yaml' should be exists.

After checking with @innobead , we will keep it the way it is. Closing.

Issue #3907, opened 2022-04-29
[IMPROVEMENT] liveness and readiness probes with longhorn csi plugin daemonset


sharanbabumg:
As of now after patching of our worker nodes are are rebooting every node after patching, but we ended up with pvc not able attach or mount issues and pods were stuck in container creating state only.

MountVolume.MountDevice failed for volume "pvc-d0af4c3e-355d-40db-88d5-cbec82fe3f00" : rpc error: code = Internal desc = Get "[http://longhorn-backend:9500/v1/volumes/pvc-d0af4c3e-355d-40db-88d5-cbec82fe3f00":](http://longhorn-backend:9500/v1/volumes/pvc-d0af4c3e-355d-40db-88d5-cbec82fe3f00%22:) dial tcp 172.26.151.93:9500: connect: no route to host

The above error keep on triggering and for csi plugin pods and pvc's. Can we get liveness and readiness probe configurations for longhorn-csi-plugin daemonsets so that whenever we face issue for these pods we can keep alerts for these pods and fix immediately.

driver name driver.longhorn.io not found in the list of registered csi drivers

Now the workaround we perform is restarting longhorn-cis-plugin pods.

sharanbabumg:
@PhanLe1010  Kindly reopen the task and give some fix or work around this. As we have deployed longhorn in uat and production k8s clusters. After patching activity on worker nodes the nodes get rebooted and the kubelet restart happens and we are facing issue with longhorn-csi-plugin pods to be in crashloopbackoff state so we manually need to restart pods for now as work around to fix the application pods which are in container creating state. 

This would not be a permananet solution, as customer come to know the state of their pods via alerts triggered to their emails. So if this issue can be fixed permanently this would be great help for us otherwise will loose customers.

If possible kindly add readiness and liveness probe for longhorn-csi-plugin

Kindly treat this on priority and update us asap.

innobead:
> We should also consider implementing the GRPC Probe for the csi plugin, to do an active health check.
> 
> * [https://github.com/container-storage-interface/spec/blob/master/spec.md#probe](https://github.com/container-storage-interface/spec/blob/master/spec.md?rgh-link-date=2022-07-12T01%3A20%3A35Z#probe)
[...]

Issue #4945, opened 2022-11-28
[TASK] Create longhornio liveness probe image


c3y1huang:
## What's the task? Please describe

We need a `longhornio` liveness probe v2.8.0 image, similar to other [csi images](https://github.com/longhorn/longhorn/blob/master/deploy/longhorn-images.txt#L1-L5).

## Describe the items of the task (DoD, definition of done) you'd like

- [ ] Mirror a [https://github.com/kubernetes-csi/livenessprobe](https://github.com/kubernetes-csi/livenessprobe) v2.8.0 image to a longhornio image

## Additional context

https://github.com/longhorn/longhorn/issues/3907

yangchiu:
image published: https://hub.docker.com/r/longhornio/livenessprobe/tags

PR #4947, opened 2022-11-28
feat(csi-liveness-probe): update chart and YAML

c3y1huang:
https://github.com/longhorn/longhorn/issues/3907


c3y1huang:
Depending on https://github.com/longhorn/longhorn/issues/4945.
cc @innobead

innobead (approved):

PR #4954, opened 2022-11-28
Add PSP yaml

PhanLe1010:
longhorn/longhorn#4003

### DIFF
--- a/deploy/podsecuritypolicy.yaml
+++ b/deploy/podsecuritypolicy.yaml
@@ -1,0 +1,61 @@
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: longhorn-psp
+spec:
+  privileged: true
+  allowPrivilegeEscalation: true
+  requiredDropCapabilities:
+    - NET_RAW
+  allowedCapabilities:
+    - SYS_ADMIN
+  hostNetwork: false
+  hostIPC: false
+  hostPID: true
+  runAsUser:
+    rule: RunAsAny
+  seLinux:
+    rule: RunAsAny
+  fsGroup:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  volumes:
+    - configMap
+    - downwardAPI
+    - emptyDir
+    - secret
+    - projected
+    - hostPath
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: longhorn-psp-role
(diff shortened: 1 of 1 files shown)


deploy/podsecuritypolicy.yaml
```diff
@@ -0,0 +1,61 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: longhorn-psp-role
+subjects:
+  - kind: ServiceAccount
+    name: longhorn-service-account
+    namespace: longhorn-system
+  - kind: ServiceAccount
+    name: default
```
innobead, line 60: I know this exists for a while, but just notice that why we need to bind to `default`?

innobead (approved):
LGTM

deploy/podsecuritypolicy.yaml
PhanLe1010: Binding to `default` service account means that any pod created in `longhorn-system` namespace will use this PSP if they don't specify a service account name. Probably Longhorn creates some pods that don't specify a service account?

innobead: yeah, good inference. @shuo-wu do you know that?

shuo-wu: I think it's related to the old uninstaller pod which was in namespace `default`.

PhanLe1010: > I think it's related to the old uninstaller pod which was in namespace default.

The `default` keyword in this context is default service account of the `longhorn-system` namespace. I think it shouldn't effect any pod outside of `longhorn-system` namespace. It only affects pods that are inside `longhorn-system` namespace and doesn't specify a service account name. 

Just scanned the code base and I don't find such a pod yet. Maybe we can blindly remove the default bind and see if pod fail to come up :D However, I feel it is extra work for us with not much benefit

shuo-wu: OH sorry. My bad, I misread it...

> I don't find such a pod yet
I cannot find it either.

> I feel it is extra work for us with not much benefit since we are moving away from PSP anyways.
Agree. We can leave it here for now.

PhanLe1010: Oops. No need to sorry at all. Thank you for jumping in

innobead: yeah, just keep it as is, but don't miss any chance to understand why it's here :slightly_smiling_face:

Issue #4783, opened 2022-10-26
[IMPROVEMENT] Speed up replica rebuilding by the metadata such as ctime of snapshot disk files


derekbit:
## Is your improvement request related to a feature? Please describe

The replica rebuilding process in current Longhorn is time-consuming. It walks through all snapshot disk files. Each data block is hashed in the local and the remote disk files to determine if it needs to be sent to remote side. It is io-intensive and computing-intensive process, especially if the volume is large or contains tons of snapshot disk files.

ctime and checksum of snapshot disk files are already recorded in their extended attributes after introduing the [feature](https://github.com/longhorn/longhorn/issues/4210).  To speed up the rebuilding process, we can compare the file's current ctime and the ctime in the extended attribute. If they are identical, we can skip the synchronization of this disk file. Then, a reduction in the number of unnecessary computations can speed up the entire process as well as reduce the impact of the system performance.

innobead:
Aggressively adding to 1.4.0, but do this by best efforts.

yasker:
Do we have concern over just use checksum alone? How does create time work for the replica has been rebuilt once?

derekbit:
The checksum of a snapshot disk file is calculated and is recorded in the extended attribute when checking the data integrity. The `ctime` of the disk file is recorded as well.

Then, we can simply use `ctime` or `checksum` for speeding up the rebuilding.

- ctime
  - No need to be couple with `checksum`, so it should work without `checksum`, but cannot detect silent corruption.
  - If the `ctime` in the extended attribute is same as the `ctime` of the file (got from `stat`), it means the disk file is not changed after calulating the checksum, and then the file can be directly sent to remote.
  - If the `ctime` in the extended attribute is different than the `ctime` of the file, the rebuilding will use the legacy method. The `ctime` and `checksum` will be recalculated  and set by the periodical data integrity checking later.

- checksum
  - The downside of the `ctime` method is unable to detect the silent corruption during rebuilding.
  - To detect the silent corruption during replica rebuilding, we can rely on the recorded checksum of each snapshot disk files. It means the disk files' checksumes should be calculated again and then compare with the recorded values.

yasker:
@derekbit I see. Though, should it be ctime or mtime? Not sure if we care about the snapshot file's metadata change. 

[...]

PR #4968, opened 2022-11-30
chart: add fastReplicaRebuildEnabled setting

derekbit:
[Longhorn 4783](https://github.com/longhorn/longhorn/issues/4783)

### DIFF
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -627,6 +627,12 @@ WARNING:
   group: "Longhorn Default Settings"
   type: boolean
   default: "false"
+- variable: defaultSettings.fastReplicaRebuildEnabled
+  label: Fast Replica Rebuild Enabled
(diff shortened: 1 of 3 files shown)


chart/questions.yaml
```diff
@@ -627,6 +627,12 @@ WARNING:
   group: "Longhorn Default Settings"
   type: boolean
   default: "false"
+- variable: defaultSettings.fastReplicaRebuildEnabled
+  label: Fast Replica Rebuild Enabled
+  description: "Enabling this setting will support the fast replica rebuild. The feature relies on the checksum of snapshot disk files, so the feature is setting the snapshot-data-integrity to **enable** or **fast-check** is a prerequisite."
```
innobead, line 632: "This feature supports the fast replica rebuilding. It relies on the checksum of snapshot disk files, so setting the snapshot-data-integrity to **enable** or **fast-check** is a prerequisite."

innobead (changes requested):

chart/questions.yaml
derekbit: Updated. Thank you.

innobead (approved):

PR #4982, opened 2022-12-01
exclude Kasten explicitly

Tejeev:
There are two ways I was thinking we could word the if.  With an external variable saving any kasten resources, then an if that checks if the variable has a value, or just running the grep in the if statement.  I'm not sure which you would prefer but looks like this script tends to the former, so I prioritized that and commented out the latter.  Tested on the customer cluster, but I made some minor changes while transposing here so still need to run some tests

### DIFF
--- a/scripts/longhorn_rancher_chart_migration.sh
+++ b/scripts/longhorn_rancher_chart_migration.sh
@@ -162,7 +162,18 @@ fi
 echo ""
 echo ""
 echo "Looking up existing Resources ..."
-RESOURCES=$(kubectl get-all --kubeconfig ${DOWNSTREAM_KUBECONFIG} --exclude AppRevision -o name -l io.cattle.field/appId=${RELEASE_NAME} 2>/dev/null | sort)
+
+
+KASTENRESOURCES=`kubectl api-resources --kubeconfig ${DOWNSTREAM_KUBECONFIG} | grep kasten`
+
+if [ -z "${KASTENRESOURCES}" ]; then
+#if kubectl api-resources --kubeconfig ${DOWNSTREAM_KUBECONFIG} | grep -q kasten; then
+  RESOURCES=$(kubectl get-all --kubeconfig ${DOWNSTREAM_KUBECONFIG} --exclude AppRevision,applications.apps.kio.kasten.io,passkeys.vault.kio.kasten.io -o name -l io.cattle.field/appId=${RELEASE_NAME} 2>/dev/null | sort)
+else
+  RESOURCES=$(kubectl get-all --kubeconfig ${DOWNSTREAM_KUBECONFIG} --exclude AppRevision -o name -l io.cattle.field/appId=${RELEASE_NAME} 2>/dev/null | sort)
+fi
+
+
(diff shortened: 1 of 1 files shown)


innobead:
@Tejeev Please fix the DCO check. Thanks.

innobead:
cc @longhorn/qa

scripts/longhorn_rancher_chart_migration.sh
```diff
@@ -162,7 +162,18 @@ fi
 echo ""
 echo ""
 echo "Looking up existing Resources ..."
-RESOURCES=$(kubectl get-all --kubeconfig ${DOWNSTREAM_KUBECONFIG} --exclude AppRevision -o name -l io.cattle.field/appId=${RELEASE_NAME} 2>/dev/null | sort)
+
+
+KASTENRESOURCES=`kubectl api-resources --kubeconfig ${DOWNSTREAM_KUBECONFIG} | grep kasten`
```
PhanLe1010, line 167: Bug: if there is no kasten resource in the cluster, this command will return exit code 1 which stops the whole script.

scripts/longhorn_rancher_chart_migration.sh
```diff
@@ -162,7 +162,18 @@ fi
...
+
+if [ -z "${KASTENRESOURCES}" ]; then
+#if kubectl api-resources --kubeconfig ${DOWNSTREAM_KUBECONFIG} | grep -q kasten; then
+  RESOURCES=$(kubectl get-all --kubeconfig ${DOWNSTREAM_KUBECONFIG} --exclude AppRevision,applications.apps.kio.kasten.io,passkeys.vault.kio.kasten.io -o name -l io.cattle.field/appId=${RELEASE_NAME} 2>/dev/null | sort)
+else
+  RESOURCES=$(kubectl get-all --kubeconfig ${DOWNSTREAM_KUBECONFIG} --exclude AppRevision -o name -l io.cattle.field/appId=${RELEASE_NAME} 2>/dev/null | sort)
+fi
```
PhanLe1010, line 174: If there is no KASTENRESOURCES, we should not get them right?

```suggestion
if [ -z "${KASTENRESOURCES}" ]; then
  RESOURCES=$(kubectl get-all --kubeconfig ${DOWNSTREAM_KUBECONFIG} --exclude AppRevision -o name -l io.cattle.field/appId=${RELEASE_NAME} 2>/dev/null | sort)
fi
else
  RESOURCES=$(kubectl get-all --kubeconfig ${DOWNSTREAM_KUBECONFIG} --exclude AppRevision,applications.apps.kio.kasten.io,passkeys.vault.kio.kasten.io -o name -l io.cattle.field/appId=${RELEASE_NAME} 2>/dev/null | sort)
fi
```

PhanLe1010 (changes requested):
I have some comments.

After resolving these comments, we need to do 2 tests for this PR @Tejeev 

1. Test with a cluster that has Kasten again. Can you help with this one @Tejeev because Longhorn team doesn't have this setup?
2. Test with a cluster that doesn't have Kasten. I can test this one

Issue #4558, opened 2022-09-14
[FEATURE] New setting to limit the concurrent volume restoring from backup


c3y1huang:
## Is your feature request related to a problem? Please describe

We like to support to [rolling out](https://github.com/longhorn/longhorn/pull/4388) the volume with data from backup/backing-image. However, Longhorn has no limitation on the concurrent volume restoring from backup. 

Longhorn should have another setting similar to `concurrent-replica-rebuild-per-node-limit`.

## Describe the solution you'd like

Introduce a new setting to limit the concurrent volume creating from the backup.

## Describe alternatives you've considered

Handle in rollout-load controller.

## Additional context

https://github.com/longhorn/longhorn/pull/4388#discussion_r968154357

chriscchien:
Hi @c3y1huang 

I have a 3 nodes cluster, 8 backups from different volumes and Longhorn set `Concurrent Volume Backup Restore Per Node Limit` to 1
After I select all backups in backup page then click `Restore latest Backup`
The concurrent restore number somehow will greater then 3 as below picture shown.

![Screenshot from 2022-12-09 17-03-15](https://user-images.githubusercontent.com/73337386/206665121-e303f003-2d38-48ae-95b3-a05641f77b7e.png)

In addition, I had a simple test script to calculate rebuilding numbers from backend API, when restoring, somehow it will give me number greater than 3 which reflected the situation as UI displayed.
```
def test_count_restore_limit(client):

    for i in range(1000):
        try:
            restoring = 0
            volumes = client.list_volume()
            for v in volumes:
                if v.restoreStatus and v.restoreStatus[0].progress != 0 and v.restoreStatus[0].progress != 100:
                    restoring = restoring + 1
                else:
                    pass
        except:
            print("x")
            pass

        print(restoring)
        time.sleep(1)
```

chriscchien:
Did several times of restoring volume with 1 replica and Concurrent Volume Backup Restore Per Node Limit set to 1

At begging the restore number = 3 matched the setting. But there still have chance that the restoring volume number surpassed  Concurrent Volume Backup Restore Per Node Limit value, not every time, but not very hard to reproduce ( Can reproduce every time when restoring volume with 3 replicas) 

![Screenshot_20221212_084949](https://user-images.githubusercontent.com/73337386/206939697-5c276982-add2-4a69-8238-7f80543ce8e3.png)

chriscchien:
[...]

PR #4991, opened 2022-12-05
feat(lep): concurrent backup restore design

c3y1huang:
https://github.com/longhorn/longhorn/issues/4558

### DIFF
--- a/enhancements/20221205-concurrent-backup-restore-limit.md
+++ b/enhancements/20221205-concurrent-backup-restore-limit.md
@@ -1,0 +1,78 @@
+# Concurrent Backup Restore Per Node Limit
+
+## Summary
+Longhorn has no boundary on the number of concurrent volume backup restoring.
+
+Having a new `concurrent-backup-restore-per-node-limit` setting allows the user to limit the concurring backup restoring. Setting this restriction lowers the potential risk of overloading the cluster when volumes restoring from backup concurrently. For ex: during the Longhorn system restore.
+
+### Related Issues
+https://github.com/longhorn/longhorn/issues/4558
+
+## Motivation
+### Goals
+Introduce a new `concurrent-backup-restore-per-node-limit` setting to define the boundary of the concurrent volume backup restoring.
+
+### Non-goals
+`None`
+
+## Proposal
+1. Introduce a new `concurrent-backup-restore-per-node-limit` setting.
+1. Track the number of per-node volumes restoring from backup with atomic count (thread-safe) in the engine monitor.
+
+### User Stories
+Allow the user to set the concurrent backup restore per node limit to control the risk of cluster overload when Longhorn volume is restoring from backup concurrently.
+
+### User Experience In Detail
+1. Longhorn holds the engine backup restore when the number of volume backups restoring on a node reaches the `concurrent-backup-restore-per-node-limit`.
+1. The volume backup restore continues when the number of volume backups restoring on a node is below the limit.
+
+## Design
+
+### Implementation Overview
+
+#### The `concurrent-backup-restore-per-node-limit` Setting
+
+This setting controls how many engines on a node can restore the backup concurrently.
+
+Longhorn engine monitor backs off when the volume [backup restoring count](#track-the-volume-backup-restoring-per-node) reaches the setting limit.
+
+Set the value to **0** to disable backup restore.
+
+```
+Category = SettingCategoryGeneral,
+Type     = integer
+Default  = 5  # same as the default replica rebuilding number
+```
+
+#### Track the volume backup restoring per node
+
+1. Create a new atomic counter in the engine controller.
+   ```
+   type EngineController struct {
+      restoringCounter util.Counter
+   }
+   ```
+1. Pass the restoring counter to each of its engine monitors.
+   ```
+   type EngineMonitor struct {
+      restoringCounter util.Counter
+   }
+   ```
+
+1. Increase the restoring counter before backup restore.
+   > Ignore DR volumes (volume.Status.IsStandby).
+1. Decrease the restoring counter when the backup restore caller method ends
+
+### Test plan
+
+- Test the setting should block backup restore when creating multiple volumes from the backup at the same time.
+- Test the setting should be per-node limited.
+- Test the setting should not have effect on DR volumes.
+
+### Upgrade strategy
+
+`None`
+
+## Note [optional]
+
+`None`


enhancements/20221205-concurrent-backup-restore-limit.md
```diff
@@ -0,0 +1,66 @@
...
+#### The `concurrent-backup-restore-per-node-limit` Setting
+
+This setting controls how many engines on a node can restore the backup concurrently.
+
+Longhorn blocks the backup restore once the restoring engine count exceeds the limit.
+
+Set the value to **0** to disable backup restore.
+
+```
+Category = SettingCategoryDangerZone,
```
derekbit, line 45: Why put the setting in the `SettingCategoryDangerZone` category?

enhancements/20221205-concurrent-backup-restore-limit.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Implementation Overview
+
+#### The `concurrent-backup-restore-per-node-limit` Setting
+
+This setting controls how many engines on a node can restore the backup concurrently.
+
+Longhorn blocks the backup restore once the restoring engine count exceeds the limit.
+
+Set the value to **0** to disable backup restore.
```
derekbit, line 42: At first glance of the statement, I thought it means the replica on the node cannot be restored if the value is **0**.

IIUC, `concurrent-backup-restore-per-node-limit` limits the number of engines for backup restore in the node.  **0** means the engines on the nodes canno do backup restore, although its all replicas are all on different nodes. Correct? If yes, **0** might be a bit of confusing to users.

c3y1huang: Not sure if I understand what you mean, a bit confused.

Setting it to 0 means all engines on any node cannot do the backup restore. This means regardless of where the replicas are, none will be able to restore.

enhancements/20221205-concurrent-backup-restore-limit.md
c3y1huang: I am just putting it in the same zone as the [Concurrent Replica Rebuild Per Node Limit](https://longhorn.io/docs/1.3.2/references/settings/#concurrent-replica-rebuild-per-node-limit) since those settings look similar. But just my 2 cents, we can put it somewhere else.

enhancements/20221205-concurrent-backup-restore-limit.md
```diff
@@ -0,0 +1,66 @@
...
+```
+Category = SettingCategoryDangerZone,
+Type     = integer
+Default  = 5
+```
+
+#### The `longhorn.io/backup-restore-priority` Label
+
+1. Tag Longhorn Engine when volume needs to restore from backup. The value is the time in Unix nano. Then let the engine be in the restore back off. This allows Longhorn to tag all Engines when there are concurrent volumes requested for backup restore.
+1. Tag  Longhorn Engine when backup restore started. The value is `in-progress`. The number of Longhorn Engines with this labeled value will be compared to the [concurrent-backup-restore-per-node-limit](#concurrent-backup-restore-per-node-limit) setting value.
```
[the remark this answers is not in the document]
c3y1huang: Fixed, thanks.

enhancements/20221205-concurrent-backup-restore-limit.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Motivation
+### Goals
+Introduce a new `concurrent-backup-restore-limit` setting to allow the user to set a boundary to the number of concurrent volume backup restoring. 
+
+### Non-goals
+`None`
+
+## Proposal
+1. Introduce a new `concurrent-backup-restore-limit` setting.
```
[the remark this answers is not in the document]
c3y1huang: Sure. Updated.

enhancements/20221205-concurrent-backup-restore-limit.md
derekbit: Ah, thank you. Got it. It is a global setting.

enhancements/20221205-concurrent-backup-restore-limit.md
derekbit: Not sure the definition of `General` and `DangerZone`? Any rule we can follow?
cc @shuo-wu @innobead

innobead: This is the commit message for adding this catalog (danger zone). It's a highlight to remind users when changing settings in this catalog will potentially cause some impacts on active volumes. But, there is no enforcement here, but notice purpose instead.

```
setting: recreate instance manager pod when GuaranteedEngineCPU changed

Make sure user don't change GuaranteedEngineCPU when there are active
workload running.

Also, add catalog `Danger Zone` for settings that may have big impact on
running workload.
```

enhancements/20221205-concurrent-backup-restore-limit.md
```diff
@@ -0,0 +1,66 @@
+# Concurrent Backup Restore Per Node Limit
+
+## Summary
+Currently, Longhorn does not allow the user to set a boundary on the number of concurrent volume backup restoring.
+
+Have a new `concurrent-backup-restore-per-node-limit` setting allows user to limit the concurring backup restoring. This can help to lower the potential risk of overloading cluster when volume requested to restore from backup at the same time.
```
innobead, line 6: Suggest renaming to `concurrent-volume-backup-restore-per-node-limit` to differentiate system backup restore.

enhancements/20221205-concurrent-backup-restore-limit.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Non-goals
+`None`
+
+## Proposal
+1. Introduce a new `concurrent-backup-restore-per-node-limit` setting.
+1. Introduce a new `longhorn.io/backup-restore-priority` label
+   - Tag Longhorn Engine with time in `Unix nano` when volume needs to restore from backup. This value is used to determine the priority of the volume backup restore.
+   - Tag with `in-progress` during backup restore. This value will be used to check on the current number of restoring volumes.
+
```
innobead, line 23: Can we just make it simple to have a similar mechanism in engine controller as we have done for replica rebuilding? You can create an atomic count in the controller, then pass it to each monitor. 

https://github.com/longhorn/longhorn-manager/blob/5a80438e195cdb19fabcbc2a916a1dc41a3c3272/controller/replica_controller.go#L508-L512

innobead (changes requested):

enhancements/20221205-concurrent-backup-restore-limit.md
innobead: Will this setting impact DR volume? I would expect DR volume will not be impacted, because it's delta volume restore, so should not be counted here.

enhancements/20221205-concurrent-backup-restore-limit.md
```diff
@@ -0,0 +1,83 @@
...
+This setting controls how many engines on a node can restore the backup concurrently.
+
+Longhorn engine monitor backs off when the volume [backup restoring count](#track-the-volume-backup-restoring-per-node) reaches the setting limit.
+
+Set the value to **0** to disable backup restore.
+
+```
+Category = SettingCategoryGeneral,
+Type     = integer
+Default  = 5
```
innobead, line 44: The same as the default number of replica rebuilding.

enhancements/20221205-concurrent-backup-restore-limit.md
```diff
@@ -0,0 +1,83 @@
...
+	restoringCounter util.Counter
+}
+```
+1. Pass the restoring counter to each of its engine monitors.
+```
+type EngineMonitor struct {
+       ...
+
+	restoringCounter        util.Counter
+	restoringCounterAquired bool
```
innobead, line 63: `restoringCounterAquired` purpose?

enhancements/20221205-concurrent-backup-restore-limit.md
```diff
@@ -0,0 +1,83 @@
...
+}
+```
+1. Increase the restoring counter before backup restore and set restoring counter aquired.
+1. Decrease the restoring counter when restoring counter is aquired and:
```
innobead, line 67: `Decrease the restoring counter when restoring counter is acquired when:`

innobead (changes requested):

enhancements/20221205-concurrent-backup-restore-limit.md
c3y1huang: Updated 
`> Ignore checking on DR volumes (volume.Status.IsStandby).`

enhancements/20221205-concurrent-backup-restore-limit.md
c3y1huang: Moved to `General`.

enhancements/20221205-concurrent-backup-restore-limit.md
c3y1huang: This is better. Thanks and updated.

innobead (dismissed):

enhancements/20221205-concurrent-backup-restore-limit.md
```diff
@@ -0,0 +1,88 @@
...
+   ```
+1. Pass the restoring counter to each of its engine monitors.
+   ```
+   type EngineMonitor struct {
+          ...
+   
+   	restoringCounter        util.Counter
+   
+   	# restoringCounterAquired limits 1 restoring counter per engine.
+   	# The purpose is to ensure we don't blindly increase/decrease the restoring counter.
```
innobead, line 65: Not sure, but will check the implementation PR.

c3y1huang: Updated.

Issue #5012, opened 2022-12-07
[IMPROVEMENT] Check the hostname of each node in environment_check.sh


derekbit:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

One requirement of the RWX volume feature is the hostname of of each node should be unique because of the recovery backend. `environment_check.sh` needs to support the check of the uniqueness.

innobead:
@derekbit Could you also mention this notice to https://longhorn.io/docs/1.4.0/deploy/important-notes/ and https://longhorn.io/docs/1.4.0/deploy/install/#installation-requirements?

derekbit:
Yes. I will update this on the these pages.

innobead:
ref: https://github.com/longhorn/website/pull/634

PR #5016, opened 2022-12-07
environment_check.sh: check the hostname uniqueness

derekbit:
longhorn/longhorn#5012

### DIFF
--- a/scripts/environment_check.sh
+++ b/scripts/environment_check.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+######################################################
+# Log
+######################################################
 export RED='\x1b[0;31m'
 export GREEN='\x1b[38;5;22m'
 export CYAN='\x1b[36m'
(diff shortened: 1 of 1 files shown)


scripts/environment_check.sh
```diff
@@ -233,70 +242,101 @@ check_package_installed() {
...
     error "Please install missing packages."
     exit 2
-  else
-    info "Required packages are installed."
   fi
+
+  info "Required packages are installed."
+}
+
+check_hostname_uniqueness() {
```
innobead, line 260: Why not just check the K8s node resource? there should be hostname in the status?

ref: https://kubernetes.io/docs/concepts/architecture/nodes/#addresses

innobead (changes requested):

scripts/environment_check.sh
innobead: But the change LGTM.

derekbit: Ah, right. Checking the node resources is better.

derekbit: Updated. Thank you.

innobead (approved):

innobead:
cc @longhorn/qa

PR #5088, opened 2022-12-16
chart: support customized number of replicas of webhook and recovery-backend

derekbit:
Longhorn#5087

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -218,6 +218,54 @@ longhornUI:
   #  label-key1: "label-value1"
   #  label-key2: "label-value2"
 
+longhornConversionWebhook:
+  replicas: 2
+  priorityClass: ~
+  tolerations: []
+  ## If you want to set tolerations for Longhorn conversion webhook Deployment, delete the `[]` in the line above
+  ## and uncomment this example block
+  # - key: "key"
+  #   operator: "Equal"
+  #   value: "value"
+  #   effect: "NoSchedule"
+  nodeSelector: {}
+  ## If you want to set node selector for Longhorn conversion webhook Deployment, delete the `{}` in the line above
+  ## and uncomment this example block
+  #  label-key1: "label-value1"
+  #  label-key2: "label-value2"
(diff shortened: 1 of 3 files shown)


chart/values.yaml
```diff
@@ -218,6 +218,54 @@ longhornUI:
   #  label-key1: "label-value1"
   #  label-key2: "label-value2"
 
+longhornConversionWebhook:
+  replicas: 2
+  priorityClass: ~
+  tolerations: []
```
innobead, line 224: Need to update toleration and node selector doc, like https://longhorn.io/docs/1.4.0/advanced-resources/deploy/node-selector/.

derekbit: OK. https://github.com/longhorn/longhorn/issues/5089

chart/values.yaml
```diff
@@ -218,6 +218,54 @@ longhornUI:
...
+  ## If you want to set tolerations for Longhorn UI Deployment, delete the `[]` in the line above
```
innobead, line 225: The comment is for longhorn UI, so we probably need to correct it. (even I think this kind of info should be chart README instead)

chart/values.yaml
```diff
@@ -218,6 +218,54 @@ longhornUI:
...
+  ## and uncomment this example block
+  # - key: "key"
+  #   operator: "Equal"
+  #   value: "value"
+  #   effect: "NoSchedule"
+  nodeSelector: {}
+  ## If you want to set node selector for Longhorn UI Deployment, delete the `{}` in the line above
```
innobead, line 232: The same, the comment needs to fix, the same for recovery backend and admission webhook.

innobead (changes requested):

chart/values.yaml
derekbit: Updated. Thank you.

innobead (approved):

innobead:
@derekbit Please backport to 1.4.x.

Issue #5107, opened 2022-12-20
[BUG][v1.4.0-rc1] image security scan CRITICAL issues


yangchiu:
## Describe the bug (🐛 if you encounter this issue)

There are [2 CRITICAL issues](https://ci.longhorn.io/job/private/job/image-security-scan/847/) in longhorn v1.4.0-rc1 images:

`(1) support-bundle-kit-v0.0.16.SUSE-SU-2022:0283-1.krb5-1.19.2-150300.7.7.1 [CRITICAL] (from longhornio_support-bundle-kit_v0.0.16 (suse linux enterprise server 15.3)) Security update for samba`: https://ci.longhorn.io/job/private/job/image-security-scan/847/testReport/junit/support-bundle-kit-v0.0.16/SUSE-SU-2022_0283-1/krb5_1_19_2_150300_7_7_1__CRITICAL_/

`(2) longhorn-ui-v1.4.0-rc1.SUSE-SU-2022:0283-1.krb5-1.19.2-150300.7.7.1 [CRITICAL] (from longhornio_longhorn-ui_v1.4.0-rc1 (suse linux enterprise server 15.3)) Security update for samba`: https://ci.longhorn.io/job/private/job/image-security-scan/847/testReport/junit/longhorn-ui-v1.4.0-rc1/SUSE-SU-2022_0283-1/krb5_1_19_2_150300_7_7_1__CRITICAL_/

cc @innobead 

## To Reproduce

3. Perform '....'

## Log or Support bundle

If applicable, add the Longhorn managers' log or support bundle when the issue happens. 
You can generate a Support Bundle using the link at the footer of the Longhorn UI.

## Environment

 - Longhorn version:
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl):
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version:
   - Number of management node in the cluster:
   - Number of worker node in the cluster:
 - Node config
   - OS type and version:
   - CPU per node:
   - Memory per node:
   - Disk type(e.g. SSD/NVMe):
   - Network bandwidth between the nodes:
 - Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal):
 - Number of Longhorn volumes in the cluster:

innobead:
@mantissahz  Please work with @bk201 to fix the support bundle one by releasing a new version. If having any questions, check with @c3y1huang .

[...]

PR #5113, opened 2022-12-20
build(image): bump support-bundle-kit

mantissahz:
bump support-bundle-kit version to v0.0.17

longhorn/longhorn#5107


innobead (approved):

innobead:
@mantissahz Please backport to 1.4.x which is for 1.4.0 release.

PR #5115, opened 2022-12-20
Add LEP to reimplement longhorn-engine with SPDK

DamiaSan:
Signed-off-by: Damiano Cipriani <damiano.cipriani@suse.com>


enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
```diff
@@ -0,0 +1,168 @@
...
+
+What we aim is to create an external orchestrator that, with JSON-RPC calls towards multiple instances of `spdk_tgt` app running in different machines, could manage the durability and reliability of data. Actually, not all needed functionalities to do that are already available in SPDK, so some new JSON-RPC commands will be developed over SPDK. The name of such an orchestrator could be `longhorn-spdk-agent`, taking as example [Storage Management Agent](https://spdk.io/doc/sma.html) that has some similar functionalities.
+
+* The main purpose of `longhorn-spdk-agent` is to create and export via NVMe-oF logical volumes from multiple replica nodes (one of them likely local), attach to these volumes on a controller node, use resulting bdevs to create a RAID1 bdev and exporting it via NVMe-oF locally. At this point NVMe Kernel module can be used to connect to this NVMe-oF subsystem and so to create a block device `/dev/nvmeXnY` to be used by the Longhorn CSI driver. In this way we will have multiple replica of the same data written on this block device.
```
derekbit, line 41: For a local replica, does it need any configuration in Longhorn? O, the SPDK can be automatically aware of it and select the best strategy itself?

DamiaSan: Actual SPDK RAID1 implementation (it is still under code review) do a read balancing over all available bdevs without be aware if they are attached to a remote or to a local volume. So, if we want a different strategy, for example to give priority to local volume, we will need to implement that.
Local replica will need some configuration in Longhorn because the procedure to create it is different from the remote one

derekbit: Got it. Thank you.

derekbit: Another question
Can the RAID1 only have one bdev? For example, if an user creates a volume with one replica initially, then he would like to update the replica number to 3. Is there any configuration change required for this case?

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
```diff
@@ -0,0 +1,168 @@
...
+* Below a diagram that shows the control plane of the proposal ![SPDK New Architecture](./image/spdk-control-plane.png)
+* In release 23.01, support for ublk will be added in SPDK: with this functionality we can directly create a block device without using the NVMe layer on Linux kernel versions >6.0. This will be a quite big enhancement over using NVMe-oF locally.
```
derekbit, line 43: This means `instacne-manager` or `longhorn-spdk-agent` needs to check the associated hosts' kernel version and kernel module. If the not supported, we will automatically fall back to the configuration using NVMe layer. Correct?

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
```diff
@@ -0,0 +1,168 @@
...
+
+The `longhorn-spdk-agent` will be responsible to make all others control operations, like for example creating snapshots over all replicas of the same volume. Other functionalities orchestrated by the Agent will be the remote rebuild, a complete rebuild of the entire snapshot stack of a volume needed to add or repair a replica, the backup and restore, export/import of a SPDK logical volumes to/from sparse files stored on an external storage system via S3.
```
derekbit, line 45: Curiously, why not merge `longhorn-spdk-agent` and `instance-manager` into one process?

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
```diff
@@ -0,0 +1,168 @@
...
+
+`spdk_tgt` by default starts with a single thread, but it can be configured to use multiple threads: we can have a thread per core available on the CPU. This will increase the performance but comes with the cost of an high CPU utilization. Working in polling mode instead than in interrupt mode, CPU core utilization by a single thread is always rising 100% even with no workload to handle. This could be a problem, so we can configure `spdk_tgt` with dynamic scheduler: in this way, if no workload is present, only one core will be used and only one thread will continue polling. Other thread will be put in a idle state and will become active again only when needed. Moreover, dynamic scheduler has a way to reduce the CPU frequency. (See future work section.)
```
derekbit, line 64: Maybe need to make number of threads globally configurable by users.
Any downside on utilizing the dynamic scheduler?

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
```diff
@@ -0,0 +1,168 @@
...
+
+
+### Snapshots
+
+When `longhorn-spdk-agent` receive a snapshot request from `instance-manager`, before to proceed all I/O operations over volume's block device `/dev/nvmeXnY` must be stopped to ensure that snapshots over all the replicas contains the same data.
+Actually there is no way to suspend the I/O operation over a block device, but if it is acceptable that this block device can disappear for the time needed to make the snapshot, then we can perform the operation (if the block device is created with the NVMe layer, it can be done using a null bdev namespace into the NVMe-oF subsystem used to export the volume to be snapshotted; for ublk we must wait the release of this feature to see what functionalities it will have).
```
derekbit, line 70: Does the disappearance of the block device lead to any error to the POD?
Before snapshotting, executing `fsfreeze` to flushing the page cache and freezing IO operations is a common way. Is this method doable?
Freezing the IO in the block device level means the not all data are flushed. It may differ from what a user or application expects.

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
```diff
@@ -0,0 +1,168 @@
...
+* create a new volume on the node to be rebuilt
+* export this volume via NVMe-oF
+* attach to this volume in the node where we have the source data
+* _copy snapshotA over the attached volume_
+* perform a snapshot over the exported volume
+* repeat the copy and snapshot operations for snapshotB
+
+What we have to implement is a JSON-RPC to copy a logical volume over an arbitrary bdev (that in our case will represent a remote volume exported via NVMe-oF and locally attached) _while the top layer is also being modified_ (see next section).
+
+So, in this way we can rebuild the snapshot stack of a volume. But what about the live data? Actual `longhorn-engine` make the replica rebuild in an "hot" way, i.e., during the rebuilding phase it is writing over the live data of the new volume. So, how can we reproduce this with SPDK? First of all we have to wait the conclusion of RAID1 bdev's review process to see what kind of replica rebuild will be implemented. But, supposing that the rebuild feature will not be useful for us, we will need to create a couple of additional JSON-RPC over SPDK to implement the following procedure (in _italic_ what we miss):
```
derekbit, line 93: > First of all we have to wait the conclusion of RAID1 bdev's review process to see what kind of replica rebuild will be implemented. 

Could you provide the link for reference?

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
```diff
@@ -0,0 +1,168 @@
...
+
+
+### Backup and Restore
+
+Backup will be implemented exporting a volume to a sparse file and then save this file over an external storage via S3. SPDK already has a `spdk_dd` application that can copy a bdev to a file and this app has an option to preserve bdev sparseness. But using spdk_dd has some problems: actually the sparse option works only with bdev that represent a local logical volume, not an exported one via NVMe-oF. So to backup a volume we cannot work on a remote node where to export this volume, we need to work on the node where we have the data source. But in this way, to perform a backup, we would need to stop the `spdk_tgt` app, run the `spdk_dd` and then restart the `spdk_tgt`. This operation is needed because it could not be safe to run multiple spdk applications over the same disk (even if spdk_dd would read from a read only volume) and moreover `spdk_dd` could not see the volume to export if this has been created after the last restart of `spdk_tgt` app. This because blobstore metadata, and so newly created logical volume, are saved on disk only on application exit.
```
derekbit, line 108: In the case of exporting the volume to a sparse file, does the sparse file consume additional storage space?
If yes, the sparse file might consume lots of storage space temporarilly while backing up the volume. Could there be a `virtual sparse file` for backup, and read IOs to the file gets redirected to bdev's blocks?

derekbit: Just saw the second solution, looks the additional storage space is not required. Correct?

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
```diff
@@ -0,0 +1,168 @@
...
+* clone from backup over each replica, importing the backup sparse file into a new thinly provisioned logical volume. We can perform this operation over the local node, owner of the new volume, if for the backup process we choose to develop a JSON-RPC to export/import logical volume to/from sparse files. Otherwise we can do it or over a dedicated node with `spdk_dd` application, that handle sparse file with SEEK_HOLE and SEEK_DATA functionalities of `lseek`.
+
+If we leverage the same backup & restore mechanism of `longhorn-engine`, we can restore a backup done by the actual engine to a SPDK volume.
+
+
+### Remote Control
+
+The JSON-RPC API by default is only available over the `/var/tmp/spdk.sock` Unix domain socket, but SPDK offer the sample python script [rpc_http_proxy](https://spdk.io/doc/jsonrpc_proxy.html) that provides http server which listens for JSON objects from users. Otherwise we could use the `socat` application to forward requests received from an IP socket towards a Unix socket. Both `socat` and `rpc_http_proxy` can perform user authentication with password.
```
derekbit, line 126: This is fine. But why not add a TCP/IP RPC interface in the spdk_tgt?

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Motivation
+
+These are the reasons that have driven us:
+
+* Use SPDK to improve performance of Longhorn
+* Use SPDK functionality to improve reliability and robustness
+* Use SPDK to take advantage of the new features that are continuously added to the framework
+
+### Goals
```
derekbit, line 27: Add the prerequisites and limitations, such as what underlying disks are supported and what are not in the `Goals` and `Non-goals`.

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
DamiaSan: RAID1 needs at least 2 bdevs to be created, even if at run-time it can operate with only one bdev too. So, for a volume with only one replica, we have to use directly the bdev corresponding to the logical volume. After that, if the user want to update the replica number to 3, we have to create a new RAID1 bdev and to add this one to the NVMe-oF namespace in place of the previous one. This operation doesn't need any change in offline configuration

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
DamiaSan: Yes, it is correct. I suppose it can be `longhorn-spdk-agent` that, just at the start, can check kernel version and module, so choosing which strategy to adopt for block device creation.

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
DamiaSan: After the release of first version of Longhorn supporting SPDK, we will have to handle both `longhorn-engine` and `longhorn-spdk-agent`: it will be `instance-manager` to manage this scenario.

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
DamiaSan: As described [here](https://spdk.io/doc/scheduler.html), the system may be less responsive when events occur.
I think it can happen for example in case of a big amount of data that comes after an "idle" period: the system needs some time to let reactor start polling again and to assign threads to it

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
DamiaSan: Great idea, I think `fsfreeze` is exactly what we need. `longhorn-spdk-agent` could call `fsfreeze` to suspend access to the file system created over the block device, then it would create the snapshot on all the replica of the volume and finally it can resume the access to the file system. Internally, SPDK ensure that any previous write operation will be executed before the snapshot will be done.

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
DamiaSan: https://review.spdk.io/gerrit/c/spdk/spdk/+/15272/30
This is the link to the last commit done over RAID1 (you can see previous commits of the same patch series in Relation Chain)

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
DamiaSan: We can't create virtual sparse file because information about data and holes of logical volume can be accessed only by an SPDK application. Maybe we could export the snapshot for the backup as a block device and create a JSON-RPC command to know which are the allocated clusters of the volume. So we could read only the data and skip the holes from an external application that would perform the backup via S3.
What about the second solution, it requires storage only on the dedicated node where the backup runs, so yes, it doesn't require additional storage space over the replica nodes.

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
DamiaSan: I don't know the reason, maybe to not add authentication management inside SPDK

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
l-mb: As we assemble the stack over time, it might make sense to evolve this upstream (in the future work section) to start from one bdev and then add them. This needs to be possible anyway, since we may end up in the "one bdev only" bdev state only as we replace replicas.

The IO being only round-robin is a good start. We already will need to add a "write-only" mode (for the rebuild path), so similarly we could consider a more advanced read balancing strategy *if needed*, e.g., number of outstanding IOs. Again, something for potential future work items.

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
l-mb: I don't think this is worthwhile. We'd be _starting_ in a backwards-compatible mode and efforts even before we've shipped the first version. We should not needlessly increase our QA matrix.

By the time our SPDK version ships to be production-ready, the kernel version should be widespread (or at least available via an update), as should the SPDK version itself (which we bundle).

It of course needs to be checked, but my current recommendation would be to then abort if our requirements aren't met. We should focus development and testing on one stack.

(If there's a strong argument to be made in favor of NVMeOF over ublk, then we may want to consider _only_ doing the former for the forseeable future.)

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
l-mb: Making this tunable is important. For the edge scenarios though, the power consumption of preventing CPU cores from going to sleep is likely the worse trade-off vs a brief latency spike while the system wakes up.

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
l-mb: fsfreeze is insufficient. It won't work for raw/direct block access (e.g., VMs, or other data stores).

We need this built into the frontend block layer. Either by inserting a device-mapper target (but that's fairly heavy weight) that we can suspend/resume, or - for the ublk case - this should be a "simple enough" flag to implement: do not take new IO from the submission ring, and only proceed once all currently in-flight ops have completed.

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
l-mb: With the API call in the SPDK application to export the sparse incremental snapshot, we can have the receiver of that data directly write it out to the backup target (whether that's a file somewhere on NFS or an S3 target).
That should not require any intermediate on-disk storage, only a couple of megabytes of buffer in RAM during the backup process.

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
l-mb: I'm not a fan of running a socket-to-network-RPC wrapper as a separate process. I feel we should embed this interaction into the longhorn engine process we're running on all participating nodes.

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
derekbit: Agreed. Less QA matrix is better. BTW, `ublk` is still an _experimental_ feature in latest linux kernal. For the stability, NVMeOF might be a good choice.

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
derekbit: Thanks for the explanation. @DamiaSan @l-mb

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
derekbit: > fsfreeze is insufficient. It won't work for raw/direct block access (e.g., VMs, or other data stores).

You're absolutely right.
IIUC, we need to use fsfreeze and block layer suspend method simutaneously, right?

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
DamiaSan: Actual RAID1 implementation doesn't let to add bdev and I think the reason why is the necessity to have the rebuild functionality. One of RAID1 developer replied to me that the rebuild is under development, so maybe the feature to add a bdev will be included in next commits. At that point I think that start a RAID1 with only one bdev will be possible.
What about read balancing, actual algorithm tries to evenly distribute load by sending read I/O to all base bdevs following round-robin, but skipping base bdev that processed most data so far.

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
DamiaSan: So does Longhorn support direct block access too? Because the [documentation](https://longhorn.io/docs/1.4.0/concepts/#13-csi-driver) says that _The Longhorn CSI driver takes the block device, formats it, and mounts it on the node_.

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
DamiaSan: With longhorn engine process do you mean `longhorn-spdk-agent` right?

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
DamiaSan: SPDK has no documented prerequisites,  so we are not aware of limitations on the kind of disk to use even if SPDK is designed to better operate with NVMe drives.
What about SPDK version to use, at the beginning we will use self built binaries because some functionalities we will use, like RAID1, are not still released and moreover we will need to make some development over SPDK.

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
joshimoo: I agree with Lars, it makes more sense to just focus on the traditional NVME-OF path and make it a hard requirement.

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
joshimoo: If we make this tunable, can it be done at runtime or is a restart of the spdk-tgt required?

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
joshimoo: Yes the csi driver supports the 2 volume modes: FS and Block, ref: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#volume-mode

The only difference is that in FS mode we are responsible for filesystem creation on the block device.
While in the block mode, we only bind mount the /dev/longhorn (device file)

In the case where we use encryption we wrap the raw block dev via the device mapper (dm_crypt) then replace the reference and bind mount that.

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
DamiaSan: SPDK app starts by default with static scheduler and we can change to dynamic scheduler at runtime with a JSON-RPC call. But once passed to the dynamic one we cannot come back to the static at runtime, a restart is required.

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
derekbit: > What about read balancing, actual algorithm tries to evenly distribute load by sending read I/O to all base bdevs following round-robin

Is it possible to replace the round-robin algorithm with a customized one?

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
derekbit, line 64: IIUC, one `spdk_tgt` consumes at least one CPU core. Does it limit number of `spdk_tgt` instances on one node?

joshimoo: From an architecture standpoint, let me summarize:
- 1 spdk_tgt per node, which will handle many volumes/replicas (actual io) this is required to take advantage of spdks concurrency, this differs from longhorns current architecture where there is one process per volume which only allows for optimization for a single volumes io, while with a focus on per node with many volumes you can optimize io for all volumes concurrently, you can further easily to quality of service for different volume types, since you now have a central instance that manages all io for all volumes on node X.
- for each volume there will be a longhorn-spdk-agent as a controlling program that interfaces with the longhorn-manager

derekbit: Clear. Thank you @joshimoo

joshimoo (commented):
LGTM, there are some things that are under specified:
- how to handle SPDK_TGT updates? are online updates possible or will there be interruption of service, from the node perspective, from the cluster perspective (rolling updates).
- write up current limitations, i.e. dynamically adding new replicas to the raid1 device (rebuilds)

enhancements/20221213-reimplement-longhorn-engine-with-SPDK.md
DamiaSan: At the moment there is no support for that

innobead:
> LGTM, there are some things that are under specified:
> 
> * how to handle SPDK_TGT updates? are online updates possible or will there be interruption of service, from the node perspective, from the cluster perspective (rolling updates).
> * write up current limitations, i.e. dynamically adding new replicas to the raid1 device (rebuilds)

@DamiaSan we need to add limitations and spdk_tgt upgrade strategy to this LEP as well.

innobead:
Let's update this after the first integration ready.

innobead (approved):
In general, LGTM. However, some parts have been changed or possible different in the following development.

- replica rebuilding. it seems a WO replica is not required anymore in an upstream proposal. 
- remote control. We are using our home-made go-spdk-helper instead
- backup restore, it will be implemented by creating a new RPC command exposed by spdk_tgt to do it at runtime
- v1 -> v2 migration. Live migration could be a lower priority but rather focus on v1 <->v2 backup restore

About the above changes, we should have different LEPs when the feature owner starts working on it. 

cc @longhorn/dev-data-plane

innobead:
This PR has been adopted in 1.5.

Issue #5110, opened 2022-12-20
[IMPROVEMENT] Support replica sync client timeout setting to stabilize replica rebuilding


derekbit:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

The HTTP client timeout in replica rebuilding is 5 seconds. The intermittent network disconnection or disk IO blocking over 5 seconds leads to the retries of rebuilding process and has the volume degraded for a long time. To mitigate the issue, the ticket is to increase the HTTP client timeout and make it configurable.

derekbit:
API/CLI changes, so need to update the upgrade test image. 
cc @longhorn/qa @chriscchien

innobead:
@derekbit this needs the doc update as well.

chriscchien:
Verified in longhorn master-head images and v1.4.x images
Result **Pass**

1. Setting `replica-file-sync-http-client-timeout` can have a customize value in ./chart/values.yaml when deploy Longhorn.
2. After deploy complete, command `kubectl get setting -n longhorn-system` can see `replica-file-sync-http-client-timeout` with correspond value in values.yaml
3. Use command `kubectl edit setting replica-file-sync-http-client-timeout -n longhorn-system` can modify the setting ans the value updated correctly.
4. With setting value to 5 or 35, try delete random replica, replica rebuild successfully.

innobead:
> @innobead The gRPC methods and api/cli versions need changes. Do we really need to backport the feature for a minor version?

Then, no need :smiling_face_with_tear:

PR #5121, opened 2022-12-21
chart: add replicaFileSyncHttpClientTimeout

derekbit:
Longhorn/longhorn#5110


innobead (approved):

innobead:
@derekbit Backport to 1.4.x.

PR #5166, opened 2022-12-29
feat(helm): support affinity for longhornManager

sjoukedv:
Similar to nodeSelector also support affinity.

- [x] Default value is empty object
- [x] Added commented example in `values.yaml`
- [x] Helm lint passes


sjoukedv:
Needs signed off message in commit

PR #5168, opened 2022-12-29
feat(helm): support affinity for longhornManager

sjoukedv:
Similar to nodeSelector also support affinity.

- [x] Default value is empty object
- [x] Added commented example in `values.yaml`
- [x] Helm lint passes 

Signed-off-by: Sjouke de Vries <info@sdvservices.nl>


shuo-wu (approved):

innobead (approved):

innobead:
@sjoukedv Thanks for the contribution. I think we can do the same for other deployments like driver, recovery backend, admission webhook, and UI.

innobead:
> Thanks for the contribution. I think we can do the same for other deployments like driver, recovery backend, admission webhook, and UI.

recovery backend, admission webhooks have been integrated into longhorn-manager in 1.5.0, so this is unnecessary now.

Issue #3836, opened 2022-04-11
[IMPROVEMENT] Periodically clean up volume snapshots


shuo-wu:
## Is your feature request related to a problem? Please describe

Some users may:
  1. don't care about historical data held by volume snapshots or would like purge system snapshots (different strategies)
  2. be worried about the extra space consumed by volume snapshots;
  3. forgot to clean up volume snapshots, which then may lead to the too-many-snpashot error

Then there is no need to retain any snapshots for the volumes. Having an automatic cleanup mechanism would be convenient.

## Describe the solution you'd like

We can achieve this by adding a new kind of recurring job or enhancing the current snapshot recurring job.

## Describe alternatives you've considered

N/A for now

LarsBingBong:
So this didn't come into v1.4.0? So it's going to hit v1.4.1 and v1.3.3 of Longhorn?

Thank you

innobead:
> So this didn't come into v1.4.0? So it's going to hit v1.4.1 and v1.3.3 of Longhorn?
> 
> Thank you

Yes, this is the plan if no surprise.

chriscchien:
Hi, I am not sure if we should update below test step because the last 1 system snapshot can not be deleted, from UI, command or recurring job  type snapshot-cleanup https://github.com/longhorn/longhorn/blob/e1ea3d7515e4302865944b3434591f9703e24ebe/enhancements/20230103-recurring-snapshot-cleanup.md?plain=1#L118

c3y1huang:
> Hi, I am not sure if we should update below test step because the last 1 system snapshot can not be deleted, from UI, command or recurring job type snapshot-cleanup
> 

Talked to @chriscchien , and this is the expected behavior. Ref: https://longhorn.io/docs/1.4.0/concepts/#243-deleting-snapshots
> But Longhorn cannot do that for the latest snapshot since there is no more recent snapshot to be conflated with the deleted snapshot. The next “snapshot” of the latest snapshot is the live volume (volume-head), which is being read/written by the user at the moment, so the conflation process cannot happen.
> 
> Instead, the latest snapshot will be marked as removed, and it will be cleaned up next time when possible.

chriscchien:
Verified on Longhorn master `e1ea3d` with test [steps](https://github.com/longhorn/longhorn/blob/e1ea3d7515e4302865944b3434591f9703e24ebe/enhancements/20230103-recurring-snapshot-cleanup.md)
Result **Pass**

[...]

Issue #5186, opened 2023-01-01
[FEATURE] Auto Trim via recurring job


atropos112:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)
First of all, thank you a lot for the hard work you put into 1.4.0, feels like Christmas. One of my favorite features that came was the Trim feature, I had crazy cases where my PVC had 120GB and i trimmed it down to 3GB, a game changer. I've noticed these massive trim's are more noticeable with storage which gain & lose data at high rate. This will mean that I have to "manually" (click a button that is) trim the PVC every few days/weeks. 

## Describe the solution you'd like
Longhorn has recurring jobs for snaps and backups, why not have one for trimming? Or perhaps have this automated based on how much data went through a PVC (that I imagine is a lot less feasible than the other recurring job option). Another possible alternative is to create an API endpoint/ kubectl command I can automate this with.

## Describe alternatives you've considered
Like already mentioned I could do it manually , and for RWX I could automate those too (calling `kubectl exec ...` on the share pod). I am unsure how I could automate it for RWO. 

## Additional context
To be clear, the feature you produced is fantastic and I realize me saying "here is an improvement, make it better" doesn't seem to show that but I do really mean it, thank you for your hard work.

migs35323:
+1, 
for some reason when i seen volume trimming was released on longhorn i was assuming it was done automatically based in a threshold or something.
With the recurring job i believe we get the flexibility make it to trim before a backup (so 120gb of a volume that actually holds 3gb of data doesn't take a day to backup) 
OR after a backup is done (to be better safe than sorry)

haoyixin:
+1

LP0101:
Until this is implemented, I wrote up a quick python container to do it. It's very barebones, so I'd recommend tweaking the script a bit if you want to use it, but it should give you the right idea.
https://github.com/LP0101/longhorn-auto-trim

It uses API calls to trim RWO volumes. For RWX volumes, it execs into the respective share-manager pod and runs `fstrim` manually.

atropos112:
Thank you for your script @LP0101 , much appreciated. You made me realise that `<longhorn url>/v1` is a thing, thats a gold mine i am never getting out of.

Dunge:
This is kinda a must have. Just a week of Prometheus running with default parameters and the disk space grow exponentially, fast going double the space it use for real. I can't imagine the amount of disk space wasted by people using Longhorn without trimming.

[...]

Issue #2732, opened 2021-06-24
[FEATURE] Automatically purge old unused snapshots


SvenC56:
**Question**

Hey, I am using Longhorn together with Kasten to provision PVCs (Longhorn) and do backups (Kasten). Kasten creates csi snapshots which result in a huge load of snapshots in Longhorn over time. After a week I could not create any more backups and in Longhorn I am presented with this warning:

```
Name: toomanysnapshots
Last Transition Time: 12 days ago
Message: Snapshots count is 249 over the warning threshold 100
Reason: TooManySnapshots
Status: True
```

Is there a possibility to automatically purge old unused snapshots? They are sent to S3 anyways, right? Otherwise they will bloat the filesystem over time.

**Environment:**
 - Longhorn version: 1.1.1
 - Kubernetes version: 20.10.3
 - Node config
   - OS type and version Ubuntu 20.04 LTS
   - CPU per node: 32
   - Memory per node: 252GiB
   - Disk type: HDD
   - Network bandwidth and latency between the nodes: n/a
 - Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal): Baremetal

**Additional context**
n/a

jenting:
Thanks for the feedback. Will Kasten delete the old CSI snapshots after the backup finished?

Currently, Longhorn would not automatically purge old unused snapshots through the CSI snapshot mechanism.
We could discuss it internally to check if it's doable.

jenting:
cc @joshimoo any feedback?

SvenC56:
> Thanks for the feedback. Will Kasten delete the old CSI snapshots after the backup finished?

This is what I found (https://docs.kasten.io/latest/install/storage.html#csi-snapshot-configuration):

> Setting deletionPolicy to Delete isn't sufficient either as some storage systems will force snapshot deletion if the associated volume is deleted (snapshot lifecycle is not independent of the volume). Similarly, it might be possible to force-delete snapshots through the storage array's native management interface. Enabling backups together with volume snapshots is therefore required for a durable backup.

> K10 creates a clone of the original VolumeSnapshotClass with the DeletionPolicy set to 'Retain'. When restoring a CSI VolumeSnapshot, an independent replica is created using this cloned class to avoid any accidental deletions of the underlying VolumeSnapshotContent.

> Currently, Longhorn would not automatically purge old unused snapshots through the CSI snapshot mechanism.
> We could discuss it internally to check if it's doable.

[...]

PR #5190, opened 2023-01-03
feat(lep): recurring snapshot delete design

c3y1huang:
https://github.com/longhorn/longhorn/issues/3836


shuo-wu (commented):
It seems that action `create` is responsible to create snapshots/backups only while action `delete` is responsible to delete snapshots/backups only. Then does it mean that users need to launch one `create` job + one `delete` job to achieve the effect of the previous recurring jobs?

What I am thinking is:
1. Add a new boolean field to indicate if `retain` should count the snapshots/backups not created by the current job.
2. Add a new boolean field to indicate that a job needs to create any snapshots/backups.

BTW, we plan to rely on recurring jobs to trigger snapshot purge periodically. Will we do anything about this? For example, if `retain == 0` or the current snapshot/backup count is not greater than `retain`, the job will blindly send a purge request only. I don't know if there is a separate ticket for it. [The ticket I found](https://github.com/longhorn/longhorn/issues/2732) is actually similar to what you mentioned.

c3y1huang:
> It seems that action `create` is responsible to create snapshots/backups only while action `delete` is responsible to delete snapshots/backups only. Then does it mean that users need to launch one `create` job + one `delete` job to achieve the effect of the previous recurring jobs?

There is no difference between the new create action and the current recurring job implementation: periodically create snapshot/backup + recurring job owning snapshot purge.

And the delete action periodically purges all snapshots (except the latest backup).

> 
> What I am thinking is:
> 
> 1. Add a new boolean field to indicate if `retain` should count the snapshots/backups not created by the current job.

Do you mean we should respect the current job retain?

> 2. Add a new boolean field to indicate that a job needs to create any snapshots/backups.

I prefer to deal by action. It can also work if there are more recurring job actions to come, ex: https://github.com/longhorn/longhorn/issues/5186.

> 
> BTW, we plan to rely on recurring jobs to trigger snapshot purge periodically. Will we do anything about this? For example, if `retain == 0` or the current snapshot/backup count is not greater than `retain`, the job will blindly send a purge request only. I don't know if there is a separate ticket for it. [The ticket I found](https://github.com/longhorn/longhorn/issues/2732) is actually similar to what you mentioned.

Currently, recurring job cleanup by the `cleanupSnapshotNames` then follows with the purge. The plan is RecurringJob with delete action will have its own `cleanupSnapshotNames` so it can be used for purge.

shuo-wu:
> There is no difference between the new create action and the current recurring job implementation:

OK. The action name `create` is a little bit misleading...

> Do you mean we should respect the current job retain?
I prefer to deal by action. It can also work if there are more recurring job actions to come

In your current design, when the action is `delete`, `retain` will calculate all kinds of snaps including user created or other jobs created. But when the action is `create`, `retain` just counts its own snaps, right?

> The plan is RecurringJob with delete action will have its own cleanupSnapshotNames so it can be used for purge.

For this part, the requirement is, having a mechanism triggers the snapshot purge only (without deleting existing valid snapshots). Since some operations like trim may generate `removed` or system snapshots without a following purge request, or somehow the previous purge/snapshot deletion may be interrupted, having this mechanism can help a lot. Can we achieve this here?

c3y1huang:
> In your current design, when the action is `delete`, `retain` will calculate all kinds of snaps including user created or other jobs created. But when the action is `create`, `retain` just counts its own snaps, right?

right.

> For this part, the requirement is, having a mechanism triggers the snapshot purge only (without deleting existing valid snapshots). Since some operations like trim may generate `removed` or system snapshots without a following purge request, or somehow the previous purge/snapshot deletion may be interrupted, having this mechanism can help a lot. Can we achieve this here?

Sure, added action `purge`.

shuo-wu:
> > In your current design, when the action is `delete`, `retain` will calculate all kinds of snaps including user created or other jobs created. But when the action is `create`, `retain` just counts its own snaps, right?
> 
> right.

OK. Here I am thinking if we can make it more intuitive... The naming of action `create` is misleading as well.


> > For this part, the requirement is, having a mechanism triggers the snapshot purge only (without deleting existing valid snapshots). Since some operations like trim may generate `removed` or system snapshots without a following purge request, or somehow the previous purge/snapshot deletion may be interrupted, having this mechanism can help a lot. Can we achieve this here?
> 
> Sure, added action `purge`.

Do we need one more action purge? How about asking users to set a `delete` job with a large `retain` number? I mean, a snap `delete` job will send a purge request regardless of the current snap count reaching the `retain` limit. Will this be tricky?

c3y1huang:
> OK. Here I am thinking if we can make it more intuitive... The naming of action `create` is misleading as well.
 
 Any ideas?
 Currently:
 - action `create` recurringly creates snapshot/backup and deletes and purges its own snapshots.
 - action `delete` recurringly deletes and purges all snapshots.
 - action `purge` recurringly purges all snapshots.
 
> Do we need one more action purge? 

I think there could be two use cases:
1. user just wants to periodically clean up system snapshots.
1. user wants to periodically delete all snapshots.

> How about asking users to set a `delete` job with a large `retain` number? I mean, a snap `delete` job will send a purge request regardless of the current snap count reaching the `retain` limit. Will this be tricky?

Not sure if we should consider the `delete` job cleanup the snapshot of the `create` job before the `create` job reached the retaining number. Respecting other retaining is tricky and IMO over complexing the solution.

The assumption is the delete/purge action will not be set to a frequent interval as the snapshot/backup so I prefer to leave this to the user.

enhancements/20230103-recurring-snapshot-cleanup.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/3836
+
+## Motivation
+
+### Goals
+
+Introduce recurring job actions (create, delete, purge).
```
innobead, line 16: `create`, `delete`, `purge` are general. Sometimes generalization can help abstract similar behaviors from things, but on the other side, it will also limit what we can support. 

I would suggest just extending types of `RecurringJobType`, and it will be good enough and flexible for us. Then, we will have `Snapshot`, `Backup`, `SnapshotDelete`, `BackupDelete` or even `SystemBackup` in the future.

enhancements/20230103-recurring-snapshot-cleanup.md
```diff
@@ -0,0 +1,134 @@
+# Recurring Snapshot Cleanup
+## Summary
+
+Longhorn recurring job automatically cleans up its owning expired snapshots during snapshot creation. For the non-recurring volume snapshots, the user needs to clean them manually.
```
innobead, line 4: `Longhorn recurring job automatically cleans up its owning expired snapshots during snapshot creation.` should be `Longhorn recurring job can automatically clean up older snapshots of volumes to retain no more than the desired number of snapshot defined for each volume.`

c3y1huang: This is a description of the current implementation. The suggested change doesn't look right.

I will revisit this and reword it to improve on clarity.

c3y1huang: Update.

enhancements/20230103-recurring-snapshot-cleanup.md
c3y1huang: I see, updated. Thanks!

shuo-wu:
>> How about asking users to set a delete job with a large retain number? I mean, a snap delete job will send a purge request regardless of the current snap count reaching the retain limit. Will this be tricky?
>
>Not sure if we should consider the delete job cleanup the snapshot of the create job before the create job reached the retaining number. Respecting other retaining is tricky and IMO over complexing the solution.

No. I mean, For `SnapshotDelete` jobs, it will blindly send a snapshot purge request. A purge request won't mark any valid snapshots as `Removed`, instead, it is just for cleaning up accumulated removed/system snaps. This does not involve respecting other retention.

enhancements/20230103-recurring-snapshot-cleanup.md
```diff
@@ -0,0 +1,121 @@
...
+
+https://github.com/longhorn/longhorn/issues/3836
+
+## Motivation
+
+### Goals
+
+Introduce new recurring job types:
+- `snapshot-delete`: periodically deletes and purges snapshots.
+- `snapshot-purge`: periodically purges snapshots.
```
innobead, line 18: When deleting a snapshot, it includes two steps `mark it as removed` and `purge`, so we should regard both of them as a deletion. If a snapshot has been marked remove, when deleting the snapshot, it will be just `purge`.

Basically, we don't need to disclose internal technical terms (delete or purge) to let users decide which deletion they should choose, but rather they should decide what types of snapshot they want to delete: system or all and the deletion will do the purge at the end.  

I would suggest two new job types: `snapshot-delete` and `system-snapshot-delete`. `snapshot-delete` will delete all types of snapshots including user created ones, but `system-snapshot-delete` will only delete system snapshots.

enhancements/20230103-recurring-snapshot-cleanup.md
```diff
@@ -0,0 +1,121 @@
...
+   metadata:
+     name: recurring-snap-delete-per-min
+     namespace: longhorn-system
+   spec:
+     concurrency: 1
+     cron: '* * * * *'
+     groups: []
+     labels: {}
+     name: recurring-snap-delete-per-min
+     retain: 2
```
innobead, line 54: Later, we will have different types of recurring jobs, so just need to emphasize the `retain` of the deletion jobs will supersede others. For example, if users have one snapshot recurring job to keep 5 retain, but also have a cleanup job with 2 retain. Eventually, there will be at most 2 snapshots retained. This is straight forward, but it can be briefly explained in the doc.

innobead (changes requested):

enhancements/20230103-recurring-snapshot-cleanup.md
innobead: @shuo-wu WDYT? :point_up:

innobead: > As I mentioned, I prefer to blindly send purge requests for all removing/system snapshots in `snapshot-delete` jobs.

Sounds good to me. WDYT? @c3y1huang

c3y1huang: We can keep the snapshot purge only for the recurring cleanup if the other use case is not required. However, I think naming the task as `snapshot-delete` might be misleading since the task is for periodically deleting the system snapshots only. I prefer to use `system-snapshot-delete` as the task name. WDYT @shuo-wu ?

innobead: All of you have fair points. 

How about using `snapshot-cleanup` instead? Cleanup just means the snapshots able to be cleaned up will be deleted accordingly, no matter what snapshot type each is, ex: removable or system snapshot.

c3y1huang: Sounds good to me. WDYT @shuo-wu ?

shuo-wu: OK. Let's do it this way. And the final consensus is below, right?
`snapshot-delete`: remove and purge all kinds of snapshots that exceed the retention count.
`snapshot-cleanup`: purge removed and system snapshots.

c3y1huang: I thought the consensus is to have `snapshot-cleanup` only. Do we still want `snapshot-delete`?

c3y1huang: After talking to @shuo-wu we still need the `snapshot-delete` to remove user-created snapshots.

So yes the consensus is below:
> snapshot-delete: remove and purge all kinds of snapshots that exceed the retention count.
> snapshot-cleanup: purge removed and system snapshots.

innobead: After thinking, it should be ok to have snapshot-delete and snapshot-cleanup for giving users more options for snapshot cleanup strategy.

However what we really care is snapshot-cleanup.

enhancements/20230103-recurring-snapshot-cleanup.md
```diff
@@ -0,0 +1,126 @@
...
+
+https://github.com/longhorn/longhorn/issues/3836
+
+## Motivation
+
+### Goals
+
+Introduce new recurring job types:
+- `snapshot-delete`: periodically deletes and purges snapshots.
+- `snapshot-cleanup`: periodically purges snapshots.
```
innobead, line 18: snapshot-delete: periodically remove and purge all kinds of snapshots that exceed the retention count.
snapshot-cleanup: periodically purge to remove system snapshots.

enhancements/20230103-recurring-snapshot-cleanup.md
```diff
@@ -0,0 +1,126 @@
...
+`None`
+
+## Proposal
+
+- Introduce two new `RecurringJobType`:
+  - snapshot-delete
+  - snapshot-cleanup
+- Recurring job periodically deletes and purges the snapshots for RecurringJob using the `snapshot-delete` task type. Longhorn will retain snapshots based on the given retain number.
+- Recurring job periodically purges the snapshots for RecurringJob using the `snapshot-cleanup` task type. Longhorn will retain system snapshots based on the given retain number.
+
```
innobead, line 31: Not sure if `retain` is meaningful for `snapshot-delete`, because the target snapshots are removable and system snapshots. Can make it in general first.

enhancements/20230103-recurring-snapshot-cleanup.md
```diff
@@ -0,0 +1,126 @@
...
+### User Stories
+
+- The user can create a RecurringJob with `spec.task=snapshot-delete` to instruct Longhorn periodically delete and purge snapshots.
+- The user can create a RecurringJob with `spec.task=snapshot-cleanup` to instruct Longhorn periodically purge snapshots.
```
innobead, line 35: `The user can create a RecurringJob with `spec.task=snapshot-cleanup` to instruct Longhorn periodically purge removable or system snapshots.`

enhancements/20230103-recurring-snapshot-cleanup.md
```diff
@@ -0,0 +1,126 @@
...
+     concurrency: 1
+     cron: '* * * * *'
+     groups: []
+     labels: {}
+     name: recurring-snap-delete-per-min
+     retain: 2
+     task: snapshot-delete
+   ```
+1. Assign the RecurringJob to volume.
+1. Longhorn deletes all expired snapshots. As a result of the above example, the user will see two snapshots after the job completes.
```
innobead, line 58: Should the remaining snapshots be user-created, right? or the type does not matter but the creation time of snapshot is respected instead? (this question is more about the current behavior for retaining snapshots)

enhancements/20230103-recurring-snapshot-cleanup.md
```diff
@@ -0,0 +1,126 @@
...
+1. Continue with the current implementation to purge snapshots.
+
+#### The RecurringJob `snapshot-cleanup` Task Type
+
+1. Do snapshot purge only in `doSnapshotCleanup`.
+
+### RecurringJob Mutate
+
+1. Mutate `Recurringjob.Spec.Retain` to 0.
+
```
innobead, line 100: What's the section for? Do you mean some validation for `retain` in the mutation webhook?

innobead (changes requested):
Just a few comments. Other look good.

enhancements/20230103-recurring-snapshot-cleanup.md
innobead: I think you just leverage the existing code, so the snapshot type doesn't matter.

enhancements/20230103-recurring-snapshot-cleanup.md
c3y1huang: > Not sure if retain is meaningful for snapshot-delete

I think you mean `snapshot-cleanup` right? I forgot to remove this, retain should not be effective with case `snapshot-cleanup`.

c3y1huang: Removed, thanks!

enhancements/20230103-recurring-snapshot-cleanup.md
c3y1huang: Correct.

enhancements/20230103-recurring-snapshot-cleanup.md
c3y1huang: Sorry, this was written unclearly. I mean to propose mutating the `Recurringjob.Spec.Retain` to 0 when the task type is `snapshot-cleanup` since retain value has no effect on the purge.

innobead: :100:

enhancements/20230103-recurring-snapshot-cleanup.md
```diff
@@ -0,0 +1,126 @@
...
+https://github.com/longhorn/longhorn/issues/3836
+
+## Motivation
+
+### Goals
+
+Introduce new recurring job types:
+- `snapshot-delete`: periodically remove and purge all kinds of snapshots that exceed the retention count.
+- `snapshot-cleanup`: periodically purge to remove system snapshots.
```
innobead, line 18:
```suggestion
- `snapshot-cleanup`: periodically purge removable or system snapshots.
```

innobead (changes requested):

innobead (approved):
LGTM

Issue #5189, opened 2023-01-03
[FEATURE] Introduce faster compression and multiple threads for volume backup & restore 


derekbit:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

The backup mechanism in Longhorn is not efficient as investigated in https://github.com/longhorn/longhorn/issues/1409.

Besides the refactoring, the task will cover 

**Backup**
- [x] Global backup compression setting
- [x] Per-volume backup compression setting (`volume.spec.backupCompressionMethod`)
- [x] Introduce faster compression algorithms, lz4 or zstd, into longhorn, but still need to handle backward compatibility, i.e. for the existing backups using gzip.
  - none: suitable for encoding data, e.g. encoded image and video
  - lz4: suitable for text data
  - gzip: a bit of higher compression ratio but slow. (not recommended)
- [x] Concurrent compression and transfer workers
  - Introduce producer-consumer pattern
  - Improve the compression and upload throughput

**Restore**
- [x] Concurrent decompression and transfer workers

## Additional context

https://github.com/longhorn/longhorn/issues/1409
https://github.com/longhorn/longhorn/issues/3900
https://github.com/longhorn/longhorn/issues/3175
https://github.com/longhorn/longhorn/issues/4379

derekbit:
**Setup**
Platform: Equinix
Host: Japan-Tokyo/m3.small.x86
CPU: Intel(R) Xeon(R) E-2378G CPU @ 2.80GHz
RAM: 64 GiB
Disk: Micron_5300_MTFD
OS: Ubuntu 22.04.1 LTS(kernel 5.15.0-53-generic)
Kubernetes: v1.23.6+rke2r2
Longhorn: master-branch + backup improvement

Backupstore target: external MinIO S3 (m3.small.x86)


**Single-Thread Backup and Restore**
Volume: 50 GiB containing 10 GiB random data

![Single-Thread Backup and Restore](https://user-images.githubusercontent.com/12527233/210484515-3e7aff41-eab8-447e-b007-8c05e22c0364.svg)

derekbit:
In summary, the backup throughput is increased by 15X using `lz4` and `5` concurrent threads.

**Setup**
Platform: Equinix
Host: Japan-Tokyo/m3.small.x86
CPU: Intel(R) Xeon(R) E-2378G CPU @ 2.80GHz
RAM: 64 GiB
Disk: Micron_5300_MTFD
[...]

PR #5201, opened 2023-01-04
backup: update crds

derekbit:
Longhorn/longhorn#5189

### DIFF
--- a/chart/templates/crds.yaml
+++ b/chart/templates/crds.yaml
@@ -637,6 +637,9 @@ spec:
               backupCreatedAt:
                 description: The snapshot backup upload finished time.
                 type: string
+              compressionMethod:
+                description: Compression method
+                type: string
               error:
                 description: The error message when taking the snapshot backup.
                 type: string
@@ -3240,6 +3243,12 @@ spec:
                 type: string
               baseImage:
                 description: Deprecated. Rename to BackingImage
+                type: string
+              backupCompressionMethod:
+                enum:
+                - none
+                - lz4
+                - gzip
                 type: string
               dataLocality:
                 enum:
--- a/deploy/longhorn.yaml
+++ b/deploy/longhorn.yaml
@@ -724,6 +724,9 @@ spec:
               backupCreatedAt:
                 description: The snapshot backup upload finished time.
                 type: string
+              compressionMethod:
+                description: Compression method
+                type: string
               error:
                 description: The error message when taking the snapshot backup.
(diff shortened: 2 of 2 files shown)


innobead (approved):
In general, LGTM. Merge until the manager PR merge.

deploy/longhorn.yaml
```diff
@@ -3389,6 +3392,12 @@ spec:
               baseImage:
                 description: Deprecated. Rename to BackingImage
                 type: string
+              compressionMethod:
+                enum:
+                - none
+                - lz4
+                - gzip
```
innobead, line 3399: Keeping gzip is for backward compatible, right? I assume new backup we will use `lz4` by default.

chart/templates/crds.yaml
```diff
@@ -3241,6 +3244,12 @@ spec:
               baseImage:
                 description: Deprecated. Rename to BackingImage
                 type: string
+              compressionMethod:
```
innobead, line 3247: Should we emphasize the method is for backup instead of for data persistent or even replication over replicas?

If this is a concern, suggest renaming to `backupCompressionMethod` instead.

deploy/longhorn.yaml
derekbit: Correct.

chart/templates/crds.yaml
derekbit: Yeah, `backupCompressionMethod` sounds good.

derekbit:
BTW, should we provide per-volume backup concurrent threads?

innobead:
> BTW, should we provide per-volume backup concurrent threads?

Do you mean set the number of concurrent threads for each backup from a volume? If yes, it seems not quite meaningful or too granular, because I thought it should be a thread pool for all backup operations.

derekbit:
It is sometimes useful if users want to controller which volume can run faster and which one can run slower.
We can improve it another ticket if users have feedbacks.

>  If yes, it seems not quite meaningful or too granular, because I thought it should be a thread pool for all backup operations.
For current implementation, the concurrent limit means how many worker threads per backup.

innobead:
> It is sometimes useful if users want to controller which volume can run faster and which one can run slower. We can improve it another ticket if users have feedbacks.

Sounds fair and good to me :+1:

derekbit:
cc @innobead

innobead (approved):

PR #5310, opened 2023-01-19
Made environment_check.sh POSIX compliant

Haribo112:
Replaced all instances of backtick command substitution with $()-command substitution. 
This also fixes https://github.com/longhorn/longhorn/issues/5304

Issue #5304, opened 2023-01-16
[BUG] environment_check.sh does not handle differnt kernel versions in cluster correctly


achims311:
## Describe the bug (🐛 if you encounter this issue)

The environment_check.sh uses "uname -r" to get the kernel release and kernel config file.
But the uname -r command is always run locally, not on the node . So the checks against the kernel configuration do not work in case the node is running another kernel version than the local kernel version.

## To Reproduce

1. Set up a cluster on a machine not having the kernel installed which runs on the machine running the "environment_check.sh script (having the required nfs options switched on
e.g. cluster node on server n1 with only kernel 5.19 installed and running the script on o1 with kernel 6.1 running
3. run the environment check from the documentation:
curl -sSfL https://raw.githubusercontent.com/longhorn/longhorn/v1.4.0/scripts/environment_check.sh | bash
4. warn about missing nfs setup

## Expected behavior

no warning should be reported

## Log or Support bundle

n/A

## Environment

 - Longhorn version: 1.4.0
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl): n/A
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: k3s 
   - Number of management node in the cluster: 3
   - Number of worker node in the cluster: 3
 - Node config
   - OS type and version: different!
 - Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal): baremetal
 - Number of Longhorn volumes in the cluster: 0 until now

roger-ryao:
I can reproduce it on my local
![Screenshot_20230118_114413](https://user-images.githubusercontent.com/104337648/213079245-e9c3adc7-0c29-4aff-a3ea-e9af4a7dd089.png)

After executed the [PR-5305](https://github.com/longhorn/longhorn/pull/5305)
![Screenshot_20230118_114444](https://user-images.githubusercontent.com/104337648/213079419-787d9033-8108-4d56-ac9e-f5bf893b110a.png)


## Environment

 - Longhorn version: N/A
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl): N/A
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: k3s 
   - Number of management node in the cluster: 1
   - Number of worker node in the cluster: 3
 - Node config
   - OS type and version: Ubuntu 20.04 / keneral **5.4.0-1041-aws**
 - - Local config
[...]

PR #5314, opened 2023-01-24
Fix for bug #5304 (second version including POSIX way to call subroutine)

achims311:
After #5310 the old #5305 was closed due to merge conflicts.
These are solved now, and it can be merged again automatically

### DIFF
--- a/scripts/environment_check.sh
+++ b/scripts/environment_check.sh
@@ -329,19 +329,19 @@ check_nfs_client_kernel_support() {
     declare -A nodes=()
 
     for pod in ${pods}; do
-      KERNEL_RELEASE=`detect_node_kernel_release $pod`
-      if [ x"$KERNEL_RELEASE" == x"" ]; then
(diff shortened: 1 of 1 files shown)


innobead:
ref: #5304

innobead:
I think this is not related to the POSXI compatibility fix, but rather a minor improvement.

scripts/environment_check.sh
```diff
@@ -321,13 +329,19 @@ check_nfs_client_kernel_support() {
     declare -A nodes=()
 
     for pod in ${pods}; do
+      local kernel_release=$(detect_node_kernel_release $pod)
+      if [ x"$kernel_release" == x"" ]; then
```
innobead, line 333: Is any possible to have this case, no kernel release info?

innobead (approved):
LGTM.

This PR ensures the kernel version checking should actually happen on each node instead of the node running the script.

innobead:
@Mergifyio backport v1.4.x v1.3.x

PR #5344, opened 2023-02-03
Remove the share parameter since the accessMode should be determined by the PVC

PhanLe1010:


joshimoo (approved):

joshimoo:
Just leaving some explanation here, why the share flag can be removed.

```
If I remember right the share flag is to force RWX mode, which isn't needed it's better to have the PVC determine the access mode instead. This way non RWX migratable volumes are just regular RWO volumes.

Since RWX is a requirement to have access to a volume from different nodes at the same time.
Migratable just determines whether we use the NFS path for "real" rwx or the "hacky" RWO path for 2 different engines that cannot be used concurrently.

The absence of the field is the same as having it set to false.
But I agree it's better if it's not even shown
```

Issue #5339, opened 2023-02-02
[IMPROVEMENT] Update PSP validation in the Longhorn upstream chart 


PhanLe1010:
Update PSP Helm validation so that it prevents user from enabling PSP in Kuberentes >= 1.25. The validation file should be like this:

```
#{{- if gt (len (lookup "rbac.authorization.k8s.io/v1" "ClusterRole" "" "")) 0 -}}
#{{- if .Values.enablePSP }}
#{{- if not (.Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy") }}
#{{- fail "The target cluster does not have the PodSecurityPolicy API resource. Please disable PSPs in this chart before proceeding." -}}
#{{- end }}
#{{- end }}
#{{- end }}
```

Ref: https://github.com/rancher/rancher/issues/40342

innobead:
Also, need to update the readme as well. The change in the Rancher chart PR looks general, so suggest updating as well.

PhanLe1010:
**Test plan:**

1. Create 2 downstream clusters: One 1.24 and one 1.25 Kubernetes version
1. Install Longhorn helm chart with `enablePSP` set to `true`. Verify that:
    1.  On K8s 1.24 cluster, it is installed fine
    2.  On K8s 1.25 cluster, it failed to be installed with the PSP error message 
1. For the 1.25 cluster, 
    1. Try to remove the Helm secrete and reinstall  Longhorn helm chart with `enablePSP` set to `false`.
    1. Verify that it is installed ok
1. For the 1.24 cluster:
    1. Perform the in-place upgrade of the same chart but with `enablePSP` set to `false`
    1. Upgrade K8s to 1.25
    1. Verify that the Longhorn system is healthy

yangchiu:
Verified passed on master-head (longhorn 1deb512) by following [the test steps](https://github.com/longhorn/longhorn/issues/5339#issuecomment-1420311114).

PR #5363, opened 2023-02-07
Update PSP validation

PhanLe1010:
longhorn/longhorn#5339


chart/templates/validate-psp-install.yaml
```diff
@@ -0,0 +1,7 @@
+#{{- if gt (len (lookup "rbac.authorization.k8s.io/v1" "ClusterRole" "" "")) 0 -}}
```
innobead, line 1: Feel this can be verified by Kubernetes version check directly, but it also works by checking API resource for sure. Also, not sure why not just using the same `.Capabilities.APIversions.Has` provided by helm to verify ClusterRole exists or not, the same as below.

innobead (approved):
LGTM. Thanks Phan.

innobead:
@Mergifyio backport v1.4.x

Issue #5406, opened 2023-02-16
[TASK] Longhorn SPDK Engine


shuo-wu:
## What's the task? Please describe

Launch a new engine for SPDK

## Describe the items of the task (DoD, definition of done) you'd like
> Please use a task list for items on a separate line with a clickable checkbox https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists

- [ ] Start gRPC services that handle requests from peers or the instance manager proxy.
- [ ] Start spdk-tgt, then launch and connects bdevs as the engine/replica data layer.
- [ ] Handle frontend (nvme) for the engine.
- [ ] Use SPDK to implement basics snapshot features

innobead:
cc @DamiaSan

PR #5407, opened 2023-02-16
LEP for spdk-engine

shuo-wu:
#5406


shuo-wu:
The 1st version won't touch too many details for the SPDK target. I will update it after figuring out some issues and obstacles.

enhancements/20230214-spdk-engine.md
```diff
@@ -0,0 +1,115 @@
...
+    | API | Caller | Input | Output | Comments |
+    | --- | --- | --- | --- | --- |
+    | VolumeGet              | Instance manager proxy |                     | info Volume, err error | Get in-memory fields/flags of the engine process, probably does not need to talk with SPDK target |
+    | VolumeFrontendStart    | Instance manager proxy | frontend            | endpoint string, err error | Expose the volume(SPDK RAID1 bdev) as a NVMe-oF subsystem or a block device (NVMe-oF initiator); No need to talk with SPDK Target |
+    | VolumeFrontendShutDown | Instance manager proxy | snapshotName string | err error | Stop the NVMe-oF service for the engine; No need to talk with SPDK Target |
+  - | VolumeExpand           | Instance manager proxy | size int64 | err error | TODO; nvme ns-rescan + bdev_lvol_resize + bdev_lvol_grow_lvstore? |
+  - Replica:
+    | API | Caller | Input | Output | Comments |
+    | --- | --- | --- | --- | --- |
+    | ReplicaAdd    | Instance manager proxy | snapshotName string, labels map\[string\]string | err error | Rebuilding; Need to talk with SPDK Target and replica? |
```
shuo-wu, line 54: This is actually the replica rebuilding API. Need more input from @DamiaSan later

enhancements/20230214-spdk-engine.md
```diff
@@ -0,0 +1,115 @@
...
+### API Changes
+- The engine binary cmd: 
+  - `engine`: launch a process that provides as a gRPC service. But unlike the old engine, it is not responsible for the data service.
+  - `replica`: launch a process that provides as a gRPC service. But unlike the old replica, there are no data service and probably no sync agent server.
+- The engine gRPC service:
+  - Snapshot:
+    | API | Caller | Input | Output | Comments |
+    | --- | --- | --- | --- | --- |
+    | Create | Instance manager proxy | snapshotName string, labels map\[string\]string | err error | Need to talk with replica |
+    | Delete | Instance manager proxy | snapshotName string | err error | Need to talk with replica |
```
shuo-wu, line 42: To handle snapshot requests, we need to block IO for engine RAID1 bdev first then ask all replica lvols to create/delete snapshots during the IO suspension. 
@DamiaSan Do you know if there is a built-in SPDK-related API for that? Or can we somehow achieve it?

DamiaSan: No, actually there isn't API to do that. But in some way we will achieve it, I described an idea in the Snapshots section of #5115

enhancements/20230214-spdk-engine.md
```diff
@@ -0,0 +1,115 @@
...
+### Goals
+1. Talk with SPDK Target via RPC to launch engine/replica.
+2. Expose interfaces/cmds for engine/replica launching, snapshot/backup operations, and so on.
+
+## Proposal
+1. The general architecture is similar to the current engine: Provide engine/replica process startup cmds, have a gRPC server for engine or replica to handle requests, and help handle data flow. 
+2. Unlike the old engine, the data flow will be taken over by SPDK. The new engine or replica won't directly touch the data handling. The new engine or replica is an orchestrator between Longhorn system and SPDK.
+3. Since we rely on SPDK to handle data, Launching SPDK target on each node is necessary.
+4. The way we use SPDK to handle the data flow:
+   1. A data disk will be abstracted as SPDK aio_bdev.
```
DamiaSan, line 19: SPDK can also attach directly to NVMe physical device in the system bypassing kernel driver

enhancements/20230214-spdk-engine.md
```diff
@@ -0,0 +1,115 @@
...
+    | Create | Instance manager proxy | snapshotName string, labels map\[string\]string | err error | Need to talk with replica |
+    | Delete | Instance manager proxy | snapshotName string | err error | Need to talk with replica |
+    | Clone  | Instance manager proxy | snapshotName string | err error | Need to talk with replica |
+  - Volume:
+    | API | Caller | Input | Output | Comments |
+    | --- | --- | --- | --- | --- |
+    | VolumeGet              | Instance manager proxy |                     | info Volume, err error | Get in-memory fields/flags of the engine process, probably does not need to talk with SPDK target |
+    | VolumeFrontendStart    | Instance manager proxy | frontend            | endpoint string, err error | Expose the volume(SPDK RAID1 bdev) as a NVMe-oF subsystem or a block device (NVMe-oF initiator); No need to talk with SPDK Target |
+    | VolumeFrontendShutDown | Instance manager proxy | snapshotName string | err error | Stop the NVMe-oF service for the engine; No need to talk with SPDK Target |
+  - | VolumeExpand           | Instance manager proxy | size int64 | err error | TODO; nvme ns-rescan + bdev_lvol_resize + bdev_lvol_grow_lvstore? |
```
DamiaSan, line 50: bdev_lvol_grow_lvstore should be called only if the underlying bdev, and so if the device it is created on, size is increased. So, to exapnd a lvol, only bdev_lvol_resize should be called

enhancements/20230214-spdk-engine.md
DamiaSan: For every volume the new replica will host, to complete this operation we needs to talk with 3 entities:

- new replica to add, where we have to recreate the entire snapshot stack of the volume
- the replica that will act as source for the volume data
- the target where replicas are putted together in the RAID1 bdev

enhancements/20230214-spdk-engine.md
```diff
@@ -0,0 +1,115 @@
...
+  - Snapshot:
+    | API | Caller | Input | Output | Comments |
+    | --- | --- | --- | --- | --- |
+    | Create | Engine process         | snapshotName string, labels map\[string\]string | err error | Need to talk with SPDK Target |
+    | Delete | Engine process         | snapshotName string | err error | Need to talk with SPDK Target |
+    | Clone  | Instance manager proxy | snapshotName string | err error | Need to talk with SPDK Target: bdev_lvol_clone + bdev_lvol_decouple_parent |
+  - Replica:
+    | API | Caller | Input | Output | Comments |
+    | --- | --- | --- | --- | --- |
+    | ReplicaCreate | Engine process | replicaName string, size int64 | replica Replica, err error | Need to talk with SPDK Target; launch a SPDK lvstore + lvol? need to consider the clone |
```
DamiaSan, line 68: Yes, when we create a new replica we ave to create a new lvol store and all lvol hosted by the replica

shuo-wu: OK. Thank you for the clarification! `bdev_aio` and `lvstore` should be created for the first time a Longhorn engine/replica is launched on a new disk, right?

enhancements/20230214-spdk-engine.md
```diff
@@ -0,0 +1,121 @@
...
+  - Snapshot/Backup/Replica handling. Need to talk with the replica processes and even the spdk-tgt.
+
+##### replica:
+- A replica start cmd that:
+  - Launches the gRPC service. No data server and probably no sync-agent server.
+  - Start the spdk-tgt and init the env.
+  - Create bdev_aio and lvstore for the disk
+  - Launch lvol and expose the lvol as nvmf.
+- A gRPC service that handle requests, which are mainly related to lvol + nvmf
+
```
DamiaSan, line 106: I am a bit confused about engine-controller and engine-replica. In LEP #5115 we discussed to have:
- one spdk target per node
- one longhorn-spdk-engine per volume (in the diagram it is still named longhorn-spdk-agent)
If socat will be used, this longhorn-spdk-engine can talk directly via JSON-RPC with local and remote spdk targets to create the raid1 bdev. Moreover, given that we have one longhorn-spdk-engine per volume but only one spdk target per node, maybe spdk target have to be launched by some other process

shuo-wu: Currently, I haven't planned to use socat in the new engine. 
The expected flow is like this:
1. Engine process receives requests from instance managers. Let's say rebuilding for example.
2. It will talk with the target on the current node to suspend IO of the bdev RAID1 and something else (setup write only bdev for the new replica)
3. Then it sends a gRPC request to the source replica process to ask it to do shallow copy
4. The source replica receives the requests then talks with the corresponding target...

Here each engine/replica is responsible for handling its own SPDK components. The duty is clear. And replica processes are indispensable hence I don't think we need a cross-node JSON-PRC communication.

DamiaSan: If we keep the example of shallow copy, as described in the [wiki](https://github.com/longhorn/longhorn-spdk-engine/wiki/How-to-create-a-copy-of-a-volume-on-a-different-node-with-SPDK), to perform the operation the source replica need to talk with the target on his node but also with the target of the destination node for the copy.
The same to add a write only bdev to the raid1 for the new replica: engine will have to talk with multiple targets, because the creation and the export of this bdev is made over the new replica.

shuo-wu: For the shallow copy you provided, I think the flow with longhorn engine is:
1. The dest replica process creates a new lvol and exports it as a nvmf_subsystem.
2. The src replica process attaches the dest lvol as a bdev_nvme controller.
3. The src replica process  copies one snapshot lvol content into the block device (which is actually the dest replica bdev_nvme controller)
4. The src replica process asks the dest replica process to create a snapshot lvol.
5. The dest replica process talks with its target and create a snapshot lvol.
6. Repeat step3 to step5

Will this work?

DamiaSan: Yes, if src and dest replica talk together it works.

shuo-wu: As I mentioned, each engine/replica is responsible for its own SPDK components only. Then we can make the duty/flow clearer as well as avoid any misoperations or potentially malicious requests, WDYT?

DamiaSan: Yes, I agree with that. Maybe we will have to create some new gRPC commands but in this way we make a better separation of responsibilities. And probably the architecture is more similar to the actual one, with all its advantages, right?

shuo-wu: Right. Since the gRPC communication between engines and replicas is indispensable. Relying on gRPC won't introduce extra costs except for some gRPC APIs. It's similar to the current architecture and easy to understand (others can get the flow by just checking the gRPC calls). There is not plenty of details hiding at the SPDK level.

innobead:
Let's update this after the first integration is ready.

shuo-wu:
Updated this LEP based on the actual implmenetation

innobead (approved):
In general, LGTM.

This has been implemented in 1.5.

PR #5444, opened 2023-02-23
[DNM] Architectural LEP for sharded data layout

l-mb:
This is an initial draft for discussing the architecture of sharding Longhorn volumes across multiple devices/nodes (based on top of SPDK).

### DIFF
--- a/enhancements/20230221-spdk-sharding.md
+++ b/enhancements/20230221-spdk-sharding.md
@@ -1,0 +1,375 @@
+# Title
+
+Sharded volume layout
+
+## Summary
+
+Sharding will introduce Longhorn volumes that are larger than a single
+backend storage device. This proposal discusses how to achieve this
+based on the SPDK data plane, and how key functionality provided by
+Longhorn such as snapshots and volume expansion are provided.
+
+This is achieved by introducing an aggregation layer (using SPDK's
+`raid0` bdev) on top of the existing `raid1` replication, extending
+Longhorn's data model similar to traditional RAID10. The individual
+RAID1 entities are referred to as _shards_.
+
+In a few cases, multiple options for achieving this were considered, and
+only one was chosen. These are flagged as _Rejected Design Alternatives
+(RDA)_ but hopefully provide context.
+
+This is a fairly high-level design proposal, and while a few issues need
+to be considered at this stage and level, their detailed implementations
+would make this LEP too complex. Please see the _Future Work_ section.
+
+## Motivation
+
+### Goals
+
+- Aggregate storage from multiple backend devices (including across
+  nodes) to build a larger volume than previously possible
+- Enhance space efficiency, since smaller shards can be more flexibly
+  allocated.
+- Maintain Longhorn's key features: thin provisioning, snapshots, and
+  backup/restore, ability to grow and shrink volumes, Node- or
+  device-level evacuation, recovery, and migration
+- Seamless introduction with minimal changes to the user experience
+
+### Non-goals [optional]
+
+- Load balancing for performance or storage balancing. While the
+  mechanisms laid out here would allow for this functionality, this - if
+  desired - is left for a future proposal.
+- This is meant to be introduced on top of [SPDK Engine for
+  Longhorn](./20221213-reimplement-longhorn-engine-with-SPDK.md), not
+  for LH's legacy data layout.
+- Scale out concurrent access from multiple clients. Needs such as `RWX
+  via HA NFS` or S3 must go through the one frontend.
+
+## Proposal
+
+### User Stories
+
+#### Creating a volume larger than a single storage device
+
+Previously, the size of a Longhorn volume was limited to the maximum of
+free space available on `number of replicas` backend devices.
+
+After implementing this proposal, LH will create multiple individual
+shards (sizing to be discussed below) that are each replicated on
+storage devices according to the availability/redundancy requirements,
+and then assemble a coherent single volume from this for the user to
+consume.
+
+The user can thus request the _entirety_ of the free space available in a
+Longhorn cluster.
+
+From the point of view of the user, this will happen transparently when
+a volume of a size above `shard_size` is requested.
+
+The user experience of all other management later should remain the
+same.
+
+### User Experience In Detail
+
+Beyond the former, users may have to consider the implication of setting
+`shard_size`. To allow free space of devices to be used effectively,
+this should likely be between 1-10% of the typical device's capacity.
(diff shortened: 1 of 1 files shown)


enhancements/20230221-spdk-sharding.md
```diff
@@ -0,0 +1,375 @@
...
+  end
+
+  shard1 -.- Node1
+  shard2 -.- Node1
+  shard1 -.- Node2
+  shard2 -.- Node2
+  shard1 -.- Node3
+  shard2 -.- Node3
+```
+
```
DamiaSan, line 145: Can you describe better the difference between shard and stripe? Because RAID0 knows only strip_size and so how can it aggregate different shards?

l-mb: I think that's mainly a question of terminology and what architectural level you're looking at.

Since we're referring to the feature in general as "sharding", the chunk of data sourced from one of the backend storage devices is a "shard". We then need to aggregate all of the shards together to form the full LH volume.

Since this proposal uses RAID0 as the aggregation layer, a shard is the same as a RAID0 stripe (or member device, from the RAI*D* perspective). If we used a different algorithm for combining them (say, a consistent hashing algorithm to pseudo-randomly distribute the load like Ceph does), the aggregation layer might use a different term.

Does this help?

DamiaSan: Yes thanks

enhancements/20230221-spdk-sharding.md
```diff
@@ -0,0 +1,375 @@
...
+#### Allocation
+
+- At allocation time, Longhorn instantiates the top-level `raid0` as the
+  aggregation layer. (See `Notes` section on a brief discussion of raid0
+  vs concatenation, and the `Growing a volume` section on a discussion
+  of the `stripe_size` chosen)
+- Enough shards are created to hold the intended volume size.
+  - If the `volume_size % shard_size != 0`, the last shard might only be
+    partially used.
```
derekbit, line 154: Are there any concerns about implementation and maintenance if `volume_size % shard_size != 0`?

enhancements/20230221-spdk-sharding.md
```diff
@@ -0,0 +1,375 @@
...
+this can be adjusted later to understand how careful users need to be
+with getting this right.
+
+##### Increasing
+Growing may be feasible within certain limits:
+- if the backing storage devices holding each shard still have
+  sufficient capacity.
+- If not, each shard may need to individually be migrated to a device
+  that can hold the new size.
+- This somewhat negatively impacts the allocation granularity.
```
derekbit, line 192: Can you elaborate more on the allocation granularity and the negative impact?

enhancements/20230221-spdk-sharding.md
```diff
@@ -0,0 +1,375 @@
...
+
+Since introducing a new tunable, it is important to consider whether
+this can be adjusted later to understand how careful users need to be
+with getting this right.
+
+##### Increasing
+Growing may be feasible within certain limits:
+- if the backing storage devices holding each shard still have
+  sufficient capacity.
+- If not, each shard may need to individually be migrated to a device
```
derekbit, line 190: Can it support online share migration?

enhancements/20230221-spdk-sharding.md
```diff
@@ -0,0 +1,375 @@
...
+  shard2 -.- Node2
+  shard1 -.- Node3
+  shard2 -.- Node3
+```
+
+#### Allocation
+
+- At allocation time, Longhorn instantiates the top-level `raid0` as the
+  aggregation layer. (See `Notes` section on a brief discussion of raid0
+  vs concatenation, and the `Growing a volume` section on a discussion
```
derekbit, line 150: A little confuse about the `shard_size` and `stripe_size`.
Would you be able to provide the definition of `stripe_size` and the relationship between `shard_size`?

enhancements/20230221-spdk-sharding.md
```diff
@@ -0,0 +1,375 @@
...
+```mermaid
+graph TB
+  Volume[(Longhorn Volume)]
+  Volume --> bdev[raid1]
+  bdev --> Device1{{Node1 - Device 1}}
+  bdev --> Device2{{Node2 - Device 1}}
+  bdev --> Device3{{Node3 - Device 1}}
+```
+
+Once this proposal is implemented, a more complex layout becomes
```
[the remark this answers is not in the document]
l-mb: No, RAID0 can use multiple bdevs. This is just for an example in the document so the text doesn't get too long :)

enhancements/20230221-spdk-sharding.md
```diff
@@ -0,0 +1,375 @@
...
+  can handle the need with specific configuration settings, this was
+  not considered required.
+  
+  Adding this may be considered "cleaner" versus allowing the additional
+  and removal of devices at runtime to `raid0`. We will need to see this
+  based on the SPDK discussion.
+
+## Future Work
+
+- **SPDK**: API call to add or remove a member device for `raid0`
```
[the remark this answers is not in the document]
l-mb: This is not rebalancing. This is about shrinking the volume once the space has been compacted.

enhancements/20230221-spdk-sharding.md
l-mb: I don't see any.

enhancements/20230221-spdk-sharding.md
l-mb: Compare a shard size that's 5% vs one that is 10% of our backend devices.

We can allocate a 5% shard once all our devices are 90% full, and distribute them across multiple devices as needed.

If they're at 10%, we can't.

So smaller shard sizes allow more effective use of space, at the cost of slightly more overhead. So the question is always the trade-off.

enhancements/20230221-spdk-sharding.md
l-mb: Sure, a RAID1 bdev can "migrate" a shard by allocating the new location and then removing the old one, as discussed in other places in the document

enhancements/20230221-spdk-sharding.md
l-mb: stripe_size is what RAID0 uses when distributing the data across its member devices.

shard_size is the allocation size of the member devices and the raid1s.

They can technically be different, as discussed in the LEP, which has some trade-offs.

derekbit:
@c3y1huang and I will research and design the architecture for sharded storage.

cc @innobead

Issue #5208, opened 2023-01-05
[FEATURE] Consolidate Instance Manager Engine & Replica for resource consumption reduction


innobead:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

Currently, we will have an instance manager engine & replica pods on each node, and each of them will have a default 12% Guaranteed CPU request which is configurable. After upgrading Longhorn, there will be another two new instance manager pods running, so there will be four instance managers pods before all volumes are migrated to the new version. This causes high resource requirements for the fresh install and upgrade as well. 

Engine/Replica processes running in the instance manager are process-based, so they will not impact the instance manager container. To simplify the architecture to decrease resource usage, the goal here is to consolidate the instance manager engine & replica to one pod, but continue serving all data plane operations on each node without any change. For volume migration, the same flow is as usual. 

## Describe the solution you'd like

- [x] Introduce the new architecture of the instance manager
- [x] Revamp the current automatic engine upgrade implementation
- [ ] Benchmark the resource consumption based on the number/size of volume. Have a theoretical explanation for the resource usage by instance manager. 

## Additional context

https://github.com/rancher/image-mirror/pull/329#issuecomment-1376715530

cc @longhorn/dev

c3y1huang:
Note: https://github.com/longhorn/longhorn/issues/1691#issuecomment-729633995

c3y1huang:
Initial test results:
- core tests: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3112/
- WIP

c3y1huang:
@longhorn/qa , please also do some tests using the UI, thanks!

yangchiu:
Tested on master-head, although there are volumes created and attached, the `Ref Count Volume` in `Instance Manager Image` page is always `0`, and the `Volumes` modal of each instance manager is always empty:

![Screenshot from 2023-05-04 13-52-25](https://user-images.githubusercontent.com/5709208/236133046-f96e2456-5c42-4ba5-a746-b3fd042d071e.png)

![Screenshot from 2023-05-04 13-52-41](https://user-images.githubusercontent.com/5709208/236133066-e3096366-c589-4417-9398-7af364c21bd6.png)

Also the `Instance Manager Image` in `Components` modal of each node is `N/A`:

![Screenshot from 2023-05-04 13-51-53](https://user-images.githubusercontent.com/5709208/236133350-e91282a0-9e00-4517-a689-fd7f2048264d.png)

[...]

Issue #5131, opened 2022-12-22
[FEATURE] Upgrade path enforcement & downgrade prevention


innobead:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

Currently, Longhorn does not enforce the upgrade path, even though we claim Longhorn only supports upgrading from the previous stable release, for example, upgrading to 1.4.x is only supported from 1.3.x or 1.4.0. 

Without upgrade enforcement, it allows users to upgrade from any previous version to a later version and it will cause deprecated features to be unable to remove from the later version.

## Describe the solution you'd like

- Enforce the upgrade path to disallow users to upgrade from an unsupported source version. After rejecting the user upgrade, the user longhorn setup should be intact w/o any impacts.
- Also not allow users to do the downgrade. 

## Additional context

#2865 

cc @longhorn/dev

innobead:
@mantissahz Do we have LEP? We should LEP first to see the whole design.

innobead:
Moved back to `implement` due to missing doc PR.

ejweber:
I am working on https://github.com/longhorn/longhorn/issues/4997 to remove a deprecated setting that is still referenced in an [upgrade.go file](url). I can modify that file to remove the reference to the setting or make the setting name local to the file, but these approaches are kind of hacky. Now that we are enforcing an upgrade path, it makes more sense to me to entirely remove illegal upgrade paths from the upgrade folder/package. For example, in v1.5.x, remove `upgrade/v070to080` ... `upgrade/v120to121` and remove the corresponding `UpgradeResources` calls in `upgrade/upgrade.go`. This would make current and future dead code removal easier. WDYT?

https://github.com/longhorn/longhorn-manager/tree/master/upgrade

innobead:
Exactly, we should remove unsupported paths from 1.5.0. @mantissahz Please help with that.

mantissahz:
@innobead, @ejweber 
https://github.com/longhorn/longhorn-manager/pull/1902
Removed the upgrade patch codes older than upgrading from v1.4.x to v1.5.0

derekbit:
> @innobead, @ejweber [longhorn/longhorn-manager#1902](https://github.com/longhorn/longhorn-manager/pull/1902) Removed the upgrade patch codes older than upgrading from v1.4.x to v1.5.0

There are some old codes 
```
if isCLIAPIVersionOne {
    ....
}
```
Is it possible to remove these codes after introducing the upgrade enforcement feature?

[...]

PR #5457, opened 2023-03-01
feat(lep): consolidate instance managers

c3y1huang:
https://github.com/longhorn/longhorn/issues/5208

### DIFF
--- a/enhancements/20230303-consolidate-instance-managers.md
+++ b/enhancements/20230303-consolidate-instance-managers.md
@@ -1,0 +1,598 @@
+# Consolidate Longhorn Instance Managers
+
+## Summary
+
+Longhorn architecture includes engine and replica instance manager pods on each node. After the upgrade, Longhorn adds an additional engine and replica instance manager pods. When the cluster is set with a default request of 12% guaranteed CPU, all instance manager pods will occupy 12% * 4 CPUs per node. Nevertheless, this caused high base resource requirements and is likely unnecessary.
+
+```
+NAME                STATE      E-CPU(CORES)   E-MEM(BYTES)   R-CPU(CORES)   R-MEM(BYTES)   CREATED-WORKLOADS   DURATION(MINUTES)   AGE
+demo-0 (no-IO)      Complete   8.88m          24Mi           1.55m          43Mi           5                   10                  22h
+demo-0-bs-512b-5g   Complete   109.70m        66Mi           36.46m         54Mi           5                   10                  16h
+demo-0-bs-1m-10g    Complete   113.16m        65Mi           36.63m         56Mi           5                   10                  14h
+demo-0-bs-5m-10g    Complete   114.17m        64Mi           31.37m         54Mi           5                   10                  42m
+```
+
+Aiming to simplify the architecture and free up some resource requests, this document proposes to consolidate the engine and replica instance managers into a single pod. This consolidation will not affect any data plane operations or volume migration. As the engine process is the primary consumer of CPU resources, merging the instance managers will result in a 50% reduction in CPU requests for instance managers. This is because there will only be one instance manager pod for both process types.
+
+### Related Issues
+
+Phase 1:
+- https://github.com/longhorn/longhorn/issues/5208
+
+Phase 2:
+- https://github.com/longhorn/longhorn/issues/5842
+- https://github.com/longhorn/longhorn/issues/5844
+
+## Motivation
+
+### Goals
+
+- Having single instance manager pods to run replica and engine processes.
+- After the Longhorn upgrade, the previous engine instance manager should continue to handle data plane operations for attached volumes until they are detached. And the replica instance managers should continue servicing data plane operations until the volume engine is upgraded or volume is detached.
+- Automatically clean up any engine/replica instance managers when all instances (process) get removed.
+- Online/offline upgrade volume engine should be functional. The replicas will automatically migrate to use the new `aio` (all-in-one) type instance managers, and the `engine` type instance manager will continue to serve until the first volume detachment.
+- The Pod Disruption Budget (PDB) handling for cluster auto-scaler and node drain should work as expected.
+
+### Non-goals [optional]
+
+`None`
+
+## Proposal
+
+To ensure uninterrupted upgrades, this enhancement will be implemented in two phases. The existing `engine`/`replica` instance manager may coexist with the consolidated instance manager during the transition.
+
+Phase 1:
+- Introduce a new `aio` instance manager type. The `engine` and `replica` instance manager types will be deprecated and continue to serve for the upgraded volumes until the first volume detachment.
+- Introduce new `Guaranteed Instance Manager CPU` setting, `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU` settings will be deprecated and continues to serve for the upgraded volumes until the first volume detachment.
+
+Phase 2:
+- Remove all instance manager types.
+- Remove the `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU` settings.
+
+### User Stories
+
+- For freshly installed Longhorn, the user will see `aio` type instance managers.
+- For upgraded Longhorn with all volume detached, the user will see the `engine`, and `replica` instance managers removed and replaced by `aio` type instance managers.
+- For upgraded Longhorn with volume attached, the user will see existing `engine`, and `replica` instance managers still servicing the old attached volumes and the new `aio` type instance manager servicing new volume attachments.
+
+### User Experience In Detail
+
+#### New Installation
+
+1. User creates and attaches a volume.
+   ```
+   > kubectl -n longhorn-system get volume
+   NAME     STATE      ROBUSTNESS   SCHEDULED   SIZE          NODE            AGE
+   demo-0   attached   unknown                  21474836480   ip-10-0-1-113   12s
+
+   > kubectl -n longhorn-system get lhim
+   NAME                                                STATE     TYPE   NODE            AGE
+   instance-manager-8f81ca7c3bf95bbbf656be6ac2d1b7c4   running   aio    ip-10-0-1-105   124m
+   instance-manager-7e59c9f2ef7649630344050a8d5be68e   running   aio    ip-10-0-1-102   124m
+   instance-manager-b34d5db1fe1e2d52bcfb308be3166cfc   running   aio    ip-10-0-1-113   124m
+
+   > kubectl -n longhorn-system get lhim/instance-manager-b34d5db1fe1e2d52bcfb308be3166cfc -o yaml
+   apiVersion: longhorn.io/v1beta2
+   kind: InstanceManager
+   metadata:
+     creationTimestamp: "2023-03-16T10:48:59Z"
+     generation: 1
+     labels:
+       longhorn.io/component: instance-manager
+       longhorn.io/instance-manager-image: imi-8d41c3a4
+       longhorn.io/instance-manager-type: aio
+       longhorn.io/managed-by: longhorn-manager
+       longhorn.io/node: ip-10-0-1-113
+     name: instance-manager-b34d5db1fe1e2d52bcfb308be3166cfc
+     namespace: longhorn-system
+     ownerReferences:
+     - apiVersion: longhorn.io/v1beta2
+       blockOwnerDeletion: true
+       kind: Node
+       name: ip-10-0-1-113
+       uid: 00c0734b-f061-4b28-8071-62596274cb18
+     resourceVersion: "926067"
+     uid: a869def6-1077-4363-8b64-6863097c1e26
+   spec:
+     engineImage: ""
+     image: c3y1huang/research:175-lh-im
+     nodeID: ip-10-0-1-113
+     type: aio
+   status:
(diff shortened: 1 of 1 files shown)


enhancements/20230303-consolidate-instance-managers.md
```diff
@@ -0,0 +1,641 @@
...
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/5208
+
+## Motivation
+
+### Goals
+
+- Having single instance manager pods to run replica and engine processes.
+- For the attached volumes during/after the Longhorn upgrade, the old engine and replica instance managers should continue servicing data plane operations until the first volume detachment.
```
derekbit, line 18: Should it be `last volume detachment`?

enhancements/20230303-consolidate-instance-managers.md
```diff
@@ -0,0 +1,641 @@
...
+- Automatically clean up any engine/replica instance managers when all instances (process) get removed.
+- Online/offline upgrade volume engine should be functional. The replicas will automatically migrate to use the new `aio` type instance managers, and the `engine` type instance manager will continue to serve until the first volume detachment.
```
derekbit, line 20: Just curious. What does `aio` mean? Is it an abbreviation for ...?

enhancements/20230303-consolidate-instance-managers.md
```diff
@@ -0,0 +1,641 @@
+# Consolidate Longhorn Instance Managers
+
+## Summary
+
+Longhorn architecture includes engine and replica instance manager pods on each node. After the upgrade, Longhorn adds an additional engine and replica instance manager pods. When the cluster is set with a default request of 12% guaranteed CPU, all instance manager pods will occupy 12% * 4 CPUs per node. Nevertheless, this caused high base resource requirements and is likely unnecessary.
```
derekbit, line 5: After the consolidation, how many cpu will be requested for the new instance manager?

enhancements/20230303-consolidate-instance-managers.md
c3y1huang: I've reworded this a bit to clarify the statement. PTAL.

enhancements/20230303-consolidate-instance-managers.md
c3y1huang: This is an abbreviation for all-in-one.

enhancements/20230303-consolidate-instance-managers.md
c3y1huang: The total CPU consumption for `im-e+im-r` is similar to `im-aio`, and the engine process is the main consumer of CPU resources. After consolidation, the CPU request percentage of instance managers will be reduced by half, since there will be a single instance manager pod for both process types.

enhancements/20230303-consolidate-instance-managers.md
derekbit: Thank you.

enhancements/20230303-consolidate-instance-managers.md
derekbit: Sounds great. Can we add a brief paragraph for the improvement?

innobead: Where is the info about the most resource consumption from engine processes? If having, we should add that to this LEP. 

I am also thinking if we can do more granular testing to understand if the resource consumption is linear growth, so it will be helpful for us to do the right estimation.

innobead (changes requested):

enhancements/20230303-consolidate-instance-managers.md
```diff
@@ -0,0 +1,641 @@
...
+   spec:
+     engineImage: ""
+     image: longhornio/longhorn-instance-manager:v1.5.0
+     nodeID: ip-10-0-1-113
+     type: aio
+   status:
+     apiMinVersion: 1
+     apiVersion: 3
+     currentState: running
+     instances:
```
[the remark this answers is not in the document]
c3y1huang: Not sure if I understand the question, can you elaborate a bit more?
Currently, the instances are already separated by `-e-` and `-r-` as shown in the CR.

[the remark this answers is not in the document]
innobead:
```yaml
       demo-0-e-fe37f92c:
         spec:
           name: demo-0-e-fe37f92c
         status:
           endpoint: ""
           errorMsg: ""
           listen: ""
           portEnd: 10015
           portStart: 10015
           resourceVersion: 0
           state: running
           type: ""
       demo-0-r-1605cbd6:
         spec:
           name: demo-0-r-1605cbd6
         status:
           endpoint: ""
           errorMsg: ""
           listen: ""
           portEnd: 10014
           portStart: 10000
           resourceVersion: 0
           state: running
           type: ""
```

enhancements/20230303-consolidate-instance-managers.md
c3y1huang: Updated improvement outline and some observed results (with and without IO).

As for the estimation, I think we can do more samplings after the implementation.

enhancements/20230303-consolidate-instance-managers.md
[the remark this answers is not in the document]
c3y1huang: Separated the `instance` field into `instanceEngines` and `instanceReplicas` fields.

innobead:
@shuo-wu @derekbit let's review this. thanks.

derekbit:
The LEP is straightforward. In general, LGTM.

enhancements/20230303-consolidate-instance-managers.md
```diff
@@ -0,0 +1,695 @@
...
+    podIP: 10.42.0.27
+    podIPs:
+    - ip: 10.42.0.27
+    qosClass: Burstable
+    startTime: "2023-03-01T08:13:03Z"
+  ```
+
+#### Controllers Change
+
+- Map the status of the engine/replica process to the corresponding instanceEngines/instanceReplicas fields in the InstanceManager instead of the instances field.
```
shuo-wu, line 659: Will Longhorn do this for the old engine instance managers?

enhancements/20230303-consolidate-instance-managers.md
```diff
@@ -0,0 +1,695 @@
...
+- Replace the old engine/replica InstanceManagers with the aio type instance manager during replenishment.
+
+#### New Setting
+
+- Introduce a new `Guaranteed Instance Manager CPU` setting for the new `aio` instance manager pod.
+- The `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU` will co-exist with this setting in Longhorn v1.5.x.
+
+### Phase 2 - Deprecations Overview
+
+Based on the assumption when upgrading from v1.5.x to 1.6.x, volumes should have detached at least once and migrated to `aio` type instance managers. Then the cluster should not have volume depending on `engine` and `replica` type instance managers. Therefore in this phase, remove the related types and settings.
```
shuo-wu, line 670: There is no limitation for the engine version of running volumes during the upgrade. Do we need to add this kind of check before upgrade?

c3y1huang: @mantissahz , should we cover this in https://github.com/longhorn/longhorn/issues/5131?

enhancements/20230303-consolidate-instance-managers.md
c3y1huang: The upgrade strategy section covers field transfer, which means this is applicable to older instance managers as well.

enhancements/20230303-consolidate-instance-managers.md
```diff
@@ -0,0 +1,695 @@
...
+
+Introducing a new instance manager type to have Longhorn continue to service existing attached volumes for Longhorn v1.5.x.
+
+#### New Instance Manager Type
+
+- Introduce a new `aio` (all-in-one) instance manager type to differentiate the handling of the old `engine`/`replica` instance managers and the new consolidated instance managers.
+- When getting InstanceManagers by instance of the attached volume, retrieve the InstanceManager from the instance manager list using the new `aio` type.
+
+#### InstanceManager `instances` Field Replacement
+- Separate the `instance` field into `instanceEngines` and `instanceReplicas` fields.
```
innobead, line 484: This is good. Don't change the existing property to keep backward compatibility.

enhancements/20230303-consolidate-instance-managers.md
innobead: These new properties are only for new replicas, volume live upgrade or volume reattachment.

For existing volumes, all info will be still saved in the old instances properly, right?

enhancements/20230303-consolidate-instance-managers.md
innobead: Upgrade enforcement is for longhorn version instead of engine image version for now.

We should add a way to tackle this when working on 1.6. However, it will be good to describe here.

enhancements/20230303-consolidate-instance-managers.md
```diff
@@ -0,0 +1,695 @@
...
+- Introduce a new `Guaranteed Instance Manager CPU` setting for the new `aio` instance manager pod.
+- The `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU` will co-exist with this setting in Longhorn v1.5.x.
+
+### Phase 2 - Deprecations Overview
+
+Based on the assumption when upgrading from v1.5.x to 1.6.x, volumes should have detached at least once and migrated to `aio` type instance managers. Then the cluster should not have volume depending on `engine` and `replica` type instance managers. Therefore in this phase, remove the related types and settings.
+
+#### Old Instance Manager Types
+
+- Remove the `engine`, `replica`, and `aio` instance manager types. There is no need for differentiation.
```
innobead, line 674: Oh you will do it in phase 2.

innobead (changes requested):
In general LGTM. Few questions need to answer.

enhancements/20230303-consolidate-instance-managers.md
shuo-wu: You mean we plan to add checks for the engine version of running volumes or the old instance manager existence in v1.6, right?

enhancements/20230303-consolidate-instance-managers.md
```diff
@@ -0,0 +1,695 @@
...
+https://github.com/longhorn/longhorn/issues/5208
+
+## Motivation
+
+### Goals
+
+- Having single instance manager pods to run replica and engine processes.
+- After the Longhorn upgrade, the previous engine instance manager should continue to handle data plane operations for attached volumes until they are detached. And the replica instance managers should continue servicing data plane operations until the volume engine is upgraded or volume is detached.
+- Automatically clean up any engine/replica instance managers when all instances (process) get removed.
+- Online/offline upgrade volume engine should be functional. The replicas will automatically migrate to use the new `aio` (all-in-one) type instance managers, and the `engine` type instance manager will continue to serve until the first volume detachment.
```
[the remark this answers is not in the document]
c3y1huang: > Don't need to create another type aio, but rather the type is unnecessary except for backward compatibility for the current type (engine/replica). 

We are creating a new type here so resources can [match using the instance manager labels](https://github.com/longhorn/longhorn-manager/blob/v1.4.1/controller/instance_manager_controller.go#L829). The [type is included as part of the labels](https://github.com/longhorn/longhorn-manager/blob/v1.4.1/types/types.go#L400).

> Type property should be deprecated and removed eventually.

Yes, in phase 2.

enhancements/20230303-consolidate-instance-managers.md
c3y1huang: During the upgrade process, the instance field for all existing instance managers will be moved to either `instanceEngines` or `instanceReplicas`. I've taken this approach so there is no need to handle the instance field separately.

innobead: Do you mean Longhorn upgrade no matter if volumes are still attached? I thought those fields are only after reattachment or new volume.

enhancements/20230303-consolidate-instance-managers.md
c3y1huang: https://github.com/longhorn/longhorn/issues/5842

enhancements/20230303-consolidate-instance-managers.md
```diff
@@ -0,0 +1,695 @@
...
+
+## Proposal
+
+To ensure uninterrupted upgrades, this enhancement will be implemented in two phases. The existing `engine`/`replica` instance manager may coexist with the consolidated instance manager during the transition.
+
+Phase 1:
+- Introduce a new `aio` instance manager type. The `engine` and `replica` instance manager types will be deprecated and continue to serve for the upgraded volumes until the first volume detachment.
+- Introduce new `Guaranteed Instance Manager CPU` setting, `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU` settings will be deprecated and continues to serve for the upgraded volumes until the first volume detachment.
+
+Phase 2:
```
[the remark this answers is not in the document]
innobead: Need to create an issue for this. 

ref: https://github.com/longhorn/longhorn/issues/5844

innobead (approved):
LGTM

PR #5493, opened 2023-03-06
feat: consolidate instance managers

c3y1huang:
https://github.com/longhorn/longhorn/issues/5208

### DIFF
--- a/deploy/longhorn.yaml
+++ b/deploy/longhorn.yaml
@@ -1787,6 +1787,7 @@ spec:
                 type: string
               type:
                 enum:
+                - aio
                 - engine
                 - replica
                 type: string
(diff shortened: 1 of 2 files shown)


innobead (approved):

deploy/longhorn.yaml
```diff
@@ -1835,7 +1900,7 @@ spec:
                           type: string
                       type: object
                   type: object
-                nullable: true
```
innobead, line 1838: Removing `nullable: true` will have any side effects?

innobead (changes requested):

deploy/longhorn.yaml
c3y1huang: Unintentional , let me add it back.

It shouldn't cause a side-effect because the field is optional, removing the nullable means if the list is empty it will still show the field like `instance:`.

c3y1huang: Fixed.

innobead (approved):

PR #5499, opened 2023-03-07
chore(version-bump): longhorn-instance-manager

mantissahz:
Update YAML and chart files for v1.3.x

Update `longhorn-instance-manager` version from `v1_20221003` to ~v1_20230223~ `v1_20230310`

longhorn/longhorn#5265


innobead (changes requested):
Use https://github.com/longhorn/longhorn-instance-manager/releases/tag/v1_20230310 instead, because there are some commits after v1_20221003.

mantissahz:
> Use https://github.com/longhorn/longhorn-instance-manager/releases/tag/v1_20230310 instead, because there are some commits after v1_20221003.

Updated to `v1_20221003`

innobead (approved):

Issue #3304, opened 2021-11-23
[IMPROVEMENT] Use PDB to protect Longhorn components from unexpected drains


joshimoo:
When using `kubectl drain` on a single node cluster there might be a case where it's impossible to correctly cleanup/detach a volume. Since the csi sidecars might end up evicted before the pod/va is cleaned up correctly. 

It would be useful to have PDB on the **csi-sidecars**, to prevent all sidecars from being unavailable.
We could consider removing the PDB once we have evaluated that all longhorn volumes are no longer in use.
But even as a first step just adding a PDB with a minAvailablity of 1 for the csi sidecars would be beneficial.

A similiar thing would be useful for the **share-manager**, since it's possible that the share-manager pod gets evicted before the nfs unmount happens in the csi-plugin which might stall the kubelet/csi-plugin. in single node clusters the nfs mount would get stuck since the share-manager would no longer come up, since the node is cordoned during the drain process.

--------------------------------

A separate idea is to look into providing a script to ignore the longhorn components/namespace as part of the drain.
Since if you drain all workloads that use the longhorn volumes on that node, there wouldn't be any active volumes left on that node.
This does not apply to replica and share-manager since they could be used by a different node in a multi node cluster.

Simliar to what kubevirt does in it's maintenace guide provide a selector that drains the appropriate components.
REF: https://kubevirt.io/user-guide/operations/node_maintenance/#evict-all-vms-from-a-node
REF: https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#drain

joshimoo:
potentially issues that are caused by this mismatch
https://github.com/longhorn/longhorn/issues/2933

joshimoo:
I tested the label selector idea and it works well, as a current workaround, example below:
```
# not equal x (includes absent key)
kubectl drain --pod-selector='!longhorn.io/component' jmoody-lh-work3 --ignore-daemonsets

# not equal x and not equal y selector
kubectl drain --pod-selector='!longhorn.io/component,app!=csi-attacher,app!=csi-provisioner' jmoody-lh-work1 --ignore-daemonsets
```

PhanLe1010:
Note from @innobead 
> So, I think this should be good enough for now, but probably we should also update our doc for upgrade part about the drain usage.
https://longhorn.io/docs/1.2.2/volumes-and-nodes/maintenance/#updating-the-node-os-or-container-runtime

innobead:
[...]

Issue #5296, opened 2023-01-16
[BUG] Memory leak in CSI plugin caused by stuck umount processes if the RWX volume is already gone


lazar-boradzhiev:
## Describe the bug (🐛 if you encounter this issue)

The memory usage of the CSI plugins has been growing steadily for the past few weeks after upgrading to version 1.4.0
![image](https://user-images.githubusercontent.com/73098389/212631603-93f7b860-90fd-4115-82ca-04c9d9f0ad2a.png)

## To Reproduce

No specific steps. It has been like that since upgrading to version Longhorn 1.4.0.

## Expected behavior

The CSI plugin memory usage should stay relatively small without growing steadily.

## Log or Support bundle

[supportbundle_413f9f8b-328c-47d2-a25d-4f7c903bf06f_2023-01-16T08-27-12Z.zip](https://github.com/longhorn/longhorn/files/10423824/supportbundle_413f9f8b-328c-47d2-a25d-4f7c903bf06f_2023-01-16T08-27-12Z.zip)

## Environment
Longhorn version: 1.4.0
Installation method (e.g. Rancher Catalog App/Helm/Kubectl): Helm
Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: v1.25.4+k3s1
    Number of management node in the cluster: 1
    Number of worker node in the cluster: 4
Node config
    OS type and version: Raspbian (5.15.76-v8) and Ubuntu server (22.04.1)
    CPU per node: 2-4
    Memory per node: 4-8G
    Disk type(e.g. SSD/NVMe): SSD
    Network bandwidth between the nodes: 1G
Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal): Baremetal
Number of Longhorn volumes in the cluster: 15

## Workaround
Force umount the stuck umount process by `umoun -f`.

## Additional context

Initially discussed as part of another issue https://github.com/longhorn/longhorn/issues/5195#issuecomment-1383662401

innobead:
@lazar-boradzhiev would probably need your cooperation when investigating this issue. Thanks first.

lazar-boradzhiev:
> @lazar-boradzhiev would probably need your cooperation when investigating this issue. Thanks first.

Sure, I will stay available.

derekbit:
There are 3 containers in the pod. `longhorn-csi-plugin` container eats lots of memory.
```
  containers:
  - name: longhorn-liveness-probe
[...]

PR #5516, opened 2023-03-09
Add LEP: Use PDB to protect Longhorn components from drains

PhanLe1010:
longhorn/longhorn#3304

### DIFF
--- a/enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
+++ b/enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
@@ -1,0 +1,95 @@
+# Use PDB to protect Longhorn components from drains
+
+## Summary
+
+Some Longhorn components should be available to correctly handle cleanup/detach Longhorn volumes during the draining process.
+They are: `csi-attacher`, `csi-provisioner`, `longhorn-admission-webhook`, `longhorn-conversion-webhook`, `share-manager`, `instance-manager`, and daemonset pods in `longhorn-system` namespace.
+
+This LEP outlines our existing solutions to protect these components, the issues of these solutions, and the proposal for improvement.
+
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/3304
+
+## Motivation
+
+### Goals
+
+1. Have better ways to protect Longhorn components (`csi-attacher`, `csi-provisioner`, `longhorn-admission-webhook`, `longhorn-conversion-webhook`) without demanding the users to specify the draining flags to skip these pods.
+1. Protecting `share-manager` in the single-node cluster.
+
+## Proposal
+
+Our existing solutions to protect these components are:
+* For `instance-manager`: dynamically create/delete instance manager PDB
+* For Daemonset pods in `longhorn-system` namespace: we advise the users to specify `--ignore-daemonsets` to ignore them in the `kubectl drain` command. This actually follows the [best practice](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/#:~:text=If%20there%20are%20pods%20managed%20by%20a%20DaemonSet%2C%20you%20will%20need%20to%20specify%20%2D%2Dignore%2Ddaemonsets%20with%20kubectl%20to%20successfully%20drain%20the%20node)
+* For `csi-attacher`, `csi-provisioner`, `longhorn-admission-webhook`, and `longhorn-conversion-webhook`: we advise the user to specify `--pod-selector` to ignore these pods
+* For `share-manager`, we currently don't have any advice/protection
+
+
+1. For `csi-attacher`, `csi-provisioner`, `longhorn-admission-webhook`, and `longhorn-conversion-webhook`
+
+   The problem with the existing solution is that sometime, users could not specify `--pod-selector` for the `kubectl drain` command.
+   For example, for the users that are using the project [System Upgrade Controller](https://github.com/rancher/system-upgrade-controller), they don't have option to specify `--pod-selector`.
+   Also, we would like to have a more automatic way instead of relying on the user to set kubectl drain options.
+
+   Therefore, we propose the following design:
+    * Longhorn manager automatically create PDBs for `csi-attacher`, `csi-provisioner`, `longhorn-admission-webhook`, and `longhorn-conversion-webhook` with `minAvailable` set to 1.
+      This will make sure that each of these deployment has at least 1 running pod during the draining process.
+    * Longhorn manager continuously watches the volumes and removes the PDBs once there is no attached volume.
+
+   This should work for both single-node and multi-node cluster.
+
+1. For `share-manager`
+
+   If there is at least 1 workload pod on the same node with the share manager pod, prevent it from being evicted first via PDB. 
+   Otherwise, we remove the PDB to allow the `share-manager` to be drained.
+
+### User Stories
+
+
+
+#### Story 1
+Before the enhancement, users would need to specify the drain options for drain command to exclude Longhorn pods.
+Sometimes, this is not possible when users use third-party solution to drain and upgrade kubernetes, such as System Upgrade Controller.
+Also, before the enhancement, we don't protect the `share-manager` pod in single-node cluster.
+
+After the enhancement, the user can doesn't need to specify the drain options for the drain command to exclude Longhorn pods, and we protect the `share-manager` pod in single-node cluster
+#### Story 2
+
+### User Experience In Detail
+
+After the enhancement, the user can doesn't need to specify the drain options for the drain command to exclude Longhorn pods.
+
+### API changes
+
+None
+
+## Design
+
+### Implementation Overview
+
+1. For `csi-attacher`, `csi-provisioner`, `longhorn-admission-webhook`, and `longhorn-conversion-webhook`
+
+Create a new controller inside Longhorn manager called `longhorn-pdb-controller`, the controller listens for the changes for
+`csi-attacher`, `csi-provisioner`, `longhorn-admission-webhook`, `longhorn-conversion-webhook`, and Longhorn volumes to adjust the PDB correspondingly.
+
+1. For `share-manager`
+
+We will create PDB for `share-manager` in share-manager-controller when the `share-manager` is used by workloads on the same node.
+Otherwise, when share-manager doesn't exist or when the `share-manager` is only used by workload on different nodes, remove the PDB.
+
+### Test plan
+
+Integration test plan.
+
+For engine enhancement also requires engine integration test plan.
+
+### Upgrade strategy
+
+Anything that requires if the user wants to upgrade to this enhancement.
+
+## Note [optional]
+
+Additional notes.


enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
```diff
@@ -0,0 +1,99 @@
...
+
+
+1. For `csi-attacher`, `csi-provisioner`, `longhorn-admission-webhook`, and `longhorn-conversion-webhook`
+
+   The problem with the existing solution is that sometime, users could not specify `--pod-selector` for the `kubectl drain` command.
+   For example, for the users that are using the project [System Upgrade Controller](https://github.com/rancher/system-upgrade-controller), they don't have option to specify `--pod-selector`.
+   Also, we would like to have a more automatic way instead of relying on the user to set kubectl drain options.
+
+   Therefore, we propose the following design:
+    * Longhorn manager automatically create PDBs for `csi-attacher`, `csi-provisioner`, `longhorn-admission-webhook`, and `longhorn-conversion-webhook` with `minAvailable` set to 1.
```
innobead, line 38: note: 

The context behind this is we have the below conditions for these components, so can use `minAvailable` 1.
- at least 2 replicas
- replicas/pods are pod anti-affinity provision

enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
```diff
@@ -0,0 +1,99 @@
...
+      This will make sure that each of these deployment has at least 1 running pod during the draining process.
+    * Longhorn manager continuously watches the volumes and removes the PDBs once there is no attached volume.
+
+   This should work for both single-node and multi-node cluster.
+
+1. For `share-manager`
+
+   Adding PDB would be problematic for multi-node cluster because it will block the draining of
```
innobead, line 46: Because the share manager is possibly used by multiple pods in different nodes w/o any order, it seems not easy to tackle share manager protection somehow except we have share manager failover for this case. 

Should just let the share manager be evicted if workloads using it on other nodes, then it will just be recreated to another node, then clients/workloads should still access it eventually. If the share manager is used by workloads on the same node, prevent it to evict first via PDB. 

Simply said, need to keep the share manager around if at least one dependent workload is on a different node.

WDYT? @PhanLe1010 

cc @derekbit @shuo-wu

enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
```diff
@@ -0,0 +1,99 @@
...
+   a node forever if the node contains a share manager pod and the share manager pod is being used by workload on a different node.
+
+   So, Longhorn manager should only create PDB for the `share-manager` when it sees that the cluster only has a single Longhorn node.
```
innobead, line 49: Probably should rely on the below instead of the number of nodes?

> Simply said, need to keep the share manager around if at least one dependent workload is on a different node.

enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
```diff
@@ -0,0 +1,99 @@
...
+None
+
+## Design
+
+### Implementation Overview
+
+1. For `csi-attacher`, `csi-provisioner`, `longhorn-admission-webhook`, and `longhorn-conversion-webhook`
+
+Create a new controller inside Longhorn manager called `longhorn-pdb-controller`, the controller listens for the changes for
+`csi-attacher`, `csi-provisioner`, `longhorn-admission-webhook`, `longhorn-conversion-webhook`, and Longhorn volumes to adjust the PDB correspondingly.
```
innobead, line 81: So for PDB management, there will be three places. 

- General (for CSI sidecars, etc) : longhorn-pdb-controller
- IM: instance manager controller
- SM: share manager controller

Ideally, can be placed in the same controller, but it would cause some extra efforts. 

The better design could be controllers are just interfaces, but have a PdbManager individually to handle the requests from different controllers for different PDB creation/deletion.  

cc @c3y1huang

innobead (commented):
In general, LGTM. Just a few comments or notes, especially share manager.

enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
PhanLe1010: Thank you for the context. I agree that `at least 2 replicas` is a prerequisite. However, I feel that `replicas/pods are pod anti-affinity provision` is not relevant because even if multiple csi pod lands on a single node. When we drain that node, kubectl drain will evict pod one by one. Eventually, they will be able to move and drain finishes. What do you think?

enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
PhanLe1010: > Simply said, need to keep the share manager around if at least one dependent workload is on a different node.

Assume that the share manager is on node-1, workload are on node-2. If we keep the PDB to protect the share manager in this case, user can never finish draning node-1. What do you think?

enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
PhanLe1010: > The better design could be controllers are just interfaces, but have a PdbManager individually to handle the requests from different controllers for different PDB creation/deletion.

Thank you very much for the suggestion. Could you please share a bit more details how we can structure the interface and PDBManager?

enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
PhanLe1010: I have the same concern with https://github.com/longhorn/longhorn/pull/5516#discussion_r1130347353

enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
innobead: Yeah, you are right, because the point is more than one replica. 

Do you think PDB will be helpful for concurrent draining instead of one by one? I feel probably it could be :thinking:

enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
innobead: Exactly!

enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
PhanLe1010: I think PDB should be helpful for concurrent draining. As long as there is 1 pod available somewhere, PDB will not block concurrent draining

enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
innobead: PDB management should be similar in these three controllers, so the idea is to avoid duplicated code for that if could. Having a unified help util functions for that probably should be good enough, but if need to keep some stateful management, then an independent PdbManager should be better for responsibility segregation. 

* Controller -> PDB manager (manage all PDB stuff, so we will not what PDB stuff will be handled here)

It's nice to have and depends on the implementation context, but the goal is for responsibility segregation and making code clean.

enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
innobead: Then this can be part of QA later on. 

cc @longhorn/qa

enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
derekbit: `If the share manager is used by workloads on the same node, prevent it to evict first via PDB.`
Does it mean on single node cluster or on multi-node cluster? For a multi-node cluster, is PDB protection of a share manager necessary?

innobead:
LGTM until the share manager PDB handling gets updated.

enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
derekbit: cc @innobead

innobead: Do you mean because the share manager pod will failover, don't need to protect it except for one node cluster? For example, a share manager pod can be drained before the workload on the same node, because it will failover, the workload should be able to drain in the end, right?

PhanLe1010: Fixed it. Thank you both

enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
shuo-wu: Do you mean something like InstanceHandler in pkg `controller`?

enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
```diff
@@ -0,0 +1,95 @@
...
+   Therefore, we propose the following design:
+    * Longhorn manager automatically create PDBs for `csi-attacher`, `csi-provisioner`, `longhorn-admission-webhook`, and `longhorn-conversion-webhook` with `minAvailable` set to 1.
+      This will make sure that each of these deployment has at least 1 running pod during the draining process.
+    * Longhorn manager continuously watches the volumes and removes the PDBs once there is no attached volume.
+
+   This should work for both single-node and multi-node cluster.
+
+1. For `share-manager`
+
+    If the share manager is currently used by workloads on the same node, prevent it to evict first via PDB.
```
shuo-wu, line 46: Based on the previous discussion, is the below summary for the SM pod handling right?
1. If there are multiple workload pods using the SM pod, no matter if there is one workload being on the node the SM pod resides on, no protection.
2. If there is only one workload pod using the SM pod and they are on different nodes, no protection.
3. If there is only one workload pod using the SM pod and they are on the same nodes, retain the PDB util the workload pod gets evicted.

shuo-wu (commented):
In general LGTM

enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
PhanLe1010: I am understanding that our current conclusion is:
`If there is at least 1 workload pod on the same node with the share manager pod, create PDB` so:

1. If there are multiple workload pods using the SM pod, no matter if there is one workload being on the node the SM pod resides on, no protection. -> False
2. If there is only one workload pod using the SM pod and they are on different nodes, no protection. -> True
3. If there is only one workload pod using the SM pod and they are on the same nodes, retain the PDB util the workload pod gets evicted. -> True

What do you think @shuo-wu @innobead @derekbit ?

innobead: Correct.

innobead (approved):

enhancements/20230307-pdb-for-longhon-csi-and-webhook.md
shuo-wu: In other words, we will remove the SM pod PDB when:
1. No workload using the RWX volume. (This includes the case that the last workload and the SM pod are on the same node).
2. There is one workload but it is not on the node the SM pod resides on.

With this perspective, the 2nd one looks weird...

PhanLe1010: Our current conclusion is that `we will remove the PDB when there is no workload on the same node as the share manager.` So we will remove PDB when
1. No workload using the RWX volume. -> True
2. There is one workload but it is not on the node the SM pod resides on. -> Also true

Yeah, I agree that the 2nd point feel a bit weird. However, if we don't handle like that we can have the problem: user can never finish draining node-1 which has the share manager pod while its workload pod is running on node-2. What do you think?

shuo-wu: This means, PDB will protect all kinds of RWX workloads except the only one that does not reside on the SM pod node.

For example, the SM pod is on node-1, and 2 workloads using it are on node-1 and node-2. When we try to drain 2 nodes one by one, there will be 2 cases:

Case 1:
  1. node-2 drain. The one workload on node-2 gets evicted.
  2. node-1 drain. The workload gets evicted then the SM pod. ==> All workloads are stopped gracefully.

Case 2:
  1. node-1 drain. The workload gets evicted first. Then the only workload is not on the SM pod node. Hence here k8s can evict the SM pod immediately
  2. node-2 drain. The workload gets evicted after the SM pod eviction. ==> One workload is not gracefully stopped.

innobead: It seems we merged the PR, but still can continue the discussion to clarify the use cases.

> This means, PDB will protect all kinds of RWX workloads except the only one that does not reside on the SM pod node.

It should be opposite? The PDB will only exist when there is at least a dependent workload on the same node where the SM pod is running. 

> For example, the SM pod is on node-1, and 2 workloads using it are on node-1 and node-2. When we try to drain 2 nodes one by one, there will be 2 cases:
> 
> Case 1:
> 
> 1. node-2 drain. The one workload on node-2 gets evicted.
> 2. node-1 drain. The workload gets evicted then the SM pod. ==> All workloads are stopped gracefully.

This is correct.

> Case 2:
> 
> 1. node-1 drain. The workload gets evicted first. Then the only workload is not on the SM pod node. Hence here k8s can evict the SM pod immediately
> 2. node-2 drain. The workload gets evicted after the SM pod eviction. ==> One workload is not gracefully stopped.

The SM pod should be rescheduled to other nodes after draining from node 1, so node-2 should be still evicted successfully if the SM pod is back before that, right? However, it would be a chance the workload has been drained before the SM pod failover.

We might improve this to have PDB around during the SM pod failure, so can ensure the dependent workload will not be evicted earlier during this period.  @shuo-wu @PhanLe1010 WDYT?

shuo-wu: IMO, either we use PDB to make sure all RWX workloads are being stopped gracefully, or we just directly rely on the SM pod failover mechanism. Based on the current design, as I mentioned, the behaviors can be inconsistent under the same scenario.

innobead: > Based on the current design, as I mentioned, the behaviors can be inconsistent under the same scenario.

@shuo-wu can you elaborate a bit? it's important for us to clarify.

shuo-wu: See the example I raised above. With the same init state, draining 2 nodes would lead to different result.

innobead: > See the example I raised above. With the same init state, draining 2 nodes would lead to different result.

I see, but this is inevitable if you need to rely on the SM pod availability to ensure the dependent workload/pod eviction w/o correct detachment. 

From the user's perspective, if they see all workloads can be evicted in the end, then the behavior should be consistent and don't need to understand how it actually works behind by using PDB + SM pod failover.

shuo-wu: From the user's perspective, YES they may not be able to see any differences. But it means Longhorn **partially** relies on the SM pod failover mechanism... This makes logic not as straightforward as before...
Anyway, it's not a critical issue. I am fine with either but it's better to have a comment then.

innobead: @PhanLe1010 @derekbit do you have any concerns or questions about what @shuo-wu said?

PhanLe1010: I can see @shuo-wu 's point that under the same scenario, there are 2 different possible behaviors. This might raise some inconsistency during debugging. However, I do think that both behaviors are ok from the user's perspective since the workloads will eventually be evicted successfully and the drains can finish (gracefully or not gracefully)

On the other hand,
1.  **If we don't implement this PDB logic**: We would have a deadlock in the scenario: share manager is on node-1, workload-1 is on node-1, workload-2 is node-2 -> node-2 is un-schedulable somehow -> drain node-1 -> share manager is evicted -> share manager has no place to move to -> workload-1 cannot be evicted -> drain stuck
2. **If we implement this PDB logic such that use PDB to make sure all RWX workloads are being stopped gracefully,**: We would have a deadlock in the scenario: share manager is on node-1, workload-1 is on node-1, workload-2 is node-2 -> node-2 is un-schedulable somehow -> drain node-1 -> share manager can't be evicted because the share manager's PDB is waiting for workload-2 to be gracefully stopped first -> drain stuck

shuo-wu: > If we don't implement this PDB logic: We would have a deadlock in the scenario: share manager is on node-1, workload-1 is on node-1, workload-2 is node-2 -> node-2 is un-schedulable somehow -> drain node-1 -> share manager is evicted -> share manager has no place to move to -> workload-1 cannot be evicted -> drain stuck

For this part, will Longhorn block node-2 draining when the SM pod cannot be moved to this node?

---

Yesterday I had a quick discussion with David and he corrected my misunderstanding about this LEP. The right summary should be: As long as there is a running workload staying with the SM pod on the same node, Longhorn will have a PDB protecting the SM pod, **no matter if there are other workloads running** on other nodes. (But the example/case I mentioned is still valid.)

---

My concern is, if the SM pod is evicted while there are pending IOs in the workload write buffer, will evicting this workload in the next moment would lead to unexpected data loss? If NO and the case1 you mentioned above won't happenb, why do we need to protect SM pods?

PhanLe1010: > For this part, will Longhorn block node-2 draining when the SM pod cannot be moved to this node?

Longhorn is block ing node-1 draining in this case

> Yesterday I had a quick discussion with David and he corrected my misunderstanding about this LEP. The right summary should be: As long as there is a running workload staying with the SM pod on the same node, Longhorn will have a PDB protecting the SM pod, no matter if there are other workloads running on other nodes. (But the example/case I mentioned is still valid.)

Yeah, agree. I will update the LEP

> But the example/case I mentioned is still valid
This example https://github.com/longhorn/longhorn/pull/5516#discussion_r1133429702 right? Yeah, I agree that it is still valid. However, I do think that both behaviors are ok from the user's perspective since the workloads will eventually be evicted successfully and the drains can finish (gracefully or not gracefully)

shuo-wu: > share manager has no place to move to -> workload-1 cannot be evicted
> Longhorn is block ing node-1 draining in this case. Draining of node-2 is also affected because there is no share manager to handle unmount command

Here why can't workload-1 be evicted?

---
> I am not sure but looks like yes, there can potentially data loss

If we say the workloads should re-try handling the in-fly data after the restart since the RWX volume does not report the IO success, then we are fine with the current implementation. Otherwise, we should protect SM pods for all kinds of case.

PhanLe1010: > Here why can't workload-1 be evicted?

I think share manager is evicted first so the longhorn csi plugin cannot finish the NodeUnpublishVolume call

PhanLe1010: > If we say the workloads should re-try handling the in-fly data after the restart since the RWX volume does not report the IO success, then we are fine with the current implementation. Otherwise, we should protect SM pods for all kinds of case.

Protecting all kind of RWX workload has this problem:

> If we implement this PDB logic such that use PDB to make sure all RWX workloads are being stopped gracefully: We would have a deadlock in the scenario: share manager is on node-1, workload-1 is on node-1, workload-2 is node-2 -> drain node-1 -> share manager can't be evicted because the share manager's PDB is waiting for workload-2 to be gracefully stopped first -> drain stuck

shuo-wu: > I think share manager is evicted first so the longhorn csi plugin cannot finish the NodeUnpublishVolume call

OK. I will do a test tomorrow for this part:
1. Launch 2 workload pods using a RWX volume
2. Cordon all nodes.
3. Kill the SM pod
4. Kill/Scale down one workload pod or directly drain one node.

PhanLe1010: I did the test https://github.com/longhorn/longhorn/pull/5516#discussion_r1137192884. Looks like we fixed the problem so we don't need the share manager to be around for the NodeUnpublishVolume to success (by trying to force unmount):
```
[longhorn-csi-plugin-s4zbj longhorn-csi-plugin] time="2023-03-15T23:51:07Z" level=debug msg="Trying to force unmount potential mount point /var/lib/kubelet/pods/393967fd-0317-4ec6-b1d9-636e5e6e177a/volumes/kubernetes.io~csi/pvc-84826614-7a13-489d-a509-0d4215239dc6/mount"
```

So look likes we don't need to protect the share manager at all. 
cc @innobead @derekbit

innobead: So basically the RWX volume can be always be unmounted no matter if the share manager pod around?

PhanLe1010: @innobead From testing, yes if we backport this fix to 1.3.x and 1.4.x https://github.com/longhorn/longhorn/issues/5296
Could you please help with that? I will change this LEP and the PR to move this ticket forward for the RC testing

PhanLe1010: Update LEP https://github.com/longhorn/longhorn/pull/5568

@shuo-wu @innobead @derekbit

innobead: Sounds good, then it will be easy for us to handle the PDB scenario now.

I like the clarification here, good job @shuo-wu @PhanLe1010 :+1: . 

Just a sum up (the original idea was to protect the share manager pod to prevent potentially unexpected volume detachment if it's evicted earlier than the workload pod on the same node; however, the assumption seems invalid, since the volume of the workaround pod will be able to detach w/o issues no natter if the share manager pod is around.)

innobead: @PhanLe1010 Sure, let me know when the LEP and code PR update is ready, then we can move it forward.

PhanLe1010: @innobead [Code PR](https://github.com/longhorn/longhorn-manager/pull/1764), [updated LEP](https://github.com/longhorn/longhorn/pull/5568), and [test plan](https://github.com/longhorn/longhorn/issues/3304#issuecomment-1467174481) are ready. Thank you very much

shuo-wu: Thank you for helping test it. This makes things clear!

PR #5520, opened 2023-03-09
feat(lep): recurring filesystem trim design

c3y1huang:
https://github.com/longhorn/longhorn/issues/5186

### DIFF
--- a/enhancements/20230309-recurring-filesystem-trim.md
+++ b/enhancements/20230309-recurring-filesystem-trim.md
@@ -1,0 +1,86 @@
+# Recurring Filesystem Trim
+## Summary
+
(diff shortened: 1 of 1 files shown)


shuo-wu (commented):
NIT: May need to handle `deleteSnapMode` as well since it's meaningless to this.

innobead:
> NIT: May need to handle `deleteSnapMode` as well since it's meaningless to this.

@shuo-wu can you explain a little more?

innobead (approved):
LGTM.

innobead:
@mantissahz Please review this as well.

mantissahz:
This LEP in general looks good to me and straightforward.

>I mean, the new field introduced by https://github.com/longhorn/longhorn-manager/pull/1708#event-8708770632.

I would like to clean up old snapshots that exceeds the retain number before create a new snapshot mandatorily because it is an abnormal situation. (so there is no new field introduced) as mentioned in https://github.com/longhorn/longhorn-manager/pull/1708#issuecomment-1463152193

PR #5568, opened 2023-03-16
Update the LEP: Use PDB to protect Longhorn components from drains

PhanLe1010:
longhorn/longhorn#3304


innobead (changes requested):
Can we have a note explaining why we don't need to protect the share manager pod? That will be helpful when revisiting this LEP.

innobead (approved):

Issue #5143, opened 2022-12-26
[FEATURE] Support RWX volume trim


innobead:
For a RWX volume, the `/dev/longhorn/xxxx` is mounted to `/export/xxxx` in a share-manager pod, so you cannot see the mountpoint on the host and hence got the error message `No such file or directory`.
A NFS mountpoint doesn't support discard/trim operation as well. For the RWX volumes, need to figure out a way to do trim operation instead of `nsenter .... fstrim <mountpoint>`. Let me create a ticket for the RWX trim feature.

_Originally posted by @derekbit in https://github.com/longhorn/longhorn/issues/836#issuecomment-1363794746_

cc @shuo-wu @derekbit @tommy04062019 @longhorn/qa

chriscchien:
Verified in longhorn master (longhorn-manager `7bdc99`, longhorn-share-manager `1729f0`) with test [steps](https://github.com/longhorn/longhorn/issues/5143#issuecomment-1475307800)
Result **Pass**

Dynamically provision rwx volume with deployment, after delete file in mount point then trim rwx volume from UI, `volume.status.actualSize` decreased, write data again into mount point and scale down/up deployment let volume reattach, data in mount point consistent

PR #5594, opened 2023-03-19
Update crd template and longhorn.yaml

derekbit:
https://github.com/longhorn/longhorn/issues/5143


innobead (approved):
LGTM. Merge this when longhorn-manager changes merged.

chart/templates/crds.yaml
```diff
@@ -2629,16 +2629,23 @@ spec:
...
                 type: string
             type: object
           status:
             description: ShareManagerStatus defines the observed state of the Longhorn share manager
             properties:
               endpoint:
+                description: NFS endpoint that can access the mounted filesystem of the volume
+                type: string
+              ip:
+                description: IP address of the share manager pod
```
innobead, line 2642: Do we still need this? I did not see this property in longhorn-manager PR?

innobead (changes requested):

derekbit:
This is for the old design. Not needed now.

innobead (approved):

chart/templates/crds.yaml
derekbit: Removed. Thank you.

Issue #5672, opened 2023-03-28
[IMPROVEMENT] Bump CSI sidecar components' version


PhanLe1010:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

We are currently using csi-provisioner v2.1.2 which has this upstream race bug that sometimes crashes the csi-provisioner pod [https://github.com/kubernetes-csi/external-provisioner/issues/595](https://github.com/kubernetes-csi/external-provisioner/issues/595).

 Considering bump the csi-provisioner version to version 3.x.x to avoid the upstream bug. While we are here, also consider bump other csi-xxx components' version

PhanLe1010:
This improvement should go to v1.4.x and v1.5.x only because Longhorn 1.2.x and v1.3.x have to support Kubernetes 1.18 which [csi-provisioner 3.x.x doesn't support ](https://kubernetes-csi.github.io/docs/external-provisioner.html)

ejweber:
Bumping all of the sidecars is straightforward enough and does not provide any complications in Longhorn v1.5.

Because we announced the deprecation of the v1beta1 VolumeSnapshot API in Longhorn v1.4 (for removal in a future version), we should not upgrade to versions of external-provisioner and external-snapshotter that only support the v1 VolumeSnapshot API.

I created separate PRs for the 1.4.x branches to avoid upgrading external-provisioner and external-snapshotter too far). I created separate PRs for the 1.3.x branches because I slight differences in the versions already in certain files were going to cause merge conflicts.

In general, I think it's valuable to keep the 1.3.x and 1.4.x branches updated with current sidecar images, but I'm not overly opposed to dropping these PRs if we only want to move forward in v1.5.

innobead:
> Bumping all of the sidecars is straightforward enough and does not provide any complications in Longhorn v1.5.
> 
> Because we announced the deprecation of the v1beta1 VolumeSnapshot API in Longhorn v1.4 (for removal in a future version), we should not upgrade to versions of external-provisioner and external-snapshotter that only support the v1 VolumeSnapshot API.

We probably should upgrade this to 1.5 or any concerns?

> I created separate PRs for the 1.4.x branches to avoid upgrading external-provisioner and external-snapshotter too far). I created separate PRs for the 1.3.x branches because I slight differences in the versions already in certain files were going to cause merge conflicts.
> 
[...]

PR #5716, opened 2023-04-06
Bump csi-provisioner to v3.4.1

ejweber:
#5672

### DIFF
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -113,7 +113,7 @@ questions:
     label: Longhorn CSI Provisioner Image Repository
     group: "Longhorn CSI Driver Images"
   - variable: image.csi.provisioner.tag
-    default: v2.1.2
+    default: v3.4.1
     description: "Specify CSI provisioner image tag. Leave blank to autodetect."
     type: string
     label: Longhorn CSI Provisioner Image Tag
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -50,7 +50,7 @@ image:
       tag: v3.4.0
     provisioner:
       repository: longhornio/csi-provisioner
-      tag: v2.1.2
+      tag: v3.4.1
     nodeDriverRegistrar:
       repository: longhornio/csi-node-driver-registrar
       tag: v2.5.0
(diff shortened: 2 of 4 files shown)


ejweber:
It looks like we need to update other sidecars as well (all of the versions we are using have fallen out of support). There are some interdependencies between versions that I want to be careful with (and it is probably best to handle all of it at once). I'll create a couple of new PRs tomorrow that handle this better. Apologies for the churn.

PhanLe1010:
Besides of longhorn/longhorn repo, we need to update the version in longhorn/longhorn-manager as well: https://github.com/longhorn/longhorn-manager/blob/66e31ba3e7ae77779a9ba3b5bf6695e39b9f5f68/csi/deployment.go#L23-L28

ejweber:
Good catch, @PhanLe1010, and thanks for your comment. Unfortunately, I didn't look closely enough at how the driver was being deployed on my initial pass. I assumed the Helm chart was taking the provided values and doing the deployment of the driver (using a template or templates) directly. I see now that Helm instead deploys a longhorn-manager-deployer Pod that actually does the deployment (with the "longhorn-manager csi ..." command). Eventually, on uninstall, Helm deploys a longhorn-uninstall Pod (with the "longhorn-manager uninstall" command) that reverses it.

It's clear to me now why the additional values you pointed out exist. However, it's a bit unclear why the longhorn-manager executable itself must be responsible for creating and cleaning up the CSI components instead of the overarching deployment mechanism (e.g. Helm). Do you or @innobead have any insight to share on this? Do we support a workflow in which a user might directly execute longhorn-manager to install or uninstall CSI components? (This certainly isn't necessary for me to prepare a new PR, but it might help shape the way I think about proposed changes moving forward.)

PhanLe1010:
Discussed with @ejweber: A few reasons that we are using `longhorn-manager` image in the driver-deployer to deploy the `csi-xxx` components instead of using the traditional Helm template are:
1. The driver-deployer can dynamically discover the kubelet directory and deploy the `csi-xxx` to connect to the correct the Unix domain socket
2. The driver watches the `settings.longhorn.io` like priorityClass, tolerations,... and sets them to `csi-xxx` deployment

PR #5725, opened 2023-04-07
Bump CSI sidecars v1.5

ejweber:
#5672

We can't move to these versions of external-provisioner and external-snapshotter in Longhorn < v1.5, as they no longer support the v1beta1 VolumeSnapshot API. We deprecated its use in Longhorn v1.4 (https://github.com/longhorn/longhorn/blob/master/CHANGELOG/CHANGELOG-1.4.0.md#deprecation--incompatibilities). I will create separate PRs to bump sidecar versions in the v1.4.x and v1.3.x branches without these components.

We may need a CHANGELOG notice in v1.5 that v1beta1 VolumeSnapshot functionality is completely gone.

EDIT: This comment was previously (incorrectly) missing the "<" sign.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -50,7 +50,7 @@ image:
       tag: v4.2.0
     provisioner:
       repository: longhornio/csi-provisioner
-      tag: v2.1.2
+      tag: v3.4.1
     nodeDriverRegistrar:
       repository: longhornio/csi-node-driver-registrar
       tag: v2.7.0
@@ -59,7 +59,7 @@ image:
       tag: v1.7.0
     snapshotter:
       repository: longhornio/csi-snapshotter
-      tag: v5.0.1
+      tag: v6.2.1
     livenessProbe:
       repository: longhornio/livenessprobe
       tag: v2.9.0
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -47,22 +47,22 @@ image:
   csi:
     attacher:
       repository: longhornio/csi-attacher
-      tag: v3.4.0
+      tag: v4.2.0
     provisioner:
       repository: longhornio/csi-provisioner
       tag: v2.1.2
     nodeDriverRegistrar:
       repository: longhornio/csi-node-driver-registrar
-      tag: v2.5.0
+      tag: v2.7.0
     resizer:
       repository: longhornio/csi-resizer
-      tag: v1.3.0
+      tag: v1.7.0
     snapshotter:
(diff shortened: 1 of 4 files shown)


chart/values.yaml
```diff
@@ -47,22 +47,22 @@ image:
...
-      tag: v1.3.0
+      tag: v1.7.0
     snapshotter:
       repository: longhornio/csi-snapshotter
-      tag: v5.0.1
+      tag: v6.2.1
     livenessProbe:
       repository: longhornio/livenessprobe
-      tag: v2.8.0
+      tag: v2.9.0
```
PhanLe1010, line 65: cc @c3y1huang Do you see any issue bumping the `livenessprobe` to `v2.9.0` ?

deploy/longhorn-images.txt
```diff
@@ -1,9 +1,9 @@
...
-longhornio/csi-resizer:v1.3.0
-longhornio/csi-snapshotter:v5.0.1
-longhornio/csi-node-driver-registrar:v2.5.0
-longhornio/livenessprobe:v2.8.0
+longhornio/csi-attacher:v4.2.0
+longhornio/csi-provisioner:v3.4.1
+longhornio/csi-resizer:v1.7.0
+longhornio/csi-snapshotter:v6.2.1
+longhornio/csi-node-driver-registrar:v2.7.0
+longhornio/livenessprobe:v2.9.0
```
PhanLe1010, line 6: cc @yangchiu Could you please help pushing these new image using our automation pipeline? Thank you very much

PhanLe1010 (approved):
LGTM

deploy/longhorn-images.txt
ejweber: If you don't mind providing a super high level explanation of what is involved when you do this, @yangchiu, I would appreciate it. I took a look at some of the automation early in the process and my takeaway was that this would probably happen automatically at the start of automated integration testing.

yangchiu: Yes, an extra stage can be added to automated integration testing to check if new csi images needed to be mirrored to longhorn repository before running the test.

innobead: @yangchiu can we create an issue for this? thanks.

chart/values.yaml
c3y1huang: I don't see issues bumping this up.

innobead (approved):

PR #5727, opened 2023-04-07
Bump CSI sidecars v1.3

ejweber:
https://github.com/longhorn/longhorn/issues/5672

Mostly the same as #5725 except it doesn't bump external-provisioner and external-snapshotter beyond the v1beta1 VolumeSnapshot API.


ejweber:
We decided not to update the sidecars in previously released minor versions of Longhorn. See the discussion here: https://github.com/longhorn/longhorn/issues/5672#issuecomment-1502579905.

Issue #5005, opened 2022-12-06
[FEATURE] BackingImage Management via VolumeSnapshot


innobead:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

By leveraging CSI VolumeSnapshot protocol for backing image management (create/delete), backing-image based volume backup & restore by regarding each backing image as the source of a volume (ex: VolumeSnapshot with prebuilt VolumeSnapshotContent). 

When doing backing-image based volume snapshot, it can support in-cluster snapshot (another Backing Image) or out-of-cluster snapshot backup. Both all of them can support COW over each relative base image for delta data transfer for better space efficiency.

The goal is to use CSI snapshot as  the unified interface for users to do backing image management, backing-image based volume management and backup/restore.

## Describe the solution you'd like

CSI snapshot for backing image management, BI-based volume snapshot, backup & restore

## Additional context

#4165 

cc @shuo-wu @FrankYang0529 @ChanYiLin

chriscchien:
HI @ChanYiLin ,

From test [plan](https://github.com/longhorn/longhorn/blob/master/enhancements/20230417-extend-csi-snapshot-to-support-backingimage.md#scenerios-4-restore-on-demand-provisioning-backingimage) scenerios 4: Restore on-demand provisioning BackingImage, type=export-from-volume, success scenario
Volume name in step a is `test-vol` not consistent with the name in yaml content in step b `vol-export-src`, could you update it or make the volume name not static in yaml? Thank you.

```
Events:
  Type     Reason                Age                   From                                                                                      Message
  ----     ------                ----                  ----                                                                                      -------
  Warning  ProvisioningFailed    6m55s                 driver.longhorn.io_csi-provisioner-6df8547696-fxg6n_709f40d7-dffe-4ff4-9186-5712c885324a  failed to provision volume with StorageClass "longhorn": rpc error: code = InvalidArgument desc = Bad response statusCode [500]. Status [500 Internal Server Error]. Body: [message=unable to create backing image test-bi from source type export-from-volume with parameters map[export-type:qcow2 volume-name:vol-export-src]: admission webhook "validator.longhorn.io" denied the request: failed to get volume vol-export-src before exporting backing image, code=Server Error, detail=] from [http://longhorn-backend:9500/v1/backingimages]
[...]

PR #5771, opened 2023-04-17
feat(lep): extend csi snapshot to support backingimage

ChanYiLin:
ref: longhorn/longhorn#5005

Extend csi snapshot to support BackingImage

Summary:
- Users can use VolumeSnapshot with type `bi` to create/delete BackingImage
- Users can restore a Volume with a VolumeSnapshot (create a Volume with BackingImage)
- Users can restore a Volume with an existing BackingImage (pre-provision)
    - Create the VolumeSnapshot/VolumeSnapshotContent to associate to the existing BackingImage
- Users can restore a Volume with non-existing BackingImage (on-demand provision)
    - Create the VolumeSnapshot/VolumeSnapshotContent to provide BackingImage parameters
    - Create BackingImage based on the parameters when the users create PVC from the VolumeSnapshot


enhancements/20230417-extend-csi-snapshot-to-support-backingimage.md
```diff
@@ -0,0 +1,599 @@
...
+- Create a BackingImage.
+    - Get `export-type` from VolumeSnapshotClass parameter `export-type`, default to `raw.`
+    - Encode the `snapshotId` as `bi://backing?backingImageDataSourceType=export-from-volume&backingImage=test-snapshot-backing&volume-name=${VolumeName}&export-type=raw`
+    - This `snaphotId` will be used in the later CSI CreateVolume and DeleteSnapshot call.
+
+#### CreateVolume function
+
+- If VolumeSource type is `VolumeContentSource_Snapshot`, decode the `snapshotId` to get the parameters.
+    - `bi://backing?backingImageDataSourceType=${TYPE}&backingImage=${BACKINGIMAGE_NAME}&backingImageChecksum=${backingImageChecksum}&${OTHER_PARAMETES}`
+- If BackingImage with the given name already exists, create the volume.
```
innobead, line 286: So in this case, actually users can just provide the backing image name only? `bi://backingImage={value}`

enhancements/20230417-extend-csi-snapshot-to-support-backingimage.md
```diff
@@ -0,0 +1,599 @@
...
+        apiVersion: snapshot.storage.k8s.io/v1beta1
+        kind: VolumeSnapshot
+        metadata:
+          name: test-snapshot-backing
+        spec:
+          volumeSnapshotClassName: longhorn-snapshot-vsc
+          source:
+            persistentVolumeClaimName: test-vol
+        ```
+    3. Verify it failed since the Volume `test-vol` is not attached
```
innobead, line 353: This should be tackled after https://github.com/longhorn/longhorn/issues/3715 is finished for volume auto attachment.

innobead (changes requested):
In general, LGTM. Some comments needed to answer.

Also, when testing, need to consider some resilience cases for example: if a volume is created from a bi snapshot on demand, but the BI somehow is failed to created, what's the status of the volume.

enhancements/20230417-extend-csi-snapshot-to-support-backingimage.md
```diff
@@ -0,0 +1,599 @@
...
+        kind: VolumeSnapshotContent
+        metadata:
+            name: test-on-demand-backing
+        spec:
+            volumeSnapshotClassName: longhorn-snapshot-vsc
+            driver: driver.longhorn.io
+            deletionPolicy: Delete
+            source:
+            # NOTE: change this to provide the correct parameters
+            snapshotHandle: bi://backing?backingImageDataSourceType=download&backingImage=test-bi&url=https%3A%2F%2Flonghorn-backing-image.s3-us-west-1.amazonaws.com%2Fparrot.qcow2&backingImageChecksum=bd79ab9e6d45abf4f3f0adf552a868074dd235c4698ce7258d521160e0ad79ffe555b94e7d4007add6e1a25f4526885eb25c53ce38f7d344dd4925b9f2cb5d3b
```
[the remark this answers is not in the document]
ChanYiLin: 截圖 2023-04-27 下午2 49 23

this is actually one line
as you can see in the preview, there is no indent issue

enhancements/20230417-extend-csi-snapshot-to-support-backingimage.md
```diff
@@ -0,0 +1,599 @@
...
+
+#### Restoration of an existing BackingImage (pre-provisioned)
+
+Users can request the creation of a Volume based on a prior BackingImage which was not created via the CSI VolumeSnapshot.
+
+With the BackingImage already existing, users need to create the VolumeSnapshotContent with an associated VolumeSnapshot. The `snapshotHandle` of the VolumeSnapshotContent needs to point to an existing BackingImage. Example below for a Volume named `test-restore-existing-backing` and an existing BackingImage `test-bi`
+
+- For pre-provisioning, users need to provide following query parameters:
+    - `backingImageDataSourceType`: `sourceType` of existing BackingImage, e.g. `export-from-volume`, `download`
+    - `backingImage`: Name of the BackingImage
```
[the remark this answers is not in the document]
ChanYiLin: as discussed, we can support this but will not include this in this version.
This needs to introduce a **new type** for backing image and handle this case separately in the code.

enhancements/20230417-extend-csi-snapshot-to-support-backingimage.md
ChanYiLin: As discussed, this can be supported by using `bi://backing?backingImageDataSourceType=backing&backingImage=${BACKINGIMAGE_NAME}`
so we can handle this type and skip all the existing validation.

But we are not going to include this type in this version

enhancements/20230417-extend-csi-snapshot-to-support-backingimage.md
ChanYiLin: Yes.

ChanYiLin:
> In general, LGTM. Some comments needed to answer.
> 
> Also, when testing, need to consider some resilience cases for example: if a volume is created from a bi snapshot on demand, but the BI somehow is failed to created, what's the status of the volume.

Two functions involved in this feature, they are triggered by csi sidecar
- `createSnapshot`: **When creating VolumeSnapshot -> csi create VolumeSnapshotContent**
    - This function will be triggered and create the BackingImage.
    - BackingImage creation is an async call, create and return directly. Return error when create BackingImage call fails.
- `createVolume`: **When creating PVC -> csi create PV**
    - Prepare and check the BackingImage first, it it doesn't exist, create the BackingImage
    - Create the volume, it waits until the volume is completed or **timeout**.
    - So if the BackingImage is not ready for a period of time. createVolume() returns error timeout to the csi sidecar and PV create will fail.

enhancements/20230417-extend-csi-snapshot-to-support-backingimage.md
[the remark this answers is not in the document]
innobead: I mean

```yaml
source:
    snapshotHandle: 
```

enhancements/20230417-extend-csi-snapshot-to-support-backingimage.md
innobead: cc @PhanLe1010

enhancements/20230417-extend-csi-snapshot-to-support-backingimage.md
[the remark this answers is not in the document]
ChanYiLin: fixed, thanks!

innobead (approved):

PR #5784, opened 2023-04-19
feat: use existent priority-class when upgrading

FrankYang0529:
If users give a different `priorityClass` in `values.yaml`, the `longhorn-default-setting` configmap will be updated. If there is no attached volume in the environment, `priority-class` setting will be updated, and all LH pods will be redeployed. This behavior may surprise users. I update the chart to ignore the `priorityClass` value when it's upgrading. It's better to let users update the filed through the dashboard.

Issue: https://github.com/harvester/harvester/issues/3756

### DIFF
--- a/chart/templates/default-setting.yaml
+++ b/chart/templates/default-setting.yaml
@@ -45,7 +45,15 @@ data:
       {{- end -}}
       {{- $nodeSelector := list $windowsDefaultSettingNodeSelector $defaultSettingNodeSelector }}{{ join ";" (compact $nodeSelector) -}}
     {{- end }}
-    {{ if not (kindIs "invalid" .Values.defaultSettings.priorityClass) }}priority-class: {{ .Values.defaultSettings.priorityClass }}{{ end }}
+    {{- if .Release.IsInstall -}}
+      {{- if not (kindIs "invalid" .Values.defaultSettings.priorityClass) }}
+    priority-class: {{ .Values.defaultSettings.priorityClass }}
+      {{- end }}
+    {{- else if .Release.IsUpgrade -}}
+      {{- if ne (lookup "longhorn.io/v1beta2" "Setting" "longhorn-system" "priority-class").value "" }}
+    priority-class: {{ (lookup "longhorn.io/v1beta2" "Setting" "longhorn-system" "priority-class").value }}
+      {{- end }}
+    {{- end -}}
     {{ if not (kindIs "invalid" .Values.defaultSettings.autoSalvage) }}auto-salvage: {{ .Values.defaultSettings.autoSalvage }}{{ end }}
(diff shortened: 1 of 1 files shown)


innobead:
This reminds me of some settings which are also relying on if all volumes should be detached first like the storage network. 

I feel we should fix this in a common way for all these kinds of settings, for example below:
- applied, but will not take effect until all volumes are detached first

@c3y1huang Do you know if any other settings rely on the status of all volume attachments? also, during the longhorn upgrade, if a user changes the storage network in configmap setting in values.yaml while some volumes are still attached, what will happen? will the longhorn upgrade still succeed?

FrankYang0529:
> * applied, but will not take effect until all volumes are detached first

Yeah, it works like that now, but do we want to let users decide when to restart all pods?

c3y1huang:
> @c3y1huang Do you know if any other settings rely on the status of all volume attachments?

AFAIK, there are node-selector, priority-class-, taint-toleration, and storage-network.

> also, during the longhorn upgrade, if a user changes the storage network in configmap setting in values.yaml while some volumes are still attached, what will happen? will the longhorn upgrade still succeed?

Upgrading while there are attached volumes (ex: `---set defaultSettings.storageNetwork`) won't get blocked, but the setting will not get updated either.

innobead:
> > @c3y1huang Do you know if any other settings rely on the status of all volume attachments?
> 
> AFAIK, there are node-selector, priority-class-, taint-toleration, and storage-network.
> 
> > also, during the longhorn upgrade, if a user changes the storage network in configmap setting in values.yaml while some volumes are still attached, what will happen? will the longhorn upgrade still succeed?
> 
> Upgrading while there are attached volumes (ex: `---set defaultSettings.storageNetwork`) won't get blocked, but the setting will not get updated either.

Yeah, thanks. helm upgrade will still work since the config map will be allowed to change w/o any validation. However, it looks like reconciling the config map change will fail in the below code to make the setting inconsistent with the config map and its intended. Also, the config map controller will not periodically resync the config map change, so it will just stay in an inconsistent state.

[/datastore/longhorn.go#L107-L110](https://github.com/longhorn/longhorn-manager/blob/a7c73dbaba823c438ecd225628a5987c7fdb02ca/datastore/longhorn.go#L107-L110)
```go
		if err := s.ValidateSetting(string(name), value); err != nil {
			logrus.WithError(err).Warnf("Invalid customized default setting %v with value %v, will continue applying other customized settings", name, value)
			continue
		}
```

innobead:
> > * applied, but will not take effect until all volumes are detached first
> 
> Yeah, it works like that now, but do we want to let users decide when to restart all pods?

Yes, we should keep this behavior, since all settings requiring volumes detached all following this behavior, so we should not just change the change for priority class. 

I would say the eventual effect should be acceptable for now, because it will ensure not attached volumes at least, the impact on users should be minimal. WDYT?

innobead:
As discussed, let's close since this is the correct behavior.

Issue #5794, opened 2023-04-22
[BUG] Wrong `csi.livenessProbe` image group name in `/chart/questions.yaml`


weizhe0422:
## Describe the bug (🐛 if you encounter this issue)

In `/chart/questions.yaml`, the `csi.livenessProbe` image group name is not correct and it makes the  livenessprobe image name in another independent page. And this is only happened in `v1.3.x` branch.

<img width="740" alt="截圖 2023-04-22 下午4 25 35" src="https://user-images.githubusercontent.com/17548901/233772670-f223b0db-960f-440f-b093-683c20990eab.png">

## To Reproduce

3. Perform '....'

## Expected behavior

Fix the upstream to make the group name should be `Longhorn CSI Driver Images` rather than `Longhorn CSI Liveness Probe Images`
https://github.com/longhorn/longhorn/blob/f57838cb7843e3589023e4f035e21b374ffd93f7/chart/questions.yaml#L150

## Environment

 - Longhorn version: v1.3.3
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl): Rancher Catalog App

## Additional context

https://github.com/rancher/charts/pull/2555#issuecomment-1518428200

chriscchien:
Verified **pass** on longhorn v1.8.0-dev-20241117

The group of `csi.livenessProbe` image is correct
https://github.com/longhorn/longhorn/blob/057e597f8fd733602dbe8fdc358fcaae668f3240/chart/questions.yaml#L162

PR #5795, opened 2023-04-22
fix(chart): fix livenessprobe image group name in question.yaml

weizhe0422:
[Longhorn 5794](https://github.com/longhorn/longhorn/issues/5794)

 - Fix the upstream to make the group name should be `Longhorn CSI Driver Images` rather than `Longhorn CSI Liveness Probe Images`
https://github.com/longhorn/longhorn/blob/f57838cb7843e3589023e4f035e21b374ffd93f7/chart/questions.yaml#L150

### DIFF
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -147,7 +147,7 @@ questions:
     description: "Specify CSI liveness probe image repository. Leave blank to autodetect."
     type: string
     label: Longhorn CSI Liveness Probe Image Repository
-    group: "Longhorn CSI Liveness Probe Images"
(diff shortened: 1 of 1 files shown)


PhanLe1010 (approved):
LGTM

PhanLe1010:
Good catch. Looks like we forgot to backport [this commit](https://github.com/longhorn/longhorn/commit/a7e4b23350d68ae339d0e6d6c712e5087ce1d5d0) to v1.3.x so the issue only happens in v1.3.x

innobead (approved):

Issue #4914, opened 2022-11-22
[IMPROVEMENT] Remove the global setting `mkfs-ext4-parameters`


weizhe0422:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

Since we already could let user configure mkfs parameters in storage class (https://github.com/longhorn/longhorn/issues/4642) , therefore, the setting only specific to ext4 should be deprecated. 
       
## Describe the solution you'd like

The setting and related reference in `longhorn/longhorn-manager` and `longhorn/longhorn-ui` should be deprecated, e.g., 

  - deploy files
    - `longhorn/longhorn`: `question.yaml`, `values.yaml`
  - UI
    - `setting` and `create PV/PVC` page

## Additional context

[Add any other context or screenshots about the feature request here.](https://github.com/longhorn/longhorn/issues/4642)

PR #5830, opened 2023-04-28
Remove deprecated mkfs-ext4-parameters setting

ejweber:
https://github.com/longhorn/longhorn/issues/4914

Remove deprecated mkfs-ext4-parameters setting.


innobead (changes requested):
Have you generated the new longhorn.yaml?

innobead (approved):

PR #5832, opened 2023-04-30
LEP: Object Support for Longhorn

jecluis:
We are proposing Object support for Longhorn, by integrating aquarist-labs/s3gw into Longhorn, and seamlessly provide S3-compatible object endpoints from Longhorn volumes.

Signed-off-by: Joao Eduardo Luis \<joao@suse.com>

**Please Note:** While we believe this to be a decent state, there are many things that are still unknowns and we would love the feedback, and any constructive criticism or suggestions on how to improve will be immensely appreciated!

### DIFF
--- a/enhancements/20230430-object-endpoint.md
+++ b/enhancements/20230430-object-endpoint.md
@@ -1,0 +1,263 @@
+# Object Support for Longhorn
+
+## Summary
+
+By integrating s3gw with Longhorn, we are able to provide an S3-compatible,
+Object API to clients consuming Longhorn volumes. This is achieved by creating
+an Object endpoint (using s3gw) for a Longhorn volume.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/4154
+
+
+## Motivation
+
+### Goals
+
+* Provide an Object endpoint associated with a Longhorn volume, providing an
+  S3-compatible API.
+* Multiple Object endpoints should be supported, with each endpoint being backed
+  by one single Longhorn volume.
+
+
+### Non-goals
+
+* Integration of s3gw UI for administration and management purposes. Such an
+  Enhancement Proposal should be a standalone LEP by its own right.
+* Providing Object endpoints for multiple volumes. In this proposal we limit one
+  object endpoint per Longhorn volume (see longhorn/longhorn#5444).
+* Multiple Object endpoints for a single volume, either in Read-Write Many, or
+  as active/passive for HA failover. While we believe it to be of the utmost
+  importance, we are considering it future work that should be addressed in its
+  own LEP.
+* Specify how COSI can be implemented for Longhorn and s3gw. This too should be
+  addressed in a specific LEP.
+
+
+## Proposal
+
+### User Stories
+
+#### Story
+
+Currently, Longhorn does not support object storage. Should the user want to use
+their Longhorn cluster for object storage, they have to rely on third-party
+applications.
+
+Instead we propose to enhance the user experience by allowing a Longhorn volume
+to be presented to the user as an object storage endpoint, without having the
+user to install additional dependencies or manage different applications.
+
+### User Experience In Detail
+
+* A new page view exists specifically for "Object Endpoints"
+    * Within this page there is a "Create" button, and a list of existing
+      Object Endpoints.
+* The user clicks on "Create"
+* A modal dialog is shown, with the various Object Endpoint related fields
+* User specifies the endpoint name
+* User specifies their username and password combination for the administrator
+  user
+    * This can, potentially, be randomly generated as well.
+* For publicly accessible endpoints, the user must specify a domain name to be
+  used
+* The user must provide SSL certificates to be used by the endpoint
+* Then the user clicks "Ok"
+
+### API changes
+
+The API will need a new endpoint to create an object endpoint, as well as
+listing, updating, and deleting them. We believe it's not reasonable to reuse
+the existing `/v1/volumes` API endpoints, given they are semantically distinct
+from what we are trying to achieve.
+
+We thus propose the creation of a `/v1/endpoint/object` API endpoint. This route
+could also be `/v1/object-endpoint`, but we believe that by having a
+`/v1/endpoint/...` route we can potentially future proof the API in case other
+endpoint types (not just object) are eventually added.
+
+
+## Design
+
+### Implementation Overview
+
+Integrating Longhorn with s3gw will require the creation of mechanisms that
+allow us to 1) describe an object endpoint; 2) deploy an `s3gw` pod consuming
+a volume; and 3) deploy an `s3gw-ui` pod for management and administration. For
+the purposes of this LEP, there will always be one single `s3gw-ui` pod
+associated with an `s3gw` service, hence one `s3gw-ui` pod per Object Endpoint.
+We do not exclude eventually allowing one single `s3gw-ui` pod being associated
+with multiple Object Endpoints, but that should be considered as future work.
+
+We believe we will need a new Custom Resource Definition, `ObjectEndpoint`,
+representing an Object Endpoint deployment, containing information that will be
+required to deploy an `s3gw` endpoint, as well as the `s3gw-ui` administration
+interface.
+
+Given we need backing storage for each `s3gw` instance, we will rely on
+Persistent Volume Claims to request storage from the Longhorn cluster. This
+allows us to abstract ourselves from volume creation, and rely on existing
+Kubernetes infrastructure to provide us with the needed resources.
+
+An `ObjectEndpointController` will also be necessary, responsible for creating
+and managing `s3gw` pods, both the `s3gw` endpoint and the `s3gw-ui`. This
+controller will be listening for new resources of type `objectendpoint`, and
+will create the necessary bits for proper deployment, including services and
+pods.
+
+#### Custom Resource Definition
+
+We define a new Object Endpoint Custom Resource Definition, as follows:
+
+```golang
+type ObjectEndpoint struct {
+    metav1.TypeMeta
+    metav1.ObjectMeta
+
+    Spec    ObjectEndpointSpec
+    Status  ObjectEndpointStatus
+}
+```
+
+The endpoint `Spec`, as follows, contains the information that we consider most
+relevant to the endpoint at this point. More information may be added should the
+need arise.
+
+```golang
+type ObjectEndpointSpec struct {
+    Credentials     ObjectEndpointCredentials
+    StorageClass    string
+    Size            resource.Quantity
+}
+```
+
+The `Credentials` field, as defined below, contains the access and secret keys
+that are to be used as seed credentials for the object endpoint. We need these
+to set the credentials for the default administrator user.
+
+The `StorageClass` field contains the name of the Storage Class to be used when
+obtaining a new volume to back the `s3gw` service. This needs to be specified by
+the user with an existing Storage Class; alternatively the default Storage Class
+will be used.
+
+The `Size` field represents the initial size to provision for the new volume.
+
+```golang
+type ObjectEndpointCredentials struct {
+    AccessKey   string
+    SecretKey   string
+}
+```
+
+The Object Endpoint CRD also contains a `Status` field, represented below. This
+tracks the state of the object endpoint as observed by the object endpoint controller.
+
+```golang
+type ObjectEndpointStatus struct {
+    State       ObjectEndpointState
+    Endpoint    string
+}
+
+type ObjectEndpointState string
+```
+
+The `State` field can have one of the following values: `unknown`, `starting`,
+`running`, `stopping`, or `error`. The state machine begins at `unknown`, and
+moves to `starting` once the new object endpoint is detected and resource
+creation begins. Once all resources have been created, the controller then moves
+the state to `running`, and remains there until the object endpoint is deleted,
+at which point the controller moves the state to `stopping` while waiting for
+the associated resources to be cleaned up. The state `error` means something
+went wrong.
+
+
+#### The Object Endpoint Controller
+
+The Object Endpoint Controller will be responsible for lifecycle management of
+Object Endpoints, from creation to their deletion.
+
+To a large extent, the vast majority of the logic of the controller will be in
+creating resources, or handling the `starting` and `stopping` states discussed
+before.
+
+When the Object Endpoint is created, the required resources will need to be
+created. These include,
+
+* `Service` associated with the pods being deployed
+* Various `Secret` and `ConfigMap` needed by `s3gw` and `s3gw-ui`
+* The associated `Deployment`
+* A `Persistent Volume Claim` required for `s3gw` storage.
+
+During resource creation we will also need to explicitly create a
+`Persitent Volume` that will be bound to the `Persistent Volume Claim` mentioned
+above. We need to do this so we can opinionate on the file system being used; in
+this case, XFS, which `s3gw` requires for performance reasons.
+
+While handling `starting`, the resources required to have already been created
+through the Kubernetes API, but the controller still has to wait for them to be
+ready and healthy before the transition to `running` can be performed.
+Coincidentally, this is also the behavior expected when the endpoint is in state
+`error`, given the controller will need to wait for resources to be healthy
+before being able to move the endpoint to state `running`.
+
+In turn, handling `stopping` means waiting for those same resources to be
+removed.
+
+
+#### Required changes
+
+Aside from what has been discussed previously, we believe we need to add two new
+options as arguments to `longhorn-manager`: `--object-endpoint-image`, and
+`--object-endpoint-ui-image`, both expecting their corresponding image names.
+These will be essential for us to be able to spin up the pods for the object
+endpoints being deployed.
+
+Additionally, we will require to add the new `ObjectEndpointController` to
+`StartControllers()`, in `controller/controller_manager.go`.
+
+A new informer will need to be created as well, `ObjectEndpointInformer`, adding
+it to the `DataStore`, so we can listen for `ObjectEndpoint` resources, which we
+will critically need to in the `ObjectEndpointController`.
+
+Finally, we expect to add `s3gw` images as dependencies to be downloaded by
+the Longhorn chart.
+
+Further changes may be needed as development evolves.
+
+### Backup and Restore
+
+Given the data being kept by an Object Endpoint is stored in a Longhorn volume,
+we rely on Longhorn's backup and restore capabilities.
+
+However, an Object Endpoint in this context is more than just the data held by a
+given volume: there's meta state that needs to be backed up and restored, in the
+form of Secrets, endpoint names and their associated volume, etc.
+
+At this point in time we don't yet have a solution for this, but we believe this
+should rely on whatever mechanisms Longhorn has to backup and restore its own
+state. Insights are greatly appreciated.
+
+### Test plan
+
+It is not clear at this moment how this can be tested, much due to lack of
+knowledge on how Longhorn testing works. Help on this topic would be much
+appreciated.
+
+### Upgrade strategy
+
+Upgrading to this enhancement should be painless. Once this feature is available
+in Longhorn, the user should be able to create new object endpoints without much
+else to do.
+
+At this stage it is not clear how upgrades between Longhorn versions will
+happen. We expect to be able to simply restart existing pods using new images.
+
+### Versioning
+
+Including the `s3gw` containers in the Longhorn chart means that, for a specific
+Longhorn version, only a specific `s3gw` version is expected to have been tested
+and be in working condition. We don't make assumptions as to whether other
+`s3gw` versions would correctly function.
+
+An upgrade to `s3gw` will require an upgrade to Longhorn.
--- a/enhancements/20230430-object-endpoint.md
+++ b/enhancements/20230430-object-endpoint.md
@@ -45,25 +45,62 @@ Currently, Longhorn does not support object storage. Should the user want to use
 their Longhorn cluster for object storage, they have to rely on third-party
 applications.
 
-Instead we propose to enhance the user experience by allowing a Longhorn volume
+Instead, we propose to enhance the user experience by allowing a Longhorn volume
 to be presented to the user as an object storage endpoint, without having the
 user to install additional dependencies or manage different applications.
 
 ### User Experience In Detail
 
-* A new page view exists specifically for "Object Endpoints"
-    * Within this page there is a "Create" button, and a list of existing
-      Object Endpoints.
-* The user clicks on "Create"
-* A modal dialog is shown, with the various Object Endpoint related fields
-* User specifies the endpoint name
-* User specifies their username and password combination for the administrator
-  user
-    * This can, potentially, be randomly generated as well.
-* For publicly accessible endpoints, the user must specify a domain name to be
-  used
-* The user must provide SSL certificates to be used by the endpoint
-* Then the user clicks "Ok"
+* A new page view exists specifically for *Object Storage* with the following specifications that match the existing UI behaviour:
+    * There is a *Create Object Store* to create new *Object Store*
+    * A table that lists the existing Object Endpoints. This table contains a "Delete" action button and a search field on the right side in the table header section.
+
+#### Creating a new *Object Store*
+
+* The user clicks on *Create*
+* A modal dialog is shown, with the various Object Endpoint related fields:
+  * An endpoint name.
+  * The size of the volume to be created. Defaults to `1 GiB`.
+  * The access and secret key for the administrator user. This can, potentially, be randomly generated as well.
+  * The number of replicas. Defaults to `3`.
+  * Replica Soft Anti Affinity. Defaults to global settings.
+  * Replica Zone Soft Anti Affinity. Defaults to global settings.
+  * Replica Disk Soft Anti Affinity. Defaults to global settings.
+  * Disk Selector
+  * Node Selector
+  * Data Locality. Defaults to `Disabled`.
+  * From Backup
+  * Stale Replica Timeout
+  * Recurring Job Selector
+  * Replica Auto Balance. Defaults to global settings.
+  * Revision Counter Disabled. Defaults to `False`.
+  * Unmap Mark Snap Chain Removed
+  * Backend Store Driver
+  * Target State
+  * Name of the s3gw container image.
+  * Name of the s3gw-ui container image.
+* Finally, the user clicks *Ok* to deploy the *Object Store*.
+
+#### Deleting an *Object Store*
+
+The *Object Stores* to be deleted, one or several, can be checked in the table of existing *Object Stores*.
+This is only possible for *Object Stores* whose status is not `Terminating`. After pressing the
+*Delete* button a confirmation dialog is displayed to the user before the deletion is executed.
+
+#### Administration of buckets and objects
+
+The administration of the buckets and objects is done via the standalone *s3gw*
+UI. To access that UI easily there will be a menu entry `Administration` in the
+action dropdown of each *Object Store* table row.
+
+#### Theming of standalone *s3gw* UI
+
+The standalone *s3gw* UI will use a Longhorn theme to match the look of the
+Longhorn UI.
+
+```bash
+npm run build:prod:longhorn
+```
 
 ### API changes
 


enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,181 @@
...
+        Name: <string>
+    },
+    Spec: longhorn.ObjectEndpointSpec{
+        Image: <s3gw-image-name>,
+        ImageUI: <s3gw-ui-image-name>,
+        Credentials: longhorn.ObjectEndpointCredentials {
+            AccessKey: <string>,
+            SecretKey: <string>,
+        },
+        PublicDomain: <string>,
```
l-mb, line 128: Can this be empty if we do not want to expose the endpoint externally? Perhaps directly something that more closely aligns with an `ingress` CRD (which would then likely include the SSL/TLS setup as well)?

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,181 @@
...
+
+```golang
+longhorn.ObjectEndpoint{
+    ObjectMeta: metav1.ObjectMeta{
+        Name: <string>
+    },
+    Spec: longhorn.ObjectEndpointSpec{
+        Image: <s3gw-image-name>,
+        ImageUI: <s3gw-ui-image-name>,
+        Credentials: longhorn.ObjectEndpointCredentials {
```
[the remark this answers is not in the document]
jecluis: Definitely. At this point I just know of no better way to both convey that we need them, nor on how to keep them. Suggestions welcome! :)

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,181 @@
...
+
+We believe we will need a new Custom Resource Definition, `ObjectEndpoint`,
+representing an Object Endpoint deployment, containing information that will be
+required to deploy an `s3gw` endpoint, as well as the `s3gw-ui` administration
+interface.
+
+Given we require a volume to be created before deploying an `s3gw` instance, we
+will need to first trigger volume creation ourselves, meaning creating a new
+`longhorn.VolumeSpec` using the parameters we believe to be necessary for proper
+`s3gw` execution, and calling on `VolumeManager.Create()` for volume creation.
```
l-mb, line 97: As part of these parameters, I think we may want to introduce a flag to the backend that prevents this volume from being managed independently via the LH UI (by default at least) - plus, this would enable the UI to also be able to understand that it might want to show endpoint-specific metrics.

`claimedBy` to reference the endpoint?

enhancements/20230430-object-endpoint.md
jecluis: That's a good question. We'll have to figure that out, as I don't have a good answer for this now. @m-ildefons
 and @giubacc , do you have opinions on this? :)

enhancements/20230430-object-endpoint.md
jecluis: I didn't want to specify new fields in `longhorn.Volume` because I don't know the impacts of such a thing, but that's something I'd be interested in knowing whether it is feasible. @PhanLe1010 can you please chime in here?

Also, maybe labels will do the trick, instead of adding a new field?

enhancements/20230430-object-endpoint.md
giubacc: More on this topic: my previous comment is probably more adherent with an Helm installation and could not apply with the LEP CRD implementation.
Is there btw, the possibility that the LEP CRD provisioning an s3gw installation for a LH volume actually instantiates an Helm template via the helm API invoked by the Go code of the controller.
In this case I don't see contraindications for our existing chart templates to be used by the controller too (in the hypothesis the controller goes in the Helm API usage direction); otherwise I would say this is even desirable.

Another question that comes now to my mind is: are we going in the direction to make a provisioned S3 endpoint for a LH volume as _editable_?
Or are we saying that it would be _immutable_?
In the latter case, this would mean that an user should first delete and then recreate the endpoint in the case she would change something.

innobead:
@shuo-wu @PhanLe1010 review this after 1.5.0.

l-mb:
> @shuo-wu @PhanLe1010 review this after 1.5.0.

Hi @innobead, would it be possible to give the team a casual review of the approach already? If so, we could get started on a tentative/draft PR of the actual controller already, which might drive further discussion without waiting a few weeks. We just don't want to head down the entirely wrong way :)

innobead:
@l-mb , NP. I think members will try to review it when available. I have reviewed it in general, and it should be fine in general but some detailed needs to craft over time. Having a controller is correct as described in the PRD.

enhancements/20230430-object-endpoint.md
l-mb: @giubacc I think we'd want it to be editable (as far as that's possible). Recreating an endpoint via delete/create would be very expensive with data migration etc.

If there are parameters/attributes that are indeed immutable after creation, those should be clearly flagged.

enhancements/20230430-object-endpoint.md
votdev: @PhanLe1010 ^^^^^^^^^

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,181 @@
...
+pods.
+
+#### Custom Resource Definition
+
+It is not clear at this stage what the CRD should look like in detail, but
+following the implementation for the existing `ShareManager`, something similar
+would likely be desired.
+
+```golang
+longhorn.ObjectEndpoint{
```
m-ildefons, line 117: Looking at other CRDs (e.g. longhorn-manager/k8s/pkg/apis/longhorn/v1beta2/sharemanager.go) in the longhorn manager repo, it might be necessary to also save a [status](https://book-v1.book.kubebuilder.io/basics/status_subresource.html) in this CRD as well.

E.g.:
```golang
longhorn.ObjectEndpoint {
  ObjectMeta metav1.ObjectMeta {
    Name <string>
  },
  Spec: longhorn.ObjectEndpointSpec {
    ... (as below)
  },
  Status longhorn.ObjectEndpointStatus {
    State ObjectEndpointState,
    Endpoint <string>,
  }
}
```

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,181 @@
...
+    ObjectMeta: metav1.ObjectMeta{
+        Name: <string>
+    },
+    Spec: longhorn.ObjectEndpointSpec{
+        Image: <s3gw-image-name>,
```
m-ildefons, line 122: I have my doubts that it'll be necessary to specify this, when the s3gw image to use is already specified as command line option to the longhorn-manager.
Of course having this on a per-volume basis would allow a user to upgrade the s3gw without a longhorn release or run different version of the s3gw, but I'm not sure that use case is all that valuable.

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,181 @@
...
+        ImageUI: <s3gw-ui-image-name>,
+        Credentials: longhorn.ObjectEndpointCredentials {
+            AccessKey: <string>,
+            SecretKey: <string>,
+        },
+        PublicDomain: <string>,
+        SSLCertificate: <base64-string>,
+        Volume: <string>,
```
m-ildefons, line 130: Rather than specify a (Longhorn) volume here, I think it might be better to just specify a storage class and a size, since we don't want to re-use existing volumes here anyways.
Assuming that no suitable volume exists, a new one will be generated, or if one does already exist it will be bound. Either way, this binding will be recorded in the status property. This is much more akin to the way PVCs and PVs work together.

giubacc: Maybe I'm misunderstanding, but in our current vision, a LH volume has to exist before the user decides to attach to it an s3 endpoint (?) .
From the LH gui, I'd expect you can see a LH volume and clicking on that you can add an s3 endpoint, not the contrary.
Say that you can specify only an s3 endpoint and from that, the system takes care of provisioning all the needed resources is, IMO, a step further (not that is not good, it is conceptually different).

giubacc: I'm not against this idea; I actually think it is a good feature, I only think this as a one step further in respect to what I think should be an initial implementation.
If we want to proceed incrementally, I suggest to firstly assume that we have an already allocated LH volume and from that the user can instantiate an s3 endpoint.
Once done this, in a phase-2 we can implement the other more complex scenario where the user can do all in one single step.

jecluis: There is already exploratory work towards the proposed approach. We will ascertain what is most feasible as we go. For now, the plan is as I explained above.

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,181 @@
...
+* The user clicks on "Create Object Endpoint";
+* A modal dialog is shown, with the various Object Endpoint/s3gw related fields;
+* User specifies the endpoint name;
+* User specifies their username and password combination for the administrator
+  user;
+    * This can, potentially, be randomly generated as well.
+* For publicly accessible endpoints, the user must specify a domain name to be
+  used;
+* The user must provide SSL certificates to be used by the endpoint.
+* Then the user clicks "Ok".
```
PhanLe1010, line 65: Is it correct that all of the endpoints of information here may potentially be recorded into a CRD (like `ObjectEndpoint.longhorn.io`)? 

If so, would it be better to have a separate Longhorn UI page, e.g. `/endpoints`, for all endpoints. In this page, users can create enpoints by specifying all parameters. One of the parameter will be longhorn volume name to be used as backend storage for the endpoint. WDYT?

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,181 @@
...
+
+
+## Design
+
+### Implementation Overview
+
+Integrating Longhorn with `s3gw` will require the creation of mechanisms that
+allow us to 1) describe an object endpoint; 2) create a volume to be used for
+storage; 3) deploy an `s3gw` pod consuming the created volume; and 4) deploy an
+`s3gw-ui` pod for management and administration.
```
PhanLe1010, line 87: From what I am reading on S3GW documentation, looks like one s3gw-ui instance currently can only manage one `s3gw` backend (which is equivalent to one `ObjectEndpoint` CRD in this LEP). Is it correct?

I think in general, users may have multiple EndPoints CRs in one cluster. What would we do in this case:
1. Do we also create one `s3gw-ui` instance for each endpoint? This might be too overkill
2. Will we have one `s2gw-ui` instance that can manage all endpoints like [this s3gw design document](https://s3gw-docs.readthedocs.io/en/latest/design/0001-ui-backend/#:~:text=down%20the%20line.-,Multiple%20s3gw%20instances%20managed%20by%20the%20same%20s3gw%2Dui,single%20s3gw%2Dui%20able%20to%20centrally%20manage%20multiple%20s3gw%20service%20instances.,-Considered%20Options)

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,181 @@
...
+
+We believe we will need a new Custom Resource Definition, `ObjectEndpoint`,
+representing an Object Endpoint deployment, containing information that will be
+required to deploy an `s3gw` endpoint, as well as the `s3gw-ui` administration
+interface.
```
PhanLe1010, line 92: The `ObjectEndpoint` design sounds reasonable to me

enhancements/20230430-object-endpoint.md
PhanLe1010: Why would the Longhorn volume need to be managed independently by Longhorn UI? After all, it is still a Longhorn volume that have all features like snapshots/backup like other Longhorn volumes, so Longhorn UI can/should still be able to manage it, right? 

For showing endpoint-specific metrics associated with the Longhorn volume, may be Longhorn UI just need to detect that there is an ObjectEndpoint CR that is linked to this volume and show a like to the ObjectEndpoint UI page. WDYT?

enhancements/20230430-object-endpoint.md
PhanLe1010: Yes, a `Status` seems to be needed if we want to track the status of this endpoint

PhanLe1010 (commented):
I have left some questions.

I am still learning S3GW and COSI, so please forgive me if my questions don't make sense or if they are naive

enhancements/20230430-object-endpoint.md
m-ildefons: This is a point of discussion right now within our team as well. If you look at the mockup for the UI (https://github.com/aquarist-labs/s3gw/issues/512) it proposes a workflow that requires a user to create a volume first before being able to expose it. However this has rough edges, because s3gw requires exclusive access to its volume. I'm personally in favor of a solution where object endpoints are handled independently from volumes and Longhorn will manage the creation of volumes in the background on its own, much like you suggest here, because then the user will only have to care about the amount (as in Gigabytes) of objects and the storage class.

enhancements/20230430-object-endpoint.md
m-ildefons: Currently each `s3gw-ui` instance can only service one `s3gw` backend. That is correct.

enhancements/20230430-object-endpoint.md
jecluis: Lets keep in mind that the mock up pointed to by @m-ildefons was initially created taking into account a parallel discussion that is not reflected in this LEP, so should not be considered.

The proposal of the LEP, and the path being explored, focuses on having the Longhorn Volume being magically created (i.e., without the user ever having to care about it) when creating a new Object Endpoint.

As to what @PhanLe1010 mentioned, I also believe having a specific page for all object endpoints may make sense. This way we can add more relevant information than we would otherwise via the Volumes interface. Besides, the volumes interface will nonetheless show volumes that have been claimed by the object endpoints, but that's a lower level of abstraction, and independent from the object endpoints interface.

enhancements/20230430-object-endpoint.md
jecluis: @PhanLe1010 that is definitely something being considered. We could technically use one single `s3gw-ui` (once the `s3gw-ui` backend is in place) to manage multiple `s3gw` endpoints. However, given that is not yet fully implemented and tested, nor is the workflow fully determined, I'd suggest we assume one `s3gw-ui` instance per `s3gw` endpoint for now.

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,181 @@
+# Title
+
+Object Support for Longhorn
```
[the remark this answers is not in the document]
jecluis: Will be included in the next revision of this document.

m-ildefons:
For the current POC implementation I've been playing around with different settings that may or may not be useful in the `ObjectEndpoint` CRD. This led me to the conclusion that for the features described in this LEP, the following form of a CRD would be best suited:
```
---
kind: ObjectEndpoint
metadata:
  name: test-endpoint
  namespace: longhorn-system
spec:
  storageClass: longhorn
  credentials:
    accessKey: foo
    secretKey: bar
  size: 5Gi
```
The differences to the proposed CRD are:

- The `Image` field is missing. This field is unnecessary as the Longhorn manager already knows which image it should use to deploy the s3gw, as that is given as command line option.

- `PublicDomain` and `SSLCertificate` are missing because I haven't implemented the features making use of these settings yet.

- `Volume` field has been replaced by the optional `storageClass` and `size` fields. These two fields describe the desired properties of the backing volume sufficiently. The controller can then automatically deploy the resources necessary to realize this. This makes it easier on a user to deploy an object endpoint, since they don't have to deploy a volume first and take care of the association with the object endpoint manually.

giubacc:
> For the current POC implementation I've been playing around with different settings that may or may not be useful in the `ObjectEndpoint` CRD. This led me to the conclusion that for the features described in this LEP, the following form of a CRD would be best suited:
> 
> ```
> ---
> apiVersion: longhorn.io/v1beta2
> kind: ObjectEndpoint
> metadata:
>   name: test-endpoint
>   namespace: longhorn-system
> spec:
>   storageClass: longhorn
>   credentials:
>     accessKey: foo
>     secretKey: bar
>   size: 5Gi
> ```
> 
> The differences to the proposed CRD are:
> 
> * The `Image` field is missing. This field is unnecessary as the Longhorn manager already knows which image it should use to deploy the s3gw, as that is given as command line option.
> * `PublicDomain` and `SSLCertificate` are missing because I haven't implemented the features making use of these settings yet.
> * `Volume` field has been replaced by the optional `storageClass` and `size` fields. These two fields describe the desired properties of the backing volume sufficiently. The controller can then automatically deploy the resources necessary to realize this. This makes it easier on a user to deploy an object endpoint, since they don't have to deploy a volume first and take care of the association with the object endpoint manually.

This sounds nicely. :+1: 

Some random questions:

Could it make sense to add Longhorn specific settings for controlling the volume properties?
Or here are we saying that we rely completely on the properties defined in the storage-class?
The use case here seems to suggest that is responsibility of the user to define his storage classes. Could this be automated someway when s3gw is used with LH (that is the privileged scenario)? Or is this the intended design?

m-ildefons:
> Could it make sense to add Longhorn specific settings for controlling the volume properties? Or here are we saying that we rely completely on the properties defined in the storage-class? The use case here seems to suggest that is responsibility of the user to define his storage classes. Could this be automated someway when s3gw is used with LH (that is the privileged scenario)? Or is this the intended design?

Both variants, expecting a user to pre-create storage classes or offering all settings, have their pros and cons. Expecting the user to create storage classes certainly is much simpler.
One noteworthy thing here is, that when creating a volume through the Longhorn UI, the users is asked for various settings directly, which completely circumvents the storage classes.
I'm not sure if we want to copy this behavior, but it would be a reasonable choice to be consistent with the volume creation dialogue.

We're also not necessarily confined to either/or exclusively. We could create an abstraction that can handle both, it would just be more work, e.g.:
```golang
longhorn.ObjectEndpoint{
  ObjectMeta metav1.ObjectMeta {
    Name: <string>
  },
  Spec: longhorn.ObjectEndpointSpec{
    Storage: <longhorn.StorageSettings>
  },
}

longhorn.StorageSettings{
  Size: <resource.Quantity>
  
  //+optional
  StorageClass: <string>
  
  //+optional
  VolumeSettings: <longhorn.VolumeSettings>
},

longhorn.VolumeSettings{
  numberOfReplicas: <string>
  staleReplicaTimeout: <string>
  fsType: <string>
  allowVolumeExpansion: <bool>
  [...]
},
```
This way both a storage class or volume settings could be specified directly. It's then the task of the controller to ensure that a) no conflicting settings are given (i.e. either the storage class _or_ the volume settings) and b) to provision the correct resources and connect them up.
This mechanism is certainly the most involved on the controller side, but it's within the realm of technical possibilities.

A strong argument for supporting volume settings directly instead of through a storage class is, that as far as I can tell the Longhonr UI does not have a dialogue for creating storage classes. It's up to the user to create them through the use of YAML files and `kubectl`, which is a bit of an ugly spot on the user experience IMO, especially when we expect the user to do that to make use of all the settings for the object endpoint. But a dialogue for creating storage classes from the Longhorn UI isn't in the scope of this LEP.

jecluis:
Why would we need to have the user defining the storage class? Can't we create one if there isn't one ourselves? Wouldn't that also give us more control over how the data is kept?

m-ildefons:
> Why would we need to have the user defining the storage class? Can't we create one if there isn't one ourselves? Wouldn't that also give us more control over how the data is kept?

There's a default storage class that we can fall back onto, in case the user hasn't deleted it.

Unrelated to that, another question: Should the ObjectEndpoint resource be namespace or cluster scoped?
The argument for being cluster scoped would be that this would allow us to keep the deployments, services, PVCs confined to the `longhorn-system` namespace, hiding them from the user.
Making it namespace scoped has advantages for users who utilize (cluster-)roles to limit permissions of actors on the cluster. I'm not sure how important that is.

giubacc:
> The argument for being cluster scoped would be that this would allow us to keep the deployments, services, PVCs confined to the `longhorn-system` namespace, hiding them from the user.

I haven't a strong opinion pro or cons one approach.
I would prefer technical s3gw resources to not pollute the user's namespace ... but being more pro or cons one way could become more obvious in practice.

enhancements/20230430-object-endpoint.md
votdev: > After thinking about this, and from feedback from several parties, IMO the best approach might actually be
> 
>     1. Have the volumes being used by the individual object endpoints be shown in the volumes, so, as @PhanLe1010 mentioned, we could technically perform any operations on it as with any other volume;
> 
>     2. Have an "object endpoints" page/section/whatever specific for the various object endpoints, where we can then show object endpoint-specific metrics, data, whatever.
> 
> 
> As I see it, these two live on different layers of abstraction: Volumes are the lower level bits the user should not care about when deploying an object endpoint, but that they may still care about for snapshotting or whatever. On the upper layer, we have the object endpoint, which doesn't particularly care which volume it has as long as a volume claim is associated with its deployment, and for which we can provide specific metrics that are relevant in an object world.
> 

I like this proposal because it makes everything transparent for the users. They see the volumes that are created and used by s3gw normally in the LH UI and can perform all actions that can be done with a volume.

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,181 @@
...
+appreciated.
+
+### Upgrade strategy
+
+Upgrading to this enhancement should be painless. Once this feature is available
+in Longhorn, the user should be able to create new object endpoints without much
+else to do.
+
+At this stage it is not clear how upgrades between Longhorn versions will
+happen. We expect to be able to simply restart existing pods using new images.
```
votdev, line 171: Assuming the on-disk format for the metadata store has not been changed.

jecluis: We can easily do in-place metadata upgrades now. That is no longer an issue.

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,181 @@
...
+* Integration of s3gw UI for administration and management purposes. Such an
+  Enhancement Proposal should be a standalone LEP by its own right.
+* Providing Object endpoints for multiple volumes. In this proposal we limit one
+  object endpoint per Longhorn volume (see longhorn/longhorn#5444).
+* Multiple Object endpoints for a single volume, either in Read-Write Many, or
+  as active/passive for HA failover. While we believe it to be of the utmost
+  importance, we are considering it future work that should be addressed in its
+  own LEP.
+* Specify how COSI can be implemented for Longhorn and s3gw. This too should be
+  addressed in a specific LEP.
```
PhanLe1010, line 37: While these are non-goal for this LEP. I feel that we should keep this one in mind while designing this PR to make it easier for future work

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,263 @@
...
+When the Object Endpoint is created, the required resources will need to be
+created. These include,
+
+* `Service` associated with the pods being deployed
+* Various `Secret` and `ConfigMap` needed by `s3gw` and `s3gw-ui`
+* The associated `Deployment`
+* A `Persistent Volume Claim` required for `s3gw` storage.
+
+During resource creation we will also need to explicitly create a
+`Persitent Volume` that will be bound to the `Persistent Volume Claim` mentioned
```
m-ildefons, line 193: We will also need to create the Longhorn volume explicitly.

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,263 @@
...
+```
+
+The endpoint `Spec`, as follows, contains the information that we consider most
+relevant to the endpoint at this point. More information may be added should the
+need arise.
+
+```golang
+type ObjectEndpointSpec struct {
+    Credentials     ObjectEndpointCredentials
+    StorageClass    string
```
m-ildefons, line 130: The `StorageClass` property is ignored[1], since we manage the volume explicitly ourselves in the controller. Perhaps it should be removed or replaced by other properties that are more meaningful in this context.

[1]: 
When explicitly creating a PV and PVC together, the storage class property must match or be an empty string, but all other aspects of the storage class are ignored.
https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reserving-a-persistentvolume

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,263 @@
...
+`Persitent Volume` that will be bound to the `Persistent Volume Claim` mentioned
+above. We need to do this so we can opinionate on the file system being used; in
+this case, XFS, which `s3gw` requires for performance reasons.
+
+While handling `starting`, the resources required to have already been created
+through the Kubernetes API, but the controller still has to wait for them to be
+ready and healthy before the transition to `running` can be performed.
+Coincidentally, this is also the behavior expected when the endpoint is in state
+`error`, given the controller will need to wait for resources to be healthy
+before being able to move the endpoint to state `running`.
```
m-ildefons, line 202: Is this a requirement or is this documenting what the current iteration of the controller is doing?

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,263 @@
...
+
+### Implementation Overview
+
+Integrating Longhorn with s3gw will require the creation of mechanisms that
+allow us to 1) describe an object endpoint; 2) deploy an `s3gw` pod consuming
+a volume; and 3) deploy an `s3gw-ui` pod for management and administration. For
+the purposes of this LEP, there will always be one single `s3gw-ui` pod
+associated with an `s3gw` service, hence one `s3gw-ui` pod per Object Endpoint.
+We do not exclude eventually allowing one single `s3gw-ui` pod being associated
+with multiple Object Endpoints, but that should be considered as future work.
```
m-ildefons, line 91: This LEP now _includes_ the UI integration as well?

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,263 @@
...
+
+At this point in time we don't yet have a solution for this, but we believe this
+should rely on whatever mechanisms Longhorn has to backup and restore its own
+state. Insights are greatly appreciated.
+
+### Test plan
+
+It is not clear at this moment how this can be tested, much due to lack of
+knowledge on how Longhorn testing works. Help on this topic would be much
+appreciated.
```
m-ildefons, line 245: There is an integration test suite in longhorn/longhorn-tests which we can utilize to test the controller's behavior from a K8s client point of view.
There are also unit tests in the controller it self, which can be used to test individual functions against a mocked/faked K8s API.

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,263 @@
...
+    * Within this page there is a "Create" button, and a list of existing
+      Object Endpoints.
+* The user clicks on "Create"
+* A modal dialog is shown, with the various Object Endpoint related fields
+* User specifies the endpoint name
+* User specifies their username and password combination for the administrator
+  user
+    * This can, potentially, be randomly generated as well.
+* For publicly accessible endpoints, the user must specify a domain name to be
+  used
```
m-ildefons, line 64: Is being cluster-externally accessible a requirement for this LEP?

enhancements/20230430-object-endpoint.md
jecluis: Integration, no. That presumes the administration and management UI becomes part of the Longhorn UI. That has been stated from the beginning that is not part of this LEP.

`s3gw-ui` support on the other hand has always been part of this LEP.

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,263 @@
...
+Given the data being kept by an Object Endpoint is stored in a Longhorn volume,
+we rely on Longhorn's backup and restore capabilities.
+
+However, an Object Endpoint in this context is more than just the data held by a
+given volume: there's meta state that needs to be backed up and restored, in the
+form of Secrets, endpoint names and their associated volume, etc.
+
+At this point in time we don't yet have a solution for this, but we believe this
+should rely on whatever mechanisms Longhorn has to backup and restore its own
+state. Insights are greatly appreciated.
```
PhanLe1010, line 239: I am working on designing this part

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,263 @@
...
+endpoint types (not just object) are eventually added.
+
+
+## Design
+
+### Implementation Overview
+
+Integrating Longhorn with s3gw will require the creation of mechanisms that
+allow us to 1) describe an object endpoint; 2) deploy an `s3gw` pod consuming
+a volume; and 3) deploy an `s3gw-ui` pod for management and administration. For
```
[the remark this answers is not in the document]
jecluis: I believe this has been discussed previously. There are on-going efforts to make the `s3gw-ui` not depend on a specific ingress to handle CORS. This LEP accounts for those efforts.

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,263 @@
...
+      Object Endpoints.
+* The user clicks on "Create"
+* A modal dialog is shown, with the various Object Endpoint related fields
+* User specifies the endpoint name
+* User specifies their username and password combination for the administrator
+  user
+    * This can, potentially, be randomly generated as well.
+* For publicly accessible endpoints, the user must specify a domain name to be
+  used
+* The user must provide SSL certificates to be used by the endpoint
```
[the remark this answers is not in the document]
jecluis: In-cluster deployments may also require SSL -- we know of at least such a user. We can drop this requirement and have the user deploy SSL themselves, but natively supporting this seems the path of least resistance. Is there anything you are worried about?

enhancements/20230430-object-endpoint.md
jecluis: Potentially. I don't see why not. It shouldn't be adding a lot more complexity, I think, and requiring another LEP to make stuff externally accessible seems a bit of overkill to me.

l-mb: It needs to be - otherwise we can't support the backup-from-another-LH-cluster-to-s3 use case.

enhancements/20230430-object-endpoint.md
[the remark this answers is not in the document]
l-mb: Can't we document / automate a `cert-manager` integration? We need SSL for the UIs anyway too.

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,263 @@
...
+controller will be listening for new resources of type `objectendpoint`, and
+will create the necessary bits for proper deployment, including services and
+pods.
+
+#### Custom Resource Definition
+
+We define a new Object Endpoint Custom Resource Definition, as follows:
+
+```golang
+type ObjectEndpoint struct {
```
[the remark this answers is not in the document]
l-mb: Amazon calls theirs "S3 endpoints". And, well, they are the API endpoints associated with the S3 API too, and we can have multiple ones of them. I think either phrasing works and we shouldn't get too invested in this argument :-)

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,263 @@
...
+
+* `Service` associated with the pods being deployed
+* Various `Secret` and `ConfigMap` needed by `s3gw` and `s3gw-ui`
+* The associated `Deployment`
+* A `Persistent Volume Claim` required for `s3gw` storage.
+
+During resource creation we will also need to explicitly create a
+`Persitent Volume` that will be bound to the `Persistent Volume Claim` mentioned
+above. We need to do this so we can opinionate on the file system being used; in
+this case, XFS, which `s3gw` requires for performance reasons.
```
l-mb, line 195: (As a note, we need the reflink support so we can effectively merge multipart uploads etc, which ext4 can't do.)

enhancements/20230430-object-endpoint.md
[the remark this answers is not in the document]
m-ildefons: I agree with Phan here. Using _ObjectStore_ would leave _endpoint_ free to describe TLS and ingress/domain configuration, allowing perhaps the use of multiple endpoints for one store, e.g. like so:
```
kind: ObjectStore
metadata:
[...]
spec:
  credentials:
    accessKey: ...
    secretKey: ...
  storage:
    size: 100Gi
    numberOfReplicas: 3
    diskSelector:
      - ssd
  endpoints:
    - name: public
      domainName: foobar.example.com
      tls: ...
    - name: private
      domainName: foobar.longhorn-system.svc.cluster.local
```

enhancements/20230430-object-endpoint.md
PhanLe1010: Agree to remove the `StorageClass` property. We still need to hardcode a fixed storageclass like `longhorn-objectstore` (without any parameter) so that features like PVC expansion can work. This is important because user will need to be able to resize the PVC. Ref the POC PR https://github.com/longhorn/longhorn-manager/pull/2135/files#diff-713be8c3744f9bf701a213640c0c1d0d302c1231aa21b291f7d35bdcdbd30068R433-R434

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,263 @@
...
+* Then the user clicks "Ok"
+
+### API changes
+
+The API will need a new endpoint to create an object endpoint, as well as
+listing, updating, and deleting them. We believe it's not reasonable to reuse
+the existing `/v1/volumes` API endpoints, given they are semantically distinct
+from what we are trying to achieve.
+
+We thus propose the creation of a `/v1/endpoint/object` API endpoint. This route
```
votdev, line 75: This section does not match the reality of the PR https://github.com/longhorn/longhorn-manager/pull/2136/files#diff-cb740c237beef9090d8b3a15ca8b76aac83a3af2d52deadc382b4309916a715c and needs to be adapted.

votdev (changes requested):

enhancements/20230430-object-endpoint.md
```diff
@@ -0,0 +1,285 @@
...
+### User Experience In Detail
+
+* A new page view exists specifically for *Object Storage* with the following specifications that match the existing UI behaviour:
+    * There is a *Create Object Store* to create new *Object Store*
+    * A table that lists the existing Object Endpoints. This table contains a "Delete" action button and a search field on the right side in the table header section.
+
+#### Creating a new "Object Store"
+
+* The user clicks on *Create*
+* A modal dialog is shown, with the various Object Endpoint related fields:
```
votdev, line 61: These fields are taken from https://github.com/longhorn/longhorn-manager/pull/2136/files#diff-c75a796f2c69c35ee5ab6293f23db5c8d5af9a241f4f2345e96b1fc41d72ddceR55

Is it necessary to request all that fields from the user or can we hardcode some of them?

innobead:
Closing first, and this will be revisited in 1.7.

PR #5858, opened 2023-05-04
Disable debug logging

pujan14:
Debug logging should not be enabled in production systems


innobead:
cc @ChanYiLin (we need to revisit all logs in longhorn-manager specifically, especially controllers)

pujan14:
@ChanYiLin done!

ChanYiLin:
LGTM, thanks for your contribution :+1:
cc @innobead

ChanYiLin:
> @ChanYiLin Should we remove the log level from driver deployer which doesn't have controllers running, so it will not sync the log level from the setting.

After discussion,
driver deployer is one time process, we have to review the log to change all debug level to info or warning after

ChanYiLin:
Hi @pujan14 can you help me to solve DCO issue for the last commit? Thanks!

innobead:
Thanks @pujan14 for your contribution. 

@ChanYiLin develops the log setting feature for the following 1.5, so let's use that instead. https://github.com/longhorn/longhorn/issues/5888

Issue #5738, opened 2023-04-12
[FEATURE] Environment check script for Longhorn with SPDK


derekbit:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

The script will check
1. OS distro
2. Kernel module
3. Dependent packages
4. SPDK

## Describe the solution you'd like

A clear and concise description of what you want to happen

innobead:
Assigned @yangchiu as well to make this support in the SLE micro setup provisioning. 

cc @longhorn/qa

chriscchien:
Verified **pass** in longhorn master(longhorn `15a73c`)

In 3 nodes cluster, setup environment with [instructions](https://even-pecorino-1bc.notion.site/Quick-Start-for-SPDK-Volumes-a189b3f81e23495d973e3a14b7635667#bef3578241b0409ea319c68ccb7cdc4d) and run environment check script, did not encounter problem in below distro

- Ubuntu AMD64
- Ubuntu ARM64
- SLES AMD64
- SLES ARM64

AMD64 output
```
No VM guests are running outdated hypervisor (qemu) binaries on this host.
root@ip-172-31-21-42:/home/ubuntu# bash -c "$(curl -sfL https://raw.githubusercontent.com/longhorn/longhorn/master/scripts/environment_check.sh)" -s -s
[INFO]  Required dependencies 'kubectl jq mktemp' are installed.
[INFO]  All nodes have unique hostnames.
[INFO]  Waiting for longhorn-environment-check pods to become ready (0/3)...
[INFO]  All longhorn-environment-check pods are ready (3/3).
[INFO]  MountPropagation is enabled
[INFO]  Checking iscsid...
[INFO]  Checking multipathd...
[INFO]  Checking packages...
[INFO]  Checking nfs client...
[INFO]  Checking x86-64 SSE4.2 instruction set...
[INFO]  Checking nvme-cli...
[INFO]  Checking kernel module nvme_tcp...
[INFO]  Checking kernel module uio_pci_generic...
[INFO]  Checking hugepage...
[INFO]  Cleaning up longhorn-environment-check pods...
[INFO]  Cleanup completed.
```
ARM64
```
[INFO]  Required dependencies 'kubectl jq mktemp' are installed.
[INFO]  All nodes have unique hostnames.
[INFO]  Waiting for longhorn-environment-check pods to become ready (0/3)...
[INFO]  All longhorn-environment-check pods are ready (3/3).
[INFO]  MountPropagation is enabled
[...]

Issue #5380, opened 2023-02-10
[IMPROVEMENT] Make environment check script recognize iscsid.socket enable instead of iscsid.server only  


innobead:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

Currently, in the environment_check script, it will check if iscsid.server is running or not by systemd; however, if iscsid.service is not enabled by default and not running, it doesn't mean iscsid is not working as expected. Probably it's due to it is not enabled by default, so it will be only activated by `iscsid.socket` which is enabled by default.

## Describe the solution you'd like

Have another optional check to see if `iscsid.socket` is enabled. If it's, the environment should be ready for iSCSI client operations.

innobead:
cc @longhorn/qa

jfmatth:
Agreed.  I installed Centos8-stream and just installing the iscsi-initiator-utils will not make Longhorn work.  LH will install, but won't provision volumes and fails strangely.  

Not sure how to solve on the check, but i had to enable the service and socket to make sure it was working.

oh yea, and make sure FirewallD is disabled too.

innobead:
@derekbit is helping this while working on SPDK env check.

roger-ryao:
**Verified on master-head 20230629**

- longhorn master-head https://github.com/longhorn/longhorn/commit/6b56bb2b7240aaa814b853c10e40e705cfd713a8
- longhorn-manager master-head https://github.com/longhorn/longhorn-manager/commit/9bc7e074d3b6105ab95234140ddb89b0dad9cbdd

**Result Passed**

```bash
ryao@ryao-laptop:~/Downloads/ssh_key> curl -sSfL https://raw.githubusercontent.com/longhorn/longhorn/master/scripts/environment_check.sh | bash
[INFO]  Required dependencies 'kubectl jq mktemp' are installed.
[INFO]  All nodes have unique hostnames.
[INFO]  Waiting for longhorn-environment-check pods to become ready (0/1)...
[INFO]  All longhorn-environment-check pods are ready (1/1).
[INFO]  MountPropagation is enabled
[INFO]  Checking iscsid...
[ERROR] Neither iscsid.service nor iscsid.socket is not running on ryao-master-w1-cef4a091-qtwwt
[INFO]  Checking multipathd...
[INFO]  Checking packages...
[ERROR] nfs-common is not found in ryao-master-w1-cef4a091-qtwwt.
[INFO]  Cleaning up longhorn-environment-check pods...
[...]

PR #5880, opened 2023-05-08
Environment check for SPDK support

derekbit:
Longhorn/longhorn#5738
Longhorn/longhorn#5380

### DIFF
--- a/scripts/environment_check.sh
+++ b/scripts/environment_check.sh
@@ -277,7 +277,7 @@ check_iscsid() {
     kubectl exec -t ${pod} -- nsenter --mount=/proc/1/ns/mnt -- bash -c "systemctl status --no-pager iscsid.socket" > /dev/null 2>&1
       if [ $? -ne 0 ]; then
       node=$(kubectl get ${pod} --no-headers -o=custom-columns=:.spec.nodeName)
-      error "Nither iscsid.service nor iscsid.socket is not running on ${node}"
+      error "Neither iscsid.service nor iscsid.socket is not running on ${node}"
       return 1
     fi
   fi
--- a/scripts/environment_check.sh
+++ b/scripts/environment_check.sh
@@ -73,8 +73,7 @@ error() {
 ######################################################
 # Check logics
 ######################################################
-set_packages_and_check_cmd()
-{
+set_packages_and_check_cmd() {
   case $OS in
   *"debian"* | *"ubuntu"* )
     CHECK_CMD='dpkg -l | grep -w'
@@ -105,16 +104,14 @@ set_packages_and_check_cmd()
    esac
 }
 
-detect_node_kernel_release()
-{
+detect_node_kernel_release() {
   local pod="$1"
 
   KERNEL_RELEASE=$(kubectl exec -i $pod -- nsenter --mount=/proc/1/ns/mnt -- bash -c 'uname -r')
   echo "$KERNEL_RELEASE"
 }
 
-detect_node_os()
-{
+detect_node_os() {
   local pod="$1"
 
   OS=$(kubectl exec -i $pod -- nsenter --mount=/proc/1/ns/mnt -- bash -c 'grep -E "^ID_LIKE=" /etc/os-release | cut -d= -f2')
@@ -130,13 +127,13 @@ check_local_dependencies() {
   local all_found=true
   for ((i=0; i<${#targets[@]}; i++)); do
     local target=${targets[$i]}
-    if [ "$(which $target)" == "" ]; then
+    if [ "$(which $target)" = "" ]; then
       all_found=false
       error "Not found: $target"
     fi
   done
 
-  if [ "$all_found" == "false" ]; then
+  if [ "$all_found" = "false" ]; then
     msg="Please install missing dependencies: ${targets[@]}."
     info "$msg"
     exit 2
@@ -195,7 +192,7 @@ wait_ds_ready() {
     local numberReady=$(echo $ds | jq .status.numberReady)
     local desiredNumberScheduled=$(echo $ds | jq .status.desiredNumberScheduled)
 
-    if [ "$desiredNumberScheduled" == "$numberReady" ] && [ "$desiredNumberScheduled" != "0" ]; then
+    if [ "$desiredNumberScheduled" = "$numberReady" ] && [ "$desiredNumberScheduled" != "0" ]; then
       info "All longhorn-environment-check pods are ready ($numberReady/$desiredNumberScheduled)."
       return
     fi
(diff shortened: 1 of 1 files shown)


derekbit:
Some values such as default hugepage size will be discussed later.
cc @innobead @shuo-wu @DamiaSan

scripts/environment_check.sh
```diff
@@ -205,7 +207,8 @@ wait_ds_ready() {
   done
 }
 
-check_mount_propagation() {
+check_mount_propagation()
+{
```
innobead, line 211: Make the same change to other functions for function closure consistent.

scripts/environment_check.sh
```diff
@@ -283,82 +253,211 @@ check_hostname_uniqueness() {
...
-    warn "multipathd would probably result in the Longhorn volume mount failure. Please refer to https://longhorn.io/kb/troubleshooting-volume-with-multipath for more information."
+  if [ "$all_passed" == "false" ]; then
+    return 1
   fi
 }
 
-check_iscsid() {
-  local pods=$(kubectl get pods -o name -l app=longhorn-environment-check)
-  local all_found=true
+check_iscsid()
```
innobead, line 276: Can we also tackle https://github.com/longhorn/longhorn/issues/5380 together? just need to have another else to check if `iscsid.socket` enabled or not.

scripts/environment_check.sh
```diff
@@ -283,82 +253,211 @@ check_hostname_uniqueness() {
...
 
-    for pod in ${pods}; do
-      local kernel_release=$(detect_node_kernel_release $pod)
-      if [ x"$kernel_release" == x"" ]; then
-        error "Unable to detect kernel release on node $node."
-        exit 2
-      fi
+  local kernel=$(detect_node_kernel_release ${pod})
+  if [ x"$kernel" == x"" ]; then
+    warn "Unable to detect kernel release on node ${node}"
```
innobead, line 341: Change the log to `"Failed to check NFS client installation, because unable to detect kernel release on node ${node}"`

Can you also update other error/warning to `Failed to.. ` if they are suitable.

scripts/environment_check.sh
```diff
@@ -283,82 +253,211 @@ check_hostname_uniqueness() {
...
-
-    if [ ${#nodes[@]} != 0 ]; then
-      warn ""${config}" kernel config is not enabled on nodes ${nodes[*]}."
     fi
   done
 
-  if [[ ${all_found} == false ]]; then
-    warn "NFS client kernel support, ${nfs_client_kernel_configs[*]}, is not enabled on Longhorn nodes. Please refer to https://longhorn.io/docs/1.4.0/deploy/install/#installing-nfsv4-client for more information."
+  if [ ${#enable_configs[@]} == 0 ]; then
+    error "NFS clients ${configs[*]} should be enabled at least one."
```
innobead, line 367: `configs_enabled` instead?

scripts/environment_check.sh
```diff
@@ -283,82 +253,211 @@ check_hostname_uniqueness() {
...
+    kubectl exec -t ${pod} -- nsenter --mount=/proc/1/ns/mnt -- bash -c "[ -f /boot/config-${kernel} ]" > /dev/null 2>&1
+    if [ $? != 0 ]; then
+      warn "Unable to check kernel config $config on node ${node} since /boot/config-${kernel} does not exist"
+      continue
+    fi
+
+    kubectl exec -t ${pod} -- nsenter --mount=/proc/1/ns/mnt -- bash -c "grep -E \"^# ${config} is not set\" /boot/config-${kernel}" > /dev/null 2>&1
+    if [[ $? == 0 ]]; then
+        configs_enabled["${config}"]="${config}"
+    else
```
innobead, line 356: It's not found, why added it to the `configs_enabled`?

scripts/environment_check.sh
```diff
@@ -375,10 +474,16 @@ trap cleanup EXIT
 create_ds
 wait_ds_ready
 
-check_nfs_client_kernel_support
-check_package_installed
-check_iscsid
-check_multipathd
 check_mount_propagation
+check_nodes check_iscsid
+check_nodes check_multipathd
```
innobead, line 479: nit: Can we add an info log for each function call? for example:

`Chekcing mount prepgation` when running `check_mount_propagation`
`Checking iscsid on each node` when running `check_nodes check_iscsid`

It's nice to have, but it will help the output of this script more clear.

innobead (changes requested):
In general, LGTM.

Some parts needs to update though.

scripts/environment_check.sh
derekbit: Addressed. Thank you.

scripts/environment_check.sh
derekbit: Addressed. Thank you.

scripts/environment_check.sh
derekbit: Wrong variable. Fixed. Thank you.

scripts/environment_check.sh
derekbit: Addressed. Thank you.

scripts/environment_check.sh
derekbit: Addressed. Thank you.

scripts/environment_check.sh
derekbit: Added. Thank you.

scripts/environment_check.sh
```diff
@@ -375,10 +463,17 @@ trap cleanup EXIT
 create_ds
 wait_ds_ready
 
-check_nfs_client_kernel_support
-check_package_installed
-check_iscsid
-check_multipathd
 check_mount_propagation
+check_nodes "iscsid" check_iscsid
+check_nodes "iscsid.socket" check_iscsid_socket
```
innobead, line 468: These two are tricky. At least one needs to be enabled. 

So probably can make `check_nodes` support running more than one command, then it will pass if one of them succeeds. (we can make it as or for now, since we don't have any other cases)

innobead (changes requested):
Almost there, just one reminaing comment.

scripts/environment_check.sh
derekbit: Addressed. Thank you.

scripts/environment_check.sh
```diff
@@ -283,82 +246,199 @@ check_hostname_uniqueness() {
...
 
-    if [ $? != 0 ]; then
-      all_found=false
+  kubectl exec -t ${pod} -- nsenter --mount=/proc/1/ns/mnt -- bash -c "systemctl status --no-pager iscsid.service" > /dev/null 2>&1
+  if [ $? -ne 0 ]; then
+    kubectl exec -t ${pod} -- nsenter --mount=/proc/1/ns/mnt -- bash -c "systemctl status --no-pager iscsid.socket" > /dev/null 2>&1
+      if [ $? -ne 0 ]; then
       node=$(kubectl get ${pod} --no-headers -o=custom-columns=:.spec.nodeName)
-      error "iscsid is not running on $node."
+      error "Nither iscsid.service nor iscsid.socket is not running on ${node}"
```
innobead, line 280:
```suggestion
      error "Neither iscsid.service nor iscsid.socket is not running on ${node}"
```

innobead (approved):
LGTM

Issue #5862, opened 2023-05-05
[IMPROVEMENT] Remove privilege requirement from lifecycle jobs


innobead:
Good question. It was enabled by this commit https://github.com/longhorn/longhorn/commit/eb767033390cc31c2a01ec3f93dfd2eb3ff05b1a

But I just walked through the post-upgrade and uninstall job. They are just doing operations against kubeAPI. Not sure why we need them to be privileged

_Originally posted by @PhanLe1010 in https://github.com/longhorn/longhorn/pull/5567#discussion_r1185639986_

innobead:
@mantissahz Try to fix this together with upgrade enforcement. Thanks.

cc @PhanLe1010

innobead:
@mantissahz need to check if we need to update the doc.

chriscchien:
Hi @mantissahz ,

From test [steps](https://github.com/longhorn/longhorn/issues/5862#issuecomment-1537864310), I can install/uninstall Longhorn by `kebectl`, but when I try upgrade Longhorn from v1.4.2-rc1 to master by `helm`, I got below error

```
it checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
root@ip-172-31-81-81:/home/ubuntu/k9s/longhorn# helm upgrade longhorn ./chart/ -n longhorn-system
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /etc/rancher/k3s/k3s.yaml
WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /etc/rancher/k3s/k3s.yaml
Error: UPGRADE FAILED: pre-upgrade hooks failed: 1 error occurred:
	* job failed: BackoffLimitExceeded
```

pre upgrade pod log
```
│ panic: unrecognized command: pre-upgrade                                                                                                                                                                │
│                                                                                                                                                                                                         │
│ goroutine 1 [running]:                                                                                                                                                                                  │
│ main.cmdNotFound(0xc00065eba0?, {0x7ffd4ed1b0fc?, 0xc00065ebc8?})                                                                                                                                       │
│     /go/src/github.com/longhorn/longhorn-manager/main.go:15 +0x67                                                                                                                                       │
│ github.com/urfave/cli.ShowCommandHelp(0xc0000fdb80, {0x7ffd4ed1b0fc, 0xb})                                                                                                                              │
[...]

PR #5881, opened 2023-05-08
fix: remove `privileged` from lifecycle jobs

mantissahz:
Remove `privileged` requirement from lifecycle jobs `post-upgrade` and `uninstall`.

Ref: longhorn/longhorn#5862


innobead (approved):
LGTM. Please see if we need to update the doc.

innobead:
@mantissahz have you re-generated the deploy.yaml?

Issue #5235, opened 2023-01-09
[IMPROVEMENT] Collect volume, system, feature info for metrics for better usage awareness


innobead:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

https://metrics.longhorn.io/ is open and public to let users know how Longhorn is being used by users. This is helpful info and reference for users when they start using Longhorn or even decide to upgrade Longhorn or underlying Kubernetes to which version. 

To understand more current usage from users to indirectly generate usage best practices, should collect more info about volume, host system, and feature. This would also give us more ideas on how to enhance Longhorn in the right direction.

## Describe the solution you'd like

- Host system-related info: Kernel version, Linux distro, Longhorn disk type (SSD/NVMe/HDD)
- Volume: access mode, volume mode, data locality, volume size, etc
- Cluster/node: cluster size
- Turnover rate compared with adoption rate (cluster id)

innobead:
cc @shuo-wu @derekbit feel free to update the issue if having any useful info able to add.

shuo-wu:
Volume: snapshot count; rebuilding count/frequency

(Will update this comment if there are more ideas occurring to me)

derekbit:
Metrics for backup CRUD operations, especially for the S3 backup store.

c3y1huang:
> Metrics for PUT, GET and DELETE operations of the backups, especially for the S3 backup store.

Longhorn sent the info at [per-hour intervals](https://github.com/longhorn/longhorn-manager/blob/v1.4.1/controller/setting_controller.go#L45). In my opinion, it might not be ideal to cache those real-time operations.

derekbit:
Oh, I misunderstood the purpose of the metrics collection before. What I mentioned is for the longhorn-system's metrics.

derekbit:
host CPU and memory might be useful for understanding users' environments and use cases.

innobead:
> host CPU and memory might be useful for understanding users' environments and use cases.

I assume this means Longhorn components instead of the whole system.

c3y1huang:
> > host CPU and memory might be useful for understanding users' environments and use cases.
> 
> I assume this means Longhorn components instead of the whole system?

When the host CPU and memory are likely to be shared with other applications, collecting this information might be meaningless. As suggested by @innobead, it would probably be more beneficial to collect the average CPU and memory usage of Longhorn components.

What do you think @derekbit ?

innobead:
[...]

PR #5908, opened 2023-05-10
feat(upgrade-responder): setup dev environment

c3y1huang:
https://github.com/longhorn/longhorn/issues/5235

### DIFF
--- a/dev/upgrade-responder/README.md
+++ b/dev/upgrade-responder/README.md
@@ -1,0 +1,55 @@
+## Overview
+
+### Install
+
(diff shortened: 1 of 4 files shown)


innobead:
This is good, and it will be helpful for QA testing. 

BTW, for any feature test which requires some extra env setup, engineers need to provide a setup manifest in longhorn/longhorn/dev for QA to do the following testing. 

cc @longhorn/dev @longhorn/qa

innobead (approved):

Issue #5845, opened 2023-05-03
[IMPROVEMENT] Longhorn-engine processes should refuse to serve requests not intended for them


ejweber:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

This is a follow-up to #5709. In that issue, we identified (and are fixing) race condition that allowed an expansion request to be issued to the wrong replica. The completion of this new issue should make it impossible for any such request to succeed.

## Describe the solution you'd like

All instances should, at the very least, know the volume name they are associated with (e.g. pvc-####). If it is practical, they should also know their actual instance name (e.g. pvc-####-r-####). Enhanced clients should pass the volume name and/or instance name they think they they are communicating with in each request. If the instance serving a request does not agree with the identifying information that is passed, it should refuse to serve the request.

## Describe alternatives you've considered

We could potentially make this unnecessary by reworking instance port allocation logic to better ensure it is only ever possible for the replica we expect to be listening on a particular port (see https://github.com/longhorn/longhorn/issues/5709#issuecomment-1513800543 for details). It was the opinion of the team that the approach outlined in this issue was equally valuable and probably easier to implement.

## Additional context

The initial proposal for this issue can be found here: https://github.com/longhorn/longhorn/issues/5709#issuecomment-1514935548.

The creation of this issue is intended to limit #5709 to just solving the discovered race condition. This issue will handle the more complicated new behavior previously proposed there.

ejweber:
@innobead @PhanLe1010 

The LEP (https://github.com/longhorn/longhorn/pull/5918) is ready for your initial review. I have also opened some draft PRs with POC code that demonstrates the workability of this approach for the case in which the longhorn-manager engine controller executes a ReplicaAdd command. Please let me know who else I should @ (or feel free to do @ others yourselves).

ejweber:
I realized as I was fixing some things up to day that many of the benefits of this improvement will not be available after a live engine upgrade. It is instance manager that is generally making gRPC requests to engines and replicas, so we need an upgraded instance-manager to add validation metadata. We will need to document this in the upgrade notes.

ejweber:
Keeping this in the Review phase until I fix https://github.com/longhorn/longhorn/issues/6522.

chriscchien:
Verified **pass** on longhorn master(longhorn-manager `aef322`,  longhorn-instance-manager `c5287c`) with test [steps](https://github.com/longhorn/longhorn/issues/5845#issuecomment-1626193884)

[...]

Issue #5709, opened 2023-04-06
[BUG] Rebuild rebuilding is possibly issued to a wrong replica


derekbit:
## Describe the bug (🐛 if you encounter this issue)

Engine controller iterates and compares the `e.Status.ReplicaModeMap` and `e.Status.CurrentReplicaAddressMap`, and then triggers a replica rebuilding if a replica is not in `e.Status.CurrentReplicaAddressMap`.

The `e.Status.ReplicaModeMap` and `e.Status.CurrentReplicaAddressMap` might not be the latest result from the engine monitor, so there is a chance that a replica's IP address is wrong. The replica rebuilding is issued to a wrong replica and leads to a volume expansion ([code](https://github.com/longhorn/longhorn-engine/blob/master/pkg/sync/sync.go#L501)).

## To Reproduce

3. Perform '....'

## Log or Support bundle

If applicable, add the Longhorn managers' log or support bundle when the issue happens. 
You can generate a Support Bundle using the link at the footer of the Longhorn UI.

## Environment

 - Longhorn version: v1.3.1+
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl):
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version:
   - Number of management node in the cluster:
   - Number of worker node in the cluster:
 - Node config
   - OS type and version:
   - CPU per node:
   - Memory per node:
   - Disk type(e.g. SSD/NVMe):
   - Network bandwidth between the nodes:
 - Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal):
 - Number of Longhorn volumes in the cluster:

PhanLe1010:
##A small code analysis:

### Normal replica rebuilding flow:
1.  Somehow, the replica process failed (reported by IM manager or `engine.status.replicaModeMap`) so Longhorn manager marked the replica as failed by setting `replica.Spec.FailedAt` and set `r.Spec.DesireState = longhorn.InstanceStateStopped`: [link](https://github.com/longhorn/longhorn-manager/blob/c4e7942684cc1f8ece900854d09126a7b1f8c0b6/controller/volume_controller.go#L580-L625) 
[...]

PR #5918, opened 2023-05-11
Engine Identity Validation LEP

ejweber:
#5845

See the following draft PRs for a POC implementation of this general approach.
- https://github.com/longhorn/longhorn-engine/pull/902
- https://github.com/longhorn/longhorn-instance-manager/pull/229
- https://github.com/longhorn/longhorn-manager/pull/1897

### DIFF
--- a/enhancements/20230420-engine-identity-validation.md
+++ b/enhancements/20230420-engine-identity-validation.md
@@ -79,7 +79,7 @@ After this proposal:
 
 As an administrator, after an intentional or unintentional node reboot, my volumes work as expected. If I choose to dig
 through logs, I may see some messages about refused requests to incorrect components, but this doesn't seem to
-negatively affected anything.
+negatively affect anything.
 
 #### Story 2
 
@@ -122,14 +122,14 @@ Add a `volume-name` flag and an `instance-name` flag to the `longhorn replica <d
 replica <directory> -volume-name <volume-name> -instance-name <instance-name>`). The longhorn-engine sync-agent server
 remembers its volume and instance name.
 
-Add a `volume-name` and `controller-instance-name` flag to every CLI command that launches an asynchronous task (e.g.
-`longhorn ls-replica -volume-name <volume-name> -controller-instance-name <controller-instance-name>`). All such
+Add a `volume-name` and `engine-instance-name` flag to every CLI command that launches an asynchronous task (e.g.
+`longhorn ls-replica -volume-name <volume-name> -engine-instance-name <engine-instance-name>`). All such
 commands create a controller client and these flags allow appropriate gRPC metadata to be injected into every client
 request. Requests that reach the wrong longhorn-engine controller server are rejected.
 
 Add an additional `replica-instance-name` flag to CLI commands that launch asynchronous tasks that communicate directly
 with the longhorn-engine replica server (e.g. `longhorn add-replica <address> -size <size> -current-size <current-size>
--volume-name <volume-name> -controller-instance-name <controller-instance-name> -replica-instance-name
+-volume-name <volume-name> -engine-instance-name <engine-instance-name> -replica-instance-name
 <replica-instance-name>`). All such commands create a replica client and these flags allow appropriate gRPC metadata to
 be injected into every client request. Requests that reach the wrong longhorn-engine replica server are rejected.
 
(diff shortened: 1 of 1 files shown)


innobead (commented):
In general, LGTM. 

This is about identicy check/authoration, but how about the action after an invalid operation call failed? Will we do any replica correction by removing this invalid replica from the engine?

ejweber:
> In general, LGTM.
> 
> This is about identicy check/authoration, but how about the action after an invalid operation call failed? Will we do any replica correction by removing this invalid replica from the engine?

The goal is to make it impossible to add the invalid replica to the engine in the first place (e.g. an AddReplica fails so that the engine never successfully adds a replica as a backend if it isn't supposed to).

If I'm following you correctly, though, you're interested in what happens in the management plane (i.e. does the replica address get removed from `engine.spec.replicaAddressMap`). In the case of https://github.com/longhorn/longhorn/issues/5709:

1. ReplicaAdd fails more immediately than before and with a different error ("communication with the replica is not allowed" instead of "can't add a dirty backend"), but the result is the same. https://github.com/longhorn/longhorn-manager/blob/26a13251251129a656b328234a76a2ed9f81b222/controller/engine_controller.go#L1818
1. If we aren't in backoff, we mark the replica as failed. https://github.com/longhorn/longhorn-manager/blob/26a13251251129a656b328234a76a2ed9f81b222/controller/engine_controller.go#L1854
1. If we are in backoff, we may rereconcile, hopefully with a better `engine.spec.replicaAddressMap` this time (in the case of https://github.com/longhorn/longhorn/issues/5709, the `engine.spec.replicaAddressMap` was self-correcting given sufficient time).

In general, I think we should rely on existing mechanisms like the above to handle failures unless/until we identify issues in testing. There may be other scenarios in which distinguishing between this new error and a general error would be beneficial, but I'm not sure what they are at the moment. If the general strategy here is approved, I can continue to move the code forward to the point where we can do some E2E testing before we merge the LEP. I'm also interesting in any specific scenarios reviewers can think of along these lines.

enhancements/20230420-engine-identity-validation.md
```diff
@@ -0,0 +1,400 @@
...
+Add a `volume-name` flag and an `instance-name` flag to the `longhorn replica <directory>` command (e.g. `longhorn
+replica <directory> -volume-name <volume-name> -instance-name <instance-name>`). The longhorn-engine replica server
+remembers its volume and instance name.
+
+Add a `volume-name` flag and an `instance-name` flag to the `longhorn replica <directory>` command (e.g. `longhorn
+replica <directory> -volume-name <volume-name> -instance-name <instance-name>`). The longhorn-engine sync-agent server
+remembers its volume and instance name.
+
+Add a `volume-name` and `controller-instance-name` flag to every CLI command that launches an asynchronous task (e.g.
+`longhorn ls-replica -volume-name <volume-name> -controller-instance-name <controller-instance-name>`). All such
```
shuo-wu, line 126: NIT: You can directly name the option to `engine-instance-name`

enhancements/20230420-engine-identity-validation.md
```diff
@@ -0,0 +1,400 @@
...
+		md, ok := metadata.FromIncomingContext(ctx)
+		if ok {
+			incomingVolumeName, ok := md["volume-name"]
+			// Only refuse to serve if both client and server provide validation information.
+			if ok && volumeName != "" && incomingVolumeName[0] != volumeName {
+				return nil, status.Errorf(codes.InvalidArgument, "Incorrect volume name; check controller address")
+			}
+		}
+
+		if ok {
```
shuo-wu, line 186: NIT: Redundant line

enhancements/20230420-engine-identity-validation.md
```diff
@@ -0,0 +1,400 @@
...
+			},
+		},
+		// Rest of implementation.
+	}
+}
+```
+
+#### Instance-Manager Integration
+
+Modify the ProxyEngineService server functions so that they can make correct use of the changes in longhorn-engine.
```
shuo-wu, line 299: May need to bump up `InstanceManagerProxyAPIVersion`

shuo-wu (approved):
LGTM

enhancements/20230420-engine-identity-validation.md
ejweber: Good catch. Added a paragraph for this in my latest commit.

shuo-wu (approved):

innobead (approved):

enhancements/20230420-engine-identity-validation.md
```diff
@@ -0,0 +1,405 @@
...
+The user will get benefit from this behavior automatically, but only after they have upgraded all associated components
+to a supporting version (longhorn-manager, longhorn-engine, and CRITICALLY instance-manager).
+
+We will only provide volume name and instance name information to longhorn-engine controller and replica processes on a
+supported version (as governed by the `CLIAPIVersion`). Even if other components are upgraded, when they send gRPC
+metadata to non-upgraded processes, it will be ignored.
+
+We will only populate extra ProxyEngineService fields when longhorn-manager is running with an update ProxyEngineService
+client. 
+- RPCs from an old client to a new ProxyEngineService server will succeed, but without the extra fields,
```
PhanLe1010, line 386: Just a tiny note. I think it is good to include this case in the LEP, but in practice, this should not happen as the Longhorn manager (the only one who is using ProxyClient) is always upgraded first. So it is not possible to have old client talk to new ProxyEngineService server

ejweber: Good note. @shuo-wu mentioned something similar in the longhorn-engine review. It does affect how I think about compatibility issues.

PR #5923, opened 2023-05-12
fix(csi): no need to check if volume is attached when creating backing image type snapshot

ChanYiLin:
ref: [longhorn/longhorn#5005](https://github.com/longhorn/longhorn/issues/5005)

### DIFF
--- a/enhancements/20230417-extend-csi-snapshot-to-support-backingimage.md
+++ b/enhancements/20230417-extend-csi-snapshot-to-support-backingimage.md
@@ -267,7 +267,6 @@ spec:
 
 We do:
 - Get the name of the Volume
-- Check if the volume is in attached state. If not, return `codes.FailedPrecondition` since we can not export the Volume to a BackingImage.
 - The name of the BackingImage will be same as the VolumeSnapshot `test-snapshot-backing`.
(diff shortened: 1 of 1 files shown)


enhancements/20230417-extend-csi-snapshot-to-support-backingimage.md
```diff
@@ -267,7 +267,6 @@ spec:
 
 We do:
 - Get the name of the Volume
-- Check if the volume is in attached state. If not, return `codes.FailedPrecondition` since we can not export the Volume to a BackingImage.
```
PhanLe1010, line 270: Sounds good! This will be consistent will our new AD refactoring

ChanYiLin: Yes there are several case we will do auto attachment 👍 
https://github.com/longhorn/longhorn-manager/blob/master/controller/volume_controller.go#L2845

ChanYiLin:
Hi @innobead the conflict has resolved, thanks!

innobead (approved):
LGTM

innobead (approved):

Issue #5917, opened 2023-05-11
[TASK] Remove `Guaranteed Engine Manager CPU`, `Guaranteed Replica Manager CPU`, and `Guaranteed Engine CPU` settings.


c3y1huang:
## Is your improvement request related to a feature? Please describe

After consolidating engine and replica instance managers, the `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU` settings serve no other purpose than to trigger the instance manager pod deletion.

The `Guaranteed Engine CPU` is ineffective after introducing the above two settings.

## Describe the solution you'd like

Remove `Guaranteed Engine Manager CPU`, `Guaranteed Replica Manager CPU`, and `Guaranteed Engine CPU` settings.

## Describe alternatives you've considered

Handle setting removal in https://github.com/longhorn/longhorn/issues/5844.

## Additional context

https://github.com/longhorn/longhorn/issues/5208

roger-ryao:
**Verified on master-head 20230530**

- longhorn master-head ([7b3b230](https://github.com/longhorn/longhorn/commit/7b3b230f471071f08b9621a165fab7887e087491))
- longhorn-manager master-head ([023dd3c](https://github.com/longhorn/longhorn-manager/commit/023dd3cbd5009f390608ea9cdd9f31bd397cf8a2))


**Result Passed**

- [x] - doc: https://github.com/longhorn/website/pull/671
- [x] The UI on the master branch no longer includes the options for `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU`.
- [x] After upgrading Longhorn from version `1.4.x` to the `master` branch, the options for `Guaranteed Engine Manager CPU` and `Guaranteed Replica Manager CPU` were removed from the UI.
- [] We have a typo [in the PR](https://github.com/longhorn/website/pull/671/files#r1209705291). Once the typo is fixed, I will close this ticket.

innobead:
I like the checklist which will prevent any missing criteria for closing the issue. 

cc @longhorn/qa

Issue #5876, opened 2023-05-08
[TASK][UI] support new aio typed instance managers


c3y1huang:
## What's the task? Please describe

Starting from version 1.5.x, the engine/replica instance managers will be consolidated into a new `aio` type instance manager. However, for backward compatibility, the deprecated engine and replica instance manager types will still be supported until removed in version 1.6.0.

As a result of this change, the Longhorn UI must be able to handle both the new aio type instance manager as well as the older engine and replica instance managers.

Example payload: 
```
{
    "data": [
        {
            "actions": {},
            "currentState": "running",
            "id": "instance-manager-e03ff9da49874cdb435b1aec4762ee06",
            "image": "longhornio/longhorn-instance-manager:master-head",
            "instanceEngines": {
                "new-aio-volume-e-549cf291": {
                    "spec": {
                        "name": "new-aio-volume-e-549cf291"
                    },
                    "status": {
                        "endpoint": "",
                        "errorMsg": "",
                        "listen": "",
                        "portEnd": 10015,
                        "portStart": 10015,
                        "resourceVersion": 0,
                        "state": "running",
                        "type": "engine"
                    }
                }
            },
            "instanceReplicas": {
                "new-aio-volume-r-aa955e8c": {
                    "spec": {
                        "name": "new-aio-volume-r-aa955e8c"
                    },
                    "status": {
                        "endpoint": "",
                        "errorMsg": "",
                        "listen": "",
                        "portEnd": 10014,
                        "portStart": 10000,
                        "resourceVersion": 0,
                        "state": "running",
                        "type": "replica"
                    }
                }
            },
            "instances": null,
            "links": {
                "self": "http://13.215.24.180:30111/v1/instancemanagers/instance-manager-e03ff9da49874cdb435b1aec4762ee06"
            },
            "managerType": "aio",
            "name": "instance-manager-e03ff9da49874cdb435b1aec4762ee06",
            "nodeID": "ip-10-0-1-102",
            "type": "instanceManager"
        },
        {
            "actions": {},
            "currentState": "running",
            "id": "instance-manager-e-509924af6f4b599dc258165df2bb37ab",
            "image": "longhornio/longhorn-instance-manager:v1.4.1",
            "instanceEngines": null,
            "instanceReplicas": null,
            "instances": {
                "pvc-857d4176-1bd3-45dc-b293-6fa45154b448-e-3e0c8a85": {
                    "spec": {
[...]

PR #5924, opened 2023-05-12
chore: remove Engine/ReplicaManagerCPURequest in Node CRD

c3y1huang:
https://github.com/longhorn/longhorn/issues/5917


c3y1huang:
We also need to remove it from the UI.

Ref: https://github.com/longhorn/longhorn/issues/5876

innobead (approved):

PR #5940, opened 2023-05-15
Update chart for new AD refactoring

PhanLe1010:
longhorn/longhorn#3715

### DIFF
--- a/chart/templates/crds.yaml
+++ b/chart/templates/crds.yaml
@@ -3322,14 +3322,14 @@ spec:
                 type: string
               backingImage:
                 type: string
-              baseImage:
-                description: Deprecated. Rename to BackingImage
-                type: string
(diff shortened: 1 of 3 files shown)


chart/templates/crds.yaml
```diff
@@ -3322,15 +3322,15 @@ spec:
...
               backupCompressionMethod:
                 enum:
                 - none
                 - lz4
                 - gzip
                 type: string
+              baseImage:
+                description: Deprecated. Rename to BackingImage
+                type: string
               dataLocality:
```
PhanLe1010, line 3334: The position of this code blog changed due to copy the generated code from `longhorn-manager/k8s/crds.yaml`.

Similar for `restoreVolumeRecurringJob`

innobead (approved):

Issue #5403, opened 2023-02-16
[FEATURE] Install Opt-In NetworkPolicies 


innobead:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

We have network policy manifest examples for Longhorn workloads already, but for the least privilege, it would be better to install them by default; however, they will only take effect after a policy network plugin gets installed and need to consider different installed ingress reverse proxy. It would be good to provide an opt-in policy installation.

For user-facing components like longhorn-ui, there could be a reverse proxy ahead, so should allow users to configure the ingress setting (ex: from, ports) of the installed policy based on their installed provider. (ex: https://github.com/longhorn/longhorn/pull/5378/files)

## Describe the solution you'd like

Install network policy manifests by default with customizable ingress setting for user-facing components. 

## Additional context

cc @longhorn/dev

innobead:
cc @PhanLe1010

dgiebert:
@innobead I have added it to the base helm chart as an optional field and worked on the questions.

innobead:
> @innobead I have added it to the base helm chart as an optional field and worked on the questions.

Thanks. @ChanYiLin will help review that.

ChanYiLin:
Verified

### longhorn-manager
can only be accessed by following pods
- [x] 1. longhorn-manager
- [x] 2. longhorn-ui
- [x] 3. longhorn-csi-plugin
- [x] 4. recurring-job.longhorn.io/longhorn.io/job-task (verified that recurring job works)
- [x] 6. longhorn-driver-deployer

It cannot be accessed by following pods
- instance-manager-e/r
  - [x] verified volume create/attach/detach works
- backing-image-manager/backing-image-data-source
  - [x] verified backingimage upload/export works

Note: 
Would like to know if we should allow instance-manager and backing image related pods to access longhorn-manager?

### lonhghorn-ui
can be accessed by traefik (networkPolicies.type == "k3s")

Note:
Need to do some modification if users deploy traefik by themselves using helm
the namespace would be default


This PR is basically copy the examples to the helm chart and add a opt flag
Nothing change from this issue: https://github.com/longhorn/longhorn/issues/1805#issuecomment-964724830

cc @shuo-wu @PhanLe1010 @innobead

shuo-wu:
@ChanYiLin Can you help review this based on the test result?

yangchiu:
Installed Longhorn `master-head` with networkPolicies `enabled`, recurring job seems not working:

[...]

PR #5950, opened 2023-05-17
feat(network policy): add network policy into chart

ChanYiLin:
ref: [longhorn/longhor#5403](https://github.com/longhorn/longhorn/issues/5403)


innobead (changes requested):
Do we still need recover backend for share manager and webhook network policies for 1.5?

innobead (approved):
LGTM.

PR #5955, opened 2023-05-17
prerequisite: set PCI_ALLOWED to none when setting up spdk environment

derekbit:
Currently, longhorn uses SPDK AIO feature.

### DIFF
--- a/deploy/prerequisite/longhorn-spdk-setup.yaml
+++ b/deploy/prerequisite/longhorn-spdk-setup.yaml
@@ -29,11 +29,13 @@ spec:
         image: alpine:3.12
         env:
         - name: SPDK_DIR
-          value: /tmp/spdk
+          value: "/tmp/spdk"
         - name: SPDK_OPTION
           value: ""
         - name: HUGEMEM
           value: "2048"
+        - name: PCI_ALLOWED
+          value: "none"
         - name: DRIVER_OVERRIDE
           value: "uio_pci_generic"
         securityContext:


shuo-wu:
Could help explain more about the difference between AIO and PCI?

derekbit:
> Could help explain more about the difference between AIO and PCI?

The SPDK AIO bdev driver enables access to the Linux kernel block devices via the SPDK block layer. This means disks are still managed by the Linux kernel rather than SPDK. However, upon setting PCI_ALLOW, SPDK takes on the responsibility of managing the NVMe disks, and these disks are not managed by Kernel and are not shown in `fdisk -l`.

innobead:
Very similar to DPDK to manage network cards.

derekbit:
> Very similar to DPDK to manage network cards.
> 
> So setting `PCI_ALLOWED to none` will make SPDK can manage the block devices directly, right?

No, setting `PCI_ALLOWED to none` will make PCI devices not managed by SPDK and still managed by Kernel. SPDK will access PCI devices by AIO bdev driver.

derekbit:
> > Very similar to DPDK to manage network cards.
> > So setting `PCI_ALLOWED to none` will make SPDK can manage the block devices directly, right?
> 
> No, setting `PCI_ALLOWED to none` will make PCI devices not managed by SPDK and still managed by Kernel. SPDK will access PCI devices by AIO bdev driver.

We can set `PCI_ALLOWED` in the future if longhorn supports the kernel bypass in the future.

innobead:
Sounds good to me for now.

innobead (approved):

DamiaSan:
With `PCI_ALLOWED` the SPDK NVMe driver is used, it should provide better performances.

innobead:
@derekbit Let's create an issue for this.

I am curious what performance gap between AIO and PCI.

> We can set `PCI_ALLOWED` if longhorn supports the kernel bypass (I mean using other SPDK bdev driver rather than AIO bdev driver) in the future.

derekbit:
> @derekbit Let's create an issue for this.
> 
> I am curious what performance gap between AIO and PCI.
> 
> > We can set `PCI_ALLOWED` if longhorn supports the kernel bypass (I mean using other SPDK bdev driver rather than AIO bdev driver) in the future.

Tracked in https://github.com/longhorn/longhorn/issues/5956

derekbit:
> With `PCI_ALLOWED` the SPDK NVMe driver is used, it should provide better performances. It could be interesting to make some tests

@DamiaSan Do you have numbers for the two methods?

innobead:
@derekbit AIO another purpose is to let users can try the new stack by using loop device on fs, correct?

derekbit:
> @derekbit AIO another purpose is to let users can try the new stack by using loop device on fs, correct?

Exactly.
The requirement of for NVMe driver is relatively high.

innobead:
Probably can be adaptive by depending on the type of disks later. Let's add this to the ticket you just created.

DamiaSan:
> > With `PCI_ALLOWED` the SPDK NVMe driver is used, it should provide better performances. It could be interesting to make some tests
> 
> @DamiaSan Do you have numbers for the two methods?

Actually not, I'm sorry

Issue #5011, opened 2022-12-07
[FEATURE] Support backup volumes during system backup


c3y1huang:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

Currently, Longhorn system backup doesn't backup volume data.
- If the volume in the system backup doesn't contain the last backup then that volume restored from the Longhorn system backup will not contain any data.
- The data in the volume last backup might not be the same data existing during the Longhorn system backup.

## Describe the solution you'd like

Have a setting to allow to create volume backup during Longhorn system backup.

## Describe alternatives you've considered

`None`

## Additional context

`None`

khushboo-rancher:
Just a suggestion for the solution, we can also have an option while creating the system backup to ask user whether they want to trigger the volume backup. 

A checkbox to trigger volume backup in the below screen.
<img width="549" alt="Screenshot 2022-12-19 at 4 43 40 PM" src="https://user-images.githubusercontent.com/60111667/208555379-0f5e8bc1-ec17-4001-8c69-a632d2b509fd.png">

chriscchien:
I faced a situation that a system backup will be deleted by specific operation, below is the reproduce steps:  

0. Create 2 system backups `test-backup-policy` and `test-backup-policy2` with volumeBackupPolicy set to `always` or `if-not-present`, after system backup ready, delete all backups in backup store:
1. Delete `test-backup-policy` and create again, after this system backup complete, `test-backup-policy2` will be deleted.
2. After step 0, do not delete any exist system backup and create another 1 `test-backup-policy3`, no system backup deleted.

Otherwise all volume backup create followed the setting in system backup. (Tested in longhorn master (longhorn `ffaa3d` longhorn-manager `437990`))

chriscchien:
Verified **pass** in longhorn master (longhorn `985634`, longhorn-manager `663185`, longhorn-ui `51e912`)

Create systembackup form both UI and manifest, below 3 types of `volumeBackupPolicy` worked as [expected](https://github.com/longhorn/longhorn/blob/master/enhancements/20230526-volume-backup-policy-for-longhorn-system-backup.md).

- diabled
   - No volume backup created
- always
   - Volume backup always created, after restore, volume data intact
- if-not-present
   - Volume backup created if no backup in backup store, after restore, volume data were the same when backup created.

PR #6008, opened 2023-05-26
feat(lep): volume backup policy for Longhorn system backup design

c3y1huang:
https://github.com/longhorn/longhorn/issues/5011


innobead (dismissed):
In general, LGTM.

If backups take long time, how the system backup will work? will it timeout? what's the timeout now?

If yes, should make a note for this case in the doc.

c3y1huang:
> If backups take long time, how the system backup will work? will it timeout? what's the timeout now?

Currently, the timeout is limited to 24 hours. Any of the backups fail will lead to the system backup failure.

> If yes, should make a note for this case in the doc.

Updated.

innobead (approved):

Issue #1512, opened 2020-06-22
[FEATURE] Allow users to directly activate a restoring/DR volume as long as there is one ready replica.


shuo-wu:
**Is your feature request related to a problem? Please describe.**
Some replicas may get stuck in the restoring process then the restore volume is no longer useable. We can provide an API to forcibly activate the volume as long as there is a ready replica.

**Describe the solution you'd like**
1. Marking all replicas that are still in the restore progress as failed.
2. Disable rebuild.
3. Indicate the volume is usable now and do auto detachment.

shuo-wu:
Design Doc(draft): https://docs.google.com/document/d/1k3vYA68ej09F9h5qaoLhdO0wuWm7kNAsIDNyVPH2u-A/edit?usp=sharing

yasker:
Moved with #1279 to v1.1.0.

innobead:
https://github.com/longhorn/longhorn/issues/3069

weizhe0422:
1. Create a volume and write some data
   ![image](https://github.com/longhorn/longhorn/assets/17548901/8a6540a9-2eee-4031-ac5b-6deae80a11ea)
2. Create a backup
   ![image](https://github.com/longhorn/longhorn/assets/17548901/886143c9-d24e-42f2-8943-9147bb8cefde)
3. Create a DR volume
   ![image](https://github.com/longhorn/longhorn/assets/17548901/ee400a86-f95a-4391-8268-aeea40cbaf96)
4. Disable rebuilding by setting concurrent-replica-rebuild-per-node-limit to be 0
   ![image](https://github.com/longhorn/longhorn/assets/17548901/c266b8d6-79ee-4fbb-8d81-8083607f120d)
5. Failed a replica
   ![image](https://github.com/longhorn/longhorn/assets/17548901/36712847-96d5-4e0b-ac07-5530abc084f4)
6. Forcibly activate the DR volume and volume could be activated
   ![image](https://github.com/longhorn/longhorn/assets/17548901/0684aaea-b164-4e4d-95a3-b92f8a5a457f)
7. Volume could be attached again and data is correct
   ![image](https://github.com/longhorn/longhorn/assets/17548901/1b5f3bd7-501a-406f-9b45-8c49cddd76ac)
   ![image](https://github.com/longhorn/longhorn/assets/17548901/52fae09e-484c-4034-a977-2afcd866ff10)

PR #6017, opened 2023-05-29
feat(dr-volume): force activate a DR volume LEP

mantissahz:
LEP for activating a DR volume forcibly as long as there is a ready replica even though a DR volume had been activated and stuck in attaching state.

Ref: longhorn/longhorn#1512

### DIFF
--- a/enhancements/20230601-forcibly-activate-a-restoring-dr-volume.md
+++ b/enhancements/20230601-forcibly-activate-a-restoring-dr-volume.md
@@ -1,0 +1,77 @@
+# Forcibly Activate A Restoring/DR Volume
+
+## Summary
+
+When users try to activate a restoring/DR volume with some replicas failed for some reasons, the volume will be stuck in attaching state and users can not do anything except deleting the volume. However the volume still can be rebuilt back to be normal as long as there is a healthy replica.
+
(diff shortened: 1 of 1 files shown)


enhancements/20230529-forcibly-activate-a-restoring-dr-volume.md
```diff
@@ -0,0 +1,103 @@
...
+Allow users to activate a restoring/DR volume forcibly as long as there is a healthy replica and the volume works well.
+
+### Non-goals [optional]
+
+`None`
+
+## Proposal
+
+Provide a `Force` flag for the API `activate` to allow users to activate a restoring/DR volume as long as there is a healthy replica by adding a volume status parameter `ForciblyActivating`.
+
```
innobead, line 25: Why not just `force` by default if `degraded` volume is allowed, so we don't need an extra property? Do we have any concerns to activate it even if there is only one healthy replica?

shuo-wu: LGTM.

mantissahz: Modified to use the global setting `allow-volume-creation-with-degraded-availability`
Thanks for your advice.

innobead (approved):

PR #6018, opened 2023-05-30
feat(dr-volume) forcibly avtivate a DR volume

mantissahz:
Add a volume status parameter to support forcibly activating a DR volume as long as there is a ready replica.

Ref: longhorn/longhorn#1512


mantissahz:
This new field is not required anymore.

Issue #5191, opened 2023-01-03
[TASK] Revert affinity for Longhorn user deployed components


innobead:
@sjoukedv added affinity support for longhorn manager-based related component deployment. This can be applied to other user-deployed components as well to have a consistent scheduling mechanism. 

https://github.com/longhorn/longhorn/pull/5168

The only thing that needs to pay attention to is whether this change would potentially impact Longhorn deployment on a mixed Linux/Windows cluster. https://github.com/longhorn/longhorn/blob/76eaa3d3c17f2711855bf9594f4db5554dc242fc/chart/values.yaml#L4-L22 needs to be respected.

ejweber:
@PhanLe1010 and I were discussing https://github.com/longhorn/longhorn/pull/5168 and have a concern:

Since we can now use node affinity to control which nodes longhorn-manager can deploy to (instead of just node selector), but system-managed components only respect node selector, there is the potential for some confusing behavior. If I use node affinity to express a complex requirement (e.g., run on nodes that ARE labeled key1:value1 and NOT labeled key1:value2), but cannot express that complex requirement with the setting `system-managed-components-node-selector`, there will be some nodes where `longhorn-manager` cannot run but pods like `longhorn-csi-plugin` do.

TLDR, it may be necessary to create a new `system-managed-components-node-affinity` if we want to enable users to use node affinity properly.

ejweber:
Moving this to Review so I can start filling out testing steps. However, it can't be completed until we resolve the comment above. Please let us know your thoughts @innobead.

innobead:
@ejweber exactly, we should make the behavior the same as what we support for node selector and taint toleration. 

Either fix the current issue with limitation without supporting system managed components which will be improved later, or do all things together.

If unable to achieve this in 1.5.0, let's move it to 1.6.0 instead.

weizhe0422:
Test steps:
**Install within manifest**
  1. Install Longhorn with command: `kubectl apply -f https://raw.githubusercontent.com/longhorn/longhorn/v1.5.0-rc1/deploy/longhorn.yaml`
  2. Verify the Longhorn is installed properly
     ![image](https://github.com/longhorn/longhorn/assets/17548901/b5c48c94-00de-4281-89a5-1e415a0b626a)
  3. Check volume could be created and attached
     ![image](https://github.com/longhorn/longhorn/assets/17548901/3bd3a0cc-064e-4032-b392-e64fa36e7ed4)
[...]

PR #6025, opened 2023-05-30
Node affinity for user-deployed components

ejweber:
#5191 

The previous work done in https://github.com/longhorn/longhorn/pull/5168 contained a typo that rendered it ineffective. In addition, it broadly supported affinity in general. This PR:

- Fixes the typo.
- Narrows down to NodeAffinity because:
    - It simplifies the longhorn-ui case, in which we already have a PodAntiAffinity and now we don't have to merge a user-specified value and our default value.
    - IMO NodeAffinity is what we want to be specifying for a storage system (it doesn't make sense to try to schedule the storage system to be on the same node as or avoid arbitrary workloads).
- Enabled NodeAffinity on all user-deployed components. 

There is a HUGE caveat here. Even though user-deployed components are limited by node affinity, system-deployed components are not. We must discuss this before merging.

- [ ] Resolve https://github.com/longhorn/longhorn/issues/5191#issuecomment-1561821738

### DIFF
--- a/chart/templates/daemonset-sa.yaml
+++ b/chart/templates/daemonset-sa.yaml
@@ -122,7 +122,7 @@ spec:
         {{- end }}
       {{- end }}
       {{- if .Values.longhornManager.affinity }}
-      affinitiy:
+      affinity:
 {{ toYaml .Values.longhornManager.affinity | indent 8 }}
       {{- end }}      
       serviceAccountName: longhorn-service-account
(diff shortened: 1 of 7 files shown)


ejweber:
Converting this back to a draft. After discussing with @PhanLe1010 and @innobead, we are going to go a different direction for v1.5.0 and this will not be merged as is.

PR #6032, opened 2023-05-31
SPDK charts and examples

derekbit:


examples/spdk/storageclass.yaml
```diff
@@ -0,0 +1,24 @@
...
+metadata:
+  name: longhorn-spdk
+provisioner: driver.longhorn.io
+allowVolumeExpansion: true
+reclaimPolicy: Delete
+volumeBindingMode: Immediate
+parameters:
+  numberOfReplicas: "2"
+  staleReplicaTimeout: "2880"
+  fromBackup: ""
```
innobead, line 12: If not supported yet, we can comment out. Also, add a comment next to `parameters` to mention those parameters not supported yet.

chart/templates/crds.yaml
```diff
@@ -1772,6 +1782,8 @@ spec:
                   properties:
                     spec:
                       properties:
+                        backendStoreDriver:
```
innobead, line 1785: Do we need this for deprecated `instances`?

innobead (changes requested):
In general, LGTM. Some comments needed to resolve.

chart/templates/crds.yaml
derekbit: Yes, instances in old IMs need the field as well.

examples/spdk/storageclass.yaml
derekbit: Addressed, thank you.

innobead:
Pending until manager PRs merged

chart/templates/crds.yaml
```diff
@@ -1678,6 +1684,8 @@ spec:
...
```
shuo-wu, line 1687: Why do not some `backendStoreDriver` fields include `enum`?

derekbit: Not sure if we need to include `enum` for the `backendStoreDriver` are in status.

shuo-wu: If it's allowed and there is compatibility issue, I think it's better to have

PR #6040, opened 2023-06-01
spdk: reduce hugepage size to 1024MiB and persistent the hugepage setting

derekbit:


derekbit:
Yeah, I found 8GB machine is sometimes not able to allocate enough hugepages if there are some applications on it.
Reducing to 1GB is easy for users.

innobead:
Need to document this in the doc about the SPDK data engine get-started guide/page.

derekbit:
Sure. In progress.

PR #6066, opened 2023-06-05
spdk: use 1024 MiB huge pages by default

derekbit:
Forgot updating the HUGEMEM in https://github.com/longhorn/longhorn/pull/6040/files


innobead (approved):

derekbit:
@mergify backport v1.5.x

Issue #5739, opened 2023-04-12
[FEATURE] Deployment manifests for helping install SPDK dependencies, utilities and libraries


derekbit:
## Describe the solution you'd like

A clear and concise description of what you want to happen

derekbit:
Yeah. I'd update the manifest to making hugepage persistent on each node, so the hugepage is automatically allocated after reboot. QA can retest it.

yangchiu:
@derekbit Is the control plane also needed to install SPDK dependencies even though it's tainted with `NoSchedule` and `NoExecute` so there's no workload can run on this node?

I created a cluster with 1 control plane and 3 worker nodes, and followed the quick start guide to get 3 worker nodes configured for SPDK, and then installed Longhorn.
When I tried to enable SPDK by checking `Enable SPDK Data Engine`, an error shown up:
```
failed to set the setting spdk with invalid value true: 
not enough hugepages-2Mi capacity for node ip-10-0-1-239, requested 1Gi, capacity 0
```
But actually `ip-10-0-1-239` is a control plane with taint `NoSchedule` and `NoExecute`so there's no Longhorn related pods running on it:
```
$ kubectl get nodes
NAME            STATUS   ROLES                  AGE   VERSION
ip-10-0-1-239   Ready    control-plane,master   86m   v1.24.7+k3s1

$ kubectl describe node ip-10-0-1-239
Name:               ip-10-0-1-239
Roles:              control-plane,master
Taints:             node-role.kubernetes.io/master=true:NoExecute
                    node-role.kubernetes.io/master=true:NoSchedule
```
I'm not sure if it's the expected behavior. Thanks!

innobead:
Good finding. @derekbit We should check nodes running instance manager pod instead, right?

cc @shuo-wu

yangchiu:
Compared https://even-pecorino-1bc.notion.site/Quick-Start-for-SPDK-Volumes-a189b3f81e23495d973e3a14b7635667 with https://longhorn.io/docs/1.5.0/spdk/quick-start/. The `Install Extra Kernel Modules` part:
```
apt install -y linux-modules-extra-`uname -r`
```
is missing in https://longhorn.io/docs/1.5.0/spdk/quick-start/.

I'm not sure if it's necessary to manually install `linux-modules-extra-`, but if I didn't do that, the `longhorn-nvme-cli-installation` daemonset will fail on Ubuntu 22.04 because it expects there is `nvme-tcp` but there isn't:
```
[...]

PR #6068, opened 2023-06-06
spdk: update expected-nr-hugepages to 512 in environment_check.sh

derekbit:
Longhorn/longhorn#5739


innobead:
@derekbit The reason to lower the value again is still for encouraging users to try, right? Have we tested the performance if the default is 512v

derekbit:
> @derekbit The reason to lower the value again is still for encouraging users to try, right? Have we tested the performance if the default is 512?

Not yet. Will test it if have time, but I think only one volume should not be impacted.

innobead (approved):

innobead:
@mergify backport v1.5.x

Issue #5744, opened 2023-04-13
[FEATURE] Implement Disk gRPC Service in Instance Manager for collecting SPDK disk statistics from SPDK gRPC service 


derekbit:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

According to the design and discussion with @shuo-wu, we need to implement a Disk gRPC Service in Instance Manager for
1. Collecting SPDK disk (lvstore) statistics from SPDK gRPC service
2. Requesting to create a lvstore
3. ...

```
node controller -> disk monitor -> disk service -> spdk service
```

![image](https://user-images.githubusercontent.com/12527233/231639458-f22d0742-d767-414f-bf82-9a5c3a2ae379.png)

## Describe the solution you'd like

A clear and concise description of what you want to happen

innobead:
cc @shuo-wu

chriscchien:
Hi @derekbit , 

I tested with SLES `ami-0dd7f664cbd69840b`, following instructions the environment check script passed, but when I try attach a SPDK volume to a node, the volume stuck at attaching status (no matter created from UI or dynamic provision) .


<details>
  <summary>Environment check result</summary>

```bash
ip-172-31-88-0:/home/ec2-user # bash -c "$(curl -sfL https://raw.githubusercontent.com/longhorn/longhorn/master/scripts/environment_check.sh)" -s -s
[INFO]  Required dependencies 'kubectl jq mktemp' are installed.
[INFO]  All nodes have unique hostnames.
[INFO]  Waiting for longhorn-environment-check pods to become ready (0/3)...
[INFO]  All longhorn-environment-check pods are ready (3/3).
[INFO]  MountPropagation is enabled
[INFO]  Checking iscsid...
[INFO]  Checking multipathd...
[INFO]  Checking packages...
[INFO]  Checking nfs client...
[INFO]  Checking x86-64 SSE4.2 instruction set...
[INFO]  Checking nvme-cli...
[INFO]  Checking kernel module nvme_tcp...
[INFO]  Checking kernel module uio_pci_generic...
[INFO]  Checking hugepage...
[INFO]  Cleaning up longhorn-environment-check pods...
[INFO]  Cleanup completed.
```  
</details>

<details>
  <summary>Longhorn SPDK setting</summary>

```bash
k get setting -A | grep spdk
longhorn-system   spdk-hugepage-limit                                               1024                                               24m
longhorn-system   spdk                                                              true                                               24m
```  
</details>

<details>
[...]

PR #6083, opened 2023-06-08
spdk: nvme-cli should be between 1.12 and 1.16

derekbit:
Longhorn/longhorn#5744


innobead (approved):

innobead:
@mergify backport v1.5.x

Issue #2997, opened 2021-09-13
[BUG] Volume UI displays only the last backup when using the recurring job


mfbrj:
**Describe the bug**
Volume UI is displaying only the last backup

**To Reproduce**
1. Create a backup recurring job with retain >= 2, and a group
2. Associate this recurring job group to some volume
3. Wait for 2 or 3 backups be performed
4. See that volume UI displays only the last backup. Screenshots below.

**Expected behavior**
Volume UI displays 2 or more backups, within retain range.

**Log**

**Environment:**
 - Longhorn version:  1.2.0
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl): Rancher Catalog App
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: RKE v1.20.9
   - Number of management node in the cluster: 3
   - Number of worker node in the cluster: 8
 - Node config
   - OS type and version: Ubuntu 20.04.3 LTS   kernel 5.4.0-84-generic
   - CPU per node: 4
   - Memory per node: 16 GB
   - Disk type(e.g. SSD/NVMe): SSD
   - Network bandwidth between the nodes:
 - Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal): VMWare
 - Number of Longhorn volumes in the cluster: 23

**Additional context**
Wrong behaviour:
![image](https://user-images.githubusercontent.com/835834/133016894-b9005c4b-910f-44e1-8d2a-23e6a29658c0.png)

Wrong behaviour:
![image](https://user-images.githubusercontent.com/835834/133017271-0bab5436-c258-4fd0-9b86-c9f86261d3d7.png)

Right behaviour:
![image](https://user-images.githubusercontent.com/835834/133017487-292d7b78-4052-4d0b-b0fb-edf4e6f41938.png)

PhanLe1010:
@smallteeths Do you know if the snapshot chain graph and the table of the snapshot are getting data from the same source? We are wondering if this problem is because of UI parsing the timestamp differently or if we need to fix from backend.

mfbrj:
@PhanLe1010 as far I can tell, there is only one source, because I have only one minio backup server configured on Longhorn settings.

PhanLe1010:
Thanks @mfbrj ! Sorry! My bad! the above comment was supposed for this issue https://github.com/longhorn/longhorn/issues/2994 😂

[...]

PR #6087, opened 2023-06-08
feat(setting): keep snapshots for the backup

mantissahz:
Add a new global setting that allows users to keep the retain number of snapshots when the recurring job does a backup.

Ref: longhorn/longhorn#2997


innobead:
@mantissahz the message should not be `chore` which is used for non-prod code change. Suggest using `feat` instead.

innobead (approved):

innobead:
@mantissahz Please help update the chart doc update as well by referring to https://github.com/longhorn/longhorn/wiki/Longhorn-Setting-Update.

PR #6112, opened 2023-06-13
spdk: nvme-cli should be equal to or greater than 1.12

derekbit:
go-spdk-helper can support nvme-cli v2.0+.

Longhorn/longhorn#5744


innobead (approved):

innobead:
@mergify backport v1.5.x

Issue #6126, opened 2023-06-15
[IMPROVEMENT] Minor UX changes for Longhorn SPDK


innobead:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

Some changes for a better user experience (only user-facing part, we don't need to change anything internally). Also, for all changes below, need to update the doc accordingly if needed. 

## Describe the solution you'd like

- [x] Use v1 for the current engine, and v2 for the SPDK-based engine instead of Longhorn vs SPDK
- [x] Change `DO NOT CHANGE THIS SETTING WITH ATTACHED VOLUMES. Longhorn will try to block this setting update when there are attached volumes.`  Remove `try`, just said `Longhorn will block ..`.
- [x] Add a known issue about the actual size unable to display in the doc. Also, need a corresponding issue for later SPDK part improvement. 
  - tracked in https://github.com/longhorn/longhorn/issues/5947
- [x] Change replica number to 3 https://github.com/longhorn/longhorn/blob/master/examples/spdk/storageclass.yaml#L11, also the original storage class example for the old stack. 
- [x] We already highlight CPU usage for SPDK in doc but suggest emphasizing it to the setting as well to notice users actively. 

## Additional context

cc @longhorn/dev-data-plane

derekbit:
>  Use v1 for the current engine, and v2 for the SPDK-based engine instead of Longhorn vs SPDK

We need to change instance-manager, longhorn-manager, longhorn-ui and chart.

derekbit:
@smallteeths
Not sure if UI needs changes, can you help check it on UI part?
longhorn->v1
spdk->v2

derekbit:
@innobead 

`SPDK Data Engine` -> do we need to rename it to `V2 Data Engine`? Sounds a bit weird...

innobead:
> @innobead
> 
> `SPDK Data Engine` -> do we need to rename it to `V2 Data Engine`? Sounds a bit weird/confusing...

Let's discuss this later.

shuo-wu:
Found one more thing that may be improved:
When adding a new disk via UI, the UI will automatically add `/` at the end of the path if the disk type is not set to `Block` in advance.
I think this behavior is redundant... Can we remove it?

derekbit:
> Found one more thing that may be improved: When adding a new disk via UI, the UI will automatically add `/` at the end of the path if the disk type is not set to `Block` in advance. I think this behavior is redundant... Can we remove it?

cc @smallteeths

[...]

PR #6127, opened 2023-06-15
Rename BackendStoreDrivers

derekbit:
https://github.com/longhorn/longhorn/issues/6126

### DIFF
--- a/examples/v2/storageclass.yaml
+++ b/examples/v2/storageclass.yaml
@@ -1,14 +1,14 @@
 kind: StorageClass
 apiVersion: storage.k8s.io/v1
 metadata:
-  name: longhorn-spdk
+  name: longhorn-v2-data-engine
 provisioner: driver.longhorn.io
 allowVolumeExpansion: true
 reclaimPolicy: Delete
 volumeBindingMode: Immediate
 parameters:
   # backup, backingImage and snapshot related parameters are not supported in SPDK preview stage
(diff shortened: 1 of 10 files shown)


examples/v2/storageclass.yaml
```diff
@@ -1,17 +1,17 @@
...
-  name: longhorn-spdk
+  name: longhorn-v2-data-engine
 provisioner: driver.longhorn.io
 allowVolumeExpansion: true
 reclaimPolicy: Delete
 volumeBindingMode: Immediate
 parameters:
   # backup, backingImage and snapshot related parameters are not supported in SPDK preview stage
-  numberOfReplicas: "2"
+  numberOfReplicas: "3"
```
innobead, line 11: Please also help update https://github.com/longhorn/longhorn/blob/master/examples/storageclass.yaml

innobead (changes requested):
LGTM, just one comment.

examples/v2/storageclass.yaml
derekbit: It is for legacy volumes. Need to update as well?

innobead: yeah. just make it consistent.

derekbit: Renamed.

innobead (approved):

innobead:
@mergify backport v1.5.x

PR #6151, opened 2023-06-19
update: update instance-manager image version

mantissahz:
from v1_20230407 to v1_20230616

Ref: longhorn/longhorn#4927


innobead (approved):

innobead:
@mantissahz Also https://github.com/longhorn/longhorn-manager/tree/v1.3.x/deploy

PR #6193, opened 2023-06-26
feat(upgrade-responder): support requestSchema in setup script

c3y1huang:
https://github.com/longhorn/upgrade-responder/issues/47
https://github.com/longhorn/longhorn/issues/5235

### DIFF
--- a/dev/upgrade-responder/install.sh
+++ b/dev/upgrade-responder/install.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 UPGRADE_RESPONDER_REPO="https://github.com/longhorn/upgrade-responder.git"
+UPGRADE_RESPONDER_REPO_BRANCH="master"
 UPGRADE_RESPONDER_VALUE_YAML="upgrade-responder-value.yaml"
 UPGRADE_RESPONDER_IMAGE_REPO="longhornio/upgrade-responder"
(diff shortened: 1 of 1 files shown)


dev/upgrade-responder/install.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 UPGRADE_RESPONDER_REPO="https://github.com/longhorn/upgrade-responder.git"
+UPGRADE_RESPONDER_REPO_BRANCH="master"
```
innobead, line 4: nit: It's better to lock on a specific branch or release tag later.

dev/upgrade-responder/install.sh
```diff
@@ -59,12 +60,28 @@ secret:
...
+        "releaseDate": "2020-05-18T12:30:00Z",
+        "tags": ["latest"]
+      }]
+    }
+  requestSchema: |-
+    {
+      "appVersionSchema": {
+        "dataType": "string",
+        "maxLen": 200
+      }
```
innobead, line 77: Should we add extra tag/field schema here? otherwise they will not be validated and saved to the database?

innobead (approved):

innobead:
cc @PhanLe1010 @roger-ryao

innobead:
@mergify backport v1.5.x

Issue #6096, opened 2023-06-11
[TASK] Release 1.5.0


innobead:
**What's the task? Please describe.**
Action items for releasing v1.5.0

**Describe the items of the task (DoD, definition of done) you'd like.**
  - Pre-Release
    - [x] Regression test plan (manual) - @khushboo-rancher 
    - [x] Run e2e regression for pre-GA milestones (`install`, `upgrade`) - @yangchiu 
    - [x] Run security testing of container images for pre-GA milestones - @yangchiu 
    - [x] Verify longhorn chart PR to ensure all artifacts are ready for GA (`install`, `upgrade`)  @chriscchien 
    - [x] Run core testing (install, upgrade) for GA build from the previous patch and the last patch of the previous feature release (1.4.2). - @yangchiu 
  - Release
    - [x] Release longhorn/chart from the release branch to publish to ArtifactHub
    - [x] Release note
	     - [x] Deprecation note
	     - [x] Upgrade notes including highlighted notes, deprecation, compatible changes, and others impacting the current users
  - Post-Release
    - [x] Create a new release branch of manager/ui/tests/engine/longhorn instance-manager/share-manager/backing-image-manager when creating the RC1
    - [ ] Update https://github.com/longhorn/longhorn/blob/master/deploy/upgrade_responder_server/chart-values.yaml @PhanLe1010 
    - [ ] Add another request for the rancher charts for the next patch release (`1.5.1`) @rebeccazzzz  
  - Rancher charts: verify the chart is able to install & upgrade - @khushboo-rancher 
    - [x] rancher/image-mirrors update @weizhe0422 (@PhanLe1010 )
        - https://github.com/rancher/image-mirror/pull/412
    - [x] rancher/charts 2.7 branches for rancher marketplace @weizhe0422 (@PhanLe1010)
        - `dev-2.7`: https://github.com/rancher/charts/pull/2766

cc @longhorn/qa @longhorn/dev

yangchiu:
Verify longhorn chart PR to ensure all artifacts are ready for GA verified :

For both longhorn/longhorn v1.5.0-rc1 and longhorn/chart v1.5.0-rc1:

1. Fresh install - Pass
3. Upgrade from v1.4.2 to v1.5.0-rc1 - Pass
4. After install/upgrade to v1.5.0-rc1, below test items passed
   - Create a statefulset and write data
   - Upgrade to v1.5.0-rc1
   - Upgrade engine image
   - Scale down/up workload
[...]

PR #6214, opened 2023-06-28
release: 1.5.0

innobead:


chriscchien:
This PR verified pass and [here](https://github.com/longhorn/longhorn/issues/6096#issuecomment-1611121431) is the test result for reference, thank you.

shuo-wu (approved):

PR #6228, opened 2023-06-29
Modify Engine Identity Validation LEP for changes to longhorn-engine PR

ejweber:
#5845 

FYI @PhanLe1010. Made some changes to the LEP based on my final implementation in https://github.com/longhorn/longhorn-engine/pull/902.

### DIFF
--- a/enhancements/20230420-engine-identity-validation.md
+++ b/enhancements/20230420-engine-identity-validation.md
@@ -110,32 +110,32 @@ in this LEP are backwards compatible. All gRPC metadata validation is by demand 
 upgraded) client only injects some metadata (e.g. `volume-name` but not `instance-name`), the server only validates the
 metadata provided.
 
-Add an `instance-name` flag to the `longhorn controller <volume-name>` command (e.g. `longhorn controller <volume-name>
---instance-name <instance-name>`). This command already accepts the volume name, so an additional `volume-name` flag
-would be redundant. The longhorn-engine controller server remembers its volume and instance name.
+Add a global `volume-name` flag and a global `engine-instance-name` flag to the engine CLI (e.g. `longhorn -volume-name
+<volume-name> -engine-instance-name <engine-instance-name> <command> <args>`). Virtually all CLI commands create a
+controller client and these flags allow appropriate gRPC metadata to be injected into every client request. Requests
+that reach the wrong longhorn-engine controller server are rejected.
 
-Add a `volume-name` flag and an `instance-name` flag to the `longhorn replica <directory>` command (e.g. `longhorn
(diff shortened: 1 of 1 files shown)


enhancements/20230420-engine-identity-validation.md
```diff
@@ -318,7 +318,7 @@ validation.
 
 #### Longhorn-Manager Integration
 
-Ensure the engine and replica controllers launch engine and replica processes with `--volume-name` and `--instance-name`
+Ensure the engine and replica controllers launch engine and replica processes with `-volume-name` and `-instance-name`
```
shuo-wu, line 321: NIT: and `-engine-instance-name`/`replica-instance-name`?
Similar below. There is no flag name exactly `instance-name`, right? Otherwise, it's confusing sometimes...

enhancements/20230420-engine-identity-validation.md
```diff
@@ -367,10 +370,26 @@ This is really a better fit for a negative testing scenario (do something that w
...
 
-I'm unfamiliar with our engine integration tests, but it should be fairly easy to create negative tests that verify the
-expected behavior.
+- All controller and replica processes are created with the information needed for identity validation.
+- It is convenient to create controller and replica clients with the information needed for identity validation.
+- gRPC metadata is automatically injected into controller and replica client requests when clients have the necessary
+  information.
+
+Do not modify the behavior of existing tests. Since these tests were using clients with identity validation information,
+no identity validation is performed.
```
shuo-wu, line 383: Sorry... For existing tests, will the client contain meta info but somehow the processes do not validate it?

ejweber: Right. Identity validation only takes effect when the new flags are passed to a server AND metadata is injected by the client. While I modified some of the functions/fixtures in the integration tests to allow the new flags to be passed and created an interceptor that can inject metadata client-side, I only use these features in the new identity integration tests.

In most existing integration tests, even though the server-side interceptor is enabled (by the changes to longhorn-engine), it does not reject any requests, either because the appropriate flags weren't used when launching a process or the client used to communicate with the process didn't inject metadata.

In practice, we will always launch processes with the flags (because we will modify longhorn-manager to do so) and always inject metadata (because the longhorn-engine Go client has appropriate fields).

enhancements/20230420-engine-identity-validation.md
ejweber: Good catch. This has evolved a bit since I originally wrote the LEP. Changed as advised in a few places (although the actual gRPC metadata we inject is labelled `instance-name` generically, so I did not change those mentions.)

ejweber:
@longhorn/dev Please re-review this when you are able.

innobead (approved):

PR #6243, opened 2023-07-06
Cleanup

james-munson:
#6316 
Bring scripts/cleanup.sh up to date.


shuo-wu (commented):
I just realized that this is an outdated script, which is launched during Longhorn v0.3.x. Does this still work? Now we have setting `Deleting Confirmation Flag` that protects the deletion. And we have introduced more new CRDs like recurring jobs, backups, snapshots, etc. 
Besides, there are lots of finalizers blocking the CR removal. Not sure if directly deleting CRDs will let the corresponding CRs be cleaned up correctly.

james-munson:
I was doing a lot of installs and uninstalls on a cluster, and I noticed that even after I remembered to change the confirmation setting, when I ran an uninstall via uninstall.yaml there were a lot of things left over, including some that caused a subsequent installation using Helm to fail.  In my case, even ignoring events, settings, and leases, there were still these left around:

`NAME                                               READY   STATUS      RESTARTS   AGE
pod/longhorn-admission-webhook-7c9b76b495-28jjg    1/1     Running     0          36h
pod/longhorn-admission-webhook-7c9b76b495-l4kbj    1/1     Running     0          36h
pod/longhorn-conversion-webhook-5dfc7849d4-4cv4s   1/1     Running     0          36h
pod/longhorn-conversion-webhook-5dfc7849d4-fbrp4   1/1     Running     0          36h
pod/longhorn-recovery-backend-6777c85b75-kp6p7     1/1     Running     0          36h
pod/longhorn-recovery-backend-6777c85b75-mnvjt     1/1     Running     0          36h
pod/longhorn-ui-7b7c86c7cf-jvcs6                   1/1     Running     0          36h
pod/longhorn-ui-7b7c86c7cf-lhwj9                   1/1     Running     0          36h
pod/longhorn-uninstall-xncz6                       0/1     Completed   0          2m5s

NAME                                  TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
service/longhorn-admission-webhook    ClusterIP   10.43.124.143   <none>        9443/TCP   36h
service/longhorn-backend              ClusterIP   10.43.53.91     <none>        9500/TCP   36h
service/longhorn-conversion-webhook   ClusterIP   10.43.2.171     <none>        9443/TCP   36h
service/longhorn-engine-manager       ClusterIP   None            <none>        <none>     36h
service/longhorn-frontend             ClusterIP   10.43.60.26     <none>        80/TCP     36h
service/longhorn-recovery-backend     ClusterIP   10.43.78.99     <none>        9600/TCP   36h
service/longhorn-replica-manager      ClusterIP   None            <none>        <none>     36h

NAME                                          READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/longhorn-admission-webhook    2/2     2            2           36h
deployment.apps/longhorn-conversion-webhook   2/2     2            2           36h
deployment.apps/longhorn-recovery-backend     2/2     2            2           36h
deployment.apps/longhorn-ui                   2/2     2            2           36h

NAME                                                     DESIRED   CURRENT   READY   AGE
replicaset.apps/longhorn-admission-webhook-7c9b76b495    2         2         2       36h
replicaset.apps/longhorn-conversion-webhook-5dfc7849d4   2         2         2       36h
replicaset.apps/longhorn-recovery-backend-6777c85b75     2         2         2       36h
replicaset.apps/longhorn-ui-7b7c86c7cf                   2         2         2       36h`

I figured that it would be a good idea to have a foolproof "hammer" script that would clean up everything regardless, and that seemed to be the purpose of cleanup.sh.  Admittedly, it hadn't been touched for a while.

If you think this is a bad idea, I can drop the PR, and then the right thing to do is delete the obsolete script.

ejweber:
I've been thinking about this a bit @james-munson, and it seems your bad uninstall experience might be attributable to using one version of `uninstall.yaml` to clean up another version of Longhorn.

For example, if you used `uninstall.yaml v1.5.0+` to clean up `Longhorn v1.4.2-`, you would be missing the logic to delete `v1.4.2`'s recovery backend, admission webook, and conversion webhook deployments, as these no longer exist in `v1.5.0`.

https://github.com/longhorn/longhorn-manager/commit/2a0f924060e57252f218b23aacbf522cd509baa6#diff-9809693b91d35e4d0e3e4658840003aa5555c1f73772b141cb227add23dd8bfb

Since this cleanup script deletes the longhorn-system namespace, it is likely capable of getting rid of these deployments, which are deleted in a cascading fashion once the namespace is deleted.

PhanLe1010:
This PR maybe helpful for situation like this https://github.com/longhorn/longhorn/issues/6259#issuecomment-1627470852

scripts/cleanup.sh
```diff
@@ -51,14 +50,35 @@ remove_workloads() {
...
+
 remove_crd_instances
 remove_driver
 remove_workloads
 remove_crds
+remove_unnamespaced_resources
+
+# Last, remove the namespace itself (and implicitly some remaining resource types such as events and leases.)
+# Note: this will still leave some items behind, such as persistentvolumeclaim (in default namespace)
+# TODO: ensure this does not toast backups.  See longhorn-manager/controller/uninstall_controller.go#L405-L407
```
ejweber, line 82: We should probably resolve this TODO before merging. I think it would be simple enough to update the backup-target setting at the top of the script (or potentially to just delete it, since we'll be blowing it away momentarily anyway).

scripts/cleanup.sh
```diff
@@ -20,14 +20,14 @@ remove_and_wait() {
...
-  remove_and_wait settings.longhorn.rancher.io
+  remove_and_wait volumes.longhorn.io
+  # engines and replicas should be no-ops, already done by volume delete.  See https://github.com/rancher/longhorn/issues/273
+  remove_and_wait engines.longhorn.io
+  remove_and_wait replicas.longhorn.io
+  remove_and_wait engineimages.longhorn.io
+  remove_and_wait settings.longhorn.io
   # do this one last; manager crashes
-  remove_and_wait nodes.longhorn.rancher.io
+  remove_and_wait nodes.longhorn.io
```
ejweber, line 30: There are many more CRDs than this in Longhorn today. I am guessing in many cases we don't need to delete their instances in this function because they typically all disappear when volumes are deleted?

What about BackingImage and BackingImageDataSource? These CRDs have finalizers and are not associated with any particular volume, so I think they will remain. Once longhorn-manager is deleted in remove_workloads, I think there will be nothing left to remove the finalizers and the script will be stuck when it tries to delete the namespace.

ejweber (changes requested):
In general, this certainly looks like an improvement on the existing script (which is clearly broken today given these changes). I think we need to resolve a couple of issues to be safe to merge.

That being said, I don't see how this script achieves anything substantially different from uninstall.yaml. It seems that usually when we (or customers) have uninstallation issues, we are stuck in a half-uninstalled state in which longhorn-manager is no longer available on the cluster, but there are resources remaining. Without longhorn-manager to remove finalizers, there is nothing we can do but attempt to track down all resources with finalizers and remove them manually (a frustrating process!). It looks like this script would also be unable to proceed in this situation (e.g. we will wait for volumes to delete forever because the volumes have finalizers). In my opinion, we should remove this script from the repo and make improvements to uninstall.yaml if we feel it needs them.

I would be interested in a _different_ script that could, for example, find all terminating Longhorn CRs with finalizers and remove the finalizer from each. This would have value in situations we (or customers) just want to blow Longhorn away, but are stuck with a failed uninstallation.

scripts/cleanup.sh
```diff
@@ -51,14 +50,35 @@ remove_workloads() {
...
+remove_unnamespaced_resources() {
+  remove_resource storageclass
+  remove_resource clusterrolebinding
+  remove_resource clusterrole
+}
+
 remove_crd_instances
 remove_driver
 remove_workloads
 remove_crds
```
PhanLe1010, line 77: After removing the longhorn-manger daemonset, at this moment if any resource in `longhorn-system` namespace has a finalizer we can directly remove it because no one will take care of them now. 

If we don't remove the finalizers at this step, the `remove_unnamespaced_resources` function below might stuck forever

scripts/cleanup.sh
PhanLe1010: Agree. We should clear the backup-target setting at the beginning of the script to disconnect from backup store

PhanLe1010 (changes requested):

scripts/cleanup.sh
ejweber: This would do a lot to address my overarching concern (that this script can't handle broken uninstallations).

shuo-wu:
> It looks like this script would also be unable to proceed in this situation (e.g. we will wait for volumes to delete forever because the volumes have finalizers). In my opinion, we should remove this script from the repo and make improvements to uninstall.yaml if we feel it needs them.

Agree. This script does not make too much sense since it cannot handle the broken uninstallation/Longhorn system cleanup. I prefer to remove it then open a separate ticket to improve the uninstallation.

james-munson:
Revised to simply remove the obsolete script.

shuo-wu (approved):

PhanLe1010 (approved):
LGTM

Agree that we should remove this script and improve the `uninstall.yaml` so that user can just re-run it

ejweber (approved):
LGTM. Thanks for your work on this! Hopefully we can leverage it when working on `uninstall.yaml` for a quick and effective turnaround.

Issue #6257, opened 2023-07-07
[QUESTION] Repetetive warnings and errors in a new longhorn setup


jaqua:
I've setup a completely new k8s cluster with longhorn 1.5.0.
There is nothing special running on that cluster, but I do get multiple warnings like:

    time="2023-07-07T14:51:16Z" level=warning msg="Ignoring recurring job labels on Volume pvc-1e1565de-60a0-4dd8-9e7b-30b5d5a90772 PVC prometheus-community-server due to missing source label" controller=longhorn-volume node=worker3

and repetitive errors like this:

    time="2023-07-07T14:49:46Z" level=info msg="Forwarding request to 10.42.2.155:9500"
    time="2023-07-07T14:49:46Z" level=info msg="CreateSnapshot: req: {\"name\":\"snapshot-275ba4a0-10aa-4024-87d8-2f82a37fb5b0\",\"parameters\":{\"type\":\"bak\"},\"source_volume_id\":\"pvc-7c074207-69ba-44de-af57-fba9cc564130\"}"
    time="2023-07-07T14:49:46Z" level=info msg="GetPluginInfo: rsp: {\"name\":\"driver.longhorn.io\",\"vendor_version\":\"v1.5.0\"}"
    time="2023-07-07T14:49:46Z" level=info msg="GetPluginInfo: req: {}"
    I0707 14:49:46.203466       1 snapshot_controller.go:291] createSnapshotWrapper: Creating snapshot for content snapcontent-275ba4a0-10aa-4024-87d8-2f82a37fb5b0 through the plugin ...
    E0707 14:49:46.203420       1 snapshot_controller_base.go:283] could not sync content "snapcontent-275ba4a0-10aa-4024-87d8-2f82a37fb5b0": failed to take snapshot of the volume pvc-7c074207-69ba-44de-af57-fba9cc564130: "rpc error: code = NotFound desc = volume pvc-7c074207-69ba-44de-af57-fba9cc564130 not found"
    E0707 14:49:46.203323       1 snapshot_controller.go:124] checkandUpdateContentStatus [snapcontent-275ba4a0-10aa-4024-87d8-2f82a37fb5b0]: error occurred failed to take snapshot of the volume pvc-7c074207-69ba-44de-af57-fba9cc564130: "rpc error: code = NotFound desc = volume pvc-7c074207-69ba-44de-af57-fba9cc564130 not found"
[...]

PR #6267, opened 2023-07-10
fix(chart): update default setting log level

c3y1huang:
https://github.com/longhorn/longhorn/issues/6257

### DIFF
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -574,7 +574,7 @@ Set the value to **0** to disable backup restore."
     description: "The log level Panic, Fatal, Error, Warn, Info, Debug, Trace used in longhorn manager. By default Debug."
     group: "Longhorn Default Settings"
     type: string
-    default: "Debug"
+    default: "Info"
 - variable: defaultSettings.kubernetesClusterAutoscalerEnabled
   label: Kubernetes Cluster Autoscaler Enabled (Experimental)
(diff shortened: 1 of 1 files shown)


derekbit (approved):

shuo-wu:
Do we need to change this after fixing the PVC recurring job label issue?

c3y1huang:
> Do we need to change this after fixing the PVC recurring job label issue?

We've updated the log level to "info" so the recurring job source label log at the "debug" level doesn't overwhelm when the cluster is at the default log level setting.

So this is to reflect on the log level change. Any concerns?

shuo-wu:
OK. Just noticed that the log level is changed to `debug` rather than `trace`.

As we discussed before, some of the important logs were level `debug`. Suppressing them would make the support bundle debugging more difficult. 
Not sure we already tune all logs and it's fine to change the default level to `debug` now. cc @derekbit

derekbit:
Most of debug level logs are adjusted in https://github.com/longhorn/longhorn-manager/pull/2059 and previous commits.
The remaining ones are "requeue....", "...backoff" and so on.
cc @shuo-wu 

image

image

ChanYiLin (approved):

innobead:
@shuo-wu any concern here?

innobead:
@mergify backport v1.5.x

Issue #3161, opened 2021-10-16
[FEATURE] Talos support


runningman84:
**Is your feature request related to a problem? Please describe.**
Talos does not support longhorn due to missing iscsi support:
https://github.com/talos-systems/talos/issues/3057

**Describe the solution you'd like**
Other supported storage providers in talos use the nvmf protocol to overcome this issue.

**Describe alternatives you've considered**
I would really like to continue using longhorn, the other solutions have their our limitations like needing a seperated disk.

**Additional context**
With an uncertain future of k3os many users might take a look at talos.

HaveFun83:
iscsi-tools are now support via system extensions on the talos side. 
Maybe the settings how longhorn communicate with the iscsi-daemon can be made configurable?

https://www.talos.dev/v1.1/kubernetes-guides/configuration/replicated-local-storage-with-openebs-jiva/

skandragon:
I fear it's more than just the settings.  I fear it's also that bash and other tools are used.  There's a somewhat long list of tools:

innobead:
ref: https://www.talos.dev/v1.2/kubernetes-guides/configuration/replicated-local-storage-with-openebs-jiva/

c3y1huang:
Note:
- The `iscsiadm` still needs to be executed using `nsenter` approach from its ancestor namespace.
- We should replace all dependent binary invocations. To achieve this, we need to implement namespace entering using `setns` or alternatives.
- We also need to replace the binary dependencies in the automation tests.

I've written a PoC to test the feasibility of the changes. And will revisit this issue later to provide a proper LEP.

japtain-cack:
I'm really looking forward to this. I've been trying out linstor, via the piraeus operator, and it just doesn't work as well as longhorn. Sure, it might theoretically be faster, and that does seem to be the case, but it's much more complex and I've not had good luck with loosing a disk during testing. My pod shouldn't be stopped when a single disk is lost. Longhorn never failed me in the past, but I have not been able to use it since I moved to Talos/Sidero.

c3y1huang:
Regression test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4813/

c3y1huang:
Tests results (full):
- Regression: WIP
- Talos Linux: WIP

camaeel:
Is there any estimated time when Longhorn 1.6 can be released?

c3y1huang:
> Is there any estimated time when Longhorn 1.6 can be released?

[...]

PR #6325, opened 2023-07-14
feat(lep): Talos Linux support design

c3y1huang:
https://github.com/longhorn/longhorn/issues/3161

### DIFF
--- a/enhancements/20230814-talos-linux-support.md
+++ b/enhancements/20230814-talos-linux-support.md
@@ -1,0 +1,108 @@
+# Talos Linux Support
+
+## Summary
+
+[Talos Linux is not based on X distro](https://www.talos.dev/v1.4/learn-more/philosophy/#not-based-on-x-distro)
+
+>Talos Linux isn’t based on any other distribution. We think of ourselves as being the second-generation of container-optimised operating systems, where things like CoreOS, Flatcar, and Rancher represent the first generation (but the technology is not derived from any of those.)
+>
+>Talos Linux is actually a ground-up rewrite of the userspace, from PID 1. We run the Linux kernel, but everything downstream of that is our own custom code, written in Go, rigorously-tested, and published as an immutable, integrated image. The Linux kernel launches what we call machined, for instance, not systemd. There is no systemd on our system. There are no GNU utilities, no shell, no SSH, no packages, nothing you could associate with any other distribution.
+
+Currently, Longhorn (at version v1.5.x) does not support Talos Linux as one of the [operating systems (OS)](https://longhorn.io/docs/1.5.0/best-practices/#operating-system) due to reliance on host binaries like BASH and iscsiadm.
+
+The goal of this proposal is to enable Longhorn to be installed and functional on Talos Linux clusters.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/3161
+
+## Motivation
+
+We have been approached by and observed a number of Talos Linux users who wish to utilize Longhorn. However, the lack of support for Talos Linux prevents them from using Longhorn as their storage solution.
+
+### Goals
+
+The primary goal of this proposal is to introduce support for Talos Linux, allowing users to installed and operate Longhorn on Talos Linux clusters.
+
+### Non-goals [optional]
+
+`None`
+
+## Proposal
+
+Whenever possible, replace host binary dependencies with alternative solutions:
+- Develop a common thread namespace switch package that can be utilized by projects requiring interaction with the host. This is necessary due to the absence of GNU utilities in Talos Linux. For example, Longhorn is unable to utilize nsenter and execute binaries in [lockCmd](https://github.com/longhorn/nsfilelock/blob/2315476ea52e13b4112a1cdb930626f8aa848d09/nsfilelock.go#L67-L68).
+- Modify the invocation of the `iscsiadm` binary to execute within the `iscsid` process namespace,
+- In cases where replacing binary invocations is not feasible, leverage the Talos Linux `kubelet` namespace specifically for Talos Linux. For other operating systems, maintain the existing approach.
+
+### User Stories
+
+As a user running Talos Linux, I want to be able to use Longhorn as my storage solution.
+
+Currently, Longhorn does not support Talos Linux, so I am unable to utilize its features.
+
+With this enhancement, I will be able to install and use Longhorn on my Talos Linux cluster.
+
+### User Experience In Detail
+
+1. Provision a Talos Linux cluster.
+1. Apply the required machine configurations to the cluster nodes:
+   - Install the iscsi-tool system extension.
+   - Install the util-linux system extension.
+   - Add the /var/lib/longhorn extra mount.
+1. Install Longhorn on the Talos Linux cluster.
+1. Once Longhorn is successfully installed, access and utilize Longhorn just like users on other supported operating systems.
+
+### API changes
+
+`None`
+
+## Design
+
+### Thread Namespace Switch
+
+- Create a package in the [longhorn/go-common-lib](https://github.com/longhorn/go-common-libs) repository that can be imported by other projects.
+- The targeting function execute within a goroutine and utilize the `unix.Setns` to switch to different namespaces.
+- As Golang being a multi-threaded language, to ensure exclusive use of the thread, lock the thread when switching to a different namespace.
+- Once the targeting function completes its execution, the thread will switch back to the original namespace and unlock the thread to allow other goroutines to execute.
+
+> The primary goal of this implementation is to replace the usage of GNU utilities with Go-based implementations where applicable, particularly in areas such as file handling.
+
+### Binary Dependencies
+
+#### Isciadm
+
+Longhorn currently assumes that `iscsid` is running on the host. However, in Talos Linux, `iscsid` runs as a Talos extension service in a different namespace.
+
+To address this, modify the `iscsiadm` binary invocation to execute within the `iscsid` namespace using `nsenter`.
+
+#### fstrim
+
+~Instead of rewriting the `fstrim` binary dependency, leverage the existing `fstrim` binary within the `kubelet` namespace. However, considering that the `kubelet` namespace might not be present in other operating systems, Longhorn needs to switch to the host namespace to retrieve the OS distribution and determine if it is a Talos Linux. If the host is identified as Talos Linux, Longhorn can then switch to the `kubelet` namespace to utilize the existing `fstrim` binary.~
+
+Keep the binary execution in the host namespace, following a discussion with @frezbo from @siderolabs. The Talos team has proposed an approach to include the fstrim in the host as part of `util-linux` extension.
+
+#### cryptsetup
+
+Keep the current implementation of the binary execution in the host namespace, as the `cryptsetup` binary comes [pre-installed](https://github.com/siderolabs/pkgs/blob/release-1.4/Makefile#L40) in Talos Linux, for the [Disk Encryption](https://www.talos.dev/v1.4/talos-guides/configuration/disk-encryption/).
+
+### Project Adpaptions
+
+- Replace the `nsenter` path with the `iscsid` proc path when invoking `iscsiadm` binary.
+- Replace the `nsenter` path with the `kubelet` proc path when invoking other dependency binaries. This change will be applied only if the operating system distribution is identified as Talos Linux, as other distributions may not have a running `kubelet` process.
+- For binary invocations are not part of the [binary dependencies](#binary-dependencies), replace them with appropriate Golang libraries.
+- Utilize the [thread namespace switch implementation](#thread-namespace-switch) to handle the necessary namespace switching when interacting with the host.
+
+### Test plan
+
+1. Update existing test cases that rely on `nsenter` to either correct the namespace or replace them with appropriate Python libraries.
+1. Perform regression testing by running existing test cases to ensure that no regressions are introduced.
+1. Introduce a new pipeline for testing Longhorn on Talos Linux clusters in https://ci.longhorn.io/.
+
+### Upgrade strategy
+
+`None`
+
+## Note [optional]
+
+`None`


shuo-wu (approved):
LGTM

enhancements/20230714-talos-linux-support.md
```diff
@@ -0,0 +1,93 @@
...
+
+The primary goal of this proposal is to introduce support for Talos Linux, allowing users to installed and operate Longhorn on Talos Linux clusters.
+
+### Non-goals [optional]
+
+`None`
+
+## Proposal
+
+- Develop a common thread namespace switch package that can be utilized by projects requiring interaction with the host.
```
derekbit, line 27: Can you elaborate more on why we need a thread namespace switch rather than use the existing `nsenter`? I remember you explained it before, but I forgot the details.

enhancements/20230714-talos-linux-support.md
```diff
@@ -0,0 +1,93 @@
+# Talos Linux Support
```
derekbit, line 1: If possible, can you add a short section or paragraph for introducing the difference between Talos and OS and why we need to do some changes.

enhancements/20230714-talos-linux-support.md
```diff
@@ -0,0 +1,103 @@
...
+### API changes
+
+`None`
+
+## Design
+
+### Thread Namespace Switch
+
+- Create a package in the [longhorn/go-common-lib](https://github.com/longhorn/go-common-libs) repository that can be imported by other projects.
+- The targeting function execute within a goroutine and utilize the `unix.Setns` to switch to different namespaces.
```
innobead, line 64: This is interesting to use `setns` at runtime.

enhancements/20230714-talos-linux-support.md
```diff
@@ -0,0 +1,103 @@
...
+- As Golang being a multi-threaded language, to ensure exclusive use of the thread, lock the thread when switching to a different namespace.
+- Once the targeting function completes its execution, the thread will switch back to the original namespace and unlock the thread to allow other goroutines to execute.
+
+### Binary Dependencies
+
+#### Isciadm
+
+Longhorn currently assumes that `iscsid` is running on the host. However, in Talos Linux, `iscsid` runs as a Talos extension service in a different namespace.
```
innobead, line 72: v2 data engine is relying on nvmecli on host, so we will have same need here. How about NFS, CIFS client here? do we need to have another Talos extension?

enhancements/20230714-talos-linux-support.md
```diff
@@ -0,0 +1,103 @@
...
+
+To address this, modify the `iscsiadm` binary invocation to execute within the `iscsid` namespace using `nsenter`.
+
+#### fstrim
+
+Instead of rewriting the `fstrim` binary dependency, leverage the existing `fstrim` binary within the `kubelet` namespace. However, considering that the `kubelet` namespace might not be present in other operating systems, Longhorn needs to switch to the host namespace to retrieve the OS distribution and determine if it is a Talos Linux. If the host is identified as Talos Linux, Longhorn can then switch to the `kubelet` namespace to utilize the existing `fstrim` binary.
```
innobead, line 78: or we don't need to be that intelligent, but rather rely on the info provided by users like specify this setup is talos, so we will run it differently. 

This could be helpful if we need to support other specific purpose OS.

innobead (changes requested):
What's happening if the iscsid extention pod unexpectedly restarted? Would it cause the unexpected volume detached?

enhancements/20230714-talos-linux-support.md
c3y1huang: > v2 data engine is relying on nvmecli on host, so we will have same need here.
Yes, I assume we will have the same need for the v2 data engine. However, I think the support for v2 data engine should be evaluated later since the feature is still in preview phase.

> How about NFS, CIFS client here? do we need to have another Talos extension?
In regards to the NFS client and CIFS client, [they are preinstalled in the kubelet image](https://github.com/siderolabs/kubelet/blob/v1.27.4/Dockerfile#L33-L52).

enhancements/20230714-talos-linux-support.md
c3y1huang: As there is only one OS required for this kind of handling, I prefer not to add a new setting for user to manually indicate the OS distro. Given that we already gather OS distribution information for Longhorn's usage data collection, my preferred approach would be to cache the OS distro and reuse it for the necessary dependency binaries.

enhancements/20230814-talos-linux-support.md
```diff
@@ -0,0 +1,105 @@
...
+
+#### Isciadm
+
+Longhorn currently assumes that `iscsid` is running on the host. However, in Talos Linux, `iscsid` runs as a Talos extension service in a different namespace.
+
+To address this, modify the `iscsiadm` binary invocation to execute within the `iscsid` namespace using `nsenter`.
+
+#### fstrim
+
+Instead of rewriting the `fstrim` binary dependency, leverage the existing `fstrim` binary within the `kubelet` namespace. However, considering that the `kubelet` namespace might not be present in other operating systems, Longhorn needs to switch to the host namespace to retrieve the OS distribution and determine if it is a Talos Linux. If the host is identified as Talos Linux, Longhorn can then switch to the `kubelet` namespace to utilize the existing `fstrim` binary.
```
frezbo, line 80: from Team talos, just saw this, fstrim could be included in the host as an extension though, if that makes things easier rather than switching to the kubelet container namespace

frezbo: yes, sure, I can put an example tomorrow, basically something based of this in the talos extensions repo https://github.com/siderolabs/pkgs/blob/main/util-linux/pkg.yaml, just have to enable the binaries

innobead: Thanks @frezbo :+1:

c3y1huang: @frezbo sounds good to me.

@innobead Having the fstrim execute on the host rather than within the kubelet namespace is a small change on the feature PRs. This shouldn't be a blocker for the PR reviews.

frezbo: I've created a draft PR here: https://github.com/siderolabs/extensions/pull/216

Also pushed the extensions here: `ghcr.io/frezbo/util-linux:2.39.1`, this provides `fstrim` at `/usr/local/sbin/fstrim`

c3y1huang: The extension image works for me.
```
> talosctl --talosconfig talosconfig -e 54.169.155.201 -n 10.0.28.181 get extensions
NODE          NAMESPACE   TYPE              ID                                          VERSION   NAME          VERSION
10.0.28.181   runtime     ExtensionStatus   000.ghcr.io-siderolabs-iscsi-tools-v0.1.4   1         iscsi-tools   v0.1.4
10.0.28.181   runtime     ExtensionStatus   001.ghcr.io-frezbo-util-linux-2.39.1        1         util-linux    $VERSION
> k exec -it longhorn-test -- ./run.sh -k test_filesystem_trim
Defaulted container "longhorn-test" out of: longhorn-test, longhorn-test-report
================================================ test session starts =================================================
platform linux -- Python 3.9.17, pytest-5.3.1, py-1.11.0, pluggy-0.13.1 -- /usr/bin/python3.9
cachedir: .pytest_cache
rootdir: /integration, inifile: pytest.ini
plugins: order-1.0.1, repeat-0.9.1
collected 349 items / 347 deselected / 2 selected                                                                    

test_basic.py::test_filesystem_trim[ext4] PASSED                                                               [ 50%]
test_basic.py::test_filesystem_trim[xfs] PASSED                                                                [100%]

================================================== warnings summary ==================================================
/usr/local/lib/python3.9/site-packages/_pytest/mark/structures.py:323
  /usr/local/lib/python3.9/site-packages/_pytest/mark/structures.py:323: PytestUnknownMarkWarning: Unknown pytest.mark.volume_backup_restore - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/latest/mark.html
    warnings.warn(

-- Docs: https://docs.pytest.org/en/latest/warnings.html
============================== 2 passed, 347 deselected, 1 warning in 149.90s (0:02:29) ==============================
```

@frezbo When can we expect this to be released?

c3y1huang: LEP updated.

frezbo: This would be part of Talos 1.6, I guess Dec, after k8s 1.29 release

c3y1huang: Thanks @frezbo !

@innobead , here are the 2 options before Talos 1.6:
1. Stick with the previous kubelet proc for now and later change to use the host namespace once Talos 1.6 is released.
1. Implement directly with the host namespace. In the interim period before Talos v1.6 becomes available, we can test using ghcr.io/frezbo/util-linux:2.39.1.

frezbo: since `ghcr.io/frezbo/util-linux:2.39.1` is not talos version dependent, once I get this merged, that image could be used with older talos too. 

Since util-linux is not talos version dependent (unlike kernel modules, it can be used after merging. I'll get this merged in today.

frezbo: Sorry for the confusion

frezbo: `ghcr.io/siderolabs/util-linux-tools:v1.6.0-alpha.0-4-gd9145f9` this is published and can be used

c3y1huang: Experimenting with the latest Talos version (v1.5.5), I've encountered challenges. The approach of utilizing `.machine.install` for patching node system extensions no longer seems effective.

And, utilizing the [image factory](https://www.talos.dev/v1.5/talos-guides/install/boot-assets/#image-factory) is not possible due to the absence of the `util-linux-tool`. 

So my last option is to use the [imager](https://www.talos.dev/v1.5/talos-guides/install/boot-assets/#imager), and it leading errors while generating the AWS disk image with extensions:
```
error validating extension "iscsi-tools": path "/usr/include/libopeniscsiusr.h" is not allowed in extensions
Error: error validating extension "iscsi-tools": path "/usr/include/libopeniscsiusr.h" is not allowed in extensions
```

frezbo: https://github.com/siderolabs/extensions/pull/265 sorry, already a fix is underway

c3y1huang: @frezbo is there a way for v1.5.x to achieve `.machine.install` like v1.4.x so we don't need a customized Talos AWS disk image with the extensions?

frezbo: There's [factory.talos.dev](https://factory.talos.dev/), that would be the official way going forward, also the aws image downloaded from factory with iscsi-tools is not a broken one

c3y1huang: Is there a plan to include `util-linux-tool` in the factory?

c3y1huang: > There's [factory.talos.dev](https://factory.talos.dev/), that would be the official way going forward, also the aws image downloaded from factory with iscsi-tools is not a broken one

To my knowledge, to use the AWS disk image in the factory, we need to upload it to AMI, is there are way to avoid that?

c3y1huang: Wipe my previous comment, I see v1.5.x now have `@sha256:<digest>` included.

frezbo: > Is there a plan to include util-linux-tool in the factory?

should be with Talos 1.6, you can replace the version in the link with `1.6.0-alpha.0`

> To my knowledge, to use the AWS disk image in the factory, we need to upload it to AMI

yes, no way around for now

frezbo: >  I see v1.5.x now have @sha256:<digest> included

yes, the `extensions` oci image has the list of all extension images and their sha's

c3y1huang: @frezbo I see. Is there a plan to have the extensions hot-pluggable, without the need to rebuild the image every time we add or remove an extension?

frezbo: it's part of initramfs, so it needs to be rebuild

enhancements/20230814-talos-linux-support.md
```diff
@@ -0,0 +1,108 @@
...
+`None`
+
+## Design
+
+### Thread Namespace Switch
+
+- Create a package in the [longhorn/go-common-lib](https://github.com/longhorn/go-common-libs) repository that can be imported by other projects.
+- The targeting function execute within a goroutine and utilize the `unix.Setns` to switch to different namespaces.
+- As Golang being a multi-threaded language, to ensure exclusive use of the thread, lock the thread when switching to a different namespace.
+- Once the targeting function completes its execution, the thread will switch back to the original namespace and unlock the thread to allow other goroutines to execute.
```
innobead, line 67: This should mean each namespace-based operation should be atomic instead of wrongly switching to another namespace during an operation.

enhancements/20230814-talos-linux-support.md
```diff
@@ -0,0 +1,108 @@
...
+
```
innobead, line 68: Would this cause performance issues if the function runs for a long time? Using a lock should occur soon to prevent any blocking operations? Are there any calls we need to worry about?

enhancements/20230814-talos-linux-support.md
```diff
@@ -0,0 +1,108 @@
...
+Longhorn currently assumes that `iscsid` is running on the host. However, in Talos Linux, `iscsid` runs as a Talos extension service in a different namespace.
+
+To address this, modify the `iscsiadm` binary invocation to execute within the `iscsid` namespace using `nsenter`.
+
+#### fstrim
+
+~Instead of rewriting the `fstrim` binary dependency, leverage the existing `fstrim` binary within the `kubelet` namespace. However, considering that the `kubelet` namespace might not be present in other operating systems, Longhorn needs to switch to the host namespace to retrieve the OS distribution and determine if it is a Talos Linux. If the host is identified as Talos Linux, Longhorn can then switch to the `kubelet` namespace to utilize the existing `fstrim` binary.~
+
+Keep the binary execution in the host namespace, following a discussion with @frezbo from @siderolabs. The Talos team has proposed an approach to include the fstrim in the host as part of `util-linux` extension.
+
```
innobead, line 84: This is cool. We probably require more system CLIs in the future, especially v2.

enhancements/20230814-talos-linux-support.md
```diff
@@ -0,0 +1,108 @@
...
+Keep the current implementation of the binary execution in the host namespace, as the `cryptsetup` binary comes [pre-installed](https://github.com/siderolabs/pkgs/blob/release-1.4/Makefile#L40) in Talos Linux, for the [Disk Encryption](https://www.talos.dev/v1.4/talos-guides/configuration/disk-encryption/).
+
+### Project Adpaptions
+
+- Replace the `nsenter` path with the `iscsid` proc path when invoking `iscsiadm` binary.
+- Replace the `nsenter` path with the `kubelet` proc path when invoking other dependency binaries. This change will be applied only if the operating system distribution is identified as Talos Linux, as other distributions may not have a running `kubelet` process.
+- For binary invocations are not part of the [binary dependencies](#binary-dependencies), replace them with appropriate Golang libraries.
+- Utilize the [thread namespace switch implementation](#thread-namespace-switch) to handle the necessary namespace switching when interacting with the host.
+
```
innobead, line 95: As mentioned above, need to see if longer-locking would impact any performance.

innobead (approved):
In general, LGTM. A few comments to think through, and see if anything we need to adjust/improve later.

Issue #6274, opened 2023-07-10
[RELEASE] 1.5.1 


innobead:
**What's the task? Please describe.**
Action items for releasing v1.5.1

**Describe the items of the task (DoD, definition of done) you'd like.**
  - Pre-Release
    - [x] Regression test plan (manual) - @khushboo-rancher 
    - [x] Run e2e regression for pre-GA milestones (`install`, `upgrade`) - @yangchiu 
    - [x] Run security testing of container images for pre-GA milestones - @yangchiu 
    - [x] Verify longhorn chart PR to ensure all artifacts are ready for GA (`install`, `upgrade`)  @chriscchien 
    - [x] Run core testing (install, upgrade) for GA build from the previous patch and the last patch of the previous feature release (1.4.3, 1.5.0). - @roger-ryao 
  - Release
    - [x] Release longhorn/chart from the release branch to publish to ArtifactHub
    - [x] Release note
	     - [x] Deprecation note
	     - [x] Upgrade notes including highlighted notes, deprecation, compatible changes, and others impacting the current users
  - Post-Release
    - [x] Create a new release branch of manager/ui/tests/engine/longhorn instance-manager/share-manager/backing-image-manager when creating the RC1
    - [x] Update https://github.com/longhorn/longhorn/blob/master/deploy/upgrade_responder_server/chart-values.yaml @PhanLe1010 
    - [x] Add another request for the rancher charts for the next patch release (`1.5.1`) @rebeccazzzz  
  - Rancher charts: verify the chart is able to install & upgrade - @khushboo-rancher 
    - [x] rancher/image-mirrors update @weizhe0422 (@PhanLe1010 )
        - https://github.com/rancher/image-mirror/pull/418
    - [x] rancher/charts 2.7 branches for rancher marketplace @weizhe0422 (@PhanLe1010)
        - `dev-2.7`: https://github.com/rancher/charts/pull/2797

cc @longhorn/qa @longhorn/dev

yangchiu:
Run security testing of container images for pre-GA milestones:

[v1.5.1-rc1](https://ci.longhorn.io/job/private/job/image-security-scan/1070/): 37 HIGH, 0 CRITICAL

yangchiu:
Run e2e regression for pre-GA milestones (install, upgrade):

v1.5.1-rc1:
[amd64](https://ci.longhorn.io/job/public/job/v1.5.x/job/v1.5.x-longhorn-tests-sles-amd64/62/): 4 failures
[...]

Issue #6102, opened 2023-06-12
[RELEASE] 1.4.3


innobead:
**What's the task? Please describe.**
Action items for releasing v1.4.3

**Describe the items of the task (DoD, definition of done) you'd like.**
  - Pre-Release
    - [x] Regression test plan (manual) - @khushboo-rancher 
    - [x] Run e2e regression for pre-GA milestones (`install`, `upgrade`) - @yangchiu 
    - [x] Run security testing of container images for pre-GA milestones - @roger-ryao  
    - [x] Verify longhorn chart PR to ensure all artifacts are ready for GA (`install`, `upgrade`)  @chriscchien   
    - [x] Run core testing (install, upgrade) for GA build from the previous patch and the last patch of the previous feature release (1.4.2 & 1.3.3). - @chriscchien 
  - Release
    - [x] Release longhorn/chart from the release branch to publish to ArtifactHub
    - [x] Release note
	     - [x] Deprecation note
	     - [x] Upgrade notes including highlighted notes, deprecation, compatible changes, and others impacting the current users
  - Post-Release
    - [ ] ~Create a new release branch of manager/ui/tests/engine/longhorn instance-manager/share-manager/backing-image-manager when creating the RC1~
    - [x] Update https://github.com/longhorn/longhorn/blob/master/deploy/upgrade_responder_server/chart-values.yaml @PhanLe1010 
    - [x] Add another request for the rancher charts for the next patch release (`1.4.3`) @rebeccazzzz  
  - Rancher charts: verify the chart is able to install & upgrade - @khushboo-rancher 
    - [x] rancher/image-mirrors update @weizhe0422 (@PhanLe1010 )
       - https://github.com/rancher/image-mirror/pull/416
    - [x] rancher/charts 2.7 branches for rancher marketplace @weizhe0422 (@PhanLe1010)
        - `dev-2.7`: https://github.com/rancher/charts/pull/2780

cc @longhorn/qa @longhorn/dev

roger-ryao:
Run security testing of container images for pre-GA milestones:

[v1.4.3-rc1](https://ci.longhorn.io/job/private/job/image-security-scan/1051/): 71 HIGH, 0 CRITICAL

yangchiu:
Run e2e regression for pre-GA milestones (install, upgrade):

v1.4.3-rc1:
[...]

PR #6384, opened 2023-07-25
[upgrade-responder] release v1.5.1 and v.1.4.3 Longhorn verions

PhanLe1010:
* Release v1.5.1 and v.1.4.3 Longhorn verions
* Increase the CPU and RAM limit for the upgrade responder server:
    From the monitoring, sometimes the server is killed by OOM which
    leads to some deep decreases in the graphs. As we are getting more
    Longhorn nodes now, we should increase the server CPU and RAM

longhorn/longhorn#6274
longhorn/longhorn#6102

### DIFF
--- a/deploy/upgrade_responder_server/chart-values.yaml
+++ b/deploy/upgrade_responder_server/chart-values.yaml
@@ -14,6 +14,14 @@ secret:
   # Set this to false if you don't want to manage these secrets with helm
   managed: false
 
+resources:
+  limits:
+    cpu: 400m
+    memory: 512Mi
+  requests:
+    cpu: 200m
+    memory: 256Mi
+
 # This configmap contains information about the latest release
 # of the application that is using this Upgrade Responder
 configMap:
--- a/deploy/upgrade_responder_server/chart-values.yaml
+++ b/deploy/upgrade_responder_server/chart-values.yaml
@@ -19,22 +19,339 @@ secret:
 configMap:
   responseConfig: |-
     {
-      "Versions": [
+      "versions": [
         {
-          "Name": "v1.3.3",
-          "ReleaseDate": "2023-04-19T00:00:00Z",
-          "Tags": [
+          "name": "v1.3.3",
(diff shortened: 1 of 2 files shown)


deploy/upgrade_responder_server/chart.yaml
```diff
@@ -1,5 +1,5 @@
 url: https://github.com/longhorn/upgrade-responder.git
-commit: 3c78890f5415744af1923eac01f98636ac52a113
+commit: a0c69c986fa6a258bf732469d61244d176da8e30
```
c3y1huang, line 2: Any reason for not using the latest commit?

c3y1huang (commented):
Just 1 question. Others LGTM.

deploy/upgrade_responder_server/chart-values.yaml
```diff
@@ -14,27 +14,353 @@ secret:
...
 
+resources:
+  limits:
+    cpu: 400m
+    memory: 512Mi
+  requests:
+    cpu: 200m
+    memory: 256Mi
+
+
```
c3y1huang, line 25:
```suggestion
resources:
  limits:
    cpu: 400m
    memory: 512Mi
  requests:
    cpu: 200m
    memory: 256Mi

```

c3y1huang, line 25: nit: redundant blank line.

deploy/upgrade_responder_server/chart.yaml
PhanLe1010: The PR was created before. Good catch. It should be the v.0.2.0 tag commit https://github.com/longhorn/upgrade-responder/commit/116f807836c29185038cfb005708f0a8d41f4d35


Fixed. Thank you

deploy/upgrade_responder_server/chart-values.yaml
PhanLe1010: Fixed. Thank you

c3y1huang (approved):

Issue #6392, opened 2023-07-26
[BUG] Fix errors in questions.yaml


james-munson:
## What's the task? Please describe
I noticed a few small errors from the addition of storage class node selector in https://github.com/longhorn/longhorn/pull/4576

## Describe the items of the task (DoD, definition of done) you'd like
> Please use a task list for items on a separate line with a clickable checkbox https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists

- [ ] `item 1`

## Additional context

Add any other context or screenshots about the document request here.

ejweber:
Small nit. I think this is not a [DOC] type issue since it actually fixes functionality that was broken.

innobead:
@james-munson Let's make sure the issues we are handling need to add to the current milestone. Also, add backport labels if needed.

cc @longhorn/dev

innobead:
@longhorn/qa for any issues with review-coverage label mean we need to see why the bug happens. Is it duo to some missing test case coverage? If yes, need to improve our test case.

chriscchien:
Verified with test [steps](https://github.com/longhorn/longhorn/issues/6392#issuecomment-1652415733)

Can see `nodeSelector` set correctly in storageclass `longhorn` 
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    longhorn.io/last-applied-configmap: |
      kind: StorageClass
      apiVersion: storage.k8s.io/v1
      metadata:
        name: longhorn
        annotations:
          storageclass.kubernetes.io/is-default-class: "true"
      provisioner: driver.longhorn.io
      allowVolumeExpansion: true
      reclaimPolicy: "Delete"
      volumeBindingMode: Immediate
      parameters:
        numberOfReplicas: "3"
        staleReplicaTimeout: "30"
        fromBackup: ""
        fsType: "ext4"
        dataLocality: "disabled"
        nodeSelector: "ip-172-31-34-163"
```
Following test steps, it will test `rancher/chart` to install longhorn `v1.4.2`, but it can not guarantee `question.yaml` will work  properly on` v1.6.x` too because rancher does not support to install longhorn `v1.6.x` now.

I would like to keep this ticket open until `v1.6.x` rancher chart ready and test again to make sure that the `question.yaml` by install longhorn v1.6.x from Rancher

chriscchien:
[...]

Issue #4776, opened 2022-10-24
[DOC] Explanation of storage class parameters


weizhe0422:
## Is your improvement request related to a feature? Please describe

Add a doc to explain parameters in the storage class supported by Longhorn.

[storage class parameters](https://github.com/longhorn/longhorn/blob/53c9c407ed8238a9b7217a4713b3e17fd84c6c29/chart/templates/storageclass.yaml#L20-L41)
```yaml
      numberOfReplicas: "{{ .Values.persistence.defaultClassReplicaCount }}"
      staleReplicaTimeout: "30"
      fromBackup: ""
      {{- if .Values.persistence.defaultFsType }}
      fsType: "{{ .Values.persistence.defaultFsType }}"
      {{- end }}
      {{- if .Values.persistence.migratable }}
      migratable: "{{ .Values.persistence.migratable }}"
      {{- end }}    
      {{- if .Values.persistence.backingImage.enable }}
      backingImage: {{ .Values.persistence.backingImage.name }}
      backingImageDataSourceType: {{ .Values.persistence.backingImage.dataSourceType }}
      backingImageDataSourceParameters: {{ .Values.persistence.backingImage.dataSourceParameters }}
      backingImageChecksum: {{ .Values.persistence.backingImage.expectedChecksum }}
      {{- end }}
      {{- if .Values.persistence.recurringJobSelector.enable }}
      recurringJobSelector: '{{ .Values.persistence.recurringJobSelector.jobList }}'
      {{- end }}
      dataLocality: {{ .Values.persistence.defaultDataLocality | quote }}
      {{- if .Values.persistence.defaultNodeSelector.enable }}
      nodeSelector: "{{ .Values.persistence.defaultNodeSelector.selector }}"
      {{- end }}
```

## Additional context

https://github.com/longhorn/website/pull/591#pullrequestreview-1142295370
[Issue 4574](https://github.com/longhorn/longhorn/issues/4574)

james-munson:
Screenshot from Rancher UI install of Longhorn, showing the expanded set of questions to guide StorageClass parameter selection.
![StorageClassParamsQuestions](https://github.com/longhorn/longhorn/assets/59452343/659f691e-f886-4483-8e27-4231aa93aa93)

roger-ryao:
**Verified on master-head 20230913**

- longhorn master-head https://github.com/longhorn/longhorn/commit/9948983b15f7d82af44a45917c270313668c3b2c
[...]

PR #6393, opened 2023-07-26
Fix some small errors on StorageClass NodeSelector.

james-munson:
For ticket #6392

### DIFF
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -761,18 +761,18 @@ Set the value to **0** to disable backup restore."
     group: "Longhorn Storage Class Settings"
     type: string
     default:
-- variable: defaultSettings.defaultNodeSelector.enable
-  description: "Enable recurring Node selector for Longhorn StorageClass"
+- variable: persistence.defaultNodeSelector.enable
+  description: "Enable Node selector for Longhorn StorageClass"
   group: "Longhorn Storage Class Settings"
   label: Enable Storage Class Node Selector
   type: boolean
   default: false
   show_subquestion_if: true
   subquestions:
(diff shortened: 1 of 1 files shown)


ejweber (commented):
LGTM

I would be curious to see a screenshot of what this looks like in the Rancher UI with the changes.

james-munson:
Here's a screen shot from installing the amended Longhorn chart from my branch (note the "whith" changed to "with".  Most of the changes are less visible, but are for agreement with values.yaml.

Screenshot-Rancher-6392-questions-yaml

ejweber (approved):
Screenshot looks good and verified with @james-munson that the selector is now applied to the default storage class correctly when installing via Rancher.

Please update the [test steps](https://github.com/longhorn/longhorn/issues/6392#issuecomment-1652415733) to include that verification after installation.

PhanLe1010:
Discussed with @james-munson, the node selector here is for volume's replica scheduling. The current description `We use NodeSelector when we want to bind PVC via StorageClass into desired mountpoint on the nodes tagged with its value` is a bit confusing. May be we should change it to mention that this is for replica scheduling logic

PhanLe1010:
> Discussed with @james-munson, the node selector here is for volume's replica scheduling. The current description `We use NodeSelector when we want to bind PVC via StorageClass into desired mountpoint on the nodes tagged with its value` is a bit confusing. May be we should change it to mention that this is for replica scheduling logic

Let's do this in the ticket https://github.com/longhorn/longhorn/issues/4776

PhanLe1010 (approved):
LGTM

Issue #6415, opened 2023-07-29
[IMPROVEMENT] Remove or Change Helm pre-upgrade hook to support ArgoCD


belmeopmenieuwesim:
As of now, longhorn stopped working with ArgoCD, due to ArgoCD implementing helms post-upgrade hooks as `PreSync` (and they have no plans to change this, since they believe Helm charts should not use any pre/post upgrade hooks), causing it to run even the first time during installation. This causes the pre-upgrade job of Longhorn to fail, since the service account is not yet created.

It would be nice if this helm chart would be working in ArgoCD.

Also see this issue: https://github.com/argoproj/argo-cd/issues/7536#issuecomment-1656763866

acelinkio:
Relates to: https://github.com/longhorn/charts/pull/126

c3y1huang:
@acelinkio can you also create a PR to [longhorn/longhorn](https://github.com/longhorn/longhorn/tree/master/chart)?

acelinkio:
@c3y1huang
Additional pr opened https://github.com/longhorn/longhorn/pull/6421

frg:
I can see that this was added to the v1.6.0 milestone. Are there any plans for a v1.15 release? or a workaround we can use until then?

Thanks

acelinkio:
@frg there is a partial solution to just apply the manifests manually via helm template--which is unable to apply any hooks.  From there hooks will have the resources to run properly.

frg:
@acelinkio Thanks for the suggestion but wouldn't this also disable the [pre delete hook](https://github.com/longhorn/charts/blob/3b57266ab050578a858b53b53c355eb3110e12c2/charts/longhorn/templates/uninstall-job.yaml#L5) making uninstallation risky?

I think we'll have to go for v1.4, then a volume migration when we upgrade to v1.6. Can't find simple ways we can go around this.

acelinkio:
Yeah... preselected hook is not supported.   I really don't know why these hooks are being used because none of these capabilities are available when doing the kubectl apply install method.  https://longhorn.io/docs/1.5.1/deploy/install/install-with-kubectl/

innobead:
This will be backported to 1.5.

innobead:
The pre-upgrade hook is only for the helm to prevent any invalid upgrade after upgrading the Longhorn setup to the new version, so it can ensure the current setup with the old version is intact if the upgrade hook failed. 

[...]

PR #6421, opened 2023-07-31
conditional to disable helm preUpgrade job

acelinkio:
fixes: https://github.com/longhorn/longhorn/issues/6415
relates to:  https://github.com/longhorn/charts/pull/126

The preUpgrade job was added in 1.15 and introduces a breaking change for any ArgoCD users when creating a new environment. ArgoCD does not have a method for differentiating between a helm install or a helm upgrade. https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#helm-hooks

Not being able to differentiate between an install or upgrade is not an problem in most cases because hooks are only available to the helm binary. Kubernetes was not built around Helm and hooks are not part of the core kubernetes api. Hooks are at best a makeshift shim. Relying upon hooks introduce undesired complexity when deploying applications and should be avoided because they lack scalability/flexibility.

It appears the preUpgrade job was created to avoid single pod per node nature of daemonsets. Having a daemonset pod fail will result in that nodes storage being unavailable. The hook was a preventative measure to catch a incompatibilities before there is an issue.

This pull request is for adding a parameter to make the helm hook job optional. Long term this would be best moved the application through the use of a controller or operator.

### DIFF
--- a/chart/templates/preupgrade-job.yaml
+++ b/chart/templates/preupgrade-job.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.helmPreUpgradeCheckerJob.enabled }}
 apiVersion: batch/v1
 kind: Job
 metadata:
@@ -54,3 +55,4 @@ spec:
 {{ toYaml .Values.longhornManager.nodeSelector | indent 8 }}
         {{- end }}
       {{- end }}
+{{- end }}
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -158,6 +158,9 @@ persistence:
   # -- Allow automatically removing snapshots during filesystem trim for Longhorn StorageClass. Options: `ignored`, `enabled`, `disabled`
   removeSnapshotsDuringFilesystemTrim: ignored
 
+helmPreUpgradeCheckerJob:
+  enabled: true
+
 csi:
   # -- Specify kubelet root-dir. Leave blank to autodetect
   kubeletRootDir: ~


innobead:
> fixes: #6415 relates to: [longhorn/charts#126](https://github.com/longhorn/charts/pull/126)
> 
> The preUpgrade job was added in 1.15 and introduces a breaking change for any ArgoCD users when creating a new environment. ArgoCD does not have a method for differentiating between a helm install or a helm upgrade. https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#helm-hooks

This is definitely good info, thanks for that sharing. 

> 
> Not being able to differentiate between an install or upgrade is not an problem in most cases because hooks are only available to the helm binary. Kubernetes was not built around Helm and hooks are not part of the core kubernetes api. Hooks are at best a makeshift shim. Relying upon hooks introduce undesired complexity when deploying applications and should be avoided because they lack scalability/flexibility.

The detailed explanation is https://github.com/longhorn/longhorn/issues/6415#issuecomment-1667718648, the logic of the hook is actually also part of the longhorn control plane; however, the hook can provide another benefit before upgrading to the new version through an invalid upgrade path.

mantissahz (approved):
LGTM

QuinnBast:
Would love to get this in

innobead (approved):

innobead:
@mergify backport v1.5.x

PR #6423, opened 2023-07-31
Make auto-created issues only triggered by members

innobead:
The primary change is to add the below step to create-issue.yaml as the pre-criteria for creating backport and test issues. 

```yaml
    - name: Is Longhorn Member
      uses: tspascoal/get-user-teams-membership@v1.0.4
      id: is-longhorn-member
      with:
        username: ${{ github.event.issue.user.login }}
        organization: longhorn
        GITHUB_TOKEN: ${{ secrets.CUSTOM_GITHUB_TOKEN }}
```

### DIFF
--- a/.github/workflows/add-to-projects.yml
+++ b/.github/workflows/add-to-projects.yml
@@ -1,32 +1,40 @@
 name: Add-To-Projects
 on:
   issues:
-    types: [opened, labeled]
+    types: [ opened, labeled ]
 jobs:
   community:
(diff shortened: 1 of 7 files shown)


innobead:
@yangchiu Please help review this.

yangchiu:
Does it check the result of `Is Longhorn Member` step? Otherwise it just executed the function, and didn't use the result to decide it should execute following steps or not.

Issue #3823, opened 2022-04-06
[QUESTION] (Single node setup) Disk affinity


artur9010:
## Question
I have a single node setup with a lot of disks. Is there an option to set an affinity on disks to make sure that none disk will have two or more replicas of volume?

As you can see below, 2 of 5 replicas are located on /mnt/sdb

![image](https://user-images.githubusercontent.com/4835677/162053535-59773c90-2630-4ed3-82a5-552140d14b4d.png)

shuo-wu:
Unfortunately, right now there is no disk-level affinity.

artur9010:
So if there is no option to set disk affinity, can I run multiple longhorn instances on one node and give them different disks?

shuo-wu:
IIRC, Longhorn will randomly choose disks for replicas, no guarantee of scheduling each replica to a different disk. You may need to handle/adjust it manually.

usma0118:
@innobead is this something planned?

ejweber:
https://github.com/longhorn/longhorn/issues/5149#issuecomment-1559585527

roger-ryao:
**Verified on master-head 20231207**

- longhorn master-head https://github.com/longhorn/longhorn/commit/6627b0c55c15adb0831546ec2412ce2623462ce7
- longhorn-manager master-head https://github.com/longhorn/longhorn-manager/commit/266210f7c2ac6e4d8b80381d25daa673dd9577d5
- longhorn-ui master-head https://github.com/longhorn/longhorn-ui/commit/41f43dceb0b1c0dc1367a104a60384eb7ffb0b25
- website master https://github.com/longhorn/website/commit/41f43dceb0b1c0dc1367a104a60384eb7ffb0b25

**The test steps**

1. https://github.com/longhorn/longhorn/issues/3823#issuecomment-1659169095
2. Manual test cases : https://longhorn.github.io/longhorn-tests/manual/release-specific/v1.6.0/test-replica-disk-soft-anti-affinity/

**Result Passed**

- [x] 1. Test initial behavior of global Replica Disk Soft Anti-Affinity setting
- [x] 2. Test initial behavior of global Replica Disk Soft Anti-Affinity setting after upgrade
- [x] 3. Test behavior of volume Replica Disk Soft Anti-Affinity setting
- [x] 4. Test initial behavior of volume Replica Disk Soft Anti-Affinity setting after upgrade
- [x] 5. Test effect of Replica Disk Soft Anti-Affinity on replica scheduling
[...]

PR #6431, opened 2023-07-31
Add LEP for disk anti-affinity

ejweber:
#3823

### DIFF
--- a/enhancements/20230718-disk-anti-affinity.md
+++ b/enhancements/20230718-disk-anti-affinity.md
@@ -1,0 +1,155 @@
+# Disk Anti-Affinity
+
+## Summary
+
+Longhorn supports multiple disks per node, but there is currently no way to ensure that two replicas for the same
+volume that schedule to the same node end up on different disks. In fact, the replica scheduler currently doesn't make
+any attempt achieve this goal, even when it is possible to do so.
+
+With the addition of a Disk Anti-Affinity feature, the Longhorn replica scheduler will attempt to schedule two replicas
+for the same volume to different disks when possible. Optionally, the scheduler will refuse to schedule a replica to a
+disk that has another replica for the same volume.
+
+Although the comparison is not perfect, this enhancement can be thought of as enabling RAID 1 for Longhorn (mirroring
+across multiple disks on the same node).
+
+See the [Motivation section](#motivation) for potential benefits.
+
+### Related Issues
+
+- https://github.com/longhorn/longhorn/issues/3823
+- https://github.com/longhorn/longhorn/issues/5149
+
+### Existing Related Features
+
+#### Replica Node Level Soft Anti-Affinity
+
+Disabled by default. When disabled, prevents the scheduling of a replica to a node with an existing healthy replica of
+the same volume.
+
+Can also be set at the volume level to override the global default.
+
+#### Replica Zone Level Soft Anti-Affinity
+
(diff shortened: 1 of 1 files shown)


enhancements/20230718-disk-anti-affinity.md
```diff
@@ -0,0 +1,154 @@
...
+  - Increased data durability. If a single disk fails, a healthy replica will still exist on an disk that
+    has not failed.
+  - Increased data availability. If a single disk on a node becomes unavailable, but the node itself remains
+    healthy, at least one replica remains healthy. On a single-node cluster, this can directly prevent a volume crash.
+    On a small, multi-node cluster, this can prevent a future volume crash due to the loss of a different node.
+
+### Goals
+
+- In all situations, the Longhorn replica scheduler will make a best effort to ensure two replicas for the same volume
+  do not schedule to the same disk.
```
james-munson, line 53: Kudos to you for this.

enhancements/20230718-disk-anti-affinity.md
```diff
@@ -0,0 +1,154 @@
...
+available disks as desired replicas, I want scheduling to fail obviously. It is important that I know high-priority
+volume isn't being protected so I can take action.
+
+### User Experience In Detail
+
+### API changes
+
+Introduce a new Replica Disk Level Soft Anti-Affinity setting with the following definition. By default, set it to
+`true`. While it is generally desirable to schedule replicas to different disks, it would break with existing behavior
+to refuse to schedule replicas when different disks are not available.
```
james-munson, line 86: I personally think this should default to "false", but I understand the arguments for leaving it as is.  At least the chances are reduced of falling into it accidentally.

james-munson (approved):
A good step forward.

enhancements/20230718-disk-anti-affinity.md
```diff
@@ -0,0 +1,154 @@
...
+
+1. In a cluster that includes nodes with multiple available disks, create a volume with
+   `spec.replicaSoftAntiAffinity = true`, `spec.replicaDiskSoftAntiAffinity = true`, and `numberOfReplicas` equal to the
+   total number of disks in the cluster. Confirm that each replica schedules to a different disk. It may be necessary
+   to tweak additional factors. For example, ensure that one disk has enough free space that the old scheduling
+   behavior would assign two replicas to it instead of distributing the replicas evenly among the disks.
+1. In a cluster that includes nodes with multiple available disks, create a volume with
+   `spec.replicaSoftAntiAffinity = false`, `spec.replicaDiskSoftAntiAffinity = true`, and `numberOfReplicas` equal to one more than the total number of disks in the cluster. Confirm that a replica fails to schedule. Previously,
+   multiple replicas would have scheduled to the same disk and no error would have occurred.
+
```
innobead, line 147: `spec.replicaSoftAntiAffinity = false` means each replica should be scheduled to a different node, so `multiple replicas would have scheduled to the same disk and no error would have occurred` is not right, correct?

What you want to test should be `spec.replicaSoftAntiAffinity = true, spec.replicaDiskSoftAntiAffinity = false`?

innobead (changes requested):
In general, LGTM. Just one comment needs to clarify.

enhancements/20230718-disk-anti-affinity.md
ejweber: You are correct. Good catch. Fixed in an amend.

innobead (approved):

PR #6453, opened 2023-08-03
chore(ghaction): fix invalid member check

innobead:
https://github.com/tspascoal/get-user-teams-membership/blob/master/src/index.js#L57C1-L59C11 `isTeamMember` requires a team variable, so we should just simply check if the reporter is from the org as the original behavior.


yangchiu (approved):
LGTM

Issue #6530, opened 2023-08-15
[TASK] Create a CIFS backup store example in longhorn repo


innobead:
## What's the task? Please describe

Add a CIFS example manifest in https://github.com/longhorn/longhorn/tree/master/deploy/backupstores

## Additional context

cc @longhorn/qa

innobead:
@chriscchien try this. You can consult with @derekbit .

innobead:
Done.

PR #6547, opened 2023-08-18
add cifs backupstore manifest

chriscchien:
ref: [6530](https://github.com/longhorn/longhorn/issues/6530)

After deploy the manifest, set backuptarget = `cifs://longhorn-test-cifs-svc.default/backupstore` and currently the samba image is using `derekbit/samba:latest`

cc @derekbit

### DIFF
--- a/deploy/backupstores/cifs-backupstore.yaml
+++ b/deploy/backupstores/cifs-backupstore.yaml
@@ -1,0 +1,87 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cifs-secret
+  namespace: longhorn-system
+type: Opaque
+data:
+  CIFS_USERNAME: bG9uZ2hvcm4tY2lmcy11c2VybmFtZQ== # longhorn-cifs-username
+  CIFS_PASSWORD: bG9uZ2hvcm4tY2lmcy1wYXNzd29yZA== # longhorn-cifs-password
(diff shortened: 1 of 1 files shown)


deploy/backupstores/cifs-backupstore.yaml
```diff
@@ -0,0 +1,87 @@
...
+    metadata:
+      labels:
+        app: longhorn-test-cifs
+    spec:
+      volumes:
+      - name: cifs-volume
+        emptyDir: {}
+      containers:
+      - name: longhorn-test-cifs-container
+        image: derekbit/samba:latest
```
shuo-wu, line 43: Notice that this is an image from Derek's docker hub. Do we have any modifications applied to the official image?

shuo-wu (approved):
In general LGTM

deploy/backupstores/cifs-backupstore.yaml
derekbit: Agree.
BTW, @chriscchien can you help fork https://github.com/derekbit/samba/tree/longhorn to longhornio org? Then, we can maintain the forked on and build an official image from it. Thank you.

innobead (approved):
LGTM.

We can change the image later when having a longhorn one. Chris, please help with that image build.

Issue #2166, opened 2021-01-08
[BUG] Longhorn does not launch on systems with a non-standard path


duckfullstop:
**Describe the bug**
Linux distributions such as NixOS use a very non-standard path (e.g `/run/wrappers/bin:/root/.nix-profile/bin:/etc/profiles/per-user/root/bin:/nix/var/nix/profiles/default/bin:/run/current-system/sw/bi`) - in NixOS's case, this is to allow for use of the Nix package manager.

On these systems, because Longhorn uses nsenter to enter the host namespace, and nsenter uses the path of the namespaces that called it, basic system utilities such as `mount`, as well as `iscsiadm` can't be found or called (produces `Failed to execute: nsenter [–mount=/host/proc/*/ns/mnt mount], output nsenter: failed to execute mount: No such file or directory` and `Failed environment check, please make sure you have iscsiadm/open-iscsi installed on the host` during initial startup of the longhorn-manager pod).

I have built some _extremely_ scrappy derived images based on the current stable release of Longhorn at https://github.com/duckfullstop/nixos-longhorn - these simply append the necessary NixOS-specific path entries to the container's path, such that `nsenter` can find the requisite binaries properly. **This is not a complete fix** - there is, for some reason, a hardcoded lookup path in [longhorn-manager/csi/nfs/nsenter.go](https://github.com/longhorn/longhorn-manager/blob/master/csi/nfs/nsenter.go#L80) that means the share-manager pod cannot properly instantiate `ReadWriteMany` Persistent Volumes. However, even without patching that, the containers with the updated path work perfectly, and Longhorn works flawlessly (excepting the `ReadWriteMany` issue).

The affected Longhorn components are `longhorn-manager` and `longhorn-instance-manager`, as these are the only two images that make nsenter calls.

I'm not 100% sure what the best and cleanest way to resolve this is, so I'm opening this as an issue instead of a load of PRs.

**To Reproduce**
1. Bootstrap a NixOS Kubernetes cluster (see https://nixos.wiki/wiki/Kubernetes for some vague instruction, I have incantations that I'm using that I can post somewhere eventually but a basic setup will experience the same issue)
2. Ensure that `openiscsi` is available in the host environment with `config.environment.systemPackages = [ pkgs.openiscsi ];`
3. Deploy the current Longhorn manifest to the cluster
4. Be sad when longhorn-manager fails to come up

**Expected behavior**
longhorn-manager proceeds with initialisation.

**Log**

[...]

PR #6602, opened 2023-08-28
[DRAFT] Implementing NixOS support. Needs feedback

HiWhatName:
Hello everyone!~
As indicated in Issue #2166 Longhorn has a requirement to run multiple binary files located on the host machines of individual nodes. However, challenges arise when dealing with Linux distributions such as [NixOS](https://nixos.org/), which utilize a non-standard $PATH environment (e.g.,` /run/wrappers/bin`). This difference in paths results in the failure of multiple components initialization.

In this pull request (PR), I have incorporated support for 

- all installers located within the `deployment/prerequisite` directory, 
- along with the `scripts/environment_check.sh` script.

To tackle the issue of inconsistent paths, maybe we involve deploying an appropriate **ConfigMap** that aligns with the specific node in question. For example, if NixOS is the distro in use, we could consider deploying a ConfigMap like `deploy/distro_specific/nixos/configmap.yaml`.

The `path_policy.yaml` file incorporates the data from the ConfigMap into every pod that operates within the longhorn-system namespace.

As I'm pretty inexperienced with working in Kubernetes environments, I'd love to receive some feedback regarding the implementation and potential improvements ( especially about the policy) .


HiWhatName:
> Rebased to Fix DCO **sign-off**

usmcamp0811:
I've got a cluster running Longhorn that I want to migrate to NixOS and would love to see this merged! Thank you!

adrian798:
Does not seem to work when used RWX .
longhorn share manager stuck on waiting for nfs server to start
Warning Can’t find var/run/ganesha.pid

Issue #6563, opened 2023-08-21
[FEATURE] Longhorn snapshot space management


innobead:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

Currently, when creating a longhorn snapshot (in-cluster) of a volume will not respect any provision factors except the fixed upper limit of the number of volume snapshots, so it would cause unexpected disk space exhaustion which means it's unable to prevent this beforehand by having a better space plan. 

The goal is to introduce a mechanism to manage snapshot space usage when creating snapshots. 

## Describe the solution you'd like

When creating a snapshot, should respect the below factors. All factors should be considered together.
- A customized max number of snapshots allowed globally or volume-specific setting.
- A quota for snapshot usage global, node, or disk-specific. This will be another factor to impact replica scheduling.
- Introduce snapshot scheduling like replica scheduling to determine if the disk space is enough for creating a snapshot next to the replica. (cross disks allowed). This should be similar to creating a volume by respecting disk usage first.

## Additional context

cc @longhorn/dev

FrankYang0529:
My proposal to this feature is that:

* [ ] 1. Add a setting `snapshot-max-count`. The default value is `250`. This complies with [the current fixed limit](https://github.com/longhorn/longhorn-engine/blob/8b4c80ab174b4f454a992ff998b6cb1041faf63d/pkg/replica/replica.go#L33).
* [ ] 2. Add `snapshotMaxCount` and `snapshotMaxSize` to Volume spec. The default `snapshotMaxCount` is from the `snapshot-max-count` setting. The default `snapshotMaxSize` is `0`, which means no limit on size.
* [ ] 3. Enforce the limit in [the engine controller](https://github.com/longhorn/longhorn-engine/blob/8b4c80ab174b4f454a992ff998b6cb1041faf63d/pkg/controller/control.go#L205-L207). If either count/size reaches the limit, the system doesn't allow to create a snapshot.

innobead:
As discussed with @FrankYang0529 , we will focus on the first item below to introduce the fundamental limitation for snapshot creation. About the intelligent part for snapshot scheduling (item 2 and 3), they will be planned in the future depending on user feedback. 

> 1. A customized max number of snapshots allowed globally or volume-specific setting.
> 2. A quote for snapshot usage global, node, or disk specific. This will be another factor to impact replica scheduling.
[...]

PR #6634, opened 2023-09-05
feat: add snapshot space management lep

FrankYang0529:
Ref: https://github.com/longhorn/longhorn/issues/6563

### DIFF
--- a/enhancements/20230905-snapshot-space-management.md
+++ b/enhancements/20230905-snapshot-space-management.md
@@ -1,0 +1,165 @@
+# Snapshot Space Management
+
+## Summary
+
+This feature allows users to control the count and size of snapshots of a volume.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/6563
+
+## Motivation
+
+### Goals
+
+- A replica honors snapshot count and size limitations when it creates a new snapshot.
+- Users can set snapshot count and size limitation on a Volume.
+- Users can set a global default snapshot maximum count.
+
+### Non-goals [optional]
+
+- Snapshot space management on a node or disk level.
+- Freeze disk before taking snapshot.
+- Auto deleting snapshots if there is no space for creating a new system snapshot.
+
+## Proposal
+
+### User Stories
+
+With snapshot space management, snapshot space usage is controllable. Users can set snapshot limitation on a Volume and evaluate maximum space usage.
+
+### User Experience In Detail
+
+Before snapshot space management:
+The default maximum snapshot count for each volume is `250`. This is a constants value in the [source code](https://github.com/longhorn/longhorn-engine/blob/8b4c80ab174b4f454a992ff998b6cb1041faf63d/pkg/replica/replica.go#L33) and users don't have a way to control it. If a volume size is 1G, the maximum snapshot space usage will be 250G.
+
+After snapshot space management:
+There is a configurable default maximum snapshot count setting. Users can update it to overwrite the fixed value in the system.
+Users can set different maximum snapshot count and size on each volume. A more important volume can have more snapshot space usage.
+
+## Design
+
+### Implementation Overview
+
+#### Settings
+
+Add a new setting definition:
+
+```go
+const (
+    SettingNameSnapshotMaxCount = SettingName("snapshot-max-count")
+)
+
+var (
+    SettingDefinitionSnapshotMaxCount = SettingDefinition{
+		DisplayName: "Snapshot Maximum Count",
+		Description: "Maximum snapshot count for a volume. The value should be between 2 to 250",
+		Category:    SettingCategorySnapshot,
+		Type:        SettingTypeInt,
+		Required:    true,
+		ReadOnly:    false,
+        Default:     "250",
+	}
+)
+```
+
+#### Volume CRD
+
+Add two fields to `VolumeSpec`:
+
+```go
+type VolumeSpec struct {
+    // ...
+    // +optional
+    SnapshotMaxCount int `json:"snapshotMaxCount"`
+    // +kubebuilder:validation:Type=string
+    // +optional
+    SnapshotMaxSize int64 `json:"snapshotMaxSize,string"`
+}
+```
+
+- If `SnapshotMaxSize` is `0`, it means there is no snapshot size limit for a Volume.
+
+#### Volume mutator
+
+- If `SnapshotMaxCount` is `0`, using the `snapshot-max-count` setting value to update it.
+- If a volume is expanded, checking whether `snapshot-max-size` is smaller than `size * 2`. If it is, using `size * 2` to update it.
+
+#### Volume validator
+
+- The `SnapshotMaxCount` should be between `2` to `250`. This limitation includes user and system snapshot. In LH, all snapshots can be merged to one snapshot, so at least one snapshot can't be delete. To create another snapshot, we need to have enough count for it. In conclusion, the minimum value for `SnapshotMaxCount` is `2`.
+- If `SnapshotMaxSize` is't `0`. The minimum value for `SnapshotMaxSize` is same as `Size * 2` in a Volume, because a volume can have at least two snapshots.
+
+#### Replica
+
+- Add two fields to `Replica`:
+    ```go
+    type Replica struct {
+        // ...
+        snapshotMaxCount int
+        snapshotMaxSize int64
+    }
+    ```
+- Add a function `GetSnapshotCountUsage` to retrieve snapshot count usage. We should skip the volume head, backing disk, and removed disks.
+- Add a function `GetSnapshotSizeUsage` to retrieve total snapshot size usage. We should skip the volume head, backing disk, and removed disks.
+
+#### ReplicaServer
+
+- Add the `remain_snapshot_size` field to `Replica` proto message:
+    ```protobuf
+    message Replica {
+        // ...
+        int64 remain_snapshot_size = 17;
+    }
+    ```
+- Update the `getReplica` function to return `SnapshotCountUsage` and `SnapshotSizeUsage` fields.
+
+#### Remote backend
+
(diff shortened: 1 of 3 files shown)


enhancements/20230905-snapshot-space-management.md
```diff
@@ -0,0 +1,250 @@
...
+
+#### Volume CRD
+
+Add two fields to `VolumeSpec`:
+
+```go
+type VolumeSpec struct {
+    // ...
+    // +optional
+    SnapshotMaxCount int `json:"snapshotMaxCount"`
```
derekbit, line 72: Do the value include the system snapshots?

enhancements/20230905-snapshot-space-management.md
```diff
@@ -0,0 +1,250 @@
...
+- If `SnapshotMaxSize` is `0`, it means there is no snapshot size limit for a Volume.
+
+#### Volume mutator
+
+If `SnapshotMaxCount` is `0`, using the `snapshot-max-count` setting value to update it.
+
+#### Volume validator
+
+- The `SnapshotMaxCount` should be between `2` to `250`.
+- If `SnapshotMaxSize` is't `0`. The minimum value for `SnapshotMaxSize` is same as `Size * 2` in a Volume, because a volume can have at least two snapshots.
```
derekbit, line 88: > volume can have at least two snapshots.

Can you elaborate more on the two snapshots and their purposes?

enhancements/20230905-snapshot-space-management.md
```diff
@@ -0,0 +1,250 @@
...
+Enforce snapshot size and count checking in `Snapshot` function:
+```go
+func (c *Controller) Snapshot(name string, labels map[string]string) (string, error) {
+    // ...
+    if _, remainSize, err := c.backend.RemainSnapshotCountAndSize(); err != nil {
+		return "", err
+	}
+    // TODO handle remainSize
+}
+```
```
derekbit, line 219: IIUC, the check logic is in the controller in the design. Can we move the it to longhorn-manager controller?
Before calling `engineClientProxy.SnapshotCreate`, get the snapshot counts and total size and check if the creation is doable.

enhancements/20230905-snapshot-space-management.md
```diff
@@ -0,0 +1,265 @@
...
+			continue
+		}
+		// ignore error and try next one. We can deal with all error situation later
+		remainCount, remainSize, err1 := backend.backend.RemainSnapshotCountAndSize()
+		headFileSize, err2 := backend.backend.GetHeadFileSize()
+        if err1 == nil && err2 == nil {
+			if remainCount < retCount {
+				retCount = remainCOunt
+			}
+            if remainSize < headFileSize {
```
james-munson, line 207: Does this assume that the head might be completely changed from last snapshot, and therefore the current head size is an upper bound for what might need to be written?  (A number that can't be made more accurate without doing a lot of the snapshot work.)

That won't let snapshots through that might not be able to complete, but it might also prevent snapshots that could have succeeded.  There may be customer fallout from snapshots that unexpectedly fail.  But it is a simple calculation that could be done for presentation as well, so volumes could be marked as "not currently snapshot-able" in their status or in the UI.

enhancements/20230905-snapshot-space-management.md
FrankYang0529: Yes, this also includes system snapshots.

enhancements/20230905-snapshot-space-management.md
FrankYang0529: Yes, added more content for it. Thank you.

enhancements/20230905-snapshot-space-management.md
FrankYang0529: The request flow is longhorn-manager -> longhorn-instance-manager -> longhorn-engine. We don't freeze the disk before taking a snapshot. If we put the checking limitation logic in longhorn-engine, we can have more accurate data, because the time gap is lower.

enhancements/20230905-snapshot-space-management.md
FrankYang0529: Yes, we assume the head size is the next snapshot size. The `snapshotMaxSize` is a soft limit, because we don't freeze the disk before taking a snapshot.

enhancements/20230905-snapshot-space-management.md
shuo-wu: Here I am still worried about the system snapshots triggered by rebuilding or expansion. If users disable setting `autoCleanupSystemGeneratedSnapshot`, there is no purge/cleanup for the system snapshots. Then how can you guarantee the max snapshot count/size?

enhancements/20230905-snapshot-space-management.md
```diff
@@ -0,0 +1,265 @@
...
+
+- If `SnapshotMaxSize` is `0`, it means there is no snapshot size limit for a Volume.
+
+#### Volume mutator
+
+If `SnapshotMaxCount` is `0`, using the `snapshot-max-count` setting value to update it.
+
+#### Volume validator
+
+- The `SnapshotMaxCount` should be between `2` to `250`. This limitation includes user and system snapshot. In LH, all snapshots can be merged to one snapshot, so at least one snapshot can't be delete. To create another snapshot, we need to have enough count for it. In conclusion, the minimum value for `SnapshotMaxCount` is `2`.
```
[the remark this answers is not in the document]
FrankYang0529: Hi Shuo, thanks for reviewing. I update `getRemainSnapshotCounts ` and `GetRemainSnapshotSize` to handle snapshot tree. May you take a look? Thank you.

enhancements/20230905-snapshot-space-management.md
FrankYang0529: Yeah, in this case, we can't have a new system snapshot. The current system also have this problem with `maximumChainLength`. In this situation, users still can manually remove snapshots to create a new system snapshot.

shuo-wu (commented):
The remaining snapshot count calculation LGTM.
The next question is what I mentioned in the team meeting: Can we leave some buffers/extra tolerations for the system snapshot so that the rebuilding or expansion won't be blocked forever (if users don't do snapshot cleanup manually)?

FrankYang0529:
> Can we leave some buffers/extra tolerations for the system snapshot so that the rebuilding or expansion won't be blocked forever (if users don't do snapshot cleanup manually)?

Leave some conclusion note:
If users don't enable `autoCleanupSystemGeneratedSnapshot` setting, even if we leave buffers for the system snapshot, users still face the same issue when the buffer is full.
If we just allow any new system snapshot, the snapshot space management is meaningless, because the limitation doesn't reflect real usage.

enhancements/20230905-snapshot-space-management.md
innobead: This means that we should inform users about the maximum count and size limitation. They need to be aware of this enforcement as it may prevent snapshots from being created to cause the following volume operations failed (ex: rebuild/backup will fail). We need to update the documentation to reflect this.

enhancements/20230905-snapshot-space-management.md
innobead: This should be the same as recurring jobs, snapshot retention.

enhancements/20230905-snapshot-space-management.md
```diff
@@ -0,0 +1,300 @@
...
+
+### Test plan
+
+Integration test plan.
+
+1. `snapshot-max-count` setting
+    - Validate the value should be between `2` to `250`.
+    - Create a Volume with empty `SnapshotMaxCount` and mutator should replace the value with `snapshot-max-count` setting value.
+    - Create a Volume with nonempty `SnapshotMaxCount` and mutator shouldn't update the value.
+
```
innobead, line 284: If less than 2, it will be updated to 2.

innobead (commented):
In general, LGTM.

innobead:
Please help update this after the implementation gets merged, as the code could be changed. 

For LEP, it is not mandatory to include implementation details. However, it is necessary to include the interface, spec, code flow, and e2e behavior to help reviewers understand the design. Pseudo-code is still helpful to have.

cc @longhorn/dev

enhancements/20230905-snapshot-space-management.md
```diff
@@ -0,0 +1,300 @@
...
+    ```
+
+#### Remote backend
+
+- Add `RemainSnapshotSize` to `ReplicaInfo`:
+    ```go
+    type ReplicaInfo struct {
+        // ...
+        RemainSnapshotSize int `json:"remainsnapshotSize"`
+    }
```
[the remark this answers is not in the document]
FrankYang0529: We used `RemainSnapshots` in `Replica` structure. Do we want to make it consistent with current naming?

https://github.com/longhorn/longhorn-engine/blob/7ea91cba9f25e5d290d73a739c2e0fce02cdcacb/pkg/types/resource.go#L14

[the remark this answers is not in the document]
innobead: Yeah, I noticed that.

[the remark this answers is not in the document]
innobead: No, just keep it as is, not a point now. However, we should make it clear in the future by some refactoring.

enhancements/20230905-snapshot-space-management.md
PhanLe1010: Let's say the volume is created with Spec.Size = 2GB and Spec.SnapshotMaxSize = 4GB. Then volume is expanded to 10GB. Will we dynamically increase the Spec.SnapshotMaxSize to 10GB? If not, then this may violate our expectation that Spec.SnapshotMaxSize >= 2*Spec.Size.

FrankYang0529: Yeah, good catch. I think we should use webhook mutator to update it too.

innobead (approved):

Issue #5455, opened 2023-02-28
[IMPROVEMENT] List of Longhorn Helm Chart Flags


zackbradys:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

Difficult to easily view a list of all possible helm chart flags. Of course, you are able to manually determine these using the helm chart values, but like other rancher products, there usually is a list of the flags within the docs or chart readme.

## Describe the solution you'd like

Add the list of helm chart flags to the [longhorn docs](https://longhorn.io/docs/) and [helm chart readme](https://github.com/longhorn/longhorn/tree/master/chart).

## Describe alternatives you've considered

Current alternative to this feature improvement request is to manually go through the helm chart values file to determine to helm install flags.

## Additional context

I'd be will to submit a draft of these changes/additions. No other additional context to provide to this request!

zackbradys:
Yes, the helm values. Most helm chart repos have the values represented in a table within the readme. You are correct, you're able to view the values.yaml, but having the values in documentation.

It would be nice to have it implemented for longhorn to match the rest of the rancher products (MCM, NV, RKE2, etc..)

zackbradys:
Hey @innobead / @ChanYiLin, are we still on track to complete this in the Longhorn `v1.6.0` release?

ChanYiLin:
Hi @zackbradys 
Yes, this feature request will be in Longhorn v1.6.0 which will be released in December.

PR #6639, opened 2023-09-06
doc(chart): add table of helm values

ChanYiLin:
ref: [longhorn/longhorn#5455](https://github.com/longhorn/longhorn/issues/5455)

- use `helm-doc` to generate table of values in readme
- for `default settings`, modify the description in `questions.yaml` of each setting to one sentence.
    - make the descriptions in `values.yaml`  same as those  in `questinos.yaml`
    - for details, user can check the website

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -3,30 +3,30 @@
 # Declare variables to be passed into your templates.
 global:
   cattle:
-    # -- system default registry
+    # -- System default registry
     systemDefaultRegistry: ""
     windowsCluster:
       # -- Enable this to allow Longhorn to run on the Rancher deployed Windows cluster
       enabled: false
-      # -- Tolerate Linux node taint
+      # -- Tolerate Linux nodes to run Longhorn user deployed components
       tolerations:
       - key: "cattle.io/os"
         value: "linux"
         effect: "NoSchedule"
         operator: "Equal"
-      # -- Select Linux nodes
+      # -- Select Linux nodes to run Longhorn user deployed components
       nodeSelector:
         kubernetes.io/os: "linux"
       defaultSetting:
-        # -- toleration for Longhorn run-time created components
+        # -- Toleration for Longhorn system managed components
         taintToleration: cattle.io/os=linux:NoSchedule
-        # -- node selector for Longhorn run-time created components
+        # -- Node selector for Longhorn system managed components
         systemManagedComponentsNodeSelector: kubernetes.io/os:linux
 
 networkPolicies:
   # -- Enable NetworkPolicies to limit access to the Longhorn pods
   enabled: false
-  # -- Create the policy based on your distribution to allow access for the ingress. Available types: `k3s`, `rke2`, `rke1`
+  # -- Create the policy based on your distribution to allow access for the ingress. Options: `k3s`, `rke2`, `rke1`
   type: "k3s"
 
 image:
@@ -103,12 +103,12 @@ image:
       repository: quay.io/openshift/origin-oauth-proxy
       # -- For openshift user. Specify oauth proxy image tag. Note: Use your OCP/OKD 4.X Version, Current Stable is 4.13
       tag: 4.13
-  # -- `imagePullPolicy` to apply to all containers
+  # -- Image pull policy which applies to all user deployed Longhorn Components. e.g, Longhorn manager, Longhorn driver, Longhorn UI
   pullPolicy: IfNotPresent
 
 service:
   ui:
-    # -- Define Longhorn UI service type. Available types: `ClusterIP`, `NodePort`, `LoadBalancer`, `Rancher-Proxy`
+    # -- Define Longhorn UI service type. Options: `ClusterIP`, `NodePort`, `LoadBalancer`, `Rancher-Proxy`
     type: ClusterIP
     # -- NodePort port number (to set explicitly, choose port between 30000-32767)
     nodePort: null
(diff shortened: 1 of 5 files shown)


ChanYiLin:
conflicts fixed

chart/values.yaml
```diff
@@ -3,273 +3,408 @@
 # Declare variables to be passed into your templates.
 global:
   cattle:
+    # -- system default registry
```
ejweber, line 6: Nit: start this with a capital letter.

chart/values.yaml
```diff
@@ -3,273 +3,408 @@
...
+      # -- Specify CSI liveness probe image tag. Leave blank to autodetect
       tag: v2.9.0
   openshift:
     oauthProxy:
+      # -- For openshift user. Specify oauth proxy image repository
       repository: quay.io/openshift/origin-oauth-proxy
-      tag: 4.13  # Use Your OCP/OKD 4.X Version, Current Stable is 4.13
+      # -- For openshift user. Specify oauth proxy image tag. Note: Use your OCP/OKD 4.X Version, Current Stable is 4.13
+      tag: 4.13
+  # -- `imagePullPolicy` to apply to all containers
```
ejweber, line 106: Nit: May want to make the difference between this and systemManagedPodsImagePullPolicy clear.

chart/values.yaml
```diff
@@ -3,273 +3,408 @@
...
     enable: false
+    # -- Recurring job selector list for Longhorn StorageClass. Please be careful of quotes of input. e.g., `[{"name":"backup", "isGroup":true}]`
     jobList: []
   backingImage:
+    # -- Set backing image for Longhorn StorageClass
     enable: false
+    # -- Specify a backing image that will be used by Longhorn volumes in Longhorn StorageClass. If not exists, the backing image data source type and backing image data source parameters should be specified so that Longhorn will create the backing image before using it
     name: ~
+    # -- Specify the data source type for the backing image used in Longhorn StorageClass.
+    # If the backing image does not exists, Longhorn will use this field to create a backing image. Otherwise, Longhorn will use it to verify the selected backing image.
```
ejweber, line 147: Nit: This appears to be the only use of `#` (instead of `# --`) in this PR. Is it intended?

chart/values.yaml
```diff
@@ -3,273 +3,408 @@
...
   registrySecret: ~
 
 longhornManager:
   log:
-    ## Allowed values are `plain` or `json`.
+    # -- Allowed values are `plain` or `json`.
     format: plain
+  # -- priorityClass for longhorn manager
   priorityClass: ~
+  # -- tolerations for longhorn manager
```
ejweber, line 313: Nit: We moved away from capital letters to start comments in this section and beyond. I suggest we maintain the standard as much as possible.

chart/values.yaml
```diff
@@ -3,273 +3,408 @@
...
 
 ingress:
-  ## Set to true to enable ingress record generation
+  # -- Set to true to enable ingress record generation
   enabled: false
 
-  ## Add ingressClassName to the Ingress
-  ## Can replace the kubernetes.io/ingress.class annotation on v1.18+
+  # -- Add ingressClassName to the Ingress
+  # Can replace the kubernetes.io/ingress.class annotation on v1.18+
```
ejweber, line 377: Nit: Here (and a few lines below) is another location where we used a bare `#`.

chart/values.yaml
ejweber, line 1: Nit: I prefer to adhere to a line length in files like this instead of letting long sentences wrap. Since we're adding a bunch of long sentence comments now, it is a good time to consider breaking them up onto separate lines.

ejweber (dismissed):
LGTM in general.

I brought up some minor formatting NITs that it might be nice to resolve while making large changes to `values.yaml`.

chart/values.yaml
ChanYiLin: Fixed, thanks

chart/values.yaml
ChanYiLin: Fixed, thanks!

chart/values.yaml
ChanYiLin: We use this tool to generate value table in readme: https://github.com/norwoodj/helm-docs
It requires adding comments on top of each value so it can retrieve those comments as descriptions.

The format requires starting with `#--` as below example
```
config:
  databasesToCreate:
    # -- Default database for storage of database metadata
    - postgres

    # -- Database for the [hashbash](https://github.com/norwoodj/hashbash-backend-go) project
    - hashbash
```

The second line doesn't have `--` because the line would be too long if I put it in the first line. It will be concatenated to the first line in the same description.

chart/values.yaml
ChanYiLin: Fixed, thanks!

chart/values.yaml
ChanYiLin: Explained above, thanks!

chart/values.yaml
ChanYiLin: I agree with you, I will try to keep the comment short or break it into multiple line.

chart/values.yaml
```diff
@@ -3,273 +3,442 @@
...
   #  label-key1: "label-value1"
   #  label-key2: "label-value2"
 
 longhornUI:
+  # -- replica count for longhorn ui
   replicas: 2
+  # -- priorityClass count for longhorn ui
   priorityClass: ~
+  # -- Tolerate nodes to run Longhorn UI
   tolerations: []
```
innobead, line 391: The same issue is as the doc PR. Just need to make the description consistent with others.

innobead (changes requested):
In general, LGTM. Just one comment needs to resolve. 

We should make questions.yaml automatically generated as well in the future. Please help create an issue for that and add to 1.7.

chart/values.yaml
ChanYiLin: Fixed.

ChanYiLin:
> In general, LGTM. Just one comment needs to resolve.
> 
> We should make questions.yaml automatically generated as well in the future. Please help create an issue for that and add to 1.7.

Issue created: https://github.com/longhorn/longhorn/issues/6788

innobead (approved):

innobead (approved):

PR #6640, opened 2023-09-06
Object Storage (LEP 20230430)

m-ildefons:
Modifications to the Longhorn helm-chart for the Object Storage gateway for Longhorn.
This includes updates to the CRDs, cluster roles and daemonsets.

Related: https://github.com/longhorn/longhorn-manager/pull/2136
Related: https://github.com/longhorn/longhorn-ui/pull/649
Related: https://github.com/longhorn/longhorn-engine/pull/934
Related: https://github.com/longhorn/backupstore/pull/148
Related: https://github.com/longhorn/longhorn-instance-manager/pull/285

LEP: https://github.com/longhorn/longhorn/pull/5832


chart/templates/storageclass.yaml
```diff
@@ -1,3 +1,18 @@
...
+  name: longhorn-objectstorage
+  labels: {{- include "longhorn.labels" . | nindent 4 }}
+  annotations:
+    storageclass.kubernetes.io/is-default-class: false
+provisioner: driver.longhorn.io
+allowVolumeExpansion: true
+reclaimPolicy: Delete
+volumeBindingMode: WaitForFirstConsumer
+parameters:
+  fsType: xfs
```
PhanLe1010, line 14: we don't use the reclaimPolicy/volumeBindingMode/parameter from this SC, should we remove them here to avoid confusion?

PhanLe1010, line 14: we don't use the reclaimPolicy/volumeBindingMode/parameter from this SC, should we remove them here to avoid confusion?

chart/templates/storageclass.yaml
```diff
@@ -1,3 +1,18 @@
+---
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: longhorn-objectstorage
```
PhanLe1010, line 5: Can we change the name to `longhorn-objectstore-static`? It makes it clear that the volume are static and this SC is just a dummy SC for expansion to work

chart/templates/storageclass.yaml
```diff
@@ -1,3 +1,18 @@
+---
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
```
PhanLe1010, line 3: Can we define a configmap instead of a storageclass directly. Later on, the configmap controller inside Longhorn manager can create the actual SC object. This is similar to how we define the `longhorn-storageclass` configmap below and the controller create the real longhorn SC [here](https://github.com/longhorn/longhorn-manager/blob/463dd8e9fe46650efd611a6d6540079bfddb5d79/controller/kubernetes_configmap_controller.go#L166). There are 2 main reasons for this:
1. SC is immutable. In the future, if we realize we need to update some parameters, we will not be able to update it if we define storageclass directly here. Kubenetes will reject it and Helm upgrade will fail very badly. We hit this problem before that was the reason we move from define SC directly in Helm to define the Configmap and then create/delete SC inside Longhorn manager controller
2. The 2nd reason is it is consistent with the way we are handling SC in Longhorn currently

chart/templates/storageclass.yaml
```diff
@@ -1,3 +1,22 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: longhorn-objectstorage-storageclass
```
PhanLe1010, line 5:
```suggestion
  name: longhorn-objectstorage-static-storageclass
```

chart/templates/storageclass.yaml
```diff
@@ -1,3 +1,22 @@
...
+  namespace: {{ include "release_namespace" . }}
+  labels: {{- include "longhorn.labels" . | nindent 4 }}
+data:
+  storageclass.yaml: |
+    apiVersion: storage.k8s.io/v1
+    kind: StorageClass
+    metadata:
+      name: longhorn-objectstorage-static
+      labels: {{- include "longhorn.labels" . | nindent 8 }}
```
PhanLe1010, line 14: Looks like we don't need to have Helm labels here because the SC is not deployed by Helm

Issue #2238, opened 2021-02-05
[IMPROVEMENT] Add a new settings that allows Longhorn to evict replicas automatically when a node is drained


shuo-wu:
**Is your feature request related to a problem? Please describe.**
See the title

**Additional context**
Derived from https://github.com/longhorn/longhorn/issues/2235

innobead:
remove stale due to the incorrect exempt label

rahul3002:
i would like to contribute on this issue

shuo-wu:
This feature is related to the manager repo only: https://github.com/longhorn/longhorn-manager

There are 2 parts you can refer to before starting:
1. Pick up a random setting and see how other guys add a setting: https://github.com/longhorn/longhorn-manager/blob/master/types/setting.go
2. Check how Longhorn controllers handle replicas when receiving eviction requests: https://github.com/longhorn/longhorn-manager/blob/a0ce2d22899f0d3a0831ca4c032383ecc6691121/controller/replica_controller.go#L342

If you need any help, feel free to contact us.

Again, thank you for your contribution!

innobead:
cc @PhanLe1010

weizhe0422:
Just for notes: 
 - Replica eviction ticket & PR: https://github.com/longhorn/longhorn/issues/298

weizhe0422:
I've read the context of this question and tried to prepare the LEP, there is a short version and wanted to ask if there was any misunderstanding.

 - Main purpose
   - Auto evict the replica on the deleted node to another schedulable node to rebuilding
 - What I need to do are
   - Having a new setting: `automatically-evict-replica-when-node-drained`
   - In node contoller reconcile, if the setting is `true`, then
     - Checks to see if any nodes are drained (Node-Status=`UnSchedulable`), and if so, sets `node.Spec.EvictionRequested` to `true` to let replicas coordinate to trigger replica eviction.
 - Remark
   - Replica eviction from the node (LEP: [Replica Eviction Support for Disks and Nodes](https://github.com/longhorn/longhorn/blob/master/enhancements/20200727-add-replica-eviction-support-for-disks-and-nodes.md))
     - If the volume is detached, it is attached first and then detached after replica eviction is complete.
     - Steps:
[...]

PR #6643, opened 2023-09-06
Automatically evict replicas while draining LEP

ejweber:
#2238

### DIFF
--- a/enhancements/20230905-automatically-evict-replicas-while-draining.md
+++ b/enhancements/20230905-automatically-evict-replicas-while-draining.md
@@ -1,0 +1,280 @@
+# Automatically Evict Replicas While Draining
+
+## Summary
+
+Currently, Longhorn allows the choice between a number of behaviors (node drain policies) when a node is cordoned or
+drained:
+
+- `Block If Contains Last Replica` ensures the `instance-manager` pod cannot be drained from a node as long as it is the
+  last node with a healthy replica for some volume.
+  
+  Benefits:
+
+  - Protects data by preventing the drain operation from completing until there is a healthy replica available for each
+    volume available on another node.
+  
+  Drawbacks:
+
+  - If there is only one replica for the volume, or if its other replicas are unhealthy, the user may need to manually
+    (through the UI) request the eviction of replicas from the disk or node.
+  - Volumes may be degraded after the drain is complete. If the node is rebooted, redundancy is reduced until it is
+    running again. If the node is removed, redundancy is reduced until another replica rebuilds.
+
+- `Allow If Last Replica Is Stopped` is similar to the above, but only prevents an `instance-manager` pod from
+  draining if it has the last RUNNING replica.
+
+  Benefits:
+
+  - Allows the drain operation to proceed in situations where the node being drained is expected to come back online
+    (data will not be lost) and the replicas stored on the node's disks are not actively being used.
+
+  Drawbacks:
+
+  - Similar drawbacks to `Block If Contains Last Replica`.
+  - If, for some reason, the node never comes back, data is lost.
+
+- `Always Allow` never prevents an `instance-manager` pod from draining.
+
+  Benefits:
+
+  - The drain operation completes quickly without Longhorn getting in the way.
+
+  Drawbacks:
+
+  - There is no opportunity for Longhorn to protect data.
+
+This proposal seeks to add a fourth and fifth behavior (node drain policy) with the following properties:
+
+- `Block For Eviction` ensures the `instance-manager` pod cannot be drained from a node as long as it contains any
+  replicas for any volumes. Replicas are automatically evicted from the node as soon as it is cordoned.
+
+  Benefits:
+
+  - Protects data by preventing the drain operation from completing until all replicas have been relocated.
+  - Automatically evicts replicas, so the user does not need to do it manually (through the UI).
+  - Maintains replica redundancy at all times.
+
+  Drawbacks:
+
+  - The drain operation is significantly slower than for other behaviors. Every replica must be rebuilt on another node
+    before it can complete.
+  - The drain operation is data-intensive, especially when replica auto balance is enabled, as evicted replicas may be
+    moved back to the drained node when/if it comes back online.
+  - Like all of these policies, it triggers on cordon, not on drain (it is not possible for Longhorn to distinguish
+    between a node that is actively being drained and one that is cordoned for some other reason). If a user
+    regularly cordons nodes without draining them, replicas will be rebuilt pointlessly.
+
+- `Block For Eviction If Contains Last Replica` ensures the `instance-manager` pod cannot be drained from a node as long
+  as it is the last node with a healthy replica for some volume. Replicas that meet this condition are automatically
+  evicted from the node as soon as it is cordoned.
+
+  Benefits:
+
+  - Protects data by preventing the drain operation from completing until there is a healthy replica available for each
+    volume available on another node.
+  - Automatically evicts replicas, so the user does not need to do it manually (through the UI).
+  - The drain operation is only as slow and data-intensive as is necessary to protect data.
+
+  Drawbacks:
+
+  - Volumes may be degraded after the drain is complete. If the node is rebooted, redundancy is reduced until it is
+    running again. If the node is removed, redundancy is reduced until another replica rebuilds.
+  - Like all of these policies, it triggers on cordon, not on drain (it is not possible for Longhorn to distinguish
+    between a node that is actively being drained and one that is cordoned for some other reason). If a user
+    regularly cordons nodes without draining them, replicas will be rebuilt pointlessly.
+
+Given the drawbacks, `Block For Eviction` should likely not be the default node drain policy moving forward. However,
+some users may find it helpful to switch to `Block For Eviction`, especially during cluster upgrade operations. See
+[user stories](#user-stories) for additional insight.
+
+`Block For Eviction If Contains Last Replica` is a more efficient behavior that may make sense as the long-term setting
+in some clusters. It has largely the same benefits and drawbacks as `Block If Contains Last Replica`, except that it
+doesn't require users to perform any manual drains. Still, when data is properly backed up and the user is planning to
+bring a node back online after maintenance, it is costlier than other options.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/2238
+
+## Motivation
+
+### Goals
+
+- Add a new `Block For Eviction` node drain policy as described in the summary.
+- Ensure that replicas automatically evict from a cordoned node when `Block For Eviction` is set.
+- Ensure a drain operation can not complete until all replicas are evicted when `Block For Eviction` is set
+- Document recommendations for when to use `Block For Eviction`.
+
+### Non-goals
+
+- Only trigger automatic eviction when a node is actively being drained. It is not possible to distinguish between a 
+  node that is only cordoned and one that is actively being drained.
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+I use Rancher to manage RKE2 and K3s Kubernetes clusters. When I upgrade these clusters, the system upgrade controller
+attempts to drain each node before rebooting it. If a node contains the last healthy replica for a volume, the drain
+never completes. I know I can manually evict replicas from a node to allow it to continue, but this eliminates the
+benefit of the automation.
+
+After this enhancement, I can choose to set the node drain policy to `Block For Eviction` before kicking off a cluster
+upgrade. The upgrade may take a long time, but it eventually completes with no additional intervention.
+
+#### Story 2
+
+I am not comfortable with the reduced redundancy `Block If Contains Last Replica` provides while my drained node is
+being rebooted. Or, I commonly drain nodes to remove them from the cluster and I am not comfortable with the reduced
+redundancy `Block If Contains Last Replica` provides while a new replica is rebuilt. It would be nice if I could drain
+nodes without this discomfort.
+
+After this enhancement, I can choose to set the node drain policy to `Block For Eviction` before draining a node or
+nodes. It may take a long time, but I know my data is safe when the drain completes.
+
+### User Experience In Detail
+
+### API changes
+
+Add `block-for-eviction` and `block-for-eviction-if-last-replica` options to the `node-drain-policy` setting. The user
+chooses these options to opt in to the new behavior.
+
+Add a `status.autoEvicting` to the `node.longhorn.io/v1beta2` custom resource. This is not a field users can/should
+interact with, but they can view it via kubectl.
+
+NOTE: We originally experimented with a new `status.conditions` entry in the `node.longhorn.io/v1beta2` custom resource
+with the type `Evicting`. However, this was a bit less natural, because:
+
+- Longhorn node conditions generally describe the state a node is in, not what the node is doing.
+- During normal operation, `Evicting` should be `False`. The Longhorn UI displays a condition in this state with a red
+  symbol, indicating an error state that should be investigated.
+
+Deprecate the `replica.status.evictionRequested` field in favor of a new `replica.spec.evictionRequested` field so that
+replica eviction can be requested by the node controller instead of the replica controller.
+
+## Design
+
+### Implementation Overview
+
+The existing eviction logic is well-tested, so there is no reason to significantly refactor it. It works as follows:
+
+- The user can set `spec.evictionRequested = true` on a node or disk.
+- When the replica controller sees `spec.evictionRequested == true` on the node or disk hosting a replica, it sets
+  `status.evictionRequested = true` on that replica.
+- The volume controller uses `replica.status.evictionRequested == true` to influence replica scheduling/deletion
+  behavior (e.g. rebuild an extra replica to replace the evicting one or delete the evicting one once rebuilding is
+  complete).
+- The user can set `spec.evictionRequested = false` on a node or disk.
+- When the replica controller sees `spec.evictionRequested == false` on the node or disk hosting a replica, it sets
+  `replica.status.evictionRequested = false` on that replica.
+- The volume controller uses `replica.status.evictionRequested == false` to influence replica scheduling/deletion
+  behavior (e.g. don't start a rebuild for a previously evicting replica if one hasn't been started already).
+- NOTE: If a new replica already started rebuilding as part of an eviction, it continues to rebuild and remains in the
+  cluster even after eviction is canceled. It can be cleaned up manually if desired.
+
+Make changes so that:
+
+- The node controller (not the replica controller) sets `replica.spec.evictionRequested = true` when:
+  - `spec.evictionRequested == true` on the replica's node (similar to existing behavior moved from the replica
+    controller), OR
+  - `spec.evictionRequested == true` on the replica's disk. (similar to existing behavior moved from the replica
+    controller), OR
+  - `status.Unschedulable == true` on the associated Kubernetes node object and the node drain policy is
+    `block-for-eviction`, OR
+  - `status.Unschedulable == true` on the associated Kubernetes node object, the node drain policy is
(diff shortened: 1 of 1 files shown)


enhancements/20230905-automatically-evict-replicas-while-draining.md
```diff
@@ -0,0 +1,208 @@
...
+- The user can set `spec.evictionRequested = true` on a node or disk.
+- When the replica controller sees `spec.evictionRequested == true` on the node or disk hosting a replica, it sets
+  `status.evictionRequested = true` on that replica.
+- The volume controller uses `replica.status.evictionRequested == true` to influence replica scheduling/deletion
+  behavior (e.g. rebuild an extra replica to replace the evicting one or delete the evicting one once rebuilding is
+  complete).
+- The user can set `spec.evictionRequested = false` on a node or disk.
+- When the replica controller sees `spec.evictionRequested == false` on the node or disk hosting a replica, it sets
+  `replica.status.evictionRequested = false` on that replica.
+- The volume controller uses `replica.status.evictionRequested == true` to influence replica scheduling/deletion
```
james-munson, line 143: This is a repeat of above, but presumably you wanted to talk about replica.status.evictionRequested == false.

enhancements/20230905-automatically-evict-replicas-while-draining.md
```diff
@@ -0,0 +1,208 @@
...
+- In addition to its pre-existing checks, if the replica controller sees `status.autoEvicting == true` on the node
+  hosting a replica, it sets `status.evictionRequested = true` on that replica.
+- The volume controller still  uses `replica.status.evictionRequested == true` to influence replica
+  scheduling/deletionbehavior (e.g. rebuild an extra replica to replace the evicting one or delete the evicting one once
+  rebuilding is complete).
+- The node controller checks `status.Unschedulable` on the appropriate Kubernetes node object. If
+  `status.Unschedulable == false` or the node drain policy is no longer `block-for-eviction`, it sets
+  `status.autoEvicting = false` on the appropriate `node.longhorn.io` object.
+- In addition to its pre-existing checks, if the replica controller sees `status.autoEvicting == false` on the node
+  hosting a replica, it may set `status.evictionRequested = false` on that replica.
```
james-munson, line 161: What happens to a rebuild in progress?  Does it finish, and then we decide which replica is the best candidate to discard?  (Might not be the one we just built...)

enhancements/20230905-automatically-evict-replicas-while-draining.md
```diff
@@ -0,0 +1,208 @@
...
+
+- Create a volume.
+- Ensure (through soft anti-affinity, high replica count, and/or not enough disks) that an evicted replica of the volume
+  can not be scheduled elsewhere.
+- Write data to the volume.
+- Drain a node one of the volume's replicas is scheduled to.
+- While the drain is ongoing:
+  - Verify that the volume becomes degraded (one of its replicas is unschedulable).
+  - Verify that `node.status.autoEvicting == true`.
+- Verify the drain never completes.
```
james-munson, line 190: How is it made clear to the user what the holdup is, and that it won't change without some action to make it possible?

james-munson (changes requested):
A couple of questions.

enhancements/20230905-automatically-evict-replicas-while-draining.md
ejweber: Yes, this is an unfortunately copy/paste error. Cleaning it up.

enhancements/20230905-automatically-evict-replicas-while-draining.md
ejweber: We currently don't do anything if an extra replica is rebuilt during a canceled eviction. The rebuild finishes and the user simply has extra replicas until they decide to delete one. (This is something we were discussing the other day offline, as I found the behavior a bit surprising.) It may be desirable to add the auto-deletion of a replica when eviction is cancelled, but we should probably do that in another ticket, as I don't think we want auto-eviction behavior to be different than `evictionRequested` behavior in that regard.

james-munson: It would be sufficient (and helpful) to put a sentence in the LEP to that effect - that it is by design, and can be remedied manually.  (I suspect that if auto-balance is enabled, it might find the extra and conclude that it should pare the most constrained replica of the set.)
I use the LEPs as the best documentation of design decisions, so it's valuable to put this kind of thing in there.  Thanks!

enhancements/20230905-automatically-evict-replicas-while-draining.md
james-munson: Reading up on node drain, I gather that it will be considered timed out if the evictions have not completed within some budget (an hour, maybe?  I think it is built into Kubernetes.)  But the non-progressing status won't just hang forever, if I understand it.  So we don't have to debug why it is hung, but rather why it failed.

That's a good thing, but it is hard to link that in a causal way with the blocking item, because the decisions are made at totally different layers.  We can use as a guide the behavior when the pod disruption budget is violated, perhaps.  Is that more than a logging message or an Event?

At the very least, I suppose we can add a knowledge base item about "how to figure out why your eviction failed" with a direct node attachment as one possibility.  I know, I'm not being much help here.

I like the idea of adding the yellow "auto-evicting" status for a node.

enhancements/20230905-automatically-evict-replicas-while-draining.md
ejweber: Added.

ejweber:
@james-munson has provided some helpful feedback on this already. @longhorn/dev-control-plane, can we get some other eyes on it so I can progress?

enhancements/20230905-automatically-evict-replicas-while-draining.md
```diff
@@ -0,0 +1,211 @@
...
+
+  Benefits:
+
+  - Protects data by preventing the drain operation from completing until all replicas have been relocated.
+  - Automatically evicts replicas, so the user does not need to do it manually (through the UI).
+  - Maintains replica redundancy at all times.
+
+  Drawbacks:
+
+  - The drain operation is significantly slower than for other behaviors. Every replica must be rebuilt on another node
```
shuo-wu, line 59: I am thinking if we can do rebuilding for the last healthy replicas on the cordon node only. I mean, we can make this the 5th option. Rebuilding is really time and resource consuming. If unnecessary, I prefer not to rebuild a new replica from zero.

enhancements/20230905-automatically-evict-replicas-while-draining.md
```diff
@@ -0,0 +1,211 @@
...
+nodes. It may take a long time, but I know my data is safe when the drain completes.
+
+### User Experience In Detail
+
+### API changes
+
+Add a `block-for-eviction` option to the `node-drain-policy` setting. The user chooses this option to opt in to the new
+behavior.
+
+Add a `status.autoEvicting` to the `node.longhorn.io/v1beta2` custom resource. This is not a field users can/should
```
shuo-wu, line 118: It seems that this field is unnecessary in the controller reconciliation either. All controllers can directly check setting  `node-drain-policy` during the eviction handling.

shuo-wu (commented):
In general LGTM

enhancements/20230905-automatically-evict-replicas-while-draining.md
james-munson: It sure would be nice to be able to see that status directly on the node when debugging, though.  Is the concern that a possibly redundant status has the chance that it might be wrong if we hit a failure window before it is updated?

james-munson (commented):
Clearing the request-changes flag, but I'm interested in Shuo's questions.

enhancements/20230905-automatically-evict-replicas-while-draining.md
ejweber: I see your point. I can just move the the logic about whether auto eviction should occur to a utility function and call it wherever we would otherwise check `node.status.autoEvicting`.

A downside is if we want to show auto eviction status in the UI, we must offload the logic for deciding whether auto eviction is occurring to it too. Additionally, there will be nowhere for a user to check whether auto eviction is occurring with kubectl. (They could look at individual replica statuses for `evictionRequested`, but not the node itself.) WDYT?

enhancements/20230905-automatically-evict-replicas-while-draining.md
ejweber: Yes, I think this is valuable. "Block For Eviction If Contains Last Replica" could be another behavior. We'll have to introduce some additional logic in the replica controller that doesn't currently exist. Today, it decides whether to evict a replica solely based on the `spec.evictionRequested` of its node or disk (and with my current changes, based on the `status.autoEvicting` of its node). It'll now need to selectively decide whether a replica should be evicted based on the status of the other replicas.

enhancements/20230905-automatically-evict-replicas-while-draining.md
ejweber: Looks like my comment was a bit redundant with @james-munson's! We must have been responding in parallel.

enhancements/20230905-automatically-evict-replicas-while-draining.md
```diff
@@ -0,0 +1,211 @@
...
+- `Block For Eviction` ensures the `instance-manager` pod cannot be drained from a node as long as it contains any
+  replicas for any volumes. Replicas are automatically evicted from the node as soon as it is cordoned.
+
+  Benefits:
+
+  - Protects data by preventing the drain operation from completing until all replicas have been relocated.
+  - Automatically evicts replicas, so the user does not need to do it manually (through the UI).
+  - Maintains replica redundancy at all times.
+
+  Drawbacks:
```
PhanLe1010, line 57: Maybe another drawback is that there will be a lot of data movement. Brand new replicas are being rebuilt on new nodes. Then possibly, the replica rebalancing feature will move them back.

PhanLe1010 (dismissed):
In general, LGTM

PhanLe1010:
Btw, looks like the checking for healthy/failed replicas to decide whether draining should be allowed may be vulnerable to a race condition: sometimes replica CRs were not up-to-date thus we may kill the last healthy replica because we incorrectly think that there are healthy replicas on other nodes. However, looks like I cannot think of a way to prevent this race. Maybe users have to rely on the auto-salvage feature for this case.

enhancements/20230905-automatically-evict-replicas-while-draining.md
shuo-wu: > It'll now need to selectively decide whether a replica should be evicted based on the status of the other replicas. How do we feel about this approach? Is it an overstep of the replica controller's responsibility?

OK. I understand your concern. Yes, that would be an overstep in the current design. Then what I am thinking is, if we can add a spec field `replica.spec.evictionRequested` rather than `replica.status.autoEvicting` to control it. This new spec field is updated by the node controller. WDYT?

ejweber: Ok, so in general, move the logic of deciding which replicas should be evicted to the node controller without touching the logic of HOW the replicas are actually evicted. I think this works. I'll prove it today and see if there are any problems.

ejweber:
I made significant changes to the LEP, mostly based on a POC I implemented in response to @shuo-wu's comment: https://github.com/longhorn/longhorn/pull/6643#pullrequestreview-1675673151. The spirit is largely the same, but the `block-for-eviction-if-contains-last-replica` behavior required some reworking. The changes are all in the latest commit, so feel free to look just at that one for an understanding of the differences.

Highlights:

- There is a new, additional `block-for-eviction-if-contains-last-replica` behavior.
- The `replica.status.evictionRequested` field is deprecated in favor of `replica.spec.evictionRequested`.
- The logic for kicking off a replica eviction is moved to the node controller (from the replica controller).

enhancements/20230905-automatically-evict-replicas-while-draining.md
james-munson: It occurs to me that if an eviction requires a full rebuild, we are capped at what we can rebuild in an hour, which is on the order of 150 GB, if I recall.  So any configuration with full-rebuild policy AND replicas larger than that is almost certain to fail, right?  Something to mention, perhaps in the docs or KB  that will accompany this setting.

enhancements/20230905-automatically-evict-replicas-while-draining.md
```diff
@@ -0,0 +1,275 @@
...
+I am not comfortable with the reduced redundancy `Block If Contains Last Replica` provides while my drained node is
+being rebooted. Or, I commonly drain nodes to remove them from the cluster and I am not comfortable with the reduced
+redundancy `Block If Contains Last Replica` provides while a new replica is rebuilt. It would be nice if I could drain
+nodes without this discomfort.
+
+After this enhancement, I can choose to set the node drain policy to `Block For Eviction` before draining a node or
+nodes. It may take a long time, but I know my data is safe when the drain completes.
+
+### User Experience In Detail
+
```
james-munson, line 138: At some point, perhaps also in a KB or best-practices doc, we might consider ranking this on a scale of speed vs safety.  WDYT?

james-munson (approved):
Still LGTM.  A couple of observations from me.

enhancements/20230905-automatically-evict-replicas-while-draining.md
```diff
@@ -145,27 +174,37 @@ The existing eviction logic is well-tested, so there is no reason to refactor it
...
+  - `status.Unschedulable == true` on the associated Kubernetes node object, the node drain policy is
+    `block-for-eviction-if-contains-last-replica`, and there are no other PDB-protected replicas for a volume.
+- Much of the logic currently used by the instance manager controller to recognize PDB-protected replicas is moved to
+  utility functions so both the node and instance manager controllers can use it.
+- The volume controller uses `replica.spec.evictionRequested == true` in exactly the same way it previously used
+  `replica.status.evictionRequested` to influence replica scheduling/deletion behavior (e.g. rebuild an extra replica
+  to replace the evicting one or delete the evicting one once rebuilding is complete).
+- The node controller sets `replica.spec.evictionRequested = false` when:
+  - `spec.evictionRequested == false` on the replica's node, AND
+  - `spec.evictionRequested == false` on the replica's disk, AND
```
shuo-wu, line 195: For most of the evicting replicas, there is no need to unset `replica.spec.evictionRequested` since they will be cleaned up after eviction complete (IIRC).
The only exception is the eviction cancellation. I am wondering if these 2 conditions are enough for the canceling. And that's the current implementation. If a kube node is still in state `cordon`, the controller always updates the corresponding node/disk `spec.evictionRequested` to the correct state. the replica field just needs to rely on node/disk `spec.evictionRequested` without duplicatedly considering the kube node.

enhancements/20230905-automatically-evict-replicas-while-draining.md
```diff
@@ -145,27 +174,37 @@ The existing eviction logic is well-tested, so there is no reason to refactor it
...
+  - `status.Unschedulable == false` on the associated Kubernetes node object, OR
+  - `status.Unschedulable == true` on the associated Kubernetes node object and the node drain policy is not
+    `block-for-eviction` or `block-for-eviction-if-contains-last-replica`, OR
+  - `status.Unschedulable == true` on the associated Kubernetes node object, the node drain policy is
+    `block-for-eviction-if-contains-last-replica`, and there is another PDB-protected replica.
+- The node controller sets `status.autoEvicting = true` when a node has evicting replicas because of the new drain
```
shuo-wu, line 201: Is `node.status.autoEvicting = true` determined by `kubenode.status.Unschedulable == true && (policy == block-for-eviction || policy == block-for-eviction-if-contains-last-replica)`, or by if containiing evicting replicas?

ejweber: In my current POC code, it is determined by "containing evicting replicas". My thinking is that we would like to transition `autoEvicting` to `false` after all replicas that require a drain are gone, even if the drain itself isn't complete (potentially for some completely unrelated reason, like some non-Longhorn-consuming pod that can't finish evicting).

enhancements/20230905-automatically-evict-replicas-while-draining.md
ejweber: > And that's the current implementation. If a kube node is still in state cordon, the controller always updates the corresponding node/disk spec.evictionRequested to the correct state

I am not able to find this behavior. We set the Longhorn node's `Schedulable` condition based on the Kubernetes node's status, but not its `spec.evictionRequested`.

In general, I think we do not want to automatically set `spec.evictionRequested = true` on nodes or disks because this is a user-specified field, and we do not want to lose track of the fact that they set it. I am treating those user specifications separately from any auto-eviction logic, so we need to check both ` `spec.evictionRequested` fields AND the node condition to determine whether to evict replicas. 

- If the user has specified eviction, we do so no matter what.
- If the user has not specified eviction, we may still evict replicas based on the new node drain policies. We need to check if the node is schedulable to activate this behavior.
- When the user is not specifying eviction AND the conditions aren't right for auto-eviction, we cancel evictions.

It does look like we may be able to use the Longhorn node's `Scheduling` condition here instead of the Kubernetes node's field, since the node controller is updating that automatically, but this is a minor optimization.

enhancements/20230905-automatically-evict-replicas-while-draining.md
ejweber: I agree. Adding that to the main issue.

enhancements/20230905-automatically-evict-replicas-while-draining.md
shuo-wu: > I think we do not want to automatically set spec.evictionRequested = true on nodes or disks because this is a user-specified field

OK, I got your point. In other words, we will set `replica.spec.evictionRequested` when node/disk `spec.evictionRequested == true`, or the auto eviction policy is correct with a cordoned kube node. Otherwise, we can unset `replica.spec.evictionRequested`.

NIT: The comment you explained here is more easy-to-understand comparing with the final condition checks listed in the LEP.

enhancements/20230905-automatically-evict-replicas-while-draining.md
shuo-wu: Got it. I am fine with the case you mentioned since `autoEvicting` only cares about Longhorn replicas

shuo-wu (dismissed):
In general LGTM

enhancements/20230905-automatically-evict-replicas-while-draining.md
ejweber: Got it. I wrote these lines after I wrote the code. I probably focused too much on making them mirror the code exactly and not enough on making them readable. I simplified them while cleaning up a couple of other things.

enhancements/20230905-automatically-evict-replicas-while-draining.md
ejweber: Closing this conversation, since `status.autoEvicting` is remaining in the feature as a "true status" now (set true or false depending on whether auto evicting is occuring).

enhancements/20230905-automatically-evict-replicas-while-draining.md
ejweber: Closing this conversation, since `status.autoEvicting` is remaining in the feature as a "true status" now (set true or false depending on whether auto evicting is occuring).

enhancements/20230905-automatically-evict-replicas-while-draining.md
ejweber: Please check https://github.com/longhorn/website/pull/794. I did not mention 150 GB specifically (not sure exactly where that number is documented), but tried to highlight the need to adjust drain timeouts accordingly.

PhanLe1010 (approved):
LGTM. Thanks for the great work!

PhanLe1010 (approved):

innobead:
LGTM. This is good.

cc @derekbit @shuo-wu

Issue #4883, opened 2022-11-16
[FEATURE] Support Encryption for VolumeMode Block


joshimoo:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

Currently we only support encryption for volume mode FS, we should also allow for encryption in block mode.
This is useful for harvester where we need to consider the impact on live migration.

## Describe the solution you'd like

Similiar to the FS encryption we can shadow the raw block dev and provide the encrypted block dev instead as part of the csi plugin.

## Additional context
- https://github.com/harvester/harvester/issues/3129

innobead:
After discussing with @joshimoo , he needs more time on testing to ensure it works w/o any side effects, so moving this to 1.5.0 instead. 

cc @longhorn/qa

innobead:
This issue will influence how online expansion support will be supported for encrypted volume. 

cc @shuo-wu

shuo-wu:
> This issue will influence how online expansion support will be supported for encrypted volume.
> 
> cc @shuo-wu

Currently, there is no side effect. Due to the upstream issue, Longhorn CSI cannot support the encryption volume expansion. In the future, once the upstream feature is ready, we still need to do resizing for the encrypted device, which is a kind of regular handling.

roger-ryao:
**Verified on master-head 20231124**

- longhorn master-head https://github.com/longhorn/longhorn/commit/57bf3edb1faa955ab077187a518d7be99ff38792
- longhorn-manager master-head https://github.com/longhorn/longhorn-manager/commit/dd663887de73664ffc3296465a0a4b9d2ccbdfb6

**Result Passed**

- [x] require/auto-e2e-test
 - https://github.com/longhorn/longhorn/issues/5020
   - https://github.com/longhorn/longhorn-tests/pull/1443

- [ ] require/doc - https://github.com/longhorn/website/pull/775
 - @derekbit [Please check the review comment from](https://github.com/longhorn/website/pull/775#issuecomment-1785056403) @jillian-maroket.

- [x] ~require/manual-test-plan~ - https://github.com/longhorn/longhorn/issues/4883#issuecomment-1619514908
  - We have the PR https://github.com/longhorn/longhorn-tests/pull/1443 so we don't need to add manual-test-plan

PR #6731, opened 2023-09-19
example: add an example for encrypted volume in block mode

derekbit:
Longhorn/longhorn#4883


examples/block/crypto/storageclass-crypto-global.yaml
```diff
@@ -0,0 +1,33 @@
...
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: longhorn-crypto-global-migratable
+provisioner: driver.longhorn.io
+allowVolumeExpansion: true
+parameters:
+  numberOfReplicas: "3"
+  staleReplicaTimeout: "2880" # 48 hours in minutes
+  fromBackup: ""
+  migratable: "true"
```
innobead, line 11: Why using `migratable: "true""`?

examples/block/crypto/deployment_with_pvc.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: longhorn-block-pvc
+  namespace: default
+spec:
+  accessModes:
+    - ReadWriteMany
+  volumeMode: Block
```
innobead, line 9: RWX + block volume, it seems a unsupported combination, even though the current implementation would just treat it as RWX no matter what type of volume mode.

innobead (changes requested):
Some comments need to resolve first.

innobead:
@derekbit Please help update this, then we can merge it.

examples/block/crypto/storageclass-crypto-global.yaml
derekbit: Just an example. Will

It can be
`migratable: "true""` + `accessMode: ReadWriteMany` (a block device for vm) or
`migratable: "false""` + `accessMode: ReadWriteOnce` (a generic block device).
I think both cases are fine. Will update the example to the second one.

examples/block/crypto/deployment_with_pvc.yaml
derekbit: > RWX + block volume, it seems a unsupported combination

Harvester uses RWX + migratable: "true"" + block volume, right?

derekbit:
> @derekbit Please help update this, then we can merge it.

Updated. Thank you.

innobead (approved):

examples/block/crypto/deployment_with_pvc.yaml
innobead: Right! I thought it was RWO + migratable.

examples/block/crypto/storageclass-crypto-global.yaml
innobead: Since `migratble` is a specific case for volume migration used in Harvester, it would be more helpful to have a general case instead.

Issue #6655, opened 2023-09-11
[IMPROVEMENT] Support both NFS `hard` and `soft` with custom `timeo` and `retrans` options for RWX volumes


derekbit:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

When a RWX volume is attached, a share-manager pod embedded with a userspace NFS server is created and the volume is exported. A remote exported share is `hard` mounted by Longhorn, and it is then provided to the workload. When the share-manager pod or embedded NFS server is somehow crashed or unreachable, the 'hard mount' option keeps the client trying to connect to NFS server and prevents data loss.

It has been reported that a reboot hangs when the NFS share is hard mounted and that the connection to the NFS server is lost during I/O operations.

- https://bugs.launchpad.net/ubuntu/+source/nfs-utils/+bug/1577120
- https://bugzilla.redhat.com/show_bug.cgi?id=1312002
- https://access.redhat.com/solutions/3339001

The root cause is the Linux kernel is trying to maintain filesystem stability. Linux kernel will not allow a filesystem to be unmounted until all its pending IO is written back to storage, and the system can't shut down until all file systems are
unmounted. Currently, the bug/issue is not resolved.

A feasible workaround is using `soft` instead. It is possible to mitigate the potential data loss by using the sync mount option. However, the `sync` option makes the RWX volumes unusable in most practical applications due to poor IO performance.

As discussed with @innobead, at the trade-off the potential data loss and IO performance, the Longhorn system can use both the "hard" option and allow users to use the "soft" option with long custom [`timeo` and `retrans`](https://linux.die.net/man/5/nfs) values.

## Additional context

cc @james-munson

james-munson:
This is for use as a mounted volume, but a similar ticket exists for backup target mount options:  #6608 

They are separate cases, and would not share a common configuration, but is the answer in either case to allow a list of mount options (or maybe even a customized mount command) and if specified, use that string instead of the default set of mount options?

derekbit:
> They are separate cases, and would not share a common configuration, but is the answer in either case to allow a list of mount options (or maybe even a customized mount command) and if specified, use that string instead of the default set of mount options?

Yes, I think we can allow a list of mount options.

derekbit:
### Options

There are two **mount options** in the storage class 
#### mountOptions

[...]

PR #6767, opened 2023-09-22
Add nfsOptions parameter to sample storageclass.yaml

james-munson:
More housekeeping for #6655


innobead:
Use backport PR instead.

Issue #6854, opened 2023-10-05
[IMPROVEMENT] Old kernel such as 3.10.0 set provisioning_mode to wrong value (writesame_16, disabled, full, ...) but not the correct value (unmap) so the trim feature doesn't work


PhanLe1010:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

Old kernels such as 3.10.0 set provisioning_mode to wrong value (writesame_16, disabled, full, ...) but not the correct value (unmap) so the trim feature doesn't work

Reproduce steps:
1. Cluster a cluster with CentOS7, kernel 3.10.0-1160.92.1.el7.x86_64
1. Install Longhorn v1.5.1
1. Deploy a deployment with a Longhorn PVC
1. Exec into the node where the Longhorn volume is currently attached to
1. Check the provisioning_mode of the Longhorn disk. Kernel somehow sets the provisioning_mode of the disk to writesame_16 while the correct value should be unmap:
    ```
    [root@phan-v147-cloudera-pool1-f1dec634-cm59v ~]# find /sys/ -name provisioning_mode -exec grep -H . {} + | sort
    /sys/devices/platform/host3/session1/target3:0:0/3:0:0:1/scsi_disk/3:0:0:1/provisioning_mode:writesame_16
    ```
1. Run fstrim in Longhorn UI for the volume. Kernel sets the provisioning_mode of the disk to `disabled` (the correct value should be unmap)
    ```
    [root@phan-v147-cloudera-pool1-f1dec634-cm59v ~]# find /sys/ -name provisioning_mode -exec grep -H . {} + | sort
    /sys/devices/platform/host3/session1/target3:0:0/3:0:0:1/scsi_disk/3:0:0:1/provisioning_mode:disabled
    ```

## Describe alternatives you've considered

1. Why old kernel doesn't set it correctly? What kernel versions are affected. Which patch version fixes the problem. Then create a knowledge base article for this.
2. Improve the check script that run before user install Longhorn https://github.com/longhorn/longhorn/blob/master/scripts/environment_check.sh to detect and warn that old kernel might affect trim feature

PhanLe1010:
**Tes steps:**

1. Create a cluster with CentOS 7 an kernel v3.10
1. Run the [environment_check.sh](https://github.com/longhorn/longhorn/pull/6860) script
[...]

PR #6860, opened 2023-10-07
Add kernel release check to environment_check.sh

PhanLe1010:
longhorn/longhorn#6854

### DIFF
--- a/scripts/environment_check.sh
+++ b/scripts/environment_check.sh
@@ -273,6 +273,28 @@ check_nodes() {
   fi
 }
 
+verlte() {
+    printf '%s\n' "$1" "$2" | sort -C -V
+}
+
+verlt() {
+    ! verlte "$2" "$1"
+}
+
+check_kernel_release() {
+  local pod=$1
+
+  recommended_kernel_release="5.8"
+
+  local kernel=$(detect_node_kernel_release ${pod})
+
+  if verlt "$kernel" "$recommended_kernel_release"  ; then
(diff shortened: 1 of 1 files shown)


PhanLe1010:
@mergify backport v1.5.x v1.4.x

PhanLe1010:
ping @derekbit @innobead

scripts/environment_check.sh
```diff
@@ -273,6 +273,28 @@ check_nodes() {
...
+verlt() {
+    ! verlte "$2" "$1"
+}
+
+check_kernel_release() {
+  local pod=$1
+
+  recommended_kernel_release="5.8"
+
+  local kernel=$(detect_node_kernel_release ${pod})
```
james-munson, line 289: shellcheck would be happier with a quoted "${pod}" both here and below, but not a big deal.

james-munson (approved):
LGTM.  This will be good to have in place.

scripts/environment_check.sh
PhanLe1010: Looks like we are not using quote in a lot of places in this script already so perhaps we should keep the same to keep it uniform unless we decided to change all of them in another ticket?

ejweber (approved):
In general, LGTM, but one minor concern that we discussed offline.

It may be difficult to remember to keep this 5.8 updated. I'm somewhat fine with relying on developers to search the code base for any uses of some constant when updating it, but this one might be problematic, as we document our recommendation in a different repo. Hopefully we don't change our recommendation often and we can keep on eye on it.

PR #6974, opened 2023-10-26
Fix app name label

timstoop:
According to Kubernetes docs, it's customary to use app.kubernetes.io/name instead of just app
(https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-name). This helps other tools (like cilium hubble) easier identify the workloads on the cluster.

### DIFF
--- a/chart/templates/daemonset-sa.yaml
+++ b/chart/templates/daemonset-sa.yaml
@@ -2,17 +2,17 @@ apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   labels: {{- include "longhorn.labels" . | nindent 4 }}
-    app: longhorn-manager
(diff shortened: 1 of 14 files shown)


derekbit (commented):
In general, LGTM.
However, we need to check the codes in longhorn-manager and other components and see if we need to keep `app: longhorn-manager`.

timstoop:
Good point, I made an additional check and found a few places where I forgot to fix it and some scripts that are using those. Adding the commits now.

I did not found any reference to the label key "app" in the actual longhorn code, but I did come across some examples and code in `deploy` that is using the app key. I'll modify those as well.

timstoop:
Actually, I messed this up. I'll start again using the correct labels and I noticed there are already several helper scripts to help with this as well.

PR #6976, opened 2023-10-26
release: v1.4.4

innobead:
Signed-off-by: davidko <dko@suse.com>

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: longhorn
-version: 1.4.4-rc2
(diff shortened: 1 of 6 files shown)


chriscchien:
This PR verified with blow test scenarios

 - [x] Fresh install v1.4.4 by helm
 - [x]  Upgrade from v1.4.3 to v1.4.4 by helm
      - Create Longhorn CRs before upgrade
      - After upgrade success, restore volume, replica rebuild, detach/attach volume by scale workload worked properly
 - [x] Uninstall
     - No `longhorn.io` CRD remain
     - Name space `longhorn-system` terminated

PR #6983, opened 2023-10-27
Fix chart labels

timstoop:
Switch the `app` label key to the well known `app.kubernetes.io/component` key instead. This helps other applications (like cilium hubble) identify the Pods easier. More information about the label can be found here: https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-component

I've remove the `app:` label from the tls-secret, as I felt like it did not serve a purpose and all important labels are already added by the helper script.

I was unable to find any reference to this label in the application code (but that said, I'm not very familiar with the application code).

This is a replacement for PR #6974 where I messed up the label key name.


innobead (approved):
LGTM. 

Some other labels are automaticaly generated for each resource created for a Longhorn instance installation. 

ref: [/deploy/longhorn.yaml#L4097-L4099](https://github.com/longhorn/longhorn/blob/548aa659738430552a95f3f94b3c6f4c7efc4f52/deploy/longhorn.yaml#L4097-L4099)
```yaml
    app.kubernetes.io/name: longhorn
    app.kubernetes.io/instance: longhorn
    app.kubernetes.io/version: v1.6.0-dev
```

innobead:
@derekbit @PhanLe1010 @shuo-wu any concerns about this change?

derekbit:
longhorn-manager codes need changes as well, e.g. https://github.com/longhorn/longhorn-manager/blob/master/app/post_upgrade.go#L108.

innobead:
> longhorn-manager codes need changes as well, e.g. [longhorn/longhorn-manager@`master`/app/post_upgrade.go#L108](https://github.com/longhorn/longhorn-manager/blob/master/app/post_upgrade.go?rgh-link-date=2023-11-09T12%3A28%3A41Z#L108).

@timstoop Would you like to contribute this change to longhorn-manager as well?

timstoop:
I'm no Go programmer, so I don't have the local toolset available to actually test this. Happy to make the change, but I have to depend on you to make sure the change actually works!

timstoop:
Yeah, I had a look at that code, but I have no idea what it is doing. In my mind, it's probably changing `pod.Labels["app"]` into `pod.Labels["app.kubernetes.io/component"]`, but that's honestly just a guess.

innobead:
We can keep the original label `app` as is to ensure not breaking the current implementation, but still can add the new `app.kubernetes.io/component` label to support external solutions integration.

derekbit:
> We can keep the original label `app` as is to ensure not breaking the current implementation, but still can add the new `app.kubernetes.io/component` label to support external solutions integration.

@timstoop  WDYT?

derekbit:
Since we haven't heard back from the author in a while, let's close the PR.
cc @innobead

Issue #7019, opened 2023-11-01
[BUG] Helm charts do not set the ipFamilies attribute of Services in dual-stack clusters with IPv6 as their default address family


lel-amri:
## Description

Helm charts do not set the `ipFamilies` attribute for the `longhorn-frontend` service. The `longhorn-ui` does not support IPv6. For dual-stack clusters that have IPv6 as their primary family, the `ipFamilies` for the `longhorn-frontend` ends up being automatically set to `["IPv6"]`, thus making the Longhorn UI inaccessible.

## To Reproduce

* Create a K8S cluster with an IPv6 range as the first first value for `kube-apiserver`'s `--service-cluster-ip-range` option.
* Install Longhorn with the Helm charts.
* Observe that the service automatically gets `ipFamilies` set to `["IPv6"]` (`kubectl get -o yaml -n longhorn-system  service/longhorn-frontend`).
* Try to access the UI via `kubectl proxy` at `http://localhost:8001/api/v1/namespaces/longhorn-system/services/longhorn-frontend:http/proxy/`.
* Get an error that looks like this:

  ```json
  {
    "kind": "Status",
    "apiVersion": "v1",
    "metadata": {},
    "status": "Failure",
    "message": "error trying to reach service: proxy error from 127.0.0.1:6443 while dialing [R:E:D:A:C:T:E:D]:8000, code 502: 502 Bad Gateway",
    "reason": "ServiceUnavailable",
    "code": 503
  }
  ```

## Expected behavior

The `longhorn-frontend` service have its `ipFamilies` attribute "manually" set to `["IPv4"]` (until https://github.com/longhorn/longhorn/issues/2259 is solved) by the Helm charts and is thus accessible.

## Support bundle for troubleshooting

N/A

## Environment

 - Longhorn version: v1.4.4
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl): Helm
 - Kubernetes distro and version: K3s v1.27.6+k3s1

Issue #2259, opened 2021-02-16
[BUG] Volume stuck in an attach/detach loop


BogdanIonesq:
**Describe the bug**
When trying to attach a volume, either by using the UI or `kubectl` commands, it remains stuck in an attach/detach loop. 

**To Reproduce**
1. Go to Longhorn UI > Volume.
2. Click on 'Create Volume'. Use a random name, size 1Gi, everything else left on default. Click 'OK'.
3. Try to attach the newly created volume to any node.
4. See error: volume stuck in attaching/detaching loop.

**Expected behavior**
The volume should be successfully attached to the desired node.

**Log**
Relevant logs appear on the `instance-manager-e` on the node to which the volume should be attaching:
```
[test-volume-e-fe077bd4] time="2021-02-15T14:40:52Z" level=info msg="Shutdown SCSI device for target iqn.2019-10.io.longhorn:test-volume"
[test-volume-e-fe077bd4] time="2021-02-15T14:40:53Z" level=info msg="device test-volume: SCSI device /dev/longhorn/test-volume shutdown"
[longhorn-instance-manager] time="2021-02-15T14:40:53Z" level=info msg="wait for gRPC service of process test-volume-e-fe077bd4 to start at localhost:10000"
[test-volume-e-fe077bd4] go-iscsi-helper: tgtd is already running
[test-volume-e-fe077bd4] time="2021-02-15T14:40:54Z" level=info msg="go-iscsi-helper: found available target id 1"
tgtd: device_mgmt(246) sz:79 params:path=/var/run/longhorn-test-volume.sock,bstype=longhorn,bsopts=size=1073741824
tgtd: bs_thread_open(409) 16
[test-volume-e-fe077bd4] time="2021-02-15T14:40:54Z" level=info msg="New data socket connnection established"
[test-volume-e-fe077bd4] time="2021-02-15T14:40:54Z" level=info msg="default: automatically rescan all LUNs of all iscis sessions"
[longhorn-instance-manager] time="2021-02-15T14:40:54Z" level=info msg="wait for gRPC service of process test-volume-e-fe077bd4 to start at localhost:10000"
[longhorn-instance-manager] time="2021-02-15T14:40:55Z" level=info msg="wait for gRPC service of process test-volume-e-fe077bd4 to start at localhost:10000"
[longhorn-instance-manager] time="2021-02-15T14:40:56Z" level=info msg="wait for gRPC service of process test-volume-e-fe077bd4 to start at localhost:10000"
[...]

PR #7020, opened 2023-11-01
Manually set Service ipFamilies where needed

lel-amri:
This is a PR that attempts to fix issues such as https://github.com/longhorn/longhorn/issues/7019 until https://github.com/longhorn/longhorn/issues/2259 is done.

This PR is currently a draft and is waiting for feedback and thorough testing until every service is appropriately configured (unless someone knowledgeable about the current state of IPv6 support can speed things up).


derekbit:
No needed. IPv6 will be supported since v1.10.
cc @c3y1huang @innobead

Issue #6798, opened 2023-09-27
[IMPROVEMENT] Use nvme-cli in instance-manager pod instead


derekbit:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

Using nvme-cli in the host environment can be frustrating due to the variations in versions and unexpected crashes encountered in some releases. Resolving these issues often involves time-consuming investigations.

To mitigate this unnecessary hassle, using nvme-cli within an instance-manager pod is a better approach.

## Additional context

Segfault in nvme 2.0 https://github.com/longhorn/longhorn/discussions/6493#discussioncomment-7078153
Weird nvme output: https://github.com/longhorn/longhorn/issues/6795
Unsupported format in old nvme (< v1.12): https://longhorn.io/docs/1.5.1/spdk/quick-start/#install-nvme-userspace-tool-and-load-nvme-tcp-kernel-module

derekbit:
cc @shuo-wu

c3y1huang:
Duplicated to https://github.com/longhorn/longhorn/issues/6517?

derekbit:
Yeah, sorry, I wasn't aware of the ticket.
Can I take over it? I can test it together.

cc @innobead

c3y1huang:
> Yeah, sorry, I wasn't aware of the ticket. Can I take over it? I can test it together.
> 
> cc @innobead

Fine with me.

derekbit:
Except for the issues mentioned in https://github.com/longhorn/longhorn/issues/6798#issue-1914718798
I also found `nvme list-subsys -o json` cannot list subsystems using nvme v2.5.
Becase nvme v2.x is more or less problematic in nvme v2.x, I will use v1.16 in the instance-manager pod.

chriscchien:
Hi @derekbit ,

Does the environment check [script](https://raw.githubusercontent.com/longhorn/longhorn/master/scripts/environment_check.sh) need to be updated too? Because the script will still check system nvme-cli version, thank you.
```
[INFO]  Checking nvme-cli...
[ERROR] Failed to check nvme-cli version on node ip-172-31-44-162
[ERROR] Failed to check nvme-cli version on node ip-172-31-44-241
[ERROR] Failed to check nvme-cli version on node ip-172-31-46-50
```

In addition, the V2 volume behavior verified **pass** in longhorn master (longhorn-manager `35beba`) with [steps](https://longhorn.io/docs/1.6.0/spdk/quick-start/)

[...]

PR #7119, opened 2023-11-16
Remove nvme-cli check

chriscchien:
Regarding https://github.com/longhorn/longhorn/issues/6798, environment script no longer need to check `nvme-cil` since instance-manager have `nvme-cil` installed.


derekbit (approved):
LGTM. Thank you.

chriscchien:
@mergify backport v1.5.x

Issue #7224, opened 2023-11-29
[TASK] Bump up the minimum supported Kubernetes version


c3y1huang:
## What's the task? Please describe

- Kubernetes version support, min 1.26 to align with upstream ([1.29 is planned to be released on 12/5](https://www.kubernetes.dev/resources/release/#summary), 1.28 and 1.27 are actively supported. 1.25 was EOS on 10/27.)
  - As per https://metrics.longhorn.io/, 1.25 and 1.24 are primary installed versions. 
  - Rancher supported downstream cluster is RKE/2 1.26
  - RKE2 is supporting 1.28 - 1.26 which is aligned with upstream.
  - For the current Longhorn-supported versions, we will not change the existing supported min version of Kubernetes. 

## Describe the sub-tasks

- [x] Run e2e tests to ensure all passed including installation and upgrade
- [x] Support the supported K8s versions without issues
- [x] Prevent upgrading from any unsupported K8s versions with notice info, especially for helm chart installation

## Additional context

- https://github.com/longhorn/longhorn/issues/7001
- https://github.com/longhorn/longhorn/issues/2757
- https://github.com/longhorn/longhorn/issues/3891
- https://kubernetes.io/docs/reference/using-api/deprecation-guide/
- https://endoflife.date/kubernetes

innobead:
https://kubernetes.io/releases/patch-releases/#support-period

c3y1huang:
Test unsupported Kubernetes version:
```shell
> helm install longhorn ./chart/ --namespace longhorn-system --create-namespace
Error: INSTALLATION FAILED: chart requires kubeVersion: >=1.26.0-0 which is incompatible with Kubernetes v1.25.15+k3s2
```

innobead:
Extra info about Kubernetes version support in public cloud providers is below:

AKS/EKS/GKE 1.25 support will soon reach out to EOL. 

- https://endoflife.date/google-kubernetes-engine
- https://endoflife.date/amazon-eks
- https://endoflife.date/azure-kubernetes-service

Also, RKE2/K3s 1.25 already reached EOL. They cover over 50% of installation from Longhorn usage metrics.

PhanLe1010:
Sorry for the late comment. Up until now, we usually follow the convention that we only raise min supported Kubernetes version if keeping the current min K8s version makes it impossible to support the newest K8s version. Reference https://github.com/longhorn/longhorn/issues/3891#issuecomment-1131258472

Is it the case here?

[...]

PR #7225, opened 2023-11-29
build: bump up the minimum supported Kubernetes version

c3y1huang:
https://github.com/longhorn/longhorn/issues/7224


innobead:
@longhorn/maintainer any concerns about updating the min version to 1.26?

Issue #6528, opened 2023-08-15
[FEATURE] Have default priorityClass to prevent unexpected longhorn pods eviction


innobead:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

Have a default priority class with the highest value to ensure longhorn pods have not been evicted when low system resources by the Kube scheduler.  Right now, the setting is optional.

https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/

## Additional context

cc @longhorn/dev

harridu:
Will existing deployments be migrated to the new default priority class on the fly? I am not sure if this is possible.

innobead:
> Will existing deployments be migrated to the new default priority class on the fly? I am not sure if this is possible.

In the current implementation, it's not possible. However, we should improve this part to have the following support to prevent any interruption for running volumes.

- When changing any volume-related Pod configuration (ex: IM pod, share manager pod, upcoming object storage/s3gw pod) like priority class, the change will only take effect when doing volume upgrade or volume reattachement. 

Extra info from @PhanLe1010 

![image](https://github.com/longhorn/longhorn/assets/899492/b34cb163-a8dc-401b-aafc-9c18e9efb057)

cc @shuo-wu @derekbit @c3y1huang

chriscchien:
Verified **pass** on longhorn v1.6.0-rc1 with test [stpes](https://github.com/longhorn/longhorn/issues/6528#issuecomment-1840847320)

After install v1.6.0-rc1, can see priority class `longhorn-critical`  in all longhorn pods
```
> k get priorityclass
NAME                      VALUE        GLOBAL-DEFAULT   AGE
system-node-critical      2000001000   false            8m54s
system-cluster-critical   2000000000   false            8m54s
longhorn-critical         1000000000   false            8m4s
> k get setting -n longhorn-system | grep priority
priority-class                                                    longhorn-critical                                 8m4s
```

PR #7252, opened 2023-12-04
feat(lep): default priority class

mantissahz:
Add a default priority class with the highest value to the setting `priority-class`.

Ref: longhorn/longhorn#6528

### DIFF
--- a/enhancements/20231204-default-priority-class.md
+++ b/enhancements/20231204-default-priority-class.md
@@ -1,0 +1,110 @@
+# Default Priority Class
+
+## Summary
+
+Pods can have priority which indicates the importance of a pod relative to other pods.
+Longhorn pods will face unexpected eviction due to insufficient cluster resources or being preempted by higher-priority pods
+A PriorityClass defines a mapping from a priority class name to the integer value of the priority, where a higher value signifies greater priority..
+
+### Related Issues
+
+[Have default priorityClass to prevent unexpected longhorn pods eviction.](https://github.com/longhorn/longhorn/issues/6528)
+
+## Motivation
+
+### Goals
+
+- Reduce the opportunity of the unexpected eviction of Longhorn pods.
+
+### Non-goals [optional]
+
+`None`
+
+## Proposal
+
+Have a default priority class with the highest value to all Longhorn pods.
+
+### User Stories
+
+After a fresh installation or an upgrade, users will have Longhorn pods with the field `priorityClassName: longhorn-critical` in `spec` to prevent Longhorn pods from unexpected eviction.
+
+- For install, the priority class will not be set using the default value if the value has been provided by users.
+- For upgrade, the priority class will not be set using the default value if the settings already exists.
+
+### User Experience In Detail
+
+1. For a fresh install, the default priority class will be set if there is no priority class set by users. The setting of the priority class will be updated.
+2. For an upgrade, if all volumes are detached and the setting of the priority class is empty, the setting should be applied including user-managed components. The setting of the priority class will be updated because the source of truth of settings is the config map (longhorn-default-setting).
+3. For an upgrade, no matter weather all volumes are detached or not, if the setting of the priority class has been set, the setting should not be updated with the default priority class.
+4. For an upgrade, if volumes are attached, the setting should be applied to the user-managed components only if they haven't been set. The setting of the priority class will not be updated.
+
+NOTE: Before modifying the setting `priority-class`, all Longhorn volumes must be detached.
+
+### API changes
+
+`None`
+
+## Design
+
+### Implementation Overview
+
+1. Adding a new template YAML file in the Longhorn chart/templates to create a default priority class.
+
+```yaml
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: longhorn-critical
+# The example value of a production is 4000000 from https://kubernetes.io/blog/2023/01/12/protect-mission-critical-pods-priorityclass
+# Most of the paragraph uses 1000000 for high priority class.
+# system default priority classes:
+#   """txt
+#     NAME                      VALUE        GLOBAL-DEFAULT   AGE
+#     system-node-critical      2000001000   false            5h30m
+#     system-cluster-critical   2000000000   false            5h30m
+#   """
+# The maximum allowed value of a user defined priority is 1000000000
+value: 1000000000
+description: "Ensure Longhorn pods have the highest priority to prevent any unexpected eviction by the Kubernetes scheduler under node pressure"
+globalDefault: false
+preemptionPolicy: PreemptLowerPriority
+```
+
+2. Set the `defaultSettings.priorityClass` to `longhorn-critical` and `priorityClass` of Longhorn components in the value.yaml of Longhorn chart.
+
+```yaml
+...
+defaultSettings:
+  ...
+  priorityClass: &defaultPriorityClassNameRef "longhorn-critical"
+  ...
+longhornManager:
+  log:
+    # -- Options: `plain`, `json`
+    format: plain
+  # -- Priority class for longhorn manager
+  priorityClass: *defaultPriorityClassNameRef
+longhornDriver:
+  # -- Priority class for longhorn driver
+  priorityClass: *defaultPriorityClassNameRef
+longhornUI:
+  # -- Priority class count for longhorn ui
+  priorityClass: *defaultPriorityClassNameRef
+```
+
+### Test plan
+
+1. Modify the test case `test_setting_priority_class` to check if there is a default priority class `longhorn-critical` existing.
+2. Add a test case where the `defaultSettings.priorityClass` has been modified by users and the setting `priority-class` should be the same to the value provided by users after a fresh install.
+3. Add a test case where the setting `priority-class` will be updated if all volumes are detached and the setting of the priority class is not set for an upgrade.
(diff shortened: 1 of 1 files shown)


innobead:
_*User cases:*_
- For fresh install, the default priority class will be set if there is no priority class set by users. The setting of the priority class will be updated. 
- For upgrade, if volumes are attached, the setting should be applied except for user-managed components. The setting of the priority class will not be updated. 
- For upgrade, if all volumes are detached, the setting should be applied including user-managed components. The setting of the priority class will be updated because the source of truth of settings is the config map (`longhorn-default-setting`). 

_*Notes for users:*_

- Because the setting of priority class will be overridden no matter if it has been set or not in the above item 3, we need to add this to the important note. 
- For item 2, because the setting will not be updated, need to notify users about this. If they want to make it take effect, need to update on their own. 

Anything missing? @mantissahz

mantissahz:
> User cases:

Add another case:
For upgrade, if all volumes are detached but the setting `priority-class` is set to a priority class, the setting should be applied except for user-managed components. The setting of the priority class will not be updated.

And add Test plans for user cases

enhancements/20231204-default-priority-class.md
```diff
@@ -0,0 +1,108 @@
...
+2. Change the priority class by modifying the setting `priority-class` as mentioned in section [Priority Class](https://longhorn.io/docs/1.6.0/advanced-resources/deploy/priority-class).
+
+### User Experience In Detail
+
+1. For fresh install, the default priority class will be set if there is no priority class set by users. The setting of the priority class will be updated.
+2. For upgrade, if all volumes are detached, the setting should be applied including user-managed components. The setting of the priority class will be updated because the source of truth of settings is the config map (longhorn-default-setting).
+3. For upgrade, if all volumes are detached and the setting of the priority class is set to a priority class name, the setting should be applied except for user-managed components. The setting of the priority class will not be updated.
+4. For upgrade, if volumes are attached, the setting should be applied except for user-managed components. The setting of the priority class will not be updated.
+
+NOTE: Before modifying the setting `priority-class`, all Longhorn volumes must be detached.\
```
innobead, line 41: `\` removed. Please review this PR again to fix some minor doc issues.

enhancements/20231204-default-priority-class.md
```diff
@@ -0,0 +1,108 @@
...
+description: "Used for Longhorn pods to ensure pods have not been evicted when low system resources by the Kube scheduler."
+globalDefault: false
+preemptionPolicy: PreemptLowerPriority
+```
+
+2. Add a new field `priorityClassName` and modify the value of the `priorityClass` setting of Longhorn components in the value.yaml of Longhorn chart.
+
+```yaml
+...
+priorityClassName: longhorn-critical
```
innobead, line 77: Suggest `defaultPriorityClass` to align with the existing variables in values.yaml. 

This is important to make things consistent for UX.

enhancements/20231204-default-priority-class.md
```diff
@@ -0,0 +1,108 @@
...
+  priorityClass: longhorn-critical
+...
+```
+
+3. Add `priorityClassName: {{ .Values.priorityClassName }}` for spec of the daemonset, deployments, or pods in templates yaml files in the Longhorn chart if necessary.
+
+### Test plan
+
+1. Modify the test case `test_setting_priority_class` to check if there is a default priority class `longhorn-critical` existing.
+2. Add a test case which the setting `priority-class` will be updated without a value that the user set or a volume attached after the upgrade.
```
innobead, line 97: Add a test case `where` the ....

enhancements/20231204-default-priority-class.md
```diff
@@ -0,0 +1,108 @@
...
+...
+longhornDriver:
+  # -- Priority class for longhorn driver
+  priorityClass: longhorn-critical
+...
+```
+
+3. Add `priorityClassName: {{ .Values.priorityClassName }}` for spec of the daemonset, deployments, or pods in templates yaml files in the Longhorn chart if necessary.
+
+### Test plan
```
innobead, line 94: Please update the test cases accordingly as per my other comments.

innobead (changes requested):

enhancements/20231204-default-priority-class.md
```diff
@@ -0,0 +1,108 @@
...
+
+Have a default priority class with the highest value to all Longhorn pods.
+
+### User Stories
+
+After a fresh installation or an upgrade, users will have Longhorn pods with a field `priorityClassName: longhorn-critical` in `spec` to prevent Longhorn pods from unexpected eviction.
+
+1. Users can install or upgrade Longhorn by `kubectl` or `Helm` and the default priority class `longhorn-critical` will be set.
+2. Change the priority class by modifying the setting `priority-class` as mentioned in section [Priority Class](https://longhorn.io/docs/1.6.0/advanced-resources/deploy/priority-class).
+
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20231204-default-priority-class.md
```diff
@@ -0,0 +1,108 @@
...
+## Proposal
+
+Have a default priority class with the highest value to all Longhorn pods.
+
+### User Stories
+
+After a fresh installation or an upgrade, users will have Longhorn pods with a field `priorityClassName: longhorn-critical` in `spec` to prevent Longhorn pods from unexpected eviction.
+
+1. Users can install or upgrade Longhorn by `kubectl` or `Helm` and the default priority class `longhorn-critical` will be set.
+2. Change the priority class by modifying the setting `priority-class` as mentioned in section [Priority Class](https://longhorn.io/docs/1.6.0/advanced-resources/deploy/priority-class).
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20231204-default-priority-class.md
```diff
@@ -0,0 +1,108 @@
...
+
+### User Experience In Detail
+
+1. For fresh install, the default priority class will be set if there is no priority class set by users. The setting of the priority class will be updated.
+2. For upgrade, if all volumes are detached, the setting should be applied including user-managed components. The setting of the priority class will be updated because the source of truth of settings is the config map (longhorn-default-setting).
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20231204-default-priority-class.md
```diff
@@ -0,0 +1,108 @@
...
+3. For upgrade, if all volumes are detached and the setting of the priority class is set to a priority class name, the setting should be applied except for user-managed components. The setting of the priority class will not be updated.
```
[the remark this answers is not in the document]
mantissahz: Yes, it is what you describe.
Modified. Thanks.

enhancements/20231204-default-priority-class.md
mantissahz: Modified. Thanks.

enhancements/20231204-default-priority-class.md
mantissahz: Updated. Thanks.

enhancements/20231204-default-priority-class.md
mantissahz: Removed and reviewed the PR again. Thanks

enhancements/20231204-default-priority-class.md
mantissahz: Modified. Thanks.

enhancements/20231204-default-priority-class.md
```diff
@@ -0,0 +1,110 @@
...
+
+- For install, the priority class will not be set using the default value if the value has been provided by users.
+- For upgrade, the priority class will not be set using the default value if the settings already exists.
+
+### User Experience In Detail
+
+1. For a fresh install, the default priority class will be set if there is no priority class set by users. The setting of the priority class will be updated.
+2. For an upgrade, if all volumes are detached and the setting of the priority class is empty, the setting should be applied including user-managed components. The setting of the priority class will be updated because the source of truth of settings is the config map (longhorn-default-setting).
+3. For an upgrade, no matter weather all volumes are detached or not, if the setting of the priority class has been set, the setting should not be updated with the default priority class.
+4. For an upgrade, if volumes are attached, the setting should be applied except for user-managed components. The setting of the priority class will not be updated.
```
innobead, line 39: This one is still confusing. 

Why if volumes are attached, the setting should be applied except for user-managed components? so the setting will be applied to the `priority-class` global setting? If yes, this will cause the system managed pods restarted, right?

Should be:
> For an upgrade, if volumes are attached, the setting should be applied to the user-managed components only if they haven't been set. The setting of the priority class will not be updated.

There are two places to update priority class:
- Each priority class of the user managed components
- Global settings for system managed components

innobead (changes requested):

enhancements/20231204-default-priority-class.md
mantissahz: Modified. Yes, it should be what you described.
>For an upgrade, if volumes are attached, the setting should be applied to the user-managed components only if they haven't been set. The setting of the priority class will not be updated.

Thanks

innobead (approved):
In general, LGTM. 

Here's a corrected version:

In summary, updating the global setting of the priority class will be a best effort, as it will depend on whether there are any running volumes during the upgrade.

PR #7262, opened 2023-12-05
feat: default priority class

mantissahz:
Add a priority class template YAML file to create a default priority class.
Add a new field `defaultPriorityClassName` in value.yaml.
Update the deploy YAML file - longhorn.yaml

Ref: longhorn/longhorn#6528

### DIFF
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -452,6 +452,12 @@ questions:
     group: "Longhorn Default Settings"
     type: boolean
     default: "false"
+  - variable: defaultSettings.priorityClass
+    label: Priority Class
+    description: "The name of the Priority Class to set on the Longhorn components. This can help prevent Longhorn components from being evicted under Node Pressure. Longhorn system contains user deployed components (e.g, Longhorn manager, Longhorn driver, Longhorn UI) and system managed components (e.g, instance manager, engine image, CSI driver, etc.) Note that this setting only sets Priority Class for system managed components. Depending on how you deployed Longhorn, you need to set Priority Class for user deployed components in Helm chart or deployment YAML file. WARNING: DO NOT CHANGE THIS SETTING WITH ATTACHED VOLUMES."
+    group: "Longhorn Default Settings"
+    type: string
+    default: "longhorn-critical"
   - variable: defaultSettings.replicaReplenishmentWaitInterval
     label: Replica Replenishment Wait Interval
     description: "In seconds. The interval determines how long Longhorn will wait at least in order to reuse the existing data on a failed replica rather than directly creating a new replica for a degraded volume."
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -256,7 +256,7 @@ questions:
     default:
   - variable: defaultSettings.allowRecurringJobWhileVolumeDetached
     label: Allow Recurring Job While Volume Is Detached
-    description: 'If this setting is enabled, Longhorn will automatically attaches the volume and takes snapshot/backup when it is the time to do recurring snapshot/backup.'
+    description: 'If this setting is enabled, Longhorn will automatically attach the volume and takes snapshot/backup when it is the time to do recurring snapshot/backup.'
     group: "Longhorn Default Settings"
     type: boolean
     default: "false"
@@ -268,7 +268,7 @@ questions:
     default: "false"
   - variable: defaultSettings.defaultDataPath
     label: Default Data Path
-    description: 'Default path to use for storing data on a host. By default "/var/lib/longhorn/"'
+    description: 'Default path to use for storing data on a host. By default, "/var/lib/longhorn/"'
     group: "Longhorn Default Settings"
     type: string
     default: "/var/lib/longhorn/"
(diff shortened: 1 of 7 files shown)


chart/templates/priority-class.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
```
innobead, line 2: let's change the file name to `priorityclass.yaml` to be consistent with `storageclass.yaml`.

P.S. We should review file name convention later and make names consistent if needed.

chart/templates/priority-class.yaml
```diff
@@ -0,0 +1,9 @@
...
+metadata:
+  name: {{ default "longhorn-critical" .Values.defaultPriorityClassName }}
+  labels: {{- include "longhorn.labels" . | nindent 4 }}
+description: "Used for Longhorn pods to ensure pods have not been evicted when low system resources by the Kube scheduler."
+globalDefault: false
+preemptionPolicy: PreemptLowerPriority
+value: 1000000000
```
innobead, line 9: ref: https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#priorityclass

chart/templates/priority-class.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: {{ default "longhorn-critical" .Values.defaultPriorityClassName }}
+  labels: {{- include "longhorn.labels" . | nindent 4 }}
+description: "Used for Longhorn pods to ensure pods have not been evicted when low system resources by the Kube scheduler."
```
innobead, line 6: `Ensure Longhorn pods have the highest priority to prevent any unexpected eviction by the Kubernetes scheduler under node pressure.`

chart/questions.yaml
```diff
@@ -446,6 +446,12 @@ questions:
       - "allow-if-replica-is-stopped"
       - "always-allow"
     default: "block-if-contains-last-replica"
+  - variable: defaultSettings.priorityClass
```
innobead, line 449: This was missed?! Good catch.

chart/values.yaml
```diff
@@ -342,12 +344,15 @@ privateRegistry:
   # -- If create a new private registry secret is true, create a Kubernetes secret with this name; else use the existing secret of this name. Use it to pull images from your private registry
   registrySecret: ~
 
+# -- The default name of the Priority Class to set on the Longhorn components
+defaultPriorityClassName: &defaultPriorityClassNameRef "longhorn-critical"
```
innobead, line 348: Suggest setting `defaultSettings.priorityClass` to this value as well, so you don't need to hard-code it in longhorn-manager.

innobead (changes requested):
In general, LGTM. Some feedback needs to resolve.

Also, need to have a test coverage to ensure don't break user envs if they already configure priorityClass.

chart/templates/priority-class.yaml
mantissahz: Renamed. Thanks.

chart/templates/priority-class.yaml
mantissahz: Modified. Thanks.

chart/values.yaml
mantissahz: Removed `defaultPriorityClassName` and the longhorn-manager upgrade process will handle the default priority class.
Thanks.

chart/templates/priority-class.yaml
mantissahz: Added into the YAML file `priorityclass.yaml`

chart/templates/deployment-ui.yaml
```diff
@@ -125,9 +125,7 @@ spec:
       imagePullSecrets:
       - name: {{ .Values.privateRegistry.registrySecret }}
       {{- end }}
-      {{- if .Values.longhornUI.priorityClass }}
-      priorityClassName: {{ .Values.longhornUI.priorityClass | quote }}
-      {{- end }}
+      priorityClassName: {{ .Values.longhornUI.priorityClass | default "longhorn-critical" .Values.defaultSettings.priorityClass | quote }}
```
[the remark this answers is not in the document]
mantissahz: Modified. Made a default value of `defaultSettings.priorityClass` to values.yaml.
And this part will revert to the original.

innobead:
@mantissahz as we discussed, review the change I made and also fix `questions.yaml` for the change of `values.yaml`. Of course, fix the conflict with the master. 

Do some tests, then we can merge this PR.

mantissahz:
Hi @innobead,
Tested 4 cases (fresh install, upgrade with a volume attached, upgrade with no volume attached, and upgrade with no volume attached and the setting `priority-class` has been customized by users) in the LEP with the [longhorn-manager PR](https://github.com/longhorn/longhorn-manager/pull/2332) and it looks working as we mentioned.

innobead (approved):
LGTM

Issue #6527, opened 2023-08-15
[TASK] Remove namespaceOverride from the chart values


innobead:
## What's the task? Please describe

Remove namespaceOverride from the chart value, since we don't officially support it can be installed in a different namespace. Enforce this to avoid any incorrect expectation. 

```
## Specify override namespace, specifically this is useful for using longhorn as sub-chart
## and its release namespace is not the `longhorn-system`
namespaceOverride: ""
```

## Describe the sub-tasks

Need to review the impact when removing this value.

roger-ryao:
**Verified on master-head 20231211**

- longhorn master-head https://github.com/longhorn/longhorn/commit/69739d8a63231934c5f7bb0c70619130be25fe23

**The test steps**

https://github.com/longhorn/longhorn/issues/6527#issuecomment-1844875954

**Result Passed**

- [x] 1. Successfully installed Longhorn.
- [x] 2. Successfully created `workloads`, `volumes`, `snapshots`, `backups`, and `system backup`.
- [x] 3. Successfully upgraded Longhorn from version `1.5.3` to `1.6.0-dev`.
- [x] 4. Successfully uninstalled Longhorn.
- [x] 5. [helm-chart-coretest](https://ci.longhorn.io/job/private/job/longhorn-helm-chart-test/10/) passed

mantissahz:
>Harvester needs the field to install LH in longhorn-system. Without it, LH will be in harvester-system . The root cause is that helm can't specify dependence chart in a different namespace, so we need a way to change it. ([ref](https://github.com/helm/helm/issues/10905))

Reopen this issue and add the field back.

cc @FrankYang0529

innobead:
Marked this issue as invalid, as we decided to revert the previous change.

roger-ryao:
**Verified on master-head/v1.6.x-head 20240125**
- longhorn `master-head` https://github.com/longhorn/longhorn/commit/c3a9f120e30287d5ba1b954a0227f637309dff81

- longhorn `v1.6.x-head` https://github.com/longhorn/longhorn/commit/b34b8735e0c98daa9e4d200d3f17c5cb6056e1f3

**The test steps**

https://github.com/longhorn/longhorn/issues/6527#issuecomment-1844875954


**Result Passed**

- [x] 1. Successfully installed Longhorn.
- [x] 2. Successfully created `workloads`, `volumes`, `snapshots`, `backups`, and `system backup`.
[...]

PR #7287, opened 2023-12-07
fix: remove the namespaceOverride value

mantissahz:
`namespaceOverride` in value.yaml was removed by https://github.com/longhorn/longhorn/pull/6639

Remove all `namespaceOverride` fields in the chart.

Ref: longhorn/longhorn#6527, https://github.com/longhorn/longhorn/pull/6639

### DIFF
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -58,9 +58,5 @@ app.kubernetes.io/version: {{ .Chart.AppVersion }}
  define the longhorn release namespace
 */ -}}
 {{- define "release_namespace" -}}
-{{- if .Values.namespaceOverride -}}
-{{- .Values.namespaceOverride -}}
-{{- else -}}
 {{- .Release.Namespace -}}
-{{- end -}}
 {{- end -}}


innobead:
the commit seems not right. it should `fix:` instead of `refacotr:`. 

BTW, will this removal cause a problem if users already installed Longhorn in a non-default namespace?

mantissahz:
>the commit seems not right. it should fix: instead of refacotr:.

Modified to `fix`. Thanks.

>BTW, will this removal cause a problem if users already installed Longhorn in a non-default namespace?

Tested to install Longhorn v1.5.3 in the namespace `lh-myspace` by Helm and created some workloads.
Then upgraded Longhorn with this PR and it looks working well.

chart/templates/_helpers.tpl
```diff
@@ -58,9 +58,5 @@ app.kubernetes.io/version: {{ .Chart.AppVersion }}
  define the longhorn release namespace
 */ -}}
 {{- define "release_namespace" -}}
-{{- if .Values.namespaceOverride -}}
-{{- .Values.namespaceOverride -}}
```
innobead, line 62: Just noticed `namespaceOverride` is not exposed in values.yaml in longhorn repo but in charts repo instead. Need to clean this one later when doing a release.

innobead (approved):

Issue #7357, opened 2023-12-17
[FEATURE] Add linear dm device on the top of v2 volume


derekbit:
## What's the task? Please describe

This is the prerequisite for the [live upgrade feature](https://github.com/longhorn/longhorn/issues/6001).

chriscchien:
Verified **pass** on longhorn master (longhorn-manager `b9e656`)

Perform testing with test [steps](https://github.com/longhorn/longhorn/issues/7357#issuecomment-1859215100) several times, the deployment and statefulset all recovered well and data intact.

PR #7358, opened 2023-12-17
chart: seperate v1 and v2 instanceManagers

derekbit:
Longhorn/longhorn#7357


derekbit:
Read for review cc @innobead

innobead (approved):

Issue #7166, opened 2023-11-22
[TASK] Enable GOCOVERDIR during e2e test cases to generate coverage profiles


FrankYang0529:
Go 1.20 has introduced a coverage-instrumented build at https://go.dev/blog/integration-test-coverage, and it will help us collect coverage profiles at exec runtime instead of only unit/integration testing via `go test`.

We would like to collect coverage profiles from all Longhorn components in Go including user-deployed or system-managed components after running the daily e2e regression.

After https://github.com/longhorn/longhorn/issues/5681 is done, we would like to make the report display next to each Jenkins build.

cc @yangchiu

FrankYang0529:
Close the issue, because we already have e2e coverage in v1.7.x and master branchs.

https://ci.longhorn.io/job/public/job/v1.7.x/job/v1.7.x-longhorn-tests-sles-amd64/
https://ci.longhorn.io/job/public/job/master/job/sles/job/amd64/job/longhorn-tests-sles-amd64/

PR #7367, opened 2023-12-18
feat: add GOCOVERDIR to deploy/longhorn.yaml

FrankYang0529:
The script in [longhorn-tests](https://github.com/longhorn/longhorn-tests/blob/6b789f9d0281aee39deefe2e6572ab7fe39105df/pipelines/utilities/longhorn_manifest.sh#L20) uses longhorn.yaml directly, so we need to set `enableGoCoverDir` when we generate `longhorn.yaml`.

https://github.com/longhorn/longhorn/issues/7166

### DIFF
--- a/deploy/longhorn.yaml
+++ b/deploy/longhorn.yaml
@@ -4180,6 +4180,8 @@ spec:
           mountPropagation: Bidirectional
         - name: longhorn-grpc-tls
           mountPath: /tls-files/
+        - name: go-cover-dir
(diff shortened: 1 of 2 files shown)


FrankYang0529:
Install LH from `longhorn.yaml` is mentioned in official document. It's not good to add test only variable. Close the PR.

https://longhorn.io/docs/1.5.3/deploy/install/install-with-kubectl/

Issue #7361, opened 2023-12-18
[IMPROVEMENT] Add guaranteed instanceManager CPU setting for v2 volume


derekbit:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

We've separate v1 and v2 instance managers. The minimal required CPU is 1+ core for an instance manager pod for v2 volumes and is different than one for v1 volumes. Hence, a new setting for guaranteed instance manager CPU setting for v2 volumes is necessary.

roger-ryao:
**Verified on master-head 20240102**

- longhorn master-head https://github.com/longhorn/longhorn/commit/38fa709ec8476215cf938e9768b4ae67fac9e383
- longhorn-manager master-head https://github.com/longhorn/longhorn-manager/commit/b63565dedcbf6bfc9d94fb2d6eb911c970a893fc


**Result Passed**

- [x] require/doc : https://github.com/longhorn/website/pull/825
- [x] The UI on the master branch includes the options for `Guaranteed Instance Manager CPU for V2 Data Engine` and `Guaranteed Instance Manager CPU for V1 Data Engine`.
- [x]  After configuring `Guaranteed Instance Manager CPU for V2 Data Engine`, the old instance manager will be deleted, and the corresponding new instance manager will be launched.

PR #7368, opened 2023-12-18
chart: introduce guaranteed instance manager CPU setting for v2 volume

derekbit:
https://github.com/longhorn/longhorn/issues/7361


innobead:
@derekbit Do we still need this PR? It seems duplicated of https://github.com/longhorn/longhorn/pull/7403.

derekbit:
> @derekbit Do we still need this PR? It seems duplicated of #7403.

Yes, we can just keep https://github.com/longhorn/longhorn/pull/7403 if both PRs are merged.

innobead:
> > @derekbit Do we still need this PR? It seems duplicated of #7403.
> 
> Yes, we can just keep #7403 if both PRs are merged.

Do you mean we don't need this PR if we merge #7403?

derekbit:
> Do you mean we don't need this PR if we merge #7403?

Yes

innobead:
Sounds good. Let's use #7403 instead.

Issue #7095, opened 2023-11-14
[IMPROVEMENT] Add global setting for enable v1 or v2 volume support


derekbit:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

Add global setting for enable v1 or v2 volume support.

yangchiu:
Verified passed on master-head (longhorn-manager be08850) following [the test steps](https://github.com/longhorn/longhorn/issues/7095#issuecomment-1862830103).

derekbit:
I think we can backport this improvement to v1.5.4. WDYT? @innobead

PR #7470, opened 2023-12-27
chart: rename backendStoreDriver to dataEngine

derekbit:
Longhorn/longhorn#7095

### DIFF
--- a/chart/templates/crds.yaml
+++ b/chart/templates/crds.yaml
@@ -1277,6 +1277,10 @@ spec:
     subresources:
       status: {}
   - additionalPrinterColumns:
+    - description: The data engine of the engine
+      jsonPath: .spec.dataEngine
+      name: Data Engine
+      type: string
     - description: The current state of the engine
       jsonPath: .status.currentState
       name: State
@@ -1315,11 +1319,14 @@ spec:
               active:
                 type: boolean
               backendStoreDriver:
+                description: 'Deprecated: Replaced by field `dataEngine`.'
+                type: string
+              backupVolume:
+                type: string
+              dataEngine:
                 enum:
                 - v1
                 - v2
-                type: string
-              backupVolume:
                 type: string
               desireState:
                 type: string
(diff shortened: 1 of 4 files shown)


derekbit:
longhorn-manager part is merged. The chart update is ready to merge as well.

PhanLe1010 (dismissed):
in general, LGTM

examples/v2/storageclass.yaml
```diff
@@ -11,7 +11,7 @@ parameters:
   numberOfReplicas: "3"
   staleReplicaTimeout: "2880"
   fsType: "ext4"
-  backendStoreDriver: "v2"
```
PhanLe1010, line 14: For user who still have the old SC with `backendStoreDriver` field, will Longhorn webhook/CSI plugin automatically translate it to `dataEngine` ?

chart/templates/crds.yaml
```diff
@@ -1315,12 +1319,15 @@ spec:
               active:
                 type: boolean
               backendStoreDriver:
+                description: Deprecated.
```
c3y1huang, line 1322:
```suggestion
                description: 'Deprecated: Replaced by field `dataEngine`.'
```

examples/v2/storageclass.yaml
derekbit: No, users need to use `dataEngine`. I will mention it in doc.
Do you think we need to help the translation?

PhanLe1010: I mean for users who already have the old SC with backendStoreDriver, they might not be aware of the new change. I am ok with either:
1. A release note
2. Auto translation

innobead (dismissed):

chart/templates/crds.yaml
derekbit: Updated. Thank you.

c3y1huang (approved):

innobead (approved):

examples/v2/storageclass.yaml
derekbit: `Auto translation` is good idea for existing storageclass, but users still need to re-created their stgorageclass using `backendStoreDriver` after we remove it in v1.7.0. @innobead Do you think we should help the translation, or just let user recreate directly?

innobead: As you mentioned, it's a preview (alpha) feature, so things could be changed. Let's have a note in the [import notes](https://longhorn.io/docs/1.6.0/deploy/important-notes/), good enough for now.

Issue #7423, opened 2023-12-22
[IMPROVEMENT] Improve handling of 16TiB+ volumes with ext4 as the underlying file system


ejweber:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

`ext4` supports a [maximum file size of 16TiB](https://en.wikipedia.org/wiki/Ext4#Features). So Longhorn volumes greater than or equal to 16TiB cannot be created on an `ext4` disk and Longhorn volumes less than 16TiB cannot be expanded to greater than or equal to 16TiB on an `ext4` disk.

The Longhorn UI and control plane allows the creation of 16TiB volumes. These volumes fail during initial attachment due to a SIGSEGV for the replica process. The failure is confusing to users, as it is not clear why the volume fails to attach.

The Longhorn UI and control plane allows the expansion of existing volumes to 16TiB. Expansion of these volumes fails due to a SIGSEGV for the replica process. The SIGSEGV does not allow a rollback, so the replica is left broken. In https://github.com/longhorn/longhorn/issues/7174, the user lost data as a result.

## Describe the solution you'd like

Maybe we don't need all of these, but combined they seem like a complete solution. Perhaps we can backport the documentation and the SIGSEV prevention while doing the nodes.longhorn.io CR work only in the latest version?

- [ ] Clearly document that the maximum size of a Longhorn volume when using an `ext4` underlying file system is 16383 GiB (or less than 16 TiB).
- [ ] Figure out where exactly the replica process SIGSEGV occurs and prevent it. For example, (though I'm sure there's a better way), execute `truncate -s <new_size>` before doing any other operations and refuse to proceed if that command fails. This prevents corruption due to a failed rollback.
- [ ] Learn and record the file system type of each disk in its corresponding `nodes.longhorn.io`. Do not schedule replicas for 16TiB+ volumes to `ext4` disks. Refuse to expand volumes to 16TiB+ if they use `ext4` disks before the expansion request reaches longhorn-engine in the first place. (The user can evict their `ext4` replicas if they want to do the expansion.)

## Additional context

- https://github.com/longhorn/longhorn/issues/7174
- https://github.com/longhorn/longhorn/issues/4768
- https://github.com/longhorn/longhorn/issues/2509

ejweber:
Instance manager logs from a 16TiB volume that fails on its initial attach:

```
[...]

PR #7474, opened 2023-12-28
fix(volume): add a field "FSType" in Node.Status.DiskStatus

mantissahz:
Ref: longhorn/longhorn#7423, https://github.com/longhorn/longhorn-manager/pull/2414

### DIFF
--- a/deploy/longhorn.yaml
+++ b/deploy/longhorn.yaml
@@ -5,6 +5,20 @@ kind: Namespace
 metadata:
   name: longhorn-system
 ---
+# Source: longhorn/templates/priorityclass.yaml
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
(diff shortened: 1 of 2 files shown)


deploy/longhorn.yaml
```diff
@@ -5,6 +5,20 @@ kind: Namespace
 metadata:
   name: longhorn-system
 ---
+# Source: longhorn/templates/priorityclass.yaml
```
innobead, line 8: forgot to generate? need to be careful next time.

innobead: wait, it was just rearranged. It's a weird behavior from helm.

innobead (approved):

deploy/longhorn.yaml
mantissahz: It seems that the generating script moves this part up from the bottom.

PR #7510, opened 2024-01-02
doc(backupbackingimage): add example yaml of backingimage backup and restore

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/4165

### DIFF
--- a/enhancements/20230807-backingimage-backup-support.md
+++ b/enhancements/20230807-backingimage-backup-support.md
@@ -69,9 +69,9 @@ This improve the user experience and reduce the operation overhead.
 1. Add a new CRD `backupbackingimage.longhorn.io`
     ```go
     type BackupBackingImageSpec struct {
-        Labels           map[string]string `json:"labels"`
-        BackingImageName string            `json:"backingImageName"`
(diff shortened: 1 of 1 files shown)


enhancements/20230807-backingimage-backup-support.md
```diff
@@ -69,9 +69,9 @@ This improve the user experience and reduce the operation overhead.
 1. Add a new CRD `backupbackingimage.longhorn.io`
     ```go
     type BackupBackingImageSpec struct {
-        Labels           map[string]string `json:"labels"`
-        BackingImageName string            `json:"backingImageName"`
```
innobead, line 73: `BackingImageName` is not required? how to know which backing image is being backed up?

innobead (changes requested):

enhancements/20230807-backingimage-backup-support.md
ChanYiLin: BackingImage is global unique
so its backup should also be global unique
I simply make them the same name.

innobead (approved):

Issue #4105, opened 2022-06-12
Auto-balancing volumes between nodes & disks


iosifnicolae2:
## Is your feature request related to a problem? 
- The actual size of volumes is very high and **disks are getting low on available disk space**.
- the performance is affected because instead of distributing the volumes across multiple disks, **one disk is used to store all the volumes**

## Describe the solution you'd like
It would be awesome if Longhorn would be able to:
1. **allocate** new volumes based on available disk space and/or based on actual size usage .
2. **evict** a volume from a disk

## Describe alternatives you've considered
1. Set `best-effort` as data locality and then constrain the deployment on different nodes (this solution does not fix the problem entirely because we can't configure which disk to be used).
2. To evict a volume, increase the replication factor to 2, wait a few hours, delete the old replica, set replication factor to 1

derekbit:
Hello @iosifnicolae2,
Could you elaborate `evict a volume from a disk` more?
What's the purpose of the feature?

iosifnicolae2:
> Hello @iosifnicolae2,
> Could you elaborate `evict a volume from a disk` more?
> What's the purpose of the feature?

hi, sure

by evicting a volume from a disk I mean to actually move it to move it to another disk because the current disk is almost full and we have other disks which have plenty of space

derekbit:
> > Hello @iosifnicolae2,
> > Could you elaborate `evict a volume from a disk` more?
> > What's the purpose of the feature?
> 
> hi, sure
> 
> by evicting a volume from a disk I mean to actually move it to move it to another disk because the current disk is almost full and we have other disks which have plenty of space

Got it. More precisely, you hope the replica can automatically move to another disk which has more free space.

joshimoo:
Epic for tracking replica scheduling improvements #3840 
I think we will be adding disk anti affinity rules as part of the volume group feature #2128 
The replica eviction request sounds reasonable, since we currently already allow for evicting replica from a node.

innobead:
cc @c3y1huang

yangchiu:
When testing by running test case `test_replica_auto_balance_disk_in_pressure` on `ubuntu 24.04` + `v1.30.0+k3s1`, I can find 2 volumes tried to move its replica from `test-disk1` to `test-disk2`. 

For the first one, the replica can be rebuilt on `test-disk2` and the old replica can be removed automatically without problem. 

But for the second one, after the new replica rebuilt on `test-disk2`, one of nodes became `Down`:
[...]

PR #7576, opened 2024-01-08
feat(lep): auto-balance pressured disks

c3y1huang:

Issue #4105

Propose to enhance replica auto-balance (best-effort) with `replica-auto-balance-disk-pressure-percentage` setting to allow automatically rebuild replica on another disk when the defined threshold is reached.

`None`

`None`

### DIFF
--- a/enhancements/20240108-replica-auto-balance-pressured-disks.md
+++ b/enhancements/20240108-replica-auto-balance-pressured-disks.md
@@ -1,0 +1,191 @@
+# Replica Auto-Balance Pressured Disks On Same Node
+
+## Summary
+
+This proposal enhances Longhorn's replica auto-balancing feature to address disk space pressure caused by growing volumes. It introduces a new setting, `replica-auto-balance-disk-pressure-percentage`, allowing users to define a percentage threshold for automatic action.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/4105
+
+## Motivation
+
+### Goals
+
+- **Reduced Manual Intervention:** Automatic rebalances replicas during disk pressure, minimizing manual effort.
+- **Improved Performance:** Potentially faster replica rebuilds within the same node using local file copy.
+
+### Non-goals [optional]
+
+- This proposal does not address auto-balancing replicas across different nodes. Longhorn prioritizes balancing across nodes and zones before disks within a node.
+- Performance evaluation of local file copy is outside the scope of this proposal.
+
+## Proposal
+
+### User Stories
+
+#### Story 1: Reduced Manual Intervention
+
+- **As a Longhorn user:** I want Longhorn to automatically handle replica movement based on disk pressure, minimizing manual intervention.
+- **Before:** I need to regularly check disk space and manually trigger replica movement to address pressure situations.
+- **After:** Longhorn automatically rebalances replica when disk pressure threshold percentage is reached, freeing me up for other tasks.
+
+#### Story 2: Improved Replica Rebuild Performance
+
+- **As a Longhorn user:** I want Longhorn to leverage local file copy for replica rebuilding within the same node, potentially improving rebuild speed.
+- **Before:** Replica rebuild relies on TCP transfer, which can be slower than local file copy.
+- **After:** When rebuilding replicas on the same node, Longhorn uses local file copy for data transfer.
+
+### User Experience In Detail
+
+1. **Setup:**
+	- You have a Longhorn cluster with multiple nodes, each containing multiple schedulable disks.
+	- Configure via Longhorn UI or Kubernetes CRD.
+		- Set `replica-auto-balance` to `best-effort`.
+		- Define the `replica-auto-balance-disk-pressure-percentage` (e.g., 90% used space triggers disk rebalancing).
+		- Set `replica-soft-anti-affinity` to `enabled` to allow replicas on the same node.
+1. **Deploy Workload:** Deploy workload utilizing Longhorn volumes.
+1. **Replica Disk Reaches Pressured Threshold:**
+	- Longhorn automatically find another disk on the same node with more available disk space.
+	- Longhorn rebuild replica on the same node using the local file copy for data transfer.
+
+
+## Design
+
+### New Setting: `replica-auto-balance-disk-pressure-percentage`
+
+This setting allows defining the disk pressure threshold (percentage of used space) that triggers automatic rebalancing. Its only functional when:
+- `replica-auto-balance` is set to `best-effort`.
+- `replica-soft-anti-affinity` is set to `enabled`.
+```go
+SettingDefinitionReplicaAutoBalanceDiskPressurePercentage = SettingDefinition{
+	DisplayName: "Replica Auto Balance Disk Pressure Threshold (%)",
+	Description: "Sets the threshold percentage of disk space utilization that triggers replica auto-balance.\n\n" +
+		"When the threshold percentage is reached, Longhorn automatically rebuilds replicas under disk pressure onto another disk within the same node.\n\n" +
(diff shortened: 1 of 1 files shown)


ejweber (commented):
IIUC correctly, there is no mechanism to select which replica on a disk would be best (whatever that means) to autobalance. Instead, it is essentially random. Whichever volume happens to reconcile first when a disk is under pressure is balanced to a different disk.

Do you think there is a significant advantage to choosing one replica versus another? If so, how difficult do you think it would be to add that kind of choice to the implementation? (My initial thought is that it would add significant complexity, since right now all replica scheduling decisions are made with only the replica to be scheduled and the aggregate available space on nodes in disks in mind.)

derekbit:
Some questions before diving into the codes:
1. There is no lock mechanism in replica scheduler. Is it possible that all volumes are triggered to be rebuilt on the same disk while reaching to the user-defined threshold?

2. `Set `replica-soft-anti-affinity` to `enabled`` is a requirement for enabling the feature. Will it introduce a side effect that multiple replicas of a volume are created on the same node?

3. Snapshot files are sparse. Will the copy method handle holes?

enhancements/20240108-replica-auto-balance-pressured-disks.md
```diff
@@ -0,0 +1,146 @@
...
+
+Before: User manually balance replicas when disk space pressures arise due to volume growth.
+
+After: User can set a disk pressure threshold, enabling Longhorn to automatically rebuild replicas to another disk with more available storage on the same node.
+
+### User Experience In Detail
+
+1. Cluster nodes have multiple schedulable disks.
+1. Set `replica-auto-balance` to `best-effort`.
+1. Set `replica-soft-anti-affinity` to `enabled`.
```
innobead, line 40: We only have node, zone and disk anti affinity. What's replica affinity here?

enhancements/20240108-replica-auto-balance-pressured-disks.md
```diff
@@ -0,0 +1,146 @@
...
+	  int32 file_sync_http_client_timeout = 5;
+	  FileLocalSync local_sync = 6;
+	}
+
+	message FileLocalSync{
+		string source = 1;
+		string target = 2;
+	}
+	```
+1. Then the sync-agent server is responsible for copying the files using `req.SyncFileInfoList` from source to the target directory when `req.FilesSyncRequest.LocalSync` is provided.
```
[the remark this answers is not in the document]
innobead: Can we leverage https://github.com/longhorn/sparse-tools/blob/master/sparse/file.go or even improve it to support spares file copy?

innobead (commented):
In general, LGTM. However, need to resolve the questions from Eric and Derek.

enhancements/20240108-replica-auto-balance-pressured-disks.md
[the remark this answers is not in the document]
c3y1huang: > How exactly are the files copied? I have experienced some strange behavior when experimenting with copying replica sparse files before, even when using options meant to preserve sparseness. The checksums always matched (and IIRC matched even if sparseness was not preserved), but it the actual space usage was different by a little bit between the source and target file. I think the exact behavior varied between copy utilities (e.g. rsync vs cp, etc.).

The files will be copied using the [CopyFile](https://github.com/longhorn/go-common-libs/blob/2133a7e73771ecdd26494a9c1ed2b31495ffd4f2/io/file.go#L102) in go-common-lib. This function should support sparse files. I have verified on my cluster and written a unit test to ensure that the space usage remains the same after the copy.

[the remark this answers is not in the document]
c3y1huang: > Can we leverage https://github.com/longhorn/sparse-tools/blob/master/sparse/file.go or even improve it to support spares file copy?

Sure, though I don't see a strong reason for it now since it's just a simple copy. However, we could evaluate the need for this enhancement later. What do you think?

enhancements/20240108-replica-auto-balance-pressured-disks.md
c3y1huang: Enabling replica-soft-anti-affinity allows Longhorn to build replicas on the same node.

https://longhorn.io/docs/1.6.2/references/settings/#replica-node-level-soft-anti-affinity

c3y1huang:
> * There is no lock mechanism in replica scheduler. Is it possible that all volumes are triggered to be rebuilt on the same disk while reaching to the user-defined threshold?

Great catch, I've updated the LEP to handle this case.

> * `Set replica-soft-anti-affinity to enabled` is a requirement for enabling the feature. Will it introduce a side effect that multiple replicas of a volume are created on the same node?

Not sure about the potential side effects. Is this related to the previous question? Could you elaborate on the scenario?

> * Snapshot files are sparse. Will the copy method handle holes?

https://github.com/longhorn/longhorn/pull/7576#discussion_r1648694213

enhancements/20240108-replica-auto-balance-pressured-disks.md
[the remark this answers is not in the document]
innobead: Sounds good. Just need to make sure the spare files are correctly copied. 

cc @shuo-wu @derekbit

derekbit:
> > * There is no lock mechanism in replica scheduler. Is it possible that all volumes are triggered to be rebuilt on the same disk while reaching to the user-defined threshold?
> 
> Great catch, I've updated the LEP to handle this case.
> 
> > * `Set replica-soft-anti-affinity to enabled` is a requirement for enabling the feature. Will it introduce a side effect that multiple replicas of a volume are created on the same node?
> 

IIRC, enabling the `replica-soft-anti-affinity` setting can result in replicas of a volume being scheduled on the same node. However, the auto-balance feature requires `replica-soft-anti-affinity` to be enabled. If a user wants to utilize the auto-balance feature but wants to avoid the possibility of replicas being placed on the same node, is this feasible?

derekbit:
> > IIRC, enabling the `replica-soft-anti-affinity` setting can result in replicas of a volume being scheduled on the same node. However, the auto-balance feature requires `replica-soft-anti-affinity` to be enabled. If a user wants to utilize the auto-balance feature but wants to avoid the possibility of replicas being placed on the same node, is this feasible?
> 
> I see, I am fine with removing the dependency on this setting.

Sound good to me.
cc @shuo-wu @innobead

enhancements/20240108-replica-auto-balance-pressured-disks.md
```diff
@@ -0,0 +1,191 @@
...
+
+### User Experience In Detail
+
+1. **Setup:**
+	- You have a Longhorn cluster with multiple nodes, each containing multiple schedulable disks.
+	- Configure via Longhorn UI or Kubernetes CRD.
+		- Set `replica-auto-blance` to `best-effort`.
+		- Define the `replica-auto-balance-disk-pressure-percentage` (e.g., 90% used space triggers disk rebalancing).
+
+	> **Note:** This feature will bypass the `replica-soft-anti-affinity` setting.
```
derekbit, line 47: Does it mean one volume's replicas are no longer unable to locate on the same node if `replica-soft-anti-affinity` is enabled?

c3y1huang: This means this feature is not affected by the `replica-soft-anti-affinity` setting.  I have updated the statement to clarify this. Thank you.

derekbit: Sounds good!

derekbit:
If @shuo-wu and @innobead has no other concerns, we can merge it.

innobead (approved):
In general, LGTM.

For reslience, need to make sure to clean up the temporary failed replica clone if the local sync is failed.

Issue #7606, opened 2024-01-10
[IMPROVEMENT] Increase the hugepage size for spdk_tgt to 2GiB


derekbit:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

To address some feedback from uses and increase better UX, increase the hugepage size for spdk_tgt to 2GiB.

chriscchien:
Verified **pass** on longhorn master(longhorn `d918ad`)

Setup v2 engine environment by [instructions](https://longhorn.io/docs/1.6.0/spdk/quick-start/) and then install longhorn by both kubectl and helm, after longhorn ready, all nodes have `hugepages-2Mi: 2Gi` value in `status.capacity` and `status.allocatable`

```
status:
  addresses:
  - address: 172.31.47.74
    type: InternalIP
  - address: ip-172-31-47-74
    type: Hostname
  allocatable:
    cpu: "4"
    ephemeral-storage: "98672037811"
    hugepages-1Gi: "0"
    hugepages-2Mi: 2Gi
    memory: 14268920Ki
    pods: "110"
  capacity:
    cpu: "4"
    ephemeral-storage: 101430960Ki
    hugepages-1Gi: "0"
    hugepages-2Mi: 2Gi
    memory: 16366072Ki
    pods: "110"
```
```
> kubectl get node -o yaml | grep huge
      hugepages-1Gi: "0"
      hugepages-2Mi: 2Gi
      hugepages-1Gi: "0"
      hugepages-2Mi: 2Gi
      hugepages-1Gi: "0"
      hugepages-2Mi: 2Gi
      hugepages-1Gi: "0"
      hugepages-2Mi: 2Gi
      hugepages-1Gi: "0"
      hugepages-2Mi: 2Gi
      hugepages-1Gi: "0"
      hugepages-2Mi: 2Gi
```

PR #7607, opened 2024-01-10
Incease default hugepage size for v2 data engine to 2GiB

derekbit:

Issue https://github.com/longhorn/longhorn/issues/7606

### DIFF
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -664,6 +664,12 @@ questions:
   group: "Longhorn V2 Data Engine (Preview Feature) Settings"
   type: boolean
   default: false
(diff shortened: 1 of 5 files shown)


innobead:
@derekbit do we need to change chart values.yaml or related files?

derekbit:
> @derekbit do we need to change chart values.yaml or related files?

Yes, I add v2-data-engine-hugepage-limit since v1.6.0.
Didn't put in the setting list on purpose before.

derekbit:
PR is updated according to the comments from @jillian-maroket and ready to review. cc @innobead

innobead (approved):

Issue #2598, opened 2021-05-18
[DOC] Reference Architecture and Sizing Guidelines


wjimenez5271:
As a user of Longhorn, I need to know the ideal implementation of Longhorn to ensure the best experience. For example, I want to know if I am risking stability by using a node or network type that is undersized for Longhorn. 

Also, need to consider https://github.com/longhorn/longhorn/issues/2246#issue-804376552. 
There may be some unexpected issues when Longhorn creates snapshots/backups on high spec nodes:
#1675
#2029

https://github.com/longhorn/longhorn/issues/2598#issuecomment-1666884745

yasker:
For the volume size limit, let's targeting for supporting up to 2TB single volume.

PhanLe1010:
Related to the slack message https://cloud-native.slack.com/archives/CNVPEL9U3/p1622728334055500

PhanLe1010:
Action items (checked item means we already know):
- [ ] Recommended volume sizes (min, max, best) 
- [ ] Recommended min CPU
- [ ] Recommended min RAM
- [ ] Recommended min Network bandwidth, min latency
- [ ] Recommended Disk min letancy, min IPOS 
- [ ] Recommended OS, Filesystem on Node
- [ ] Recommended backup target (request latency)
- [ ] Recommended cluster setup (Where in the cluster Longhorn should run)

@yasker @joshimoo @shuo-wu @wjimenez5271 Am I missing anything?

yasker:
resource requirement for volume number/total size in cluster or node.

yasker:
@joshimoo What's the biggest volume we've tested right now? 4TB?

innobead:
@PhanLe1010 when working on that, please also move to analysis or implementation pipeline. Thanks!

joshimoo:
@yasker I have never tested that big of a size, I think @meldafrawi / @khushboo-rancher are working on some size tests for backups?

khushboo-rancher:
I've tested with max 1T data while testing https://github.com/longhorn/longhorn/issues/1948. Is the data from https://github.com/longhorn/longhorn/issues/1948#issuecomment-865362997 helpful here?

yasker:
Discussed with @innobead , minimal resource investigation should be also done as a part of sizing guideline, so we will know what's our baseline is.

innobead:
We can wrap up this after 1.2.0 out.

innobead:
Need to wrap up after 1.2.1.

[...]

PR #7616, opened 2024-01-10
[LEP] Reference Setup, Performance, Scalability, and Sizing Guidelines

PhanLe1010:
longhorn/longhorn#2598

### DIFF
--- a/enhancements/20240109-reference-architecture-and-sizing-guidelines.md
+++ b/enhancements/20240109-reference-architecture-and-sizing-guidelines.md
@@ -1,0 +1,142 @@
+# Reference Architecture and Sizing Guidelines
+
+## Summary
+
+As a Longhorn user, I need to know the reference architecture and sizing to ensure the best experience.
+From these references, I may have a picture of how Longhorn would perform in a similar cluster spec.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/2598
+
+## Motivation
+
+### Goals
+
+#### Goal 1
+For each supported environment, pick a cluster spec (small, medium, large) to run Longhorn and measure Longhorn performance/scalability with that cluster spec.
+
+The supported environments are:
+1. On-Prem (using Equinix Metal for testing)
+1. Public Cloud (using AWS for testing)
+1. Edge (Not sure where to test)
+
+The cluster spec including the factors:
+1. Node's CPU
+1. Node's RAM
+1. Node's Disk IOOPs/bandwidth/latency
+1. Internode networking bandwidth/latency
+1. OS distro 
+1. Kernel
+1. The number of nodes
+1. Kubernetes version and distro 
+1. Number of controller plane nodes
+1. Number of ETCD nodes
+1. Longhorn specific setting like:
+   1. Priority class: stick to a high value as best practice
+   1. Replica balancing: stick to enable best practice
+1. There is volume rebuilding happens inside the cluster or not (we can trigger this by restarting certain number of nodes at a time)
+
+
+
+The Longhorn performance/scalability including the factors:
+1. Max/min/recommended volume size
+1. Max/recommended number of Longhorn volumes inside the cluster 
+1. Max/recommended number of Longhorn volumes attached to a node
+1. Max/recommended number of Longhorn replicas on to a node
+1. Longhorn volume IOOPs/bandwidth/latency
+
+#### Goal 2
+For the recommended cluster spec of each supported environments, figure out:
+1. If we scale up the number of node, what is the max number of nodes and Longhorn volumes that Longhorn can support
+
+#### Goal 3
+Figure out Recommended backup target.
+
+The factors to consider:
+1. Backup target type (NFS/S3/SMB/CIFS/...)
+1. Backup target latency
+
+
+#### Goal 4
+Develop some kind of automation for this test plan so that we can run this again for every new release.
+This will help us to identify if there are improvement/regression 
+
+
+### Non-goals
+
+We are not planning to find the answer for the questions:
+1. The exact formular to calculate minimum resource requirement (CPU/RAM/Network) for a specific amount of volumes 
+
+## Proposal
+
+This is where we get down to the nitty-gritty of what the proposal actually is.
+
+### User Stories
+Detail the things that people will be able to do if this enhancement is implemented. A good practice is including a comparison of what the user cannot do before the enhancement is implemented, why the user would want an enhancement, and what the user needs to do after, to make it clear why the enhancement is beneficial to the user.
+
+The experience details should be in the `User Experience In Detail` later.
+
+#### Story 1
+#### Story 2
+
+### User Experience In Detail
+
+Detail what the user needs to do to use this enhancement. Include as much detail as possible so that people can understand the "how" of the system. The goal here is to make this feel real for users without getting bogged down.
+
+## Design
+
+### Payload design 
+We are considering 5 types of payload:
+
+1. 100% read, 100% RWO volumes
+1. 100% read, 100% RWX volumes
+1. 100% write, 100% RWO volumes
+1. 100% write, 100% RWX volumes
+1. No read/write and 100% RWO volumes (this will test the scalability of Longhorn control plane)
+
+We are going to use fio to develop these payload since fio is a popular tool for testing storage. 
+We are not going to select a particular application such as a SQL DB, Kafka, Prometheus for the payload as 
+this might be too specific and many users might not be interested in the particular application we chose. 
+
+More detailed about how to implement this payload:
+* We can use our favorite [kbench](https://github.com/yasker/kbench) benchmarking tool with some modification:
+  * Right now, it is having 2 modes: `quick` mode and `normal` mode. We can add 2 more modes: `readonly` and `writeonly` modes.
+  * Then when deploying the workload pod, the yaml can select either `readonly` and `writeonly` modes according to the test's payload type
+* Then test workload pod then run the kbench benchmarking tool continuously and record the result into a file after each run
+* The workload pod can include a small webserver to export the test result as Prometheus metrics for data collection
+* On the cluster level, we can deploy Rancher monitoring to collect this data and display it on the grafana graph for realtime monitoring  
+
+### Testing passing criteria
+How to decide if cluster is able to handle the payload? 
+1. Workload pods are not crashing
+1. Volumes, engines, and replicas are not crashing
+1. IOOPs/bandwidth/latency result is not reduced. We need design a way to collect testing result data from the workload pods: 
+   1. Maybe the workload pod exports the benchmark data as Prometheus metrics then we can collect them using Rancher monitoring
+   1. Maybe another way is each pod calculate the result and output it at the end of pod's life. Then we can read and parse them using `kubectl logs`
+
+### Test steps:
+
+
+1. Pick one of the fixed cluster spec above
+1. Pick one of the payload type above
+1. Applied the payload with minimal scale (i.e., 1 pods). Make sure that pods are evenly distributed across the nodes
+1. Scale up number of workload pods until the `Testing passing criteria` above break.
+1. Publish the result:
+   1. Max number of Longhorn volumes inside the cluster
+   1. Max number of Longhorn volumes attached to a node
+   1. Max number of Longhorn replicas on to a node
+   1. Longhorn volume IOOPs/bandwidth/latency
+   1. CPU/memory used by Longhorn processes
+1. Try to scale the cluster's number of nodes and number of workload pods proportionally. 
+At which point the workload become very slow to starts, crash, IO slow down significantly 
+1. Publish the result:
+   1. The number of cluster nodes Longhorn can support with this cluster spec
+1. Scale the number of control plan node back to 1, worker nodes back to 3, workload back to 1
+1. Increase the size of the volume, trigger rebuilding, perform regular volume snapshot backup operation
+1. Publish the result:
+   1. At which size, the volume rebuilding and other basic operation are broken?
+
+## Note [optional]
+
+Additional notes.


innobead:
For the members I requested, please review this PR.

PhanLe1010:
Could we help review the defined goals/objectives in this LEP first?

The detailed testing plan will be developed after we agree on the defined goals/objectives

enhancements/20240109-reference-architecture-and-sizing-guidelines.md
```diff
@@ -0,0 +1,101 @@
...
+1. OS distro 
+1. Kernel
+1. The number of nodes
+1. Kubernetes version and distro 
+1. Number of controller plane nodes
+1. Number of ETCD nodes
+1. Longhorn specific setting like:
+   1. CPU request
+   1. Priority class
+   2. Replica balancing
```
ejweber, line 38: I recommend factoring some of these things out of these guidelines. For example, OS distro, kernel, and Kubernetes distro belong in best practices or in a supportability matrix. We cannot afford to vary the other factors and ALSO vary these. This will be expensive and result in too much data for a customer to interpret successfully.

I also think we can remove factors like CPU request, priority class, and replica balancing as factors.

- Our testing should help us understand how much CPU Longhorn uses in a variety of scenarios. Instead of varying the CPU request, we can just publish a best practice that CPU request should be (for example) 110% of this value.
- Our best practice should always be a priority class with high priority. I think we should simply not support any decision to allow Longhorn pods to be OOM killed before others. If the customer chooses this path and logs indicate OOM killing of Longhorn pods, we should recommend setting the priority class before assisting further.
- I think all of our testing should assume replicas are balanced. There really isn't any other logical way to test (and trying to vary how balanced replicas are creates too many factors to consider). Of course, this makes our results harder to apply to the real world (in which replicas probably aren't exactly evenly balanced, at least in terms of I/O). We can either add some kind of safety factor to our results or simply publicly state our methodology.

ejweber (commented):
This is a very difficult problem with many factors to consider! (You have done a good job so far of capturing many of those factors.) It is tempting to assume we can give simple answers to questions like, "How many volumes can we support?". In reality, this is impossible. What are the volumes doing? How many resources can be consumed by Longhorn volumes without starving applications? Etc... We need to be careful to limit the scope to something that is actually achievable and recognize that whatever we publish will have to be interpreted by the customer. In other words, we can provide them with a certain amount of data about workloads we've tested, but they will have to decide how well the workloads they want to run fit the test scenarios.

We are going to discuss this offline later today, but here are some initial thoughts I have after reading what you have so far.

I think application I/O load is probably the biggest factor in determining how well we perform in a particular environment. There may be upper limits to the number of volumes we can support from purely from a "numerical" standpoint (e.g. some maximum number of iSCSI targets on a node or something), but my gut tells me that the cumulative throughput and number of IOPs from workloads is most important. For example, maybe we can handle 2 volumes consumed by very intensive applications but 10 volumes that don't write much. Or maybe we can handle 10 volumes consumed by applications with heavy read throughput requirements, but only 2 that have do many small writes. We can probably only benchmark in terms of these numbers. For example, we can't tell a customer that a particular environment can support 20 volumes. Maybe we can produce a chart that says the environment supports (for example):

- 20 volumes with 50 MB/s of sustained write throughput each.
- 25 volumes with 40,000 write IOPs each.
- Etc...

Or maybe its more reasonable to produce a chart that says (for example):

- When we run maximum write bandwidth to 20 volumes simultaneously, this is how much write bandwidth we can achieve per workload.
- When we run maximum write bandwidth to 25 volumes simultaneously, this is how much write bandwidth we can achieve per workload.
- Etc...

Realistically, the customer must be able to evaluate their own needs and decide which I/O characteristics are most similar to their needs.

We need to think about what we mean when we say a certain number of volumes are supported. What threshold breaks that support?

- Obviously, if engines are crashing, a certain number of Longhorn volumes are no longer supported.
- What about I/O latency? As we increase the workload size, I/Os will slow down significantly. If a customer has latency objectives, that will come into play.

Longhorn runs on the same node as application workloads. This makes it different than many other storage solutions, as it consumes a portion of the application resources (memory, CPU, bandwidth). So we have to consider the actual application together with Longhorn when deciding whether a number of volumes can run well. In other words, Longhorn cannot be allowed to use the whole node.

I think we should de-emphasize metrics like "number of Longhorn volumes inside the cluster", as they are mostly theoretical. Maybe we can run some tests to determine how many volumes Longhorn can keep track of in its control plane, but in the real world, these volumes have actual workloads running against them, so we should be focused on volumes with particular performance characteristics per node.

It will be very important to consider rebuilding as we benchmark. A cluster making maximum use of its resources while stable will probably start to fail if it is both serving application I/O AND rebuilding replicas. Maybe this is something we can consider in the "second" version of these guidelines, but it should be on our roadmap. (For example, this is how many volumes we can support if we are actively rebuilding a node's worth of replicas at the same time.)

I know we are not considering actual testing steps yet, but I think it is helpful to consider how the testing might go in general as we define the goals. Maybe we could do something like the following:

1. Define a few standard workload pods. I'm not super familiar with benchmarking tools, but if they can be configured to deliver targeted I/O workloads (e.g. 50 MB/s writes or 40,000 IOPs) instead of just, "write as fast as you can", this could be helpful. Then a workload pod could be "this benchmark tool" configured in "this way' operating against a single Longhorn volume.
2. Define a few cluster specs. For example, small nodes with a small interconnect. Large nodes with a large interconnect. I think we cannot possibly test every combination.
3. Run the workload pods in a "minimum" cluster according to the spec. For example, a cluster with three worker nodes. Ensure the workload pods are evenly distributed.
4. Try to increase the number of workload pods until... something. Engines crash? I/O latency gets too high?
5. Repeat the test for each type of workload pod. Publish separate results for each type of workload pod.
6. Repeat these tests for each defined cluster spec.
7. Publish key values. For example: I/O latency, CPU used by Longhorn processes, memory used by Longhorn processes, bandwidth used by Longhorn processes). Hopefully, this can give an idea of the Longhorn "overhead". The benchmarking applications won't much CPU and memory like actual applications would, so it is important to know how much Longhorn actually consumes.
8. Try to scale the cluster's number of nodes and number of workload pods proportionally. (If we were running 6 pods on 3 nodes and hit some kind of threshold, lets run 8 pods on 4 nodes). Is there a point at which our ability to do this breaks?

Workload balance is a critical factor in all of this, but I think it's something we can't really consider in our tests. For example, maybe Kubernetes schedules all of a cluster's IOPs intensive workloads to one node (or volumes on one node) and not another. I think the best we can probably do is some kind of safety factor (e.g. we recommend running 25% less volumes than our testing supports) or just make a statement that the customer needs to balance their workloads well to achieve these results.

james-munson:
I concur with Eric's points.  It's true that the question of "about how big should I make my installation" is a tough one to answer, because "how big is your workload?" is.  Having a set of representative standard workloads and publishing results for those, allowing the user to estimate how close their own will be is the best approach.  Even with the number of variables for workload I/O, node count, disk type, etc cut to a minimum, there will be plenty of automation and test resources spent to get reasonable coverage.  We just need to be clear about exactly what we've done and publish it in a way that can be reproduced.

enhancements/20240109-reference-architecture-and-sizing-guidelines.md
james-munson: But also including volume type (RWO vs RWX)?

PhanLe1010: Thanks a lot for the valuable input! 

I agree with most of the points you mentioned here about:
1. Payload design should be defined so that it closely represents real-world payload:
    1. Including mix of IO types. We can utilize `fio` for this purpose 
    1. Including 15% of the volume is RWX, 85% of the volume is RWO. This ratio is what we currently observe from our real-world public metrics.
1.  Cluster spec selection: 
    1. Agree that we should always balance the replicas. 
    1. We don't need to adjust the CPU request 
    1.  We should select a fix env including OS, Kernel,.. that we believe is the best practice at the beginning.
    1. We should also select a fixed node spec CPU, RAM, Disk, Network,... 
    1. We will vary the number of nodes in the cluster to answer some scaling questions
1. Define criteria for how to consider if the test is passed
1. Agree with the detailed test plan.

PhanLe1010: @ejweber @james-munson and I had a meeting discussing this. Will updated the meeting note below

PhanLe1010:
Some of the debatable topic come up during the meeting today with @ejweber and @james-munson. Would be appricatate if others can give an opinion:
1. Payload design: Do we need to break it down into multiple types like this https://github.com/longhorn/longhorn/pull/7616/files#diff-7db4584c175029bb1576e92bd0628a2cc72516127bd0b038b6b9d9752b3d2633R90-R98. Or we can have single payload of mix IO and volume modes? 
    1. Breakdown the payload into multiple types gives a better idea of each extreme payload patterns but significantly increase the number if tests we need to perform
    1. Having a single payload of mix IO and volume modes reduces the number of test cases but we cannot choose a perfect mix payload that represents all cases in real-world.

enhancements/20240109-reference-architecture-and-sizing-guidelines.md
```diff
@@ -0,0 +1,101 @@
...
+   1. Priority class
+   2. Replica balancing 
+
+
+The Longhorn performance/scalability including the factors:
+1. Max/min/recommended volume size
+1. Max/recommended number of Longhorn volumes inside the cluster 
+1. Max/recommended number of Longhorn volumes attached to a node
+1. Max/recommended number of Longhorn replicas on to a node
+1. Longhorn volume IO bandwidth/throughput
```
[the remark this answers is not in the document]
PhanLe1010: Agree, we will defined a set of testing passing criteria to determine these value. Could you head to the section [Design](https://github.com/longhorn/longhorn/pull/7616/files#diff-7db4584c175029bb1576e92bd0628a2cc72516127bd0b038b6b9d9752b3d2633R87) to see if make senses?

enhancements/20240109-reference-architecture-and-sizing-guidelines.md
```diff
@@ -0,0 +1,101 @@
...
+#### Goal 2
+For the recommended cluster spec of each supported environments, figure out:
+1. If we scale up the number of node, what is the max number of nodes and Longhorn volumes that Longhorn can support
+
+#### Goal 3
+Figure out Recommended backup target.
+
+The factors to consider:
+1. Backup target type (NFS/S3/SMB/CIFS/...)
+1. Backup target latency
```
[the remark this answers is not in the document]
PhanLe1010: Agree that the current best practice is using s3 since we see less problem with it. However, I think yes, we gonna do some testing with s3 and publish some data like what is is the characteristic of the s3 target that is best for Longhorn

innobead:
cc @Vicente-Cheng @WebberHuang1118 join together to review this.

enhancements/20240109-reference-architecture-and-sizing-guidelines.md
```diff
@@ -0,0 +1,137 @@
...
+Detail what the user needs to do to use this enhancement. Include as much detail as possible so that people can understand the "how" of the system. The goal here is to make this feel real for users without getting bogged down.
+
+## Design
+
+### Payload design 
+We are considering 8 types of payload:
+
+1. 100% read, 100% RWO volumes
+1. 100% read, 100% RWX volumes
+1. 100% read, 15% RWX volumes, 85% RWX volumes
```
ejweber, line 94:
```suggestion
1. 100% read, 15% RWX volumes, 85% RWO volumes
```

I assume this mix is because of the metrics showing that 15% of our user's volumes are RWX? Maybe we should call out that motivation somewhere.

I was going to recommend we drop this one (as well as lines 97-98). this is mainly because I don't think we have any data to suggests that users are deploying clusters with anything like this mixture. I would guess that we see this split because some users are primarily RWX and most users are primarily RWO. (Of course, I don't know, but that's the point.) HOWEVER, we did discuss in our meeting certain sanity check payloads that would demonstrate that performance is between the extremes when payload spec is between the extremes, and these payloads fit that criteria.

enhancements/20240109-reference-architecture-and-sizing-guidelines.md
```diff
@@ -0,0 +1,137 @@
...
+
+We are going to use fio to develop these payload since fio is a popular tool for testing storage. 
+We are not going to select a particular application such as a SQL DB, Kafka, Prometheus for the payload as 
+this might be to specific and many users might not be interested in the application we chose. 
+
+### Testing passing criteria
+How to decide if cluster is able to handle the payload? 
+1. Workload pods are not crashing
+1. Volumes, engines, and replicas are not crashing
+1. IO bandwidth, latency result is not reduced. We need design a way to collect testing result data from the workload pods:
```
ejweber, line 109: I expect our latency graphs will look something like what is shown in https://www.netapp.com/media/85630-tr-4969.pdf, except with the x-axis being number of pods? If so, we will have to decide on a threshold. Clearly, there is a point in the graphs when latency starts increasing nonlinearly with workload. Maybe we can estimate where this point is? Other options include:

- Some multiplicative factor of initial latency (e.g. when latency is 2x that of the minimum workload).
- Some fixed amount (e.g. when latency exceeds x ms).

PhanLe1010: > Some multiplicative factor of initial latency (e.g. when latency is 2x that of the minimum workload)

I think this is a good idea. I will implement this in detailed testing plan

enhancements/20240109-reference-architecture-and-sizing-guidelines.md
PhanLe1010: Sounds good to me! I will drop these mixed payload from the LEP. 
We can do sanity check at the end to verify the theory if needed.

PhanLe1010:
Note: I have renamed the task from `Reference Architecture and Sizing Guidelines` to `Reference Setup, Performance, Scalability, and Sizing Guidelines` because `Reference Architecture` can be misleading that this document is provide detailed design architecture of Longhorn

enhancements/20240109-reference-architecture-and-sizing-guidelines.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Goals
+
+#### Goal 1
+For each supported environment, pick a cluster spec (small, medium, large) to run Longhorn and measure Longhorn performance/scalability with that cluster spec.
+
+The supported environments are:
+1. On-Prem (using Equinix Metal for testing)
+1. Public Cloud (using AWS for testing)
+1. Edge (Not sure where to test)
```
derekbit, line 22: What does `Edge` mean in the environments?

Vicente-Cheng:
That's a nice (tough) topic for every storage. I saw Eric and James' opinions. They already point out the valuable things. I tried to bring up something related to this topic.

It is good to define some reasonable workload pod to test. As you said, RWO volume with read/write and RWX volume with read/write. For the reality scenario, not only fully read or write the scenario. We could also define the workload pod as 70%write + 30%read and 70%read + 30%write for both RWX/RWO volumes. Maybe looks like

workload pods can be defined as:
- 100% write for RWO/RWX volume
- 100% read for RWO/RWX volume
- 70% write and 30% read for RWO/RWX volume
- 70% read and 30% write for RWO/RWX volume
I don't think the RWO+RWX mix is common, so maybe we can focus on only RWO/RWX.

Also, different IO sizes should be tested, like 4k, 64k, 128k, and 1M.
The reason is that 4k is simulated for the DB scenario, 64k/128k is more likely the general scenario, and 1M is simulated for the streaming application.

We could focus on the test result's throughput, IOPS, and latency.

The scalability could be checked in two parts.
- cluster size
We do the same test on the various cluster sizes (as you mentioned on LEP), which will help with analysis.
- workload pod
Various workload pods can give more information that we can provide. For example, one pod can provide 10000 iops. Could we provide 5000 iops with 2 pods?

I thought the customer would like how they can get in a few workload pods or many workloads. That will help them to know how to utilize it. For example, Someone has a workload that needs 10000 iops within 1 ms latency, and another has 20 workloads. Each workload only needs 200 iops with any latency. They can all get what they want from the sizing guide.

I agree that we should test with the volume degraded. However, it will introduce more factors in benchmarking.
We can first do the basic rebuilding test. For example, the default rebuild maximum is 5. We can test the 5 volumes in rebuilding at this moment.

Also, we should test the baseline on the disk for the internal data.
The baseline will help us to know how we utilize the disk.

enhancements/20240109-reference-architecture-and-sizing-guidelines.md
```diff
@@ -0,0 +1,142 @@
...
+1. Kernel
+1. The number of nodes
+1. Kubernetes version and distro 
+1. Number of controller plane nodes
+1. Number of ETCD nodes
+1. Longhorn specific setting like:
+   1. Priority class: stick to a high value as best practice
+   1. Replica balancing: stick to enable best practice
+1. There is volume rebuilding happens inside the cluster or not (we can trigger this by restarting certain number of nodes at a time)
+
```
james-munson, line 39: And number and type of disks.  It is reasonable to assume one drive per worker node, perhaps, but we still need to compare at least NVME and SSD.

james-munson: And we need to have a story about storage network (which could be to say that it has not been used for this testing.)
The note below about possibly measuring network bandwidth used (i.e. saturation) spurred this thought.

enhancements/20240109-reference-architecture-and-sizing-guidelines.md
```diff
@@ -0,0 +1,142 @@
...
+
+#### Goal 4
+Develop some kind of automation for this test plan so that we can run this again for every new release.
+This will help us to identify if there are improvement/regression 
+
+
+### Non-goals
+
+We are not planing to find the answer for the questions:
```
james-munson, line 68: ... planning ...

enhancements/20240109-reference-architecture-and-sizing-guidelines.md
```diff
@@ -0,0 +1,142 @@
...
+1. IO bandwidth, latency result is not reduced. We need design a way to collect testing result data from the workload pods: 
+   1. Maybe the workload pod exports the benchmark data as Prometheus metrics then we can collect them using Rancher monitoring
+   1. Maybe another way is each pod calculate the result and output it at the end of pod's life. Then we can read and parse them using `kubectl logs`
+
+### Test steps:
+
+
+1. Pick one of the fixed cluster spec above
+1. Pick one of the payload type above
+1. Applied the payload with minial scale (i.e., 1 pods). Make sure that pods are evenly distributed across the nodes
```
james-munson, line 123: ... minimal ...

enhancements/20240109-reference-architecture-and-sizing-guidelines.md
PhanLe1010: It was intended for edge env but now thinking about it, this might be the on-prem small spec case? cc @innobead

enhancements/20240109-reference-architecture-and-sizing-guidelines.md
```diff
@@ -0,0 +1,142 @@
...
+1. Scale up number of workload pods until the `Testing passing criteria` above break.
+1. Publish the result:
+   1. Max number of Longhorn volumes inside the cluster
+   1. Max number of Longhorn volumes attached to a node
+   1. Max number of Longhorn replicas on to a node
+   1. Longhorn volume IO bandwidth/throughput/latency
```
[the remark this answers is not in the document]
PhanLe1010: You are right! My bad. I fixed it. Thank you

PhanLe1010:
Thanks @Vicente-Cheng for the great feedback at https://github.com/longhorn/longhorn/pull/7616#issuecomment-1898201828


> It is good to define some reasonable workload pod to test. As you said, RWO volume with read/write and RWX volume with read/write. For the reality scenario, not only fully read or write the scenario. We could also define the workload pod as 70%write + 30%read and 70%read + 30%write for both RWX/RWO volumes. Maybe looks like
> 
> workload pods can be defined as:
> 
> 100% write for RWO/RWX volume
> 100% read for RWO/RWX volume
> 70% write and 30% read for RWO/RWX volume
> 70% read and 30% write for RWO/RWX volume
> I don't think the RWO+RWX mix is common, so maybe we can focus on only RWO/RWX.

After multiple time discussed with @ejweber and @james-munson, we think that we cannot choose a perfect mix payload that represents all cases in real-world so we will perform 100% read and 100% write test cases. Then user can use the 2 extreme test cases to estimate their performance when mixing read/write at a certain ratio. That being said, we will do ONE sanity test with mix of read/write to verify. 

I agree with separate RWO and RWX case. 

So what do you thinking about 5 payload types below (Ref: [link](https://github.com/longhorn/longhorn/pull/7616/files#diff-7db4584c175029bb1576e92bd0628a2cc72516127bd0b038b6b9d9752b3d2633R89-R96))
1. 100% read, 100% RWO volumes
1. 100% read, 100% RWX volumes
1. 100% write, 100% RWO volumes
1. 100% write, 100% RWX volumes
1. No read/write and 100% RWO volumes (this will test the scalability of Longhorn control plane)

> Also, different IO sizes should be tested, like 4k, 64k, 128k, and 1M.
> The reason is that 4k is simulated for the DB scenario, 64k/128k is more likely the general scenario, and 1M is simulated for the streaming application.
> 
> We could focus on the test result's throughput, IOPS, and latency.

Agree with this idea! We are going to use our favorite [kbench](https://github.com/yasker/kbench) benchmarking tool. This tool will use 4k block size for IOOPs/latency results (applications demanding big IOOPs or small latency mainly concerning smaller IO chunks) and 128k block size for bandwidth results (applications demanding big bandwidth mainly concerning bigger IO chunks). More details are at [link](https://github.com/longhorn/longhorn/pull/7616/files#diff-7db4584c175029bb1576e92bd0628a2cc72516127bd0b038b6b9d9752b3d2633R102-R108)

> The scalability could be checked in two parts.
> 
> cluster size
> We do the same test on the various cluster sizes (as you mentioned on LEP), which will help with analysis.
> workload pod
> Various workload pods can give more information that we can provide. For example, one pod can provide 10000 iops. Could we provide 5000 iops with 2 pods?

Agree!

> I thought the customer would like how they can get in a few workload pods or many workloads. That will help them to know how to utilize it. For example, Someone has a workload that needs 10000 iops within 1 ms latency, and another has 20 workloads. Each workload only needs 200 iops with any latency. They can all get what they want from the sizing guide.

This is a good idea! We (@ejweber, @james-munson, and I) also thought similar to this and will design the test in this fashion 

> I agree that we should test with the volume degraded. However, it will introduce more factors in benchmarking.
We can first do the basic rebuilding test. For example, the default rebuild maximum is 5. We can test the 5 volumes in rebuilding at this moment.

I have not thinking much about this degraded test case as I am focusing on stable test cases. I will keep this in mind when I touch this part. 

> Also, we should test the baseline on the disk for the internal data.
The baseline will help us to know how we utilize the disk.

Surely, we will run [kbench](https://github.com/yasker/kbench) with localpath provisioner to get the baseline

enhancements/20240109-reference-architecture-and-sizing-guidelines.md
```diff
@@ -0,0 +1,142 @@
...
+Figure out Recommended backup target.
+
+The factors to consider:
+1. Backup target type (NFS/S3/SMB/CIFS/...)
+1. Backup target latency
+
+
+#### Goal 4
+Develop some kind of automation for this test plan so that we can run this again for every new release.
+This will help us to identify if there are improvement/regression
```
c3y1huang, line 63: Probably out of the scope, but it would be nice to periodically run the test plan with the sprint releases, and have a visualization for the results.

enhancements/20240109-reference-architecture-and-sizing-guidelines.md
```diff
@@ -0,0 +1,142 @@
...
+
+#### Goal 1
+For each supported environment, pick a cluster spec (small, medium, large) to run Longhorn and measure Longhorn performance/scalability with that cluster spec.
+
+The supported environments are:
+1. On-Prem (using Equinix Metal for testing)
+1. Public Cloud (using AWS for testing)
+1. Edge (Not sure where to test)
+
+The cluster spec including the factors:
```
c3y1huang, line 24: Should we also consider the dependency versions such as iscsid, nfs, etc.?

enhancements/20240109-reference-architecture-and-sizing-guidelines.md
```diff
@@ -0,0 +1,142 @@
...
+1. Node's CPU
+1. Node's RAM
+1. Node's Disk IOOPs/bandwidth/latency
+1. Internode networking bandwidth/latency
```
c3y1huang, line 28: Is it worth putting different Kubernetes CNI plugins into the picture? If not, we should define which CNI plugins will be used as this could impact the cluster networking.

innobead: Consider at least the default builtin to rke2/k3s implying what Harvester is using.

PhanLe1010:
Still needs a lot more work but I have a quick update about how the testing graph would look like:
1. On top of the page, we can select global parameters corresponding to the test that we are running:
    * volume-access-mode (rwo, rwx)
    * test-mode (read-only, write-only, read-write)
    * rate-limit-type (no-rate-limit, rate-limit)
1. Then there are 2 sections:
    * Read performance: all read metrics - applicable to read or read-write test mode
    * Write performance: all write metrics - applicable to write or read-write mode
    * Each metric has a corresponding graph of number of reported pods below it. This enable us to see how performance is affected when the number of pods go up 

Please let me know if you have feedback. Thank you

screencapture-143-198-70-16-k8s-clusters-c-m-t7rvd8cj-api-v1-namespaces-cattle-monitoring-system-services-http-rancher-monitoring-grafana-80-proxy-d-2fqV-mhSz-reference-setup-scalability-performance-and-sizing-2024-02-06-13_21_57

PhanLe1010:
Update: I summited a PR into Kbench repo to add more functionality to assist this LEP https://github.com/yasker/kbench/pull/7

PhanLe1010:
For example, this is a test run on a cluster of 3 worker nodes. Each has 4CPUs, 8GiB of RAM. with the test parameters:
* Longhorn RWO volumes: each Kbench pod is running test on a Longhorn RWO volume attached to it
* The test mode is `MODE: random-read-iops`: Kbench is running `random-read-iops` job only
* `LONG_RUN: true`: Kbench will run the `random-read-iops` job repeatedly 
* `metric-exporter` container is deployed on the same pod as the `kbench` container to export test result as Prometheus metrics
* User will scale the number of workload (which is the number of kbench statefulset) from 1 and observer the performance overtime.

This is the Grafana graph for this test:

Screenshot from 2024-02-07 18-19-28


From the graph we can see that:
1. When there is only 1 workload pod, the number of random IOPs that volume can achieve is 10.7k
1. When user scales number of workload pod to 3:
    1. We can see that each volume can achieve around 7.2k random IOPs
    1. Total random IOPs of all volume is about 21.6k
1. When scale number of workload pod to 6:
    1. We can see that each volume can achieve around 3.4k random IOPs
    1. Total random IOPs of all volume is about 20.4k

We can see that there is an upper bound amount of IOPs that can be achieved in this cluster (21K). The number of IOPs each volume can achieve (x) and the number of volumes (y) form a reciprocal function: x * y = 21000. If the user keeps scaling up number of pods eventually, this reciprocal relation will no longer hold as the CPU contention and other factors kick in (i.e. x*y will be less and less) or even Longhorn instance manager pods crash which crashes Longhorn volume. 

User can use this test result as a reference for a similar cluster setup:
1. Longhorn can push up to 21K random IOPs in this cluster
1. If each of your workload pod is doing 1K random IOPs in average, you can have 18 pods (not 21K because of the overhead when having more pods)

PhanLe1010:
Update: start using the above modified Kbench and test plan to perform the test for public cloud - medium node spec. The report will be continuously update at https://github.com/longhorn/longhorn/pull/7905

PhanLe1010:
I think we don't need the LEP anymore because the actual report already has all the information about the design together with the actual data https://github.com/longhorn/longhorn/pull/7905 . I am closing this one, feel to free to reopen if you have additional concern

PR #7679, opened 2024-01-15
fix(doc): update the description in helm values

ChanYiLin:
ref: longhorn/longhorn 7667

Basically the copy/paste from the PR https://github.com/longhorn/website/pull/835


innobead:
@mergify backport v1.6.x

Issue #7399, opened 2023-12-20
[IMPROVEMENT] Remove static sessionAffinity: ClientIP set in most services if not required


innobead:
## What's the task? Please describe

Longhorn API doesn't require sessions as each API call should be independent. Although backing image download and upload are present, the internal forwarding is handled by Longhorn manager and not the load balancer layer. However, most services have sessionAffinity hard-coded with ClientIP. 

[/charts/longhorn/templates/services.yaml#L1-L16](https://github.com/longhorn/charts/blob/728dfbcad8f23cca4574db0b61937015ebadb1dd/charts/longhorn/templates/services.yaml#L1-L16)
```yaml
apiVersion: v1
kind: Service
metadata:
  labels: {{- include "longhorn.labels" . | nindent 4 }}
    app: longhorn-conversion-webhook
  name: longhorn-conversion-webhook
  namespace: {{ include "release_namespace" . }}
spec:
  type: ClusterIP
  sessionAffinity: ClientIP
  selector:
    app: longhorn-conversion-webhook
  ports:
  - name: conversion-webhook
    port: 9443
    targetPort: conversion-wh
```

ejweber:
The scope of this issue is to investigate whether or not we can remove `sessionAffinity: ClientIP` from our services, and that will be my primary area of investigation. However, the broader context from one user can be summarized as follows:

Can Longhorn run in an "multitenant" environment in which some nodes related to one tenant can only talk to other nodes related to one tenant? `sessionAffinity: ClientIP`  makes this difficult, because after the Kubernetes network control plane decides randomly which destination IP address to resolve a service to, it will not make another decision for 10800 (configured by Longhorn) seconds. If, instead, the Kubernetes network control plane might make a different decision every time a connection attempt was made, a pod in one tenant would have a reasonable chance of getting connected to a pod backing a service in the same tenant.

Before getting started with the main investigation, I decided to try to understand exactly how/why Longhorn cannot work in the "multitenant" environment today.

### Setup

In a two node cluster, temporarily ensure it is impossible for one node (`kgxdq`) to route traffic to another node (`nn4rg`) and vice-versa.

```
root@eweber-v126-worker-9c1451b4-kgxdq:~# ip route
default via 143.198.224.1 dev eth0 proto static 
10.42.0.0/24 via 10.42.0.0 dev flannel.1 onlink 
10.42.1.0/24 via 10.42.1.0 dev flannel.1 onlink
[...]

PR #7748, opened 2024-01-22
Remove sessionAffinity ClientIP from services

ejweber:

Issue #7399

Remove `sessionAffinity: ClientIP` from Longhorn services. See #7399 for justification.

We could optionally introduce a Helm variable that controls the value of `sessionAffinity` for the Longhorn services that previously used `sessionAffinity: ClientIP`. If we did this, I would argue for making `None` the default, with the option to change back to `ClientIP` if desired. However:

1. I cannot identify any situation in which we should be using `sessionAffinity: ClientIP`,
2. More Helm variables makes Longhorn deployment more difficult to understand.

I'm curious whether reviewers agree.

### DIFF
--- a/chart/templates/daemonset-sa.yaml
+++ b/chart/templates/daemonset-sa.yaml
@@ -156,7 +156,6 @@ metadata:
   {{- end }}
 spec:
   type: {{ .Values.service.manager.type }}
-  sessionAffinity: ClientIP
   selector:
     app: longhorn-manager
   ports:
--- a/chart/templates/services.yaml
+++ b/chart/templates/services.yaml
@@ -7,7 +7,6 @@ metadata:
   namespace: {{ include "release_namespace" . }}
 spec:
   type: ClusterIP
-  sessionAffinity: ClientIP
   selector:
     app: longhorn-manager
   ports:
@@ -24,7 +23,6 @@ metadata:
   namespace: {{ include "release_namespace" . }}
 spec:
   type: ClusterIP
-  sessionAffinity: ClientIP
   selector:
     app: longhorn-manager
   ports:
@@ -41,7 +39,6 @@ metadata:
   namespace: {{ include "release_namespace" . }}
 spec:
   type: ClusterIP
-  sessionAffinity: ClientIP
   selector:
     app: longhorn-manager
   ports:
--- a/deploy/longhorn.yaml
+++ b/deploy/longhorn.yaml
@@ -4208,7 +4208,6 @@ metadata:
   namespace: longhorn-system
 spec:
   type: ClusterIP
-  sessionAffinity: ClientIP
   selector:
     app: longhorn-manager
   ports:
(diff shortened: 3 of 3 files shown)


innobead (approved):
LGTM.

PhanLe1010 (approved):
LGTM

Agree that we should remove the setting completely!

Quick question. Assume that we have `sessionAffinity: ClientIP` set. In the first time the client connects to the backend service it is routed to manager pod with IP-1. Then manager pod with IP-1 crash, restarted, and get a new IP-2. When client connect to backend service for the 2nd time, will Kubernetes network control plane route the client to a new IP different than the IP-1 assume no manager pod has IP-1 anymore?

ejweber:
> Quick question. Assume that we have sessionAffinity: ClientIP set. In the first time the client connects to the backend service it is routed to pod with IP-1. Then pod with IP-1 crash, restarted, and get new IP-2. When client connect to backend service for the 2nd time, will Kubernetes network control plane route the client to a new IP different than the IP-1 assume no manager pod has IP-1 anymore?

Yes. When the longhorn-manager pods want to communicate with the longhorn-backend service, the DNS lookup always returns the same ClusterIP. So, from the perspective of all longhorn-manager pods, there is only one IP address to communicate with. However, as soon as the Kubernetes network control plane recognizes a change in available longhorn-backend pods, it updates the associated endpoint slice. This causes kube-proxy on every node to update DNAT iptables rules appropriately. The next connection attempt makes use of the updated DNAT rules (with the updated list of pod IPs).

ejweber:
> Quick question. Assume that we have sessionAffinity: ClientIP set. In the first time the client connects to the backend service it is routed to pod with IP-1. Then pod with IP-1 crash, restarted, and get new IP-2. When client connect to backend service for the 2nd time, will Kubernetes network control plane route the client to a new IP different than the IP-1 assume no manager pod has IP-1 anymore?

I think yes, though I am interested enough to test it.

From https://github.com/longhorn/longhorn/issues/7399#issuecomment-1896494098:

---
The chain that handles load balancing (`KUBE-SVC-NQL64TY2T3FWFXFQ`) has two kinds of rules:

- The first set of rules are for packets from source IPs that have already been balanced in the last configured number of seconds (`10800`). If, for example, the source IP is in the `KUBE-SEP-W3Y4QUAIHX5U3K5C` list, we jump to the `KUBE-SEP-W3Y4QUAIHX5U3K5C` chain.
- The second set of rules are for packets from source IPs that have not already been balanced. Probabilities are used to ensure we have an equal chance to jump to any related chain (e.g. `KUBE-SEP-W3Y4QUAIHX5U3K5C`).
---

The event you described should cause kube-proxy to remove the list of source IPs that previously balanced to the old destination IP and the rule that would check that list (e.g. `KUBE-SEP-W3Y4QUAIHX5U3K5C`). The source IP is obviously not in any of the other, still existing, lists, so we would eventually fall down to the set of rules that does the load balancing and DNAT to a different IP address.

In other words, I don't think `sessionAffinity: ClientIP` is likely to be the cause of any previously strange connectivity behavior we may have observed.

innobead:
@mergify backport v1.6.x v1.5.x v1.4.x

Issue #6386, opened 2023-07-26
[IMPROVEMENT] Automatically remount read-only RWO volume to read-write


derekbit:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

In the current system, if a read-write volume was accidentally changed to a read-only state when there are IO errors, user must scale down the workload using the volume and then scale up to make the volume to a read-write state.

To address this issue, we have implemented RWX volume state monitoring and automatic remount functionality in https://github.com/longhorn/longhorn/issues/5961. We can introduce a similar logic for RWO volumes. This logic will periodically monitor the state of the volume, and if it gets stuck in a read-only state, the mechanism will automatically request a remount and restart the workload that uses the volume.

## Additional context

cc @james-munson @PhanLe1010 @shuo-wu  @innobead

chriscchien:
Hi @ChanYiLin ,

Following [steps](https://github.com/longhorn/longhorn/issues/6386#issuecomment-1822340518), I can see the volume read only related infromation below , but the workload can not write data to volume, and did not observe remount after the read only happen.

According to the ticket title, the volume should remount and will be in read-write mode, could  your help to confirm if I misunderstand something? thank you.

(I set `auto-delete-pod-when-volume-detached-unexpectedly` to `false` when test with deployment. Tested on both `true` and `false` value when test with pod, the test result were all the same)

- deploymentr pod
```
root@test-deployment-754dd9fc66-lkn6d:/# echo 123123 >> /mnt/data/1
bash: /mnt/data/1: Read-only file system
```
- longhorn-manager log
```bash
time="2024-01-02T01:52:12Z" level=info msg="Volume request remount at 2024-01-02T01:52:12Z due to engine detected read-only filesystem" func="controller.(*VolumeController).requestRemountIfFileSystemReadOnly" file="volume_controller.go:1417" accessMode=rwo controller=longhorn-volume frontend=blockdev migratable=false node=ip-172-31-40-139 owner=ip-172-31-40-139 state=attached volume=pvc-431e5599-d52d-4e82-b6b2-4f4d1d0c5617
```
- engine.longhorn.io
```bash
> k get engine.longhorn.io -A -o yaml  | grep FilesystemReadOnly -C 1
    status: "True"
    type: FilesystemReadOnly
  currentImage: longhornio/longhorn-engine:master-head
```
[...]

PR #7758, opened 2024-01-23
feat(mount): process manager periodically check mount status and update engine condition

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/6386

Add conditions to instance-manager crd


ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7501

innobead (approved):

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7730

Issue #7496, opened 2023-12-29
[BUG] Rancher cannot import longhorn 1.5 charts due to "error converting YAML to JSON: yaml: line 699: did not find expected key"


iTaybb:
## Describe the bug (🐛 if you encounter this issue)
When using Rancher 2.7.X, installing longhorn 1.5.0, 1.5.1, 1.5.2 or 1.5.3 results in the following error:
```
{
"type": "error",
"links": { },
"code": "ServerError",
"message": "error converting YAML to JSON: yaml: line 699: did not find expected key",
"status": 500
}
```

## To Reproduce
1. Add https://charts.longhorn.io/ as a ClusterRepo object (rancher2_catalog_v2)
2. Try to install Longhorn 1.5.0, 1.5.1, 1.5.2 or 1.5.3

## Expected behavior

Installation should succeed.

## Environment
Longhorn 1.5.0, 1.5.1, 1.5.2 or 1.5.3
Rancher 2.7.9

 - Longhorn version:
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl):
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version:
   - Number of management node in the cluster:
   - Number of worker node in the cluster:
 - Node config
   - OS type and version:
   - Kernel version:
   - CPU per node:
   - Memory per node:
   - Disk type(e.g. SSD/NVMe/HDD):
   - Network bandwidth between the nodes:
 - Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal):
 - Number of Longhorn volumes in the cluster:
 - Impacted Longhorn resources:
   - Volume names:

samene:
Any update on this? We are facing the same issue on v1.5.2

iTaybb:
This still happens unfortunately

innobead:
What Rancher version are you using?

@khushboo-rancher please help follow up with this. cc @PhanLe1010 @mantissahz

iTaybb:
I'm using 2.7.9

samene:
we are using 2.7.9 too.

PhanLe1010:
Thanks all. I am checking the issue

PhanLe1010:
### Reproduce steps
1. Deploy Rancher 2.7.9
1. Deploy this terraform manifest 
    ```tf
    terraform {
      required_providers {
        rancher2 = {
          source = "rancher/rancher2"
[...]

PR #7777, opened 2024-01-25
Fix duplicated quotation causing Rancher terraform fails to deploy longhorn chart

PhanLe1010:
longhorn/longhorn#7776

More details at https://github.com/longhorn/longhorn/issues/7496#issuecomment-1909146497


mantissahz (approved):
LGTM

innobead (approved):

PR #7796, opened 2024-01-27
fix(environment_check): add check for globalDefault priority class.

james-munson:

Issue https://jira.suse.com/browse/SURE-7492

Make the environment_check script look for an existing priority class with globalDefault=true to head off this issue and https://github.com/longhorn/longhorn/issues/2820.

This is really only for backport to v1.4.x and v1.5.x.  In 1.6.0 and later, Longhorn defines its own `longhorn-critical` priority class and ensures that system pods are set to that.

### DIFF
--- a/scripts/environment_check.sh
+++ b/scripts/environment_check.sh
@@ -260,6 +260,18 @@ check_hostname_uniqueness() {
   info "All nodes have unique hostnames."
 }
 
+check_default_priority_class() {
+  hasDefault=$(kubectl get priorityclass -o jsonpath='{.items[*].globalDefault}' | grep true)
+
+  if [ ! -z $hasDefault ]; then
+      # See https://longhorn.io/kb/troubleshooting-instance-manager-pods-are-restarted-every-hour/#reason
+      warn "Cluster has a global default priority class."
+      return 1
+  fi
+
+  info "Cluster has no global default priority class."
+}
+
 check_nodes() {
   local name=$1
   local callback=$2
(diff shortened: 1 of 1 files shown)


scripts/environment_check.sh
```diff
@@ -260,6 +260,18 @@ check_hostname_uniqueness() {
   info "All nodes have unique hostnames."
 }
 
+check_default_priority_class() {
+  hasDefault=$(kubectl get priorityclass -o jsonpath='{.items[*].globalDefault}' | grep true)
+
+  if [ ! -z $hasDefault ]; then
+      # See https://longhorn.io/kb/troubleshooting-instance-manager-pods-are-restarted-every-hour/#reason
+      warn "Cluster has a global default priority class."
```
shuo-wu, line 268: NIT: Mention that this global default priority class may contradict the Longhorn setting then lead to unexpected Longhorn components restart

ejweber (commented):
Maybe we should target this one only for Longhorn `v1.5.x` and older. If I understand correctly, there will be no issue with a default priority class in Longhorn `v1.6.0+`. WDYT?

james-munson:
> Maybe we should target this one only for Longhorn `v1.5.x` and older. If I understand correctly, there will be no issue with a default priority class in Longhorn `v1.6.0+`. WDYT?

I do think so, and said that in the notes, but not sure how to accomplish that.  It is suggested that I just open the PR directly into 1.5.x and then backport from there, so I'll do that.

scripts/environment_check.sh
james-munson: That's what the KB article pointed to in the script comment is about.
It seems like the environment_check script is generally pretty terse, and the user is expected to look in the docs to find the full reason behind each check.  This might possibly need some doc text as well.

james-munson:
This PR has been superseded by [PR 7801](https://github.com/longhorn/longhorn/pull/7807) to go into v1.5.x and backport to 1.4.x from there.

scripts/environment_check.sh
shuo-wu: The warning message should provide either a concrete explanation or the doc link. I don't think the majority of users won't actively check the doc or know where the doc is after receiving this message.

PR #7807, opened 2024-01-29
fix(environment_check): add check for globalDefault priority class.

james-munson:
https://github.com/longhorn/longhorn/issues/7831

Make the environment_check script look for an existing priority class with globalDefault=true to head off potential issues with Longhorn.

This is really only for backport to v1.4.x and v1.5.x. 
In 1.6.0 and later, Longhorn defines its own `longhorn-critical` priority class and ensures that system pods are set to that.

This PR is directly into `v1.5.x` to take the place of [PR 7796](https://github.com/longhorn/longhorn/pull/7796), because it really doesn't need to go into master.

### DIFF
--- a/scripts/environment_check.sh
+++ b/scripts/environment_check.sh
@@ -260,6 +260,14 @@ check_hostname_uniqueness() {
   info "All nodes have unique hostnames."
 }
 
+check_default_priority_class() {
+  hasDefault=$(kubectl get priorityclass -o jsonpath='{.items[*].globalDefault}' | grep true)
+
+  if [ ! -z $hasDefault ]; then
+      warn "Cluster has a global default priority class. See potential issue and solution at https://longhorn.io/kb/troubleshooting-instance-manager-pods-are-restarted-every-hour/#reason"
+  fi
+}
+
 check_nodes() {
   local name=$1
   local callback=$2
@@ -511,7 +519,10 @@ DEPENDENCIES=("kubectl" "jq" "mktemp" "sort" "printf")
 check_local_dependencies "${DEPENDENCIES[@]}"
 
 # Check the each host has a unique hostname (for RWX volume)
+# Do this first, so we also verify we have a workable KUBECONFIG.
 check_hostname_uniqueness
+
+check_default_priority_class
 
 # Create a daemonset for checking the requirements in each node
 TEMP_DIR=$(mktemp -d)


scripts/environment_check.sh
```diff
@@ -250,6 +250,18 @@ check_hostname_uniqueness() {
   info "All nodes have unique hostnames."
 }
 
+check_default_priority_class() {
+  hasDefault=$(kubectl get priorityclass -o jsonpath='{.items[*].globalDefault}' | grep true)
+
+  if [ ! -z $hasDefault ]; then
+      # See https://longhorn.io/kb/troubleshooting-instance-manager-pods-are-restarted-every-hour/#reason
+      warn "Cluster has a global default priority class."
```
PhanLe1010, line 258: Can we link the recommended action into the log for user to see it? 
```suggestion
      warn "Cluster has a global default priority class. See potential issue and solution at https://longhorn.io/kb/troubleshooting-instance-manager-pods-are-restarted-every-hour/#reason"
```

scripts/environment_check.sh
```diff
@@ -250,6 +250,18 @@ check_hostname_uniqueness() {
...
+      return 1
+  fi
+
+  info "Cluster has no global default priority class."
```
PhanLe1010, line 262: Should we remove this log because it may sound like something negative when user see it while it is actually a good thing? If the check pass, maybe just silently move on as other checks?

ejweber (commented):
LGTM in general.

I agree with @Phan's suggestions.

scripts/environment_check.sh
james-munson: Done.

ejweber (approved):

scripts/environment_check.sh
```diff
@@ -260,6 +260,15 @@ check_hostname_uniqueness() {
   info "All nodes have unique hostnames."
 }
 
+check_default_priority_class() {
+  hasDefault=$(kubectl get priorityclass -o jsonpath='{.items[*].globalDefault}' | grep true)
+
+  if [ ! -z $hasDefault ]; then
+      warn "Cluster has a global default priority class. See potential issue and solution at https://longhorn.io/kb/troubleshooting-instance-manager-pods-are-restarted-every-hour/#reason"
```
innobead, line 267: Do we not need to check whether the priority class of Longhorn components is set or not? If there is a global priority, it does not imply that Longhorn doesn't have a configured priority class.

innobead (changes requested):

scripts/environment_check.sh
ejweber: I think we generally expect the user to run the environment check before they install Longhorn: https://longhorn.io/docs/1.5.3/deploy/install/#using-the-environment-check-script. (Though certainly we ask some users to run it after the fact.)

innobead: That's fair. Probably just make it a warning instead of returning an error.

james-munson: Right, it is just a warning and a reference article.

shuo-wu (approved):

innobead (approved):

james-munson:
@mergify backport v1.4.x

Issue #5041, opened 2022-12-12
[FEATURE] Support "ReadOnlyMany" access mode for Longhorn volumes


derekbit:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

K8s supports the `ReadOnlyMany` access modes for PersistentVolumes. This mode allows multiple pods on different nodes.
More information on ROX: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes

## Describe the solution you'd like

A clear and concise description of what you want to happen

## Additional context

https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/readonlymany-disks

innobead:
The reasonable use case behind this can refer to https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/readonlymany-disks.

PhanLe1010:
We discussed this ticket in our US sync up. @james-munson would like to take over the ticket as he has some good ideas for the implementation 

cc @innobead

james-munson:
We can already do something very close to this with existing share-manager code.  If a volume is made ReadWriteMany, but uses a storage class with nfsOptions overrides that include "ro", it is mounted read-only for all the workload pods it attaches to.
For instance (exec into the workload pod):

```
root@volume-test-1:/# cd /data
root@volume-test-1:/data# mount -l | grep nfs
10.43.253.78:/pvc-22425f85-5f70-4aa8-b55e-d48c7b9ea8df on /data type nfs4 (ro,relatime,vers=4.2,rsize=1048576,wsize=1048576,namlen=255,hard,proto=tcp,timeo=600,retrans=2,sec=sys,clientaddr=64.23.158.38,local_lock=none,addr=10.43.253.78)

root@volume-test-1:/data# ll
total 28
drwxrwxrwx 4 root root  4096 Jan 25 22:14 ./
drwxr-xr-x 1 root root  4096 Jan 24 23:29 ../
drwx------ 2 root root 16384 Jan 24 23:29 lost+found/
drwxr-xr-x 2 root root  4096 Jan 25 22:14 test_dir/

root@volume-test-1:/data# cd test_dir/
root@volume-test-1:/data/test_dir# ll
total 12
drwxr-xr-x 2 root root 4096 Jan 25 22:14 ./
drwxrwxrwx 4 root root 4096 Jan 25 22:14 ../
-rw-r--r-- 1 root root 3569 Jan 24 23:32 test_data.txt

[...]

PR #7853, opened 2024-02-05
feat(lep): support a ReadOnlyMany behavior.

james-munson:

Issue https://github.com/longhorn/longhorn/issues/5041

Make an LEP with a proposal for how to implement `ReadOnlyMany` behavior in Longhorn code.

### DIFF
--- a/enhancements/20240202-read-only-many.md
+++ b/enhancements/20240202-read-only-many.md
@@ -1,0 +1,167 @@
+# ReadOnlyMany Support
+
+## Summary
+
+Longhorn should support a ReadOnlyMany behavior, along the lines of https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/readonlymany-disks.  It should do so using the same share-manager mechanism with exposure through NFS that ReadWriteMany uses.
+
+### Related Issues
+
+https://github/longhorn/longhorn/issues/5401
+
+## Motivation
+
+### Goals
+
+Pods and multi-pod workloads such as deployments and statefulssets can mount an existing volume's data without being able to alter it.
+
+The configuration should be simple to use, to document, and to detect.
+
+### Non-goals [optional]
+
+The mounted volume need not be read-only at all levels. In particular, the CSI mount to share-manager may be read/write.
+
+## Proposal
+
+Add a parameter to storageclass and thus to volume spec named `mountReadOnly` with default value of `false`.
+
+### User Stories
+
+The user can already do something very close to this with existing share-manager code. If a volume is made ReadWriteMany, but uses a storage class with `nfsOptions` overrides that include "ro", it is mounted read-only for all the workload pods it attaches to.
+
+It's not a complete solution for a couple of reasons. First, it feels "back-door", more like a workaround than a supported feature.
+
+Second, and more critical, Longhorn code has to be able to determine that a volume that is read-only is intended to be so. Otherwise, the logic that attempts to un-mount and re-mount as writable will continually try and fail to do so, causing repeated attach/detach sequences.
+
+There are several possible implementations. They all use the share-manager to achieve multiple readers. The mount onto the sharemanager in `/export` is not itself read-only. Only the NFS mount by the workload pod(s) is read-only.
+
+The alternatives are
+
+  1. Just document to use "ro" in storageclass parameter `nfsOptions`.
+  2. Make a `ReadOnlyMany` access mode enum value (alias `rox`) and use it instead of `ReadWriteMany (rwx)` but still link it to the share-manager.
+  3. Introduce another parameter, `mountReadOnly: true` in storage class (defaults to false).  
+
+The pros and cons:
+
+#### 1. As is:
+  - Doesn't feel like a feature; more of a back-door.
+  - Auto-remount feature (volumes that are read-only due to some mishap but should not be) needs a way to check the volume's intent. It would be cumbersome to make it parse the NFS mount options to figure that out.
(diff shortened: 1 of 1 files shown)


enhancements/20240202-read-only-many.md
```diff
@@ -0,0 +1,167 @@
...
+
+The configuration should be simple to use, to document, and to detect.
+
+### Non-goals [optional]
+
+The mounted volume need not be read-only at all levels. In particular, the CSI mount to share-manager may be read/write.
+
+## Proposal
+
+Add a parameter to storageclass and thus to volume spec named `mountReadOnly` with default value of `false`.
```
innobead, line 25: We should just leverage the access mode of a PVC to determine if the volume is ROM. The mount ro option should supersede the customized options provided by users.

enhancements/20240202-read-only-many.md
```diff
@@ -0,0 +1,167 @@
...
+  3. Introduce another parameter, `mountReadOnly: true` in storage class (defaults to false).  
+
+The pros and cons:
+
+#### 1. As is:
+  - Doesn't feel like a feature; more of a back-door.
+  - Auto-remount feature (volumes that are read-only due to some mishap but should not be) needs a way to check the volume's intent. It would be cumbersome to make it parse the NFS mount options to figure that out.
+  - Since it requires using `nfsOptions`, the user is responsible for knowing all the other necessary mount options to add to "ro".
+
+#### 2. New `ReadOnlyMany` access mode enum value:
```
innobead, line 50: As an official support, we should go with option 2 to provide a better user experience via a standard interface (access mode: ROM).

innobead (changes requested):

innobead:
Would like to hear thoughts from @derekbit and @shuo-wu

enhancements/20240202-read-only-many.md
derekbit: Agree with @innobead's suggestion.
If the PVC is ROM, we automatically append `ro` option.

ejweber (commented):
I agree with the comments so far: we should leverage the Kubernetes `AccessMode = ReadOnlyMany` (set in PVC and PV) instead of adding a new field to a StorageClass and Longhorn volume.

We should consider the possibility that we will be asked to create a volume with multiple capabilities / access modes simultaneously. For example:

- ReadOnlyMany, and
- ReadWriteMany.

Then, we may be asked to publish the volume with either of these access modes.

What should we do in this case?

I think we should probably create a Longhorn volume with `AccessMode == rwx` when responding to the `CreateVolume` request. Then, we should modify behavior appropriately in `ControllerPublishVolume` (if we want to enforce read-only on the export) and `NodeStageVolume`/`NodePublishVolume` (where we will modify mount options).

ejweber:
> Instead of "only" doing this on the mount side though I believe we'll want to change the actual export to be read-only (or rather, not rw), so that the NFS server side enforces this too.

I am not sure this is strictly necessary, but it may be the "right" thing to do. It looks like we already template the Ganesha configuration, so it might not be too difficult to change the access type on the export if we can get the right information to the share manager pod (https://github.com/longhorn/longhorn-share-manager/blob/3168f43cad72c25cec704494064f3f58e28b7210/pkg/server/nfs/nfs_server.go#L148-L169). That may mean an additional API change though, since I don't think there's an existing ShareManager field to drop the read-only information into.

PhanLe1010:
As discussed with @james-munson, currently Kubernetes has 4 access modes in total:
1. RWO - ReadWriteOnce
1. ROX - ReadOnlyMany
1. RWX - ReadWriteMany
1. RWOP - ReadWriteOncePod

We are missing `ReadOnlyMany` and `ReadWriteOncePod`. Maybe `ReadWriteOncePod` can be on a different ticket but it would be great if we keep it in mind while developing this `ReadOnlyMany` mode.

PhanLe1010:
> simultaneously. For example:
> 
> ReadOnlyMany, and
> ReadWriteMany.
> Then, we may be asked to publish the volume with either of these access modes.
> 
> What should we do in this case?
> 
> I think we should probably create a Longhorn volume with AccessMode == rwx when responding to the CreateVolume request. Then, we should modify behavior appropriately in ControllerPublishVolume (if we want to enforce read-only on the export) and NodeStageVolume/NodePublishVolume (where we will modify mount options).

This is a good point. I think we can take a look at how CSI sidecar is handling the selection when there are multiple modes specified. Some combinations should be blocked in CreateVolume:  https://github.com/kubernetes-csi/csi-lib-utils/blob/f82f9de5b8aeb3c3b236d7f58fc5eeab34438078/accessmodes/access_modes.go#L36-L78

PhanLe1010:
> Instead of "only" doing this on the mount side though I believe we'll want to change the actual export to be read-only (or rather, not rw), so that the NFS server side enforces this too.

Agree with @ejweber that  this might not be strictly necessary. When PVC requests ReadOnlyMany, it is not mandatory that we have to force the underlying storage layer to be read-only https://kubernetes.io/docs/concepts/storage/persistent-volumes/#:~:text=For%20example%2C%20even%20if%20a%20PersistentVolume%20is%20created%20as%20ReadOnlyMany%2C%20it%20is%20no%20guarantee%20that%20it%20will%20be%20read%2Donly

james-munson:
Converting this to draft while I incorporate the feedback and renew with the recommended approach.  I'll either re-open or close and replace with another PR.

innobead:
@james-munson This PR has been pending for a while. Let's resume this and try to get it for 1.7.0 still.

derekbit:
ping @james-munson

derekbit:
Revisit while working on this feature.
cc @COLDTURNIP

Issue #7428, opened 2023-12-23
[IMPROVEMENT] Remove startup probe of CSI driver after liveness probe conn fix ready


innobead:
Ideas to fix (in order of preference?):

1. Patch upstream `livenessprobe` to restore previous behavior. Revert to `v2.10.0` (or `v2.9.0`, since we already have a Longhorn version of it) in the meantime.
2. Add a startup probe to the `longhorn-csi-plugin` container with a `5m` window or configure its `livenessProbe.initialDelaySeconds` to `5m`. Kubernetes will not kill this container for the first five minutes it is alive. This will keep it alive long enough for `livenessprobe` to connect with it whenever it comes back, but maintain the existing probe behavior after startup.
3. ~Rework `longhorn-csi-plugin` startup so that it does not crash if it cannot connect to the `longhorn-backend` service. This may prevent the offset CrashLoopBackoffs from occurring in the first place.~ Kubernetes liveness probes will still fail, since the container is not ready. This only works as a potential addendum to 2.

_Originally posted by @ejweber in https://github.com/longhorn/longhorn/issues/7116#issuecomment-1861835259_

ejweber:
- [ ] Backport to v1.5.x as well.

I am holding off for now because I am not sure we can get the master PRs merged and backported in time for v1.5.4 testing. If things go quickly, we can aim for v1.5.4. Otherwise, I will hold off on the backport until it is released.

Issue #7116, opened 2023-11-16
[BUG] CSI components CrashLoopBackOff, failed to connect to unix://csi/csi.sock after cluster restart


yangchiu:
## Describe the bug (🐛 if you encounter this issue)

After cluster restart (reboot all k8s nodes including control plane node), all csi components got stuck in CrashLoopBackOff:
```
╰─$ kl get pods
NAME                                                     READY   STATUS             RESTARTS          AGE
longhorn-manager-pxlcx                                   1/1     Running            16 (10h ago)      15h
longhorn-manager-gr26g                                   1/1     Running            16 (10h ago)      15h
engine-image-ei-b907910b-m6gr4                           1/1     Running            16 (10h ago)      15h
instance-manager-bb7e4a2035b50ec3896800262c56800c        1/1     Running            0                 10h
engine-image-ei-b907910b-7vfnt                           1/1     Running            16 (10h ago)      15h
instance-manager-dca4e951d12ab6ab6f714f42f1a2416e        1/1     Running            0                 10h
csi-attacher-67f8c99bcd-9fqwj                            1/1     Running            20 (10h ago)      15h
csi-snapshotter-6f6bf9c757-t68d7                         1/1     Running            19 (10h ago)      15h
csi-resizer-545b7c64f5-7l2hv                             1/1     Running            19 (10h ago)      15h
csi-attacher-67f8c99bcd-2chq8                            1/1     Running            19 (10h ago)      15h
csi-provisioner-557c7f7c44-g2x77                         1/1     Running            20 (10h ago)      15h
longhorn-driver-deployer-7f94bb668f-s7zwg                1/1     Running            16 (10h ago)      15h
longhorn-csi-plugin-h5xqw                                3/3     Running            72 (10h ago)      15h
longhorn-ui-646f4bc8df-tbmtt                             1/1     Running            29 (10h ago)      15h
csi-resizer-545b7c64f5-gzm2z                             1/1     Running            19 (10h ago)      15h
csi-provisioner-557c7f7c44-2p6lc                         1/1     Running            19 (10h ago)      15h
engine-image-ei-b907910b-9ztgm                           1/1     Running            16 (10h ago)      15h
csi-snapshotter-6f6bf9c757-jqqp9                         1/1     Running            19 (10h ago)      15h
longhorn-csi-plugin-4sqkv                                3/3     Running            69 (10h ago)      15h
longhorn-manager-dfn6d                                   1/1     Running            16 (10h ago)      15h
[...]

PR #7908, opened 2024-02-09
Use livenessprobe version with upstream fix

ejweber:

longhorn/longhorn#7428

This version of livenessprobe will not allow us to get stuck like in https://github.com/longhorn/longhorn/issues/7116. By upgrading to it, we can revert the changes associated with that ticket.

https://github.com/kubernetes-csi/livenessprobe/issues/236


PhanLe1010 (approved):

ejweber:

Issue #4648, opened 2022-10-01
[BUG] Fix the share-manager deletion failure if the confimap is not existing


derekbit:
## Describe the bug

Fix the share-manager deletion failure if the confimap is somehow deleted

## To Reproduce

1. Install Longhorn master branch
2. Update the PVC size of `longhorn/examples/rwx/rwx-nginx-deployment.yaml` to 8589934591Gi
3. Run `kubectl apply -f longhorn/examples/rwx/rwx-nginx-deployment.yaml`
4. Then, delete the rwx volume by  `kubectl delete -f longhorn/examples/rwx/rwx-nginx-deployment.yaml`
5. The deletion will be stuck because the confimap (recovery backend) is not found.

## Log or Support bundle

If applicable, add the Longhorn managers' log or support bundle when the issue happens. 
You can generate a Support Bundle using the link at the footer of the Longhorn UI.

## Environment

 - Longhorn version:
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl):
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version:
   - Number of management node in the cluster:
   - Number of worker node in the cluster:
 - Node config
   - OS type and version:
   - CPU per node:
   - Memory per node:
   - Disk type(e.g. SSD/NVMe):
   - Network bandwidth between the nodes:
 - Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal):
 - Number of Longhorn volumes in the cluster:

Lichiehyao:
**Verified on master 20221013**

- longhorn master-head ([97cb124](https://github.com/longhorn/longhorn/commit/97cb124f247e736a77c4e73e319cc4b5c2f1bb46)) 
- longhorn-manager master-head ([d0df36b](https://github.com/derekbit/longhorn-manager/commit/d0df36b1dd7112182eea0ee9abc967f2b9b08fe9)) 


**The test steps**

[#4648 (comment)](https://github.com/longhorn/longhorn/issues/4648#issue-1393184270)

**Result Passed** 
1. The rwx volume could be deleted

PR #7910, opened 2024-02-09
Add a Helm variable that disables the default StorageClass

ejweber:

longhorn/longhorn#4648

Allows the longhorn-storageclass ConfigMap to be completely omitted on Helm install by setting `enableDefaultStorageClass: false`. When combined with longhorn-manager improvements, this can have the effect of not creating and even removing the Longhorn default StorageClass.


ejweber:
We decided to close https://github.com/longhorn/longhorn/issues/4648 instead of continuing.

Issue #7887, opened 2024-02-07
[BUG] Volumes stuck upgrading after 1.5.3 -> 1.6.0 upgrade.


Eilyre:
## Describe the bug

Upgraded the system from version 1.5.3 to 1.6.0 today, to get around the RWX bug. Got to the part of upgrading engines through UI. Did that to all volumes. The amount of replicas went sky high, and the volumes are stuck in the upgrading state.

As there was a volume stuck detaching, I thought restarting the instance manager responsible for it would maybe break it out. It did not, and made things so-so much worse. Now there's even more volumes stuck detaching, all the volumes are declared degraded, and rebuilding replicas won't happen. I cannot now rollback the volumes either, with the error: `cannot do live upgrade for a unhealthy volume`

## To Reproduce

Not sure if it's reproducible.

## Expected behavior

I've done this upgrade method from longhorn version 1.0.2. I had expected it to work the same way, as it did before.

## Support bundle for troubleshooting

Support bundle attached.

## Environment

 - Longhorn version: 1.6.0
 - Impacted volume (PV): All of them.
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl): Kubectl
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: Kubeadm, 1.27.4
   - Number of control plane nodes in the cluster: 3
   - Number of worker nodes in the cluster: 3
 - Node config
   - OS type and version: Flatcar, latest
   - Kernel version: 6.1.73
   - CPU per node: 128 cores
   - Memory per node: 512 GB
   - Disk type (e.g. SSD/NVMe/HDD): NVMe
   - Network bandwidth between the nodes (Gbps): 100 gbps
 - Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal): Baremetal
 - Number of Longhorn volumes in the cluster: ~160

## Additional context

This is a production system, so I am currently fairly worried.

ejweber:
Hello @Eilyre. I am not able to find the support bundle. Can you post the link or send it along to longhorn-support-bundle@suse.com?

ejweber:
So far, I think it has something to do with this (repeated) log:

```
[...]

PR #7916, opened 2024-02-12
Explicitly set preserveUnknownFields to false

ejweber:
Longhorn 7887

longhorn/longhorn#7887

If Longhorn and Kubernetes are upgraded in a particular order from Longhorn `v1.0.2-` to the latest, `preserveUnknownFields` may remain `true`, even though the default (and intended value) is `false`. As a result, Kubernetes does not consider affected CRDs to have structural schemas, potentially resulting in unexpected behavior (including https://github.com/longhorn/longhorn/issues/7887).

### DIFF
--- a/chart/templates/crds.yaml
+++ b/chart/templates/crds.yaml
@@ -1367,7 +1367,6 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.7.0
-  creationTimestamp: null
   labels: {{- include "longhorn.labels" . | nindent 4 }}
     longhorn-manager: ""
   name: engines.longhorn.io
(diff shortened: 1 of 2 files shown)


chart/templates/crds.yaml
```diff
@@ -1367,7 +1367,6 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.7.0
-  creationTimestamp: null
```
james-munson, line 1370: You've probably explained it elsewhere, but why does this PR also remove creationTimestamp?

ejweber: It is a side effect of running `generate_code.sh` (and implicitly `kustomize edit add patch`). Cross referencing `crds.yaml` with `patches/crd`, any CRD that previously had a patch also has `createTimestamp: null`. These changes add patches, so they remove some more `creationTimestamp: null` fields. TBH I'm not too sure why this occurs, but it appears expected and harmless at this point.

james-munson (approved):

ejweber:
@mergify backport v1.6.x v1.5.x v1.4.x

Issue #7672, opened 2024-01-15
[FEATURE] Longhorn block-type disks supports different SPDK disk bdev drivers


derekbit:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

Currently, Longhorn block-type disks rely on SPDK AIO bdev driver. However, AIO bdev driver does not support UNMAP operation ([codes](https://github.com/longhorn/spdk/blob/longhorn/module/bdev/aio/bdev_aio.c#L520-L532)), and hence trimming filesystem on a v2 volume is infeasible.

To support filesystem feature, a disk can be managed by [NVMe bdev](https://github.com/longhorn/spdk/blob/longhorn/module/bdev/nvme/bdev_nvme.c#L2571) for a NVMe device or [Virtio Block](https://github.com/longhorn/spdk/blob/longhorn/module/bdev/virtio/bdev_virtio_blk.c#L224) for a virtio device.
Ref: https://spdk.io/doc/bdev.html

- [x] Support bind, unbind, disk-status, disk-driver commands in spdk setup.sh
- [x] Support disk management by SPDK NVMe or virtio bdev 
- [x] Document update

derekbit:
@innobead  Can I pick up the ticket?

innobead:
100% go ahead to take it! @derekbit

roger-ryao:
I encountered the issue https://github.com/longhorn/longhorn/issues/8846, so I could not verify this ticket.

Test Steps for Clearer Verification

1. Verify the detection of **VirtIO** disk
    1. Get the BDF of the disk by `ls -al /sys/block`
    ```
    # lsblk 
    vda                 252:0    0    50G  0 disk 
    ├─vda1              252:1    0     1M  0 part 
    ├─vda2              252:2    0     2G  0 part /boot
    └─vda3              252:3    0    48G  0 part 
      └─ubuntu--vg-ubuntu--lv
                        253:0    0    48G  0 lvm  /var/lib/kubelet/pods/e864c8e7-c552-4082-9788-158e0937625d/volume-subpaths/admin-kubeconfig/proxy/0
                                                  /var/lib/kubelet/pods/e864c8e7-c552-4082-9788-158e0937625d/  volume-subpaths/user-kubeconfig/helm/1
                                                  /
    vdb                 252:16   0    20G  0 disk

    # ls -al /sys/block/
[...]

PR #8014, opened 2024-02-23
crd: introduce disk driver in node.spec.diskSpec

derekbit:

Issue Longhorn/longhorn#7672

### DIFF
--- a/chart/templates/crds.yaml
+++ b/chart/templates/crds.yaml
@@ -2116,6 +2116,12 @@ spec:
                       - filesystem
                       - block
                       type: string
+                    diskDriver:
+                      enum:
+                      - ""
+                      - auto
+                      - aio
+                      type: string
                     evictionRequested:
                       type: boolean
                     path:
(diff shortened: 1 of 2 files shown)


chart/templates/crds.yaml
```diff
@@ -2116,6 +2116,12 @@ spec:
                       - filesystem
                       - block
                       type: string
+                    diskDriver:
```
innobead, line 2119: Can we make a default value here?

derekbit: No, it is not allowable.

innobead: I see, so we will assume still `aio` will be the default, correct?

derekbit: We will use `auto` in my current implementation.
If the diskPath is a BDF, the SPDK engine discerns whether it's an NVMe or Virtio disk. Alternatively, if the path follows a canonical format such as `/dev/sdb`, it indicates an `aio` disk.

innobead: Sounds good :muscle: , but we should show the real drive info in the status.

derekbit: Yes, the information will be recorded in the status
image

innobead (approved):

PR #8078, opened 2024-02-29
feat: Update OKD/OCP Docs for 4.15

ArthurVardevanyan:
Update Tested Versions for OKD/OCP Installation


chart/ocp-readme.md
```diff
@@ -51,11 +61,27 @@ Main changes and tasks for OCP are:
...
+    - 1.5.3
+  - 4.14.0-0.okd-2023-08-12-022330 - 4.14.0-0.okd-2024-01-26-175629:
     - Tested, No Known Issues
+- 4.15 / 1.28:
+  - Longhorn Versions Tested:
+    - 1.5.3
+    - 1.6.0
+  - 4.15.0-0.okd-2024-01-27-070424 - 4.15.0-0.okd-2024-02-23-163410:
+    - Tested, No Known Issues
+      - Note: 4.15 requires an additional yaml for UI Login, provided under ```chart/templates/deployment-ui: secret/longhorn-ui-service-account```
```
innobead, line 83: Do we have any upstream reference for this requirement?

chart/templates/deployment-ui.yaml
```diff
@@ -34,6 +34,17 @@ spec:
...
+apiVersion: v1
+kind: Secret
+metadata:
+  labels: {{- include "longhorn.labels" . | nindent 4 }}
+    app: longhorn-ui
+  name: longhorn-ui-service-account
+  namespace: {{ include "release_namespace" . }}
+  annotations:
+    kubernetes.io/service-account.name: longhorn-ui-service-account
+type: kubernetes.io/service-account-token
```
innobead, line 46: This is not a secure way mentioned in https://kubernetes.io/docs/concepts/configuration/secret/#serviceaccount-token-secrets. 

Why do we require this token persistent in a secret here? Please elaborate this.

innobead (changes requested):

chart/templates/deployment-ui.yaml
ArthurVardevanyan: I agree.
Why it is needed, I am not 100% sure. This token is not actually used by an end user to log in.  However without the secret present, the https://github.com/openshift/oauth-proxy doesn't seem to correctly front the request.

chart/ocp-readme.md
ArthurVardevanyan: I have not been able to find any documentation on this. I came across this by troubleshooting.

ArthurVardevanyan: Bug is Opened: https://issues.redhat.com/browse/OCPBUGS-30319

chart/templates/deployment-ui.yaml
innobead: @mantissahz any ideas on this as per your verification experience?

mantissahz: I did not encounter any issues about this during my verification.
Is this related to this https://docs.openshift.com/container-platform/4.15/authentication/using-service-accounts-in-applications.html#auto-generated-sa-token-secrets_using-service-accounts ?

ArthurVardevanyan: Correct

akashsemil:
Here is more info about the issue:

I deployed latest OKD version with longhorn 1.6.2 : [4.15.0-0.okd-2024-03-10-010116](https://github.com/okd-project/okd/releases/tag/4.15.0-0.okd-2024-03-10-010116)

Tried to access the UI using route "Log in with OpenShift" shows the following error:

image

Checking on the openshift-authentication pod logs

image

Eventually found out service account exists but no secret token attached to it.

Screenshot from 2024-08-13 22-22-34


As in Kubernetes v1.24 

```
The LegacyServiceAccountTokenNoAutoGeneration feature gate is beta, and enabled by default. When enabled, Secret API objects containing service account tokens are no longer auto-generated for every ServiceAccount. Use the [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) API to acquire service account tokens, or if a non-expiring token is required, create a Secret API object for the token controller to populate with a service account token by following this [guide](https://kubernetes.io/docs/concepts/configuration/secret/#service-account-token-secrets). (https://github.com/kubernetes/kubernetes/pull/108309, [@zshihang](https://github.com/zshihang))
```

In order to fix this either create a tokenRequest or a long-lived secret as mentioned in the PR

Refer: 

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#urgent-upgrade-notes
https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#auto-generated-legacy-serviceaccount-token-clean-up

innobead:
@mantissahz please help review this.

ArthurVardevanyan:
Updated to include the OKD 4.16 Builds, while not full stable channel, they are releasing "next" builds.
While not recommended it is possible to upgrade from 4.15 -> 4.16.
New Builds on 4.16 seem to be fine.

mantissahz:
Hi @ArthurVardevanyan,

Could you help test the Longhorn 1.7.0 as well?

ArthurVardevanyan:
> Hi @ArthurVardevanyan,
> 
> Could you help test the Longhorn 1.7.0 as well?

# Part 1
Started Clean on
OKD ```4.15.0-0.okd-2024-03-10-010116```
Longhorn ```1.6.2```
Screenshot from 2024-08-22 22-18-50

Screenshot from 2024-08-22 22-29-36


# Part 2
Upgraded to ```1.7.0```, deleted the two below service objects manually, since they don't appear to be in 1.7.0
```longhorn-engine-manager``` and  ```longhorn-replica-manager```
Git Delta on my Repo: https://github.com/ArthurVardevanyan/HomeLab/commit/e3efac38c3bf0854e955f5d3737cfe33d764db5f
Screenshot from 2024-08-23 10-44-40
Screenshot from 2024-08-23 10-45-12

# Part 3
Then I did the manual steps to force a 4.15->4.16 Upgrade, (again, not generally recommended yet)

Final
OKD ```4.16.0-0.okd-scos-2024-08-21-155613```
Longhorn ```1.7.0```
Screenshot from 2024-08-23 12-24-06

Screenshot from 2024-08-23 12-19-11

derekbit:
@ArthurVardevanyan 
The doc is a bit outdated. We can close the PR. You can create a new one if you have time. WDYT?

ArthurVardevanyan:
> @ArthurVardevanyan The doc is a bit outdated. We can close the PR. You can create a new one if you have time. WDYT?

Sounds good, will circle back to it.

Issue #7070, opened 2023-11-09
[FEATURE] Support periodic or on-demand full backups to enhance backup reliability


derekbit:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

In the existing Longhorn backup system, the initial backup is a full backup, while subsequent backups are incremental. If any block becomes corrupted, all backup revisions relying on that block will also be corrupted as well. An approach to address the issue might perform a full backup after every N incremental backups. This method can decreases the likelihood of backup corruption, enhancing the overall reliability of the backup process.

Current implementation
![image](https://github.com/longhorn/longhorn/assets/12527233/38aa9fd1-2dd5-4175-9460-8941c548eecc)

A possible solution
![image](https://github.com/longhorn/longhorn/assets/12527233/3b4350b1-7d4a-4812-ad39-358279242d71)

Ref: https://www.architecting.it/blog/incrementals-forever-or-synthetic-fulls/

derekbit:
I would highlight the feature that can help improve the resilience to the silent corruption of a backup server. 
cc @innobead

derekbit:
The current implementation is adding a `backup-mode` label within a recurring job. User must establish two separate recurring jobs for incremental and full backups to regulate the frequency of full backups. This method complicates management.

However, following discussions with @ChanYiLin and @c3y1huan, we can record the frequency, period, or count within recurring job or backup volume, and the solution could simplify the configuration, requiring only a single recurring job.

chriscchien:
Verified **pass** on longhorn master(longhorn `e5a1b5`) with test [steps](https://github.com/longhorn/longhorn/issues/7070#issuecomment-2008928608)

Waiting [doc](https://github.com/longhorn/website/pull/933) and UI [PR](https://github.com/longhorn/longhorn/issues/8291) mergerd then close this ticket.

ChanYiLin:
Hi @chriscchien 
both doc and UI PR are merged.
Thanks

derekbit:
Updated in https://github.com/longhorn/website/pull/951

PR #8186, opened 2024-03-14
feat(lep): add recurring and manual full backup support

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7070

Add a proposal of **recurring and manual full backup support**

### DIFF
--- a/enhancements/20240314-recurring-and-manual-full-backup-support.md
+++ b/enhancements/20240314-recurring-and-manual-full-backup-support.md
@@ -1,0 +1,221 @@
+# 20240314-recurring-and-manual-full-backup-support
+
+## Summary
+
+This feature enables Longhorn to create **recurring job** for full backup or **manually trigger** the full backup of the volume.
+
+### Related Issues
+
+- Community issue: https://github.com/longhorn/longhorn/issues/7069
+- Improvement issue: https://github.com/longhorn/longhorn/issues/7070
+
+## Motivation
+
+Longhorn always does incremental backup which only backup the newly updated blocks.
+There is a chance that the previous backup blocks on the backupstore are corrupted. In this case, users can not restore the volume anymore because Longhorn aborts the restoration when it finds those blocks have different checksum.
+
+### Goals
+
+- Add a new fields `Parameters` to `RecurringJob` and `Backup`
+  - `backup-mode`: used in `Backup` CR to trigger the full backup (Options: `"full"`, `"incremental"`, default to `"incremental"` for always incremental)
+  - `full-backup-interval`: used in `RecurringJob - Backup Type` to execute full backup every N incremental backups (default to 0 for always incremental)
+- When doing full backup, Longhorn will backup **all the current blocks** of the volume and **overwrite them** on the backupstore even if those blocks already exists on the backupstore.
+- Collect metrics of `newly upload data size` and `overwritten data size` for user to better understand the cost.
+
+## Proposal
+
+### User Stories
+
+### User Experience In Detail
+
+#### Recurring Full Backup - Always
+
+1. Create a `Backup` task type RecurringJob with the parameter `full-backup-interval: 0` and assign it to the volume
+```
+apiVersion: longhorn.io/v1beta2
+kind: RecurringJob
+metadata:
+  name: recurring-full-backup-per-min
+  namespace: longhorn-system
+spec:
+  concurrency: 1
+  cron: '* * * * *'
+  groups: []
+  labels: {}
+  parameters:
+    full-backup-interval: 0
+  name: recurring-full-backup-per-min
+  retain: 0
+  task: backup
+```
+2. The RecurringJob runs and fully backup the volume every time.
+
+#### Recurring Full Backup - Every N Incremental Backups
+
+1. Create a `Backup` task type RecurringJob with the label `full-backup-interval: 5` and assign it to the volume
+```
+apiVersion: longhorn.io/v1beta2
+kind: RecurringJob
+metadata:
+  name: recurring-full-backup-per-min
+  namespace: longhorn-system
+spec:
+  concurrency: 1
+  cron: '* * * * *'
+  groups: []
+  labels: {}
+  parameters:
+    full-backup-interval: 5
+  name: recurring-full-backup-per-min
+  retain: 0
+  task: backup
+```
+2. The RecurringJob runs and fully backup the volume every 5 incremental backups.
+
+#### Manual Full Backup
+
+1. When creating backup, users can check the checkbox `Full Backup: []` or add the parameters to the spec `backup-mode: full`.
+2. The backup will be full backup.
+3. Maybe adjust the UI to make the process more simple 
+
+## Design
+
+### Implementation Overview
+
+#### Metrics
+
+1. Add two new fields `new upload data size`, `reupload data size` to the Backup Status
+
+#### UI
+
+1. In **Volume Page** >> **Create Backup** , add a checkbox `Full Backup: []`
+  - If it is checked, automatically add the parameters `backup-mode: full` to the request payload
+  - For example:
+    ```
+    HTTP/1.1 POST /v1/volumes/${VOLUME_NAME}?action=snapshotBackup
+    Host: localhost:8080
+    Accept: application/json
+    Content-Type: application/json
+    Content-Length: 55
+
+    {
+      "parameters": {
+        "backup-mode": "full",
+      },
+      "name": ${BACKUP_NAME},
+    }
+    ```
+
+2. In **Recurring Jo** >> **Create Recurring Job**, add a new sector for user to fill in the parameters when the task is `Backup` related task.
+  - Currently only support:
+    - `full-backup-interval`
+
+3. In **Backup** >> **${BackVOlume}**, add a new field `Backup Mode`
+  - If it has the parameters `backup-mode: full`, show `full`, otherwise show `incremental`
+
+#### CRD
+
+1. **BackupVolume**: add a new status `.Status.BackupCount` to record how many backups have been created.
+
+2. **Backup**: add a new fields `parameters` to pass the backup options.
+  - `backup-mode`: `"full"` to trigger full backup. Default to `"incremental"` for incremental backup
+
+3. **RecurringJob**: add a new fields `parameters`.
+  - `full-backup-interval`: Only used in `Backup` related task. Execute full backup every N incremental backups. Default to 0 for always incremental
+
+Backup CR Example
+```yaml
+apiVersion: longhorn.io/v1beta2
+kind: Backup
+metadata:
+  name: backup-abcde1234
+  namespace: longhorn-system
+spec:
+  snapshot: fake-snapshot
+  parameters:
+    backup-mode: full
+```
+
+#### Backupstore
+1. Need to pass `parameters` through the grpc function call chain.
+2. In our implementation, if the Volume has `lastBackup`, we then always perform incremental Backup.
+3. Now, if `backup-mode: full` exists in the parameters,
+    - we then pretend the last Backup does not exist and force it to do the full Backup.
+    - overwrites the block on the backupstore even it already exists.
+4. store the `new upload data size`, `reupload data size` to the Backup Status.
+
+#### Webhook
+1. Check the parameters to prevent from typo.
+2. ReucrringJob currently only accept `full-backup-interval`
+3. Backup currently only accept `backup-mode`
+
+### Test plan
+
+#### Manually Full Backup
+1. Create a Volume 4MB and fill in the content.
+2. Create a Backup of the Volume.
+3. Intentionally replace the content of the first block(2MB) on the backupstore
+4. Restore the Volume, and will get error logs like below
+    ```
+    [pvc-XXXXXX] time="XXXX" level=error msg="Backup data restore Error Found in Server[gzip: invalid checksum]"
+    ```
+5. Create a full backup with the parameter `backup-mode: full`
+6. Restore the backup, this time should work
+
+#### Recurring Job Full Backup
+1. Create a Volume 4MB and fill in the content.
+2. Create a Backup of the Volume.
+3. Intentionally replace the content of the first block(2MB) on the backupstore
+4. Restore the Volume, and will get error logs like below
+    ```
+    [pvc-XXXXXX] time="XXXX" level=error msg="Backup data restore Error Found in Server[gzip: invalid checksum]"
+    ```
+5. Create a `backup` task type RecurringJob with the parameter `full-backup-interval: 1` and assign it to the volume
+```
+apiVersion: longhorn.io/v1beta2
+kind: RecurringJob
+metadata:
+  name: recurring-full-backup-per-min
+  namespace: longhorn-system
+spec:
+  concurrency: 1
+  cron: '* * * * *'
+  groups: []
+  labels: {}
+  parameters:
+    full-backup-interval: 1
+  name: recurring-full-backup-per-min
+  retain: 0
+  task: backup
+```
+6. Wait for the recurring job to be finished.
+7. Restore the backup, this time should work
+
+
+#### Concurrent Backup
+
+1. Create a Volume 4MB and fill in the content.
+2. Create 3 recurring job for every 1 min, two for normal incremental backup and the other one for full backup
+3. These 3 recurring job should be triggered at once.
+4. Wait for 3 backup to be finished.
+5. Restore the last backup of the BackupVolume.
+6. The content should be the same as original Volume.
+
+#### DR Volume with Full Backup
+1. Setup a backupstore.
+2. Create a pod with a volume and wait for pod to start.
+3. Write data to `/data/test1` inside the pod.
+4. Create the 1st backup for the volume.
+5. Create a DR volume based on the backup and wait for the init restoration complete.
+6. Write more data to the original volume and get the md5sum.
+7. Create a full backup recurring job.
+8. Wait for the full backup job to be finished.
+9. Activate the DR volume and check the md5sum is the same as the original volume.
+
+### Upgrade strategy
+
+No need.
+
+## Note [optional]
+
+None.


enhancements/20240314-recurring-and-manual-full-backup-support.md
```diff
@@ -0,0 +1,124 @@
...
+
+1. When creating backup, user can select the checkbox `full backup`.
+2. UI will then automatically add the label`longhorn-backup-mode: full` to the backup create request.
+3. The backup will be full backup.
+
+
+#### CRD
+
+1. **RecurringJob**: add a new `RecurringJobType: full-backup`.
+2. **Backup**: add a new reserved Longhorn label `longhorn-backup-mode: full`
```
innobead, line 72: Have you considered adding this as a new field into Backup CRD?

enhancements/20240314-recurring-and-manual-full-backup-support.md
```diff
@@ -0,0 +1,124 @@
...
+    - When the Backup CR has such label, it will perform full backup.
+    - Using label so we can filter Backup by the label to distinguish the full backup and normal backup easily.
```
innobead, line 74: Can we still leverage the field selector to achieve the same purpose if using a field?

https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/

enhancements/20240314-recurring-and-manual-full-backup-support.md
```diff
@@ -0,0 +1,124 @@
...
+    - We already store the label to the backupstore when doing backup. Thus, when we pull the Backup from the backupstore in a new cluster, the label will be pulled as well.
+    - This label is only used when the backup happens and tell the engine/replica to do the full backup.
+    - Since we already pass `Label` through the grpc call chain from `longhorn-manager`->`longhorn-instance-manager`->`longhorn-engine/replica`->`backupstore`. So we don't need to do much modification.
```
innobead, line 77: This is a sweet extension point.

enhancements/20240314-recurring-and-manual-full-backup-support.md
```diff
@@ -0,0 +1,124 @@
...
+    longhorn-backup-mode: full
+```
+
+#### Backupstore
+0. In our implementation, if the Volume has `lastBackup`, we then always perform incremental Backup.
+1. Now, if `longhorn-backup-mode: full` exists in the label,
+    - we then pretend the last Backup does not exist and force it to do the full Backup.
+    - overwrites the block on the backupstore even it already exists.
+
+### Test plan
```
innobead, line 98: Suggest adding a regression test to ensure incremental backup of concurrent recurring jobs should still work, the same as DR volume, the incremental restore should work correctly.

innobead (changes requested):
In general, LG. Just a few comments. 

BTW, we should consider adding some cases to ensure our existing use cases around back, restore and DR should work as expected.

enhancements/20240314-recurring-and-manual-full-backup-support.md
innobead: @shuo-wu @c3y1huang @ejweber @mantissahz any idea? 

We should better determine when it's appropriate to use CR fields or labels.

ChanYiLin: This is kind of like passing parameters or options to the Backup execution

If considering the future extension
I don't think adding **fields** to the CRD is a proper way, cause we already pass a lot of request parameters down the grpc call chain, we then need to change the grpc interface every time we want to pass new options.
And that will introduce grpc interface breaking change.
https://github.com/longhorn/longhorn-manager/blob/v1.6.0/engineapi/backup_monitor.go#L85-L87
```
		_, replicaAddress, err := engineClientProxy.SnapshotBackup(engine, backup.Spec.SnapshotName, backup.Name,
			backupTargetClient.URL, volume.Spec.BackingImage, biChecksum, string(compressionMethod), concurrentLimit, storageClassName,
			backup.Spec.Labels, backupTargetClient.Credential)
```

If we really don't want to add this kind of Backup options to the label 
We at least should add a general `Options map[string]string` or `Parameters map[string]string` for all this kind of Backup Options.

ChanYiLin: In addition, when adding parameters
We will have to update the UI in 
1. `Create Backup` in Volume Page
2. `Create Backup Task Recurring Job`
to add place for user to add parameters.

innobead: This is more like a general question. 

If adding a field would cause breaking changes, then we need to think about how we design the current gRPC interface, because they should be compatible always if could. gRPC or protobuf can be compatibly designed.

The code you shared is about our `wrapper client` instead of the gRPC client, the real client should be below and the payload will be wrapped in `ProxyEngineRequest` which should be backward compatible even if adding any new field there. That said, for a new Longhorn manager with a full backup support client, when it calls an older version IM, the full backup will not take effect but delta backup as usual.

[/pkg/client/proxy_backup.go#L42-L62](https://github.com/longhorn/longhorn-instance-manager/blob/d670b03db8d2445296a19f6e737c283f1ed7798d/pkg/client/proxy_backup.go#L42-L62)
```go
	req := &rpc.EngineSnapshotBackupRequest{
		ProxyEngineRequest: &rpc.ProxyEngineRequest{
			Address:    serviceAddress,
			EngineName: engineName,
			// nolint:all replaced with DataEngine
			BackendStoreDriver: rpc.BackendStoreDriver(driver),
			DataEngine:         rpc.DataEngine(driver),
			VolumeName:         volumeName,
		},
		Envs:                 envs,
		BackupName:           backupName,
		SnapshotName:         snapshotName,
		BackupTarget:         backupTarget,
		BackingImageName:     backingImageName,
		BackingImageChecksum: backingImageChecksum,
		CompressionMethod:    compressionMethod,
		ConcurrentLimit:      int32(concurrentLimit),
		StorageClassName:     storageClassName,
		Labels:               labels,
	}
	recv, err := c.service.SnapshotBackup(getContextWithGRPCTimeout(c.ctx), req)
```

ChanYiLin: Discussed with @innobead , here is the summarize

1. If this setting is used for local cluster to reconcile the CR, we should just add the field to the CRD Spec. In this case, we don't even have to store this field as metadata in the backupstore for another cluster to pull, because it is useless in new cluster.
2. If we want this setting to be preserved in the backupstore and be pulled to another cluster because it will be used in another cluster, e.g. `longhorn.io/volume-access-mode: rwo`, we can just use `Label`.
3. To conclude, for execution options, it might only used for backup execution. But since it is useful for user to reference the options in the new cluster. We may still use `Label` to store the execution options.

Note:
- But use key/value map like `Labels` or `Parameters` is not a good practice cause that will confuse the user what is the correct **"key"** of each setting and need to look up the doc. Need to keep in mind.

innobead (approved):
LGTM

enhancements/20240314-recurring-and-manual-full-backup-support.md
c3y1huang: Pros:
- Labeling provides a quick way of resource search and batch operation.

Cons:
- The user needs to be cautious when specifying the label, as a minor typo could lead to incorrect configurations. For instance, a small mistake might cause a recurring job to run incremental backups instead of full backups, potentially resulting in data loss without the user realizing it.

Is there any way to make this obvious for the user to know that a full backup has been taken? Perhaps through the status?

ejweber: > The user needs to be cautious when specifying the label, as a minor typo could lead to incorrect configurations.

It looks like we do not currently validate labels (e.g. in the webhook). Perhaps we can add label validation? Probably we cannot attempt to validate all labels, since the user may add arbitrary labels to objects. But, if we detect a label with the prefix `longhorn.io`, we can ensure the key is known. For many (like the one we are discussing), we can also ensure the value it known.

ChanYiLin: Sure I can add a status to indicate if this is a full backup
If `longhron.io/backup-mode: full` exists in the label, we then mark the `status.full = True`

ChanYiLin: Conclusion:

1. Use label instead of spec.XXX to store options
    - We still want to store these options as future metadata and show them in another cluster when restoring.
2. Since we use label to store options and these options are all with prefix `longhorn.io`. We should add validation to check those label with `longhron.io` are the options we know.
    - In this case, we don't need to add status, because we ensure the options is correct in the begining.

enhancements/20240314-recurring-and-manual-full-backup-support.md
```diff
@@ -0,0 +1,177 @@
...
+
+Longhorn always does incremental backup which only backup the newly updated blocks.
+There is a chance that the previous backup blocks on the backupstore are corrupted. In this case, users can not restore the volume anymore because Longhorn aborts the restoration when it finds those blocks have different checksum.
+
+### Goals
+
+- Add a new preserved label `longhorn.io/backup-mode: full` for the Backup CR to trigger the full backup
+- Add a new preserved label `longhorn.io/full-backup-interval` to execute full backup every N incremental backups.
+- With this label `longhorn.io/backup-mode: full`, user can create **recurring full backup job** or **manually trigger** the full backup of the volume
+- When doing full backup, Longhorn will backup **all the current blocks** of the volume and **overwrite them** on the backupstore even if those blocks already exists on the backupstore.
```
derekbit, line 22: The main purpose of the labels is to reduce the number of parameters. Is it correct?

enhancements/20240314-recurring-and-manual-full-backup-support.md
```diff
@@ -0,0 +1,177 @@
...
+
+- Community issue: https://github.com/longhorn/longhorn/issues/7069
+- Improvement issue: https://github.com/longhorn/longhorn/issues/7070
+
+## Motivation
+
+Longhorn always does incremental backup which only backup the newly updated blocks.
+There is a chance that the previous backup blocks on the backupstore are corrupted. In this case, users can not restore the volume anymore because Longhorn aborts the restoration when it finds those blocks have different checksum.
+
+### Goals
```
derekbit, line 17: Collect the metrics such as number of new blocks and number of overridden blocks in a full backup. 
The purpose is to help users evaluatee the cost growth and adjust the interval.

enhancements/20240314-recurring-and-manual-full-backup-support.md
```diff
@@ -0,0 +1,177 @@
...
+  retain: 0
+  task: backup
+```
+2. The RecurringJob runs and fully backup the volume every 5 incremental backups.
+
+#### Manual Full Backup
+
+1. When creating backup, users can fill in the label `longhorn.io/backup-mode: full`.
+2. The backup will be full backup.
+3. Maybe adjust the UI to make the process more simple
```
derekbit, line 76: Make the full and incremental backup visualized.
For example, add `full` or `incremental` tag in the backup chain.

derekbit (commented):
Will the full backup introduce impacts to DR volumes?

enhancements/20240314-recurring-and-manual-full-backup-support.md
ChanYiLin: But RecurringJob is shared with different job type, e.g. Backup, Snapshot and Trim
`full-backup-interval` will only be used in Backup

I think the concept here is more like BackingImage or StorageClass parameters (`map[string]string`)
With different type of BackingImage, it accept different set of Options.
```
sourceType: export-from-volume
sourceParameters:
  volume-name: vol-export-src
  export-type: qcow2
```

derekbit: I prefer the `spec.parameters` as wll.

enhancements/20240314-recurring-and-manual-full-backup-support.md
ChanYiLin: Discussed with @derekbit 
Since "new blocks and number of overridden blocks in a full backup" is actually the whole volume size
user can simply check the volume size for this case

ChanYiLin:
> Will the full backup introduce impacts to DR volumes?

No, I don't think so
Full Backup is just like Incremental Backup
The difference is only when doing Backup, Full Backup will pretend there is no lastBackup to collect all the blocks of the Volume.
From the metadata perspective, it is still a Backup and it will become the lastBackup after completion. 
So the DR Volume will pull the lastBackup and do incremental restore in the new cluster

enhancements/20240314-recurring-and-manual-full-backup-support.md
ChanYiLin: Yeah
But last time I discussed with @innobead ,we got the conclusion that both `Label` and `Parameters` is `map[string]string` so there is no much differences.

enhancements/20240314-recurring-and-manual-full-backup-support.md
ChanYiLin: Added

enhancements/20240314-recurring-and-manual-full-backup-support.md
ChanYiLin: Added.

I found we need to store the information in new fields in Backup CRD
because the backup is a async call, so we can't get the number in the grpc call but can only get it by syncing the Backup metadata.

enhancements/20240314-recurring-and-manual-full-backup-support.md
derekbit: I see. Tried to clarify the purposes of labels and spec.parameters.
According to the [kubernetes document](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/)

```
Labels are intended to be used to specify identifying attributes of objects that are meaningful and relevant to
users, but do not directly imply semantics to the core system. Labels can be used to organize and to select 
subsets of objects.
```

Labels in Kubernetes are designed for recording attributes and metadata that describe resources. They're not intended for controlling the behavior of a resource, which is the role of the spec.

`full-backup-interval` is a parameter that controls the behavior of a backup recurring job, so it'd better to store in the spec.parameters.

@innobead WDYT?

derekbit (approved):
LGTM

enhancements/20240314-recurring-and-manual-full-backup-support.md
```diff
@@ -0,0 +1,211 @@
...
+metadata:
+  name: recurring-full-backup-per-min
+  namespace: longhorn-system
+spec:
+  concurrency: 1
+  cron: '* * * * *'
+  groups: []
+  labels: {}
+  parameters:
+    full-backup-interval: 0
```
derekbit, line 46: Following our discussion with @shuo-wu, we recommend replacing `parameters map[string]string` with `parameters map[JobType]map[string]string`.
`JobType` could include values like `Backup`, `Snapshot` and so on. This modification ensures that parameters for different types are clearly classified.

enhancements/20240314-recurring-and-manual-full-backup-support.md
shuo-wu: As discussed with @derekbit, I think it's better to have a spec field like this:
```
type Parameters struct {
	Backup *BackupParameters `json:"backup,omitempty"`

	Snapshot *SnapshotParameters `json:"snapshot,omitempty"`

	Trim *TrimParameters `json:"trim,omitempty"`
}
```

I prefer not to rely on the labeling too much for the decision-making. As mentioned in the k8s doc, the label is mainly for identifying or filtering objects. Besides, there is no restrictive convention against the Longhorn reserved label key format and usage. Even developers like me, need to check the reference doc before using labels to operate something. It's unfriendly for users.

BTW, the idea is from [SPDK Bdev](https://github.com/longhorn/go-spdk-helper/blob/d05637da59788eeeb8927c663f0eec7c958d3f97/pkg/spdk/types/bdev.go#L124).

innobead: LGTM @shuo-wu @derekbit above comments.

enhancements/20240314-recurring-and-manual-full-backup-support.md
ChanYiLin: We have `spec.task = backup` in RecurringJob CR
So we don't need to change the parameters structure.

enhancements/20240314-recurring-and-manual-full-backup-support.md
```diff
@@ -0,0 +1,211 @@
...
+2. In **Recurring Jo** >> **Create Recurring Job**, add a new sector for user to fill in the parameters when the task is `Backup` related task.
+  - Currently only support:
+    - `full-backup-interval`
+
+3. In **Backup** >> **${BackVOlume}**, add a new field `Backup Mode`
+  - If it has the parameters `backup-mode: full`, show `full`, otherwise show `incremental`
+
+#### CRD
+
+1. **BackupVolume**: add a new status `.Status.BackupCount` to record how many backups have been created.
```
innobead, line 118: What's the purpose of this new status field?

enhancements/20240314-recurring-and-manual-full-backup-support.md
```diff
@@ -0,0 +1,211 @@
...
+    backup-mode: full
+```
+
+#### Backupstore
+1. Need to pass `parameters` through the grpc function call chain.
+2. In our implementation, if the Volume has `lastBackup`, we then always perform incremental Backup.
+3. Now, if `backup-mode: full` exists in the parameters,
+    - we then pretend the last Backup does not exist and force it to do the full Backup.
+    - overwrites the block on the backupstore even it already exists.
+4. store the `new upload data size`, `reupload data size` to the Backup Status.
```
innobead, line 145: Does `new upload data size` mean delta uploaded data? `reupload ata size` means full backup?

If yes, suggest the following naming. WDYT?
- `incrementa-backup/data-size`
- `full-backup/data-size`

enhancements/20240314-recurring-and-manual-full-backup-support.md
```diff
@@ -0,0 +1,211 @@
...
+
+#### Webhook
+1. Check the parameters to prevent from typo.
+2. ReucrringJob currently only accept `full-backup-interval`
+3. Backup currently only accept `backup-mode`
+
+### Test plan
```
innobead, line 152: Need to add a test case for DR. 

Recurring backup with a full backup period, then see if DR volume can be activated with correct data.

innobead (commented):
In general, LGTM. Just a few feedback.

enhancements/20240314-recurring-and-manual-full-backup-support.md
ChanYiLin: not exactly
one can keep doing full-backup but each time with new data block
the block is not exactly from `incrementa-backup/data-size`

enhancements/20240314-recurring-and-manual-full-backup-support.md
ChanYiLin: sure

enhancements/20240314-recurring-and-manual-full-backup-support.md
innobead: Does backup time mean the number of `full backup`? When will it be reset?

ChanYiLin: It starts when the BackupVolume is created and is stored on the backupstore
it meas how many times this Volume has backed up no matter if it is full or incremental

The parameter is `full-backup-interval: 5` in recurring backup job.
So every 5 times backup, it will do full backup
It used the backupCount in BackupVolume to check.

enhancements/20240314-recurring-and-manual-full-backup-support.md
```diff
@@ -0,0 +1,211 @@
...
+## Motivation
+
+Longhorn always does incremental backup which only backup the newly updated blocks.
+There is a chance that the previous backup blocks on the backupstore are corrupted. In this case, users can not restore the volume anymore because Longhorn aborts the restoration when it finds those blocks have different checksum.
+
+### Goals
+
+- Add a new fields `Parameters` to `RecurringJob` and `Backup`
+  - `backup-mode`: used in `Backup` CR to trigger the full backup (Options: `"full"`, `"incremental"`, default to `"incremental"` for always incremental)
+  - `full-backup-interval`: used in `RecurringJob - Backup Type` to execute full backup every N incremental backups (default to 0 for always incremental)
```
shuo-wu, line 21: NIT: It's better to provide an example in the description: If N is 5, it means that after 5 regular incremental backups, the 6th backup will be the full one.

enhancements/20240314-recurring-and-manual-full-backup-support.md
```diff
@@ -0,0 +1,211 @@
...
+- When doing full backup, Longhorn will backup **all the current blocks** of the volume and **overwrite them** on the backupstore even if those blocks already exists on the backupstore.
+- Collect metrics of `newly upload data size` and `overwritten data size` for user to better understand the cost.
```
shuo-wu, line 23: Can Longhorn collect this info if it's not recorded in Backup CR?

shuo-wu (approved):
In general LGTM

enhancements/20240314-recurring-and-manual-full-backup-support.md
ChanYiLin: the request is from @derekbit 
it is to help user understand the real data transfer amount in this backup
it count when we actually "upload" the block

so even for `incremental backup` => if the block already on the backupstore, we won't upload it, thus it is not in `new upload data size`
so for `full backup` => if there is new block needs to upload, it is in `new upload data size`

enhancements/20240314-recurring-and-manual-full-backup-support.md
```diff
@@ -0,0 +1,211 @@
...
+Backup CR Example
+```yaml
+apiVersion: longhorn.io/v1beta2
+kind: Backup
+metadata:
+  name: backup-abcde1234
+  namespace: longhorn-system
+spec:
+  snapshot: fake-snapshot
+  parameters:
```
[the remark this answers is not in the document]
ChanYiLin: you agree on @derekbit and @shuo-wu above and now change again?
Should we discussed with @derekbit and @shuo-wu agiain

enhancements/20240314-recurring-and-manual-full-backup-support.md
```diff
@@ -0,0 +1,211 @@
...
+7. Restore the backup, this time should work
+
+
+#### Concurrent Backup
+
+1. Create a Volume 4MB and fill in the content.
+2. Create 3 recurring job for every 1 min, two for normal incremental backup and the other one for full backup
+3. These 3 recurring job should be triggered at once.
+4. Wait for 3 backup to be finished.
+5. Restore the last backup of the BackupVolume.
```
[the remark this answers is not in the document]
ChanYiLin: The purpose of this test is that you wonder if concurrent backup restore work (lock issue)
But I found we actually have test for that: https://github.com/longhorn/longhorn-tests/blob/v1.5.1/manager/integration/tests/test_basic.py#L3186

And we also already have incremenatal  backup and restore e2e tests
So I don't quite understand what you want to achieve in this test?

enhancements/20240314-recurring-and-manual-full-backup-support.md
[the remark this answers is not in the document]
ChanYiLin: I have discussed with @derekbit 
He will provide conclusion we have.

enhancements/20240314-recurring-and-manual-full-backup-support.md
```diff
@@ -0,0 +1,211 @@
...
+  name: backup-abcde1234
+  namespace: longhorn-system
+spec:
+  snapshot: fake-snapshot
+  parameters:
+    backup-mode: full
+```
+
+#### Backupstore
+1. Need to pass `parameters` through the grpc function call chain.
```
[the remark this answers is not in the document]
ChanYiLin: I have discussed with @derekbit 
He will provide conclusion we have.

enhancements/20240314-recurring-and-manual-full-backup-support.md
ChanYiLin: it is not for the metrics.
it is for recurring job to check how many backup has been executed 
so it will perform full backup next time.

for example
```
kind: RecurringJob
metadata:
  name: recurring-full-backup-per-min
  namespace: longhorn-system
spec:
  concurrency: 1
  cron: '* * * * *'
  groups: []
  labels: {}
  parameters:
    full-backup-interval: 5
  name: recurring-full-backup-per-min
  retain: 0
  task: backup
```
`full-backup-interval: 5` meaning every 5 incremental backup, we execute 1 full backup
then how do we know how many backup we executed (since backup might be deleted by users.)
we have to record the number in backupvolume 
so the job can get the value and see if it needs to perform full backup.

enhancements/20240314-recurring-and-manual-full-backup-support.md
ChanYiLin: added

enhancements/20240314-recurring-and-manual-full-backup-support.md
ChanYiLin: added the example, thanks

enhancements/20240314-recurring-and-manual-full-backup-support.md
ChanYiLin: we can only know if this is newly upload data or overwritten data
when it checks if the block on the backupstore or not during upload process (and it is concurrent process) in backupstore library.

I think it is quite hard for Longhorn to collect the information in real time.

enhancements/20240314-recurring-and-manual-full-backup-support.md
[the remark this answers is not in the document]
derekbit: After discussing with @ChanYiLin, we both agree with

- Using `spec.parameters` is better for passing them to the longhorn-engine via gRPC without additional changes to the gRPC interface, but it makes the attributes in CRD implicit.

- If we adopt `spec.mode`, we need to update the gRPC interface and append parameters. This extra effort is inevitable.

To avoid the drawbacks of both methods, we propose the following solution:

- Use `spec.mode` in the backup CRD.
- Convert the spec fields to a `parameters` map and pass them to the longhorn-engine via gRPC. For example:
   ```
  parameters := getBackupParameters(spec, ...)
  err := SnapshotBackup(..., parameters)
   ```

This approach requires updating only the CRD and the `getBackupParameters` function. The gRPC interface remains unchanged when introducing a new parameter.


WDYT? @innobead @ChanYiLin

enhancements/20240314-recurring-and-manual-full-backup-support.md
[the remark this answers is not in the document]
derekbit: After discussing with @ChanYiLin, we both agree with

- Using `spec.parameters` is better for passing them to the longhorn-engine via gRPC without additional changes to the gRPC interface, but it makes the attributes in CRD implicit.

- If we adopt `spec.mode`, we need to update the gRPC interface and append parameters. This extra effort is inevitable.

To avoid the drawbacks of both methods, we propose the following solution:

- Use `spec.mode` in the backup CRD.
- Convert the spec fields to a `parameters` map and pass them to the longhorn-engine via gRPC. For example:
   ```
  parameters := getBackupParameters(spec, ...)
  err := SnapshotBackup(..., parameters)
   ```

This approach requires updating only the CRD and the `getBackupParameters` function. The gRPC interface remains unchanged when introducing a new parameter.


WDYT? @innobead @ChanYiLin

enhancements/20240314-recurring-and-manual-full-backup-support.md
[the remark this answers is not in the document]
ChanYiLin: Discussed,concurrent backup restored already covered here: https://github.com/longhorn/longhorn-tests/blob/v1.5.1/manager/integration/tests/test_basic.py#L3186
If need to imvprove can create another ticket for it

enhancements/20240314-recurring-and-manual-full-backup-support.md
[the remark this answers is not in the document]
ChanYiLin: Disscuess @innobead , we are good with the proposal

enhancements/20240314-recurring-and-manual-full-backup-support.md
innobead: `newly upload data size`: new data upload size for delta backup
`overwritten data size`: overwritten data size for full backup.

innobead (approved):

enhancements/20240314-recurring-and-manual-full-backup-support.md
ChanYiLin: - `newly upload data size`: the data size uploaded to the backupstore in this backup (skip data due to already exists on the backupstore won't count)
  - `overwritten data size`: the data size uploaded to the backupstore and overwritten the exists block on the backupstore.

Issue #8198, opened 2024-03-18
[BUG] Can't use longhorn with Generic ephemeral volumes


mueller-tobias:
## Describe the bug

We try to rollout open project via it's helm chart. The helm chart is utilizing [Generic ephemeral volumes](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes). According to the kubernetes documentation this should work out of the box with existing storage providers.

But when i try the example the pod will not be created. The volume is created correctly in longhorn and also the attach is correct, but the setup of the pod fails.

```
MountVolume.SetUp failed for volume "pvc-60ea3b05-bd1b-40ae-b647-4898fc37c3a5" : rpc error: code = Aborted desc = no Pending workload pods for volume pvc-60ea3b05-bd1b-40ae-b647-4898fc37c3a5 to be mounted: map[]
```

## To Reproduce

Use the example from the Kubernetes Documentation to rollout a pod with a generic ephermal volume:

```yaml
kind: Pod
apiVersion: v1
metadata:
  name: my-app
spec:
  containers:
    - name: my-frontend
      image: busybox:1.28
      volumeMounts:
      - mountPath: "/scratch"
        name: scratch-volume
      command: [ "sleep", "1000000" ]
  volumes:
    - name: scratch-volume
      ephemeral:
        volumeClaimTemplate:
          metadata:
            labels:
              type: my-frontend-volume
          spec:
            accessModes: [ "ReadWriteOnce" ]
            storageClassName: "longhorn"
            resources:
              requests:
                storage: 1Gi
```

## Expected behavior

The pod should be created and running.

## Support bundle for troubleshooting

[supportbundle_108a26af-f02f-4163-8c3f-a9bf48c65937_2024-03-18T09-36-04Z.zip](https://github.com/longhorn/longhorn/files/14633999/supportbundle_108a26af-f02f-4163-8c3f-a9bf48c65937_2024-03-18T09-36-04Z.zip)

## Environment

 - Longhorn version: 1.6.0.
 - Impacted volume (PV): pvc-255d1fe4-61ad-4aa2-8062-54b3b4575c60
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl): Helm
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: RKE2 v1.27.10 +rke2r1
   - Number of control plane nodes in the cluster: 5 (3 etcd, 2 cp)
[...]

PR #8200, opened 2024-03-18
Add an example that uses a generic ephemeral volume

ejweber:
Longhorn 8198

Contributes to longhorn/longhorn#8198.

We don't have an example that deploys a workload with a Longhorn generic ephemeral volume. This example can help us quickly experiment with and test generic ephemeral volume functionality. It just modifies `examples/pod_with_pvc.yaml` so that it uses a generic ephemeral volume instead of an explicit PVC.

### DIFF
--- a/examples/pod_with_gev.yaml
+++ b/examples/pod_with_gev.yaml
@@ -1,0 +1,34 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: volume-test
+  namespace: default
+spec:
+  restartPolicy: Always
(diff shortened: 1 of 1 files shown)


james-munson (approved):
LGTM.  Would it merit an instance in docs references/examples?

ejweber:
Good thinking, @james-munson. I forgot about that page.

ejweber:

ejweber:
@mergify backport v1.4.x

Issue #7740, opened 2024-01-20
[CI] Prevent members' created tickets from deleting


innobead:
## What's the CI task to develop? Please describe

Update stale setting, https://github.com/longhorn/{repo}/blob/master/.github/workflows/stale.yaml to prevent deleting stale members' created tickets.

PR #8249, opened 2024-03-25
ci: avoid closing stale issue/pr

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/7740

avoid closing stale issue/pr

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -18,8 +18,7 @@ jobs:
         close-pr-message: 'This PR was closed because it has been stalled for 10 days with no activity.'
         days-before-stale: 60
         days-before-pr-stale: 45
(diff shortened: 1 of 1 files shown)


innobead (changes requested):
@yangchiu the change will not close the issues created by non-members as well. 

We can have a pre-step to filter the issues created by non-members only before doing stale issues cleanup.

yangchiu:
> We can have a pre-step to filter the issues created by non-members only before doing stale issues cleanup.

It's a cron job batch processing all issues instead of triggered by a single issue. The workflow has no idea about what issues it will deal with. The `actions/stale` action itself gets all issue to process. The logic to filter out issues could need to be customized and put inside `actions/stale` action. More investigation needed.

Issue #6205, opened 2023-06-27
[FEATURE] Share manager HA - Experimental


innobead:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

Longhorn share manager is the backend to serve RWX volume by using NFS ganesha. Because it's a single instance, it would be SPOF, even though it has been improved with the recovery backend mechanism at https://github.com/longhorn/longhorn/issues/2293. 

The goal is to make the share manager highly available to improve availability instead of just relying on a shorter recovery time which would be uncertain, really depending on different environmental factors. 

## Additional context

Before 1.4.4 and 1.5.2, there are some Kernel issues that could cause the volume node to get stuck during the node reboot or upgrade if the share manager pod is disconnected because we use hard mode NFS mount. Thus, to resolve this problem, the soft mode will be introduced back with a longer timeout to prevent this situation in 1.4.4, 1.5.2, and 1.6. The detailed context can be checked at https://github.com/longhorn/longhorn/issues/6655#issuecomment-1726948794. However, this could be a potential risk of data loss if the timeout is not well defined. (timeout should at least consider the pod eviction timeout)

Eventually, the hard mode will be readopted together with this feature. Still, it doesn't mean the node stuck situation will not be encountered, but it's just a very rare case at least only if share manager HA nodes are all down or pods all unavailable to lose the HA. 

cc @longhorn/dev-data-plane

james-munson:
Possibly of help: https://github.com/nfs-ganesha/nfs-ganesha/wiki/NFS-Ganesha-and-High-Availability

derekbit:
@james-munson Could you update the status of the feature implementation? Thank you.

james-munson:
There are several items needed before this is ready for release.
- Port the webhook selector changes into the branch and ensure we can turn off the webhook on the node we don't trust.
- Find the remaining paths in which the re-attachment to the new share-manager still tries to communicate with that node, and pre-empt them so they won't just time out.
- If possible, do so only for any share managers with expired leases rather than all on that node, in case there are some that are not expired.
It is unlikely to be done by 1.7.0 release.

derekbit:
> There are several items needed before this is ready for release.
> 
> * Port the webhook selector changes into the branch and ensure we can turn off the webhook on the node we don't trust.
> * Find the remaining paths in which the re-attachment to the new share-manager still tries to communicate with that node, and pre-empt them so they won't just time out.
[...]

Issue #8184, opened 2024-03-14
[FEATURE] Support storage network for RWX volumes


c3y1huang:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

Previously, when the storage network was introduced, the RWX volume feature was experimental. Now that this feature has graduated from its experimental status, Longhorn should also support the storage network for RWX volumes.

## Describe the solution you'd like

- Implement storage networking for RWX volumes.

## Describe alternatives you've considered

`None`

## Additional context

`None`

c3y1huang:
Last month, Mutus decided to archive the mutlus-service repository. So, we need to explore how the share-manager can operate, with or without Kubernetes service using the secondary network interface created by Multus.

- https://github.com/k8snetworkplumbingwg/multus-service-archived
- https://github.com/k8snetworkplumbingwg/multus-cni/issues/466

c3y1huang:
> Last month, Mutus decided to archive the mutlus-service repository. So, we need to explore how the share-manager can operate, with or without Kubernetes service using the secondary network interface created by Multus.

Exploring the possibility with EndpointSlice. (WIP)

c3y1huang:
> Exploring the possibility with [EndpointSlices](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#:~:text=In%20Kubernetes%2C%20an%20EndpointSlice%20contains,that%20match%20the%20Service%20selector.).

We can leverage custom Kubernetes Endpoint, headless Service, and service FQDN to integrate storage-network with RWX volumes.

Below is the PoC supporting this approach.
```
> k -n longhorn-system get ep
NAME                                       ENDPOINTS                                         AGE
longhorn-engine-manager                    <none>                                            28m
longhorn-replica-manager                   <none>                                            28m
longhorn-frontend                          10.42.1.4:8000,10.42.2.5:8000                     28m
longhorn-backend                           10.42.0.16:9500,10.42.1.15:9500,10.42.2.15:9500   28m
longhorn-recovery-backend                  10.42.0.16:9503,10.42.1.15:9503,10.42.2.15:9503   28m
longhorn-admission-webhook                 10.42.0.16:9502,10.42.1.15:9502,10.42.2.15:9502   28m
longhorn-conversion-webhook                10.42.0.16:9501,10.42.1.15:9501,10.42.2.15:9501   28m
pvc-93213f2d-2d76-458a-ab68-485987ea60eb   192.168.3.4:2049                                  44s

[...]

Issue #5717, opened 2023-04-07
[IMPROVEMENT] Improve manageability of share manager pod behaviors


derekbit:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

The improvement is a subtask of https://github.com/longhorn/longhorn/issues/5658.
In the current implementation, the share manager pod is created and blindly mounts a volume regardless of its healthy state. The blind mounting behavior can lead to unexpected volume faulted if the volume is dead or orphaned. This ticket aims to share manager pod's behaviors more manageable and predictable.

With the improvement, the start flow will be as follows:
1. Share manager controller creates a share manager pod and bootstraps an NFS server without any filesystem exported
2. Share manager pod mounts the volume and exports it after receiving a `FilesystemMount` request
3. Share manager controller checks the mount status by `FilesystemMountStatus`. If the volume is mounted, transition the share manager state to "running"

Shutdown flow will be as follows:
1. Gracefully shut down the NFS server
2. Gracefully or forcefully unmount the volume
3. Tear down the device.

derekbit:
Will revisit the issue after introducing AD controller.

PhanLe1010:
Hi @derekbit Will this volume mounting issue be resolve if we utilize the PV/PVC native mechanism of Kubernetes instead of developing the logic by ourselves? i.e., create a PV/PVC for share-manager pod and let Kubernetes handling the attach/mount logic?

Issue #4863, opened 2022-11-12
[FEATURE] no way to control share-manager pod placement (node tags, etc)


jaysonsantos:
## To Reproduce

1. Tag a few nodes with `local` tag
2. Create a RWX (RWO works fine) using the `local` node tag
3. Create a deployment that uses a pvc with that node
4. share-manager's pod will be created in a random node, not respecting the tag

## Expected behavior

I would assume that share-manager would only be scheduled on nodes that match those tags, maybe some config on the pvc would be possible?
## Environment

 - Longhorn version: 1.3.2
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl): Kubectl
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: k3s v1.24.7+k3s1
   - Number of management node in the cluster: 3
   - Number of worker node in the cluster: 8
 - Node config
   - OS type and version: ubuntu 22.04
   - CPU per node: 6-16
   - Memory per node: 2-24GB
   - Disk type(e.g. SSD/NVMe): HDD and NVME
   - Network bandwidth between the nodes: 1Gb
 - Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal): Baremetal
 - Number of Longhorn volumes in the cluster: 20

netwarex:
Yes please, this is important as even no data locality respected by the share-manager pod.

tcoupin:
It would be nice to have a parameter such as defaultDataLocality to keep the share manager pod close to one of the replicas.
At least, a share manage pod should not be schedule on node without disk.

netwarex:
For a quick solution, it would be nice, to at least make it possible to specify the nodeAffinity so only run share manager on nodes where defined.

A more sophisticated way is enough later.

93imagine:
I assumed the data locality option would take care of this but it does not.

Danielss89:
Would love to see this implemented too. I currently have a cluster with nodes in different regions like Tokyo, USA, Europe and even though i can put my application pods and storage on the correct nodes, the share manager will run on wherever there is room.
This gives an error because of latency, and as such, if i'm unlucky and my pods in europe gets a share-manager in Tokyo, it will fail to mount the volume.

derekbit:
@innobead 
I think it is worth being added in v1.7.0 if we have enough time.

innobead:
@james-munson Please take care of the following when you start working on that. 

[...]

PR #8252, opened 2024-03-25
feature(rwx): fast failover for share-manager.

james-munson:

Issue # https://github.com/longhorn/longhorn/issues/6205


enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,216 @@
...
+
+    Which longhorn-manager is the right one to be calling this?  The one on the owning node of the pod is a poor choice, since it may also be out of commission.  Picking another one is difficult, but it isn't necessary.  Let each one do the ping.  That's not all that much network traffic for N nodes X M rwx volumes.  Whichever node reaches the threshold first is the one that pulls the trigger.
+
+    Add a new "Suspicion" Condition on the Node CR, which defaults to "false".  Along with deleting the pod, the deciding controller will set that condition to true with reason "share-manager ping failed".  That will be useful for avoiding the same node if it has not reached NotReady state when a share-manager controller is choosing a new node.  The condition is time-stamped and is cleared when the node goes NotReady (we were right) or after some interval we decide on (at least a minute).  A node with that condition is not considered for share-manager pod creation.
```
shuo-wu, line 65: Some questions for this part:
1. If everything works fine, there will be M x N ping requests and responses in each second. For example, when there are 50 nodes & 10 RWX volumes on each, each node should send 500 ping requests every second. Will this be a heavy overhead?
2. Can you describe more about the new "Suspicion" Condition on the Node CR? If the client on node A realizes that a SM pod on node B is unreachable, which node CR will be updated? (I guess it's node B. But how does node A notify node B? Typically a node controller is responsible for updating its own CR IIRC.) If multiple RWX volumes on the same node reach the threshold, how do we update the condition for the node CR?

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,216 @@
...
+
+    No changes.
+    
+
+### Risks and Mitigation
+
+- **Rescheduling on same node**
+The replacement for a terminated share-manager pod will need to avoid being scheduled on the same node.  The node itself might not yet have transitioned to NotReady and thus appear to be schedulable according to Kubernetes.  To work around that, we add a "Suspicion" Condition on the node housing a failed share-manager pod and add that as another scheduling criterion.
+
+- **Refusal to relinquish mount on dead pod.**
```
shuo-wu, line 77: Do you mean somehow the SM pod is dead before we mount the RWO volume?

Or do you mean that some workload pods may go die before the new NFS server starts the recovery?

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,216 @@
...
+Hopefully this will not be a problem.  If testing shows that it is, it may be necessary to terminate an unresponsive pod more forcefully.
+
+- **Possible delay due to image pull on the successor node**
+With a pull policy of `IfNotPresent` this should not occur more than once per node.  If it appears to be a problem, Longhorn can add a pod to pre-pull, but it seems unlikely.
+
+- **False positives**
```
shuo-wu, line 83: Will there be another false positive:
1. There is a connection/network issue between node A and node B.
2. Node A or Node B can talk with the k8s control panel (API server).
3. Node A cannot talk with the NFS server on node B and reaches the threshold. => The condition is updated.

In this case, we don't need to restart the SM pod on node B. Restarting the workload pod on node A is probably a better choice.

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,216 @@
...
+
+- **nfs-ganesha (user-space NFS server)**
+
+    No changes.
+    
+
+### Risks and Mitigation
+
+- **Rescheduling on same node**
+The replacement for a terminated share-manager pod will need to avoid being scheduled on the same node.  The node itself might not yet have transitioned to NotReady and thus appear to be schedulable according to Kubernetes.  To work around that, we add a "Suspicion" Condition on the node housing a failed share-manager pod and add that as another scheduling criterion.
```
ejweber, line 75: It looks like today the share-manager controller simply creates a pod with a node selector such that it can run on any Longhorn node.

https://github.com/longhorn/longhorn-manager/blob/3f99a66d07a5fc15081b24fc885941b7d0b49cbc/controller/share_manager_controller.go#L752-L755

What specifically will we change about the manifest to ensure it doesn't schedule to the suspicious node?

enhancements/20240325-nfs-server-ha.md
ejweber, line 83: I was thinking this ping feature might be problematic when storage network was enabled. For example, if NFS traffic is over the storage network, temporary failure to ping the share-manager pod on the management network may not be good evidence that it needs to be restarted.

However, it looks like we currently don't use the storage network for NFS traffic: https://github.com/longhorn/longhorn/issues/8184. Maybe we can keep this feature request in the back of our minds as we move forward.

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,216 @@
...
+
+- **Refusal to relinquish mount on dead pod.**
+Hopefully this will not be a problem.  If testing shows that it is, it may be necessary to terminate an unresponsive pod more forcefully.
```
ejweber, line 78: If the node is actually down, I think deleting the pod WITHOUT force will be ineffective (it will be stuck terminating). And deleting it WITH force won't get rid of the Kubernetes volume attachment. I'm not 100% sure how Longhorn will respond without testing, but I think it will refuse to attach the underlying volume to a new node until Kubernetes tries to detach it from the old node. And Kubernetes will not try to detach it from the old node until the force detach interval is exceeded. If this is the case, the new mechanism may also need to delete the Kubernetes volume attachment.

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,216 @@
...
+        
+        `( exec 7<>/data/testfile-${i}; flock -x 7; while date | dd conv=fsync >&7 ; do sleep 1; done )`
+        
+        where ${i} is the node number.
+        
+        Turn off or restart the node where share-manager is running. Once the share-manager pod is recreated on a different node, check
+        
+        - Expect
+            - In the client side, IO to the RWX volume will hang until a share-manager pod replacement is successfully created on another node.
+            - During the outage, the server rejects READ and WRITE operations and non-reclaim locking requests (i.e., other LOCK and OPEN operations) with an error of NFS4ERR_GRACE.
```
DamiaSan, line 105: If the server rejects READ and WRITE operations, how can IO to the RWX volume hang?

enhancements/20240325-nfs-server-ha.md
james-munson: For 1, it's true that the network traffic scales up quickly, although I don't think even 1000 requests on the network each second is all that large a load.  But it might be reasonable to pick a subset of nodes to issue the pings.  One way might be to use only the nodes that have RWX consumer workload pods on them.  They would be the most interested parties, anyway.
As for 2, it is node B, the node on which the unreachable pods is running, that would have the Suspicion condition set on it.  If there is a rule that only the node controller on that node can update its node CR, then that does complicate things.  In this model, the condition would be set by some other node, and only cleared by the target node, B.  All it would take is one volume's share-manager being unresponsive, on the theory that if the node is actually dying, all the share-managers running on it will behave the same.
The Suspicion condition is not a request to kill the node itself.  It's just a mechanism we hope to use to prevent rescheduling the replacement share-manager right back on the same node, which is not yet officially unschedulable.

enhancements/20240325-nfs-server-ha.md
james-munson: This is about the work necessary for Kubernetes to reuse the mount and attachments still associated with the pod we just killed for being unresponsive.  See Eric's note below.

enhancements/20240325-nfs-server-ha.md
james-munson: As the client, the workload pod actually has all the lock information.  Offhand, I would think that restarting it would increase the possibility of an I/O corruption.
But the main question is, how does Longhorn tell the difference between network partitioning and imminent node death?  That is, is there a better health-checking solution that can know in this case within, say, 5 seconds that B is fine, A should move?

enhancements/20240325-nfs-server-ha.md
james-munson: I think we will need to increase the sophistication of the node selection for share-managers anyway, for tickets such as 
[[IMPROVEMENT] Improve manageability of share manager pod behaviors](https://github.com/longhorn/longhorn/issues/5717)
[[FEATURE] no way to control share-manager pod placement (node tags, etc)](https://github.com/longhorn/longhorn/issues/4863)
[[FEATURE] Advanced Scheduling for Share Manager](https://github.com/longhorn/longhorn/issues/2335)

Those all seem like we might invent a specific selector for share-manager pods than SystemManaged.  But then to make that work, I suppose we might have to edit node labels as we wish to include or exclude them from the set to be considered.  This is a good point, and I'll need to work on it some more.

enhancements/20240325-nfs-server-ha.md
james-munson: Yes, that seems likely.  We are trying to stampede Kubernetes faster than it is inclined to go.
If we cannot achieve the forced deletion and recreation of whatever we need to, then the only other option is to allow multiple simultaneous mount and attachment, with both the outgoing and the replacement share-manager pod.  That is, fall back to the Active/Passive approach with a share-manager deployment or statefulset, and limit all traffic to the leader.

enhancements/20240325-nfs-server-ha.md
james-munson: Sorry, language cribbed from the previous LEP.  The workload will hang in that it can't progress while I/O is rejected.  I/O commands don't hang, they are failed, and the workload app is expected to deal with that.

enhancements/20240325-nfs-server-ha.md
shuo-wu: > If there is a rule that only the node controller on that node can update its node CR, then that does complicate things, especially if node B really is dead or dying.

There is no restriction rule for the CR update. And YES, the controllers on other nodes could take over the CRs on the dying node.

> Even if there are multiple share-manager pods on that node, all it would take is one volume's share-manager being unresponsive, on the theory that if the node is actually dying, all the share-managers running on it will behave the same.
The Suspicion condition is not a request to kill the node itself. It's just a mechanism we hope to use to prevent rescheduling the replacement share-manager right back on the same node, which is not yet officially unschedulable.

The question is, is the condition applied to (the rescheduling of) that problematic SM pod, or all SM pods on the node? I would assume it prevents all SM pods from being scheduled on this node even if there is only a single pod failure.

enhancements/20240325-nfs-server-ha.md
shuo-wu: Briefly say, there is probably something wrong with the server side if the majority of the clients/workload pods cannot issue IOs. Otherwise, we should check or restart the minority problematic clients. 
It's like a leader election mechanism. Considering a cluster could have multiple RWX volumes and probably thousands of clients at most, I am not sure if it's practicable in the actual implementation.

enhancements/20240325-nfs-server-ha.md
james-munson: Yes, that's true if implemented this way.  The node would be avoided for any SM pod scheduling until it either fails and returns, or stays up without failing for some period, either of which should clear the condition.

enhancements/20240325-nfs-server-ha.md
james-munson: The point is taken that Conditions are more of a reporting mechanism, and don't really play a role in scheduling calculations.  The most effective way would be to make suspicion a "Don't Schedule" taint on the suspected failing node before creating the replacement share-manager pod.  That would have an impact on other workloads in the cluster, though.

enhancements/20240325-nfs-server-ha.md
james-munson: Suspicion would have to be a taint to be effective for scheduling, not a condition.  But the revised proposal does not need it, since the successor pod already exists on another node.

enhancements/20240325-nfs-server-ha.md
james-munson: The revised proposal makes this moot.  An SM pod is considered failed if it can't update the leadership lease.

enhancements/20240325-nfs-server-ha.md
james-munson: Marking this as resolved, since it is superseded by the revised proposal.

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,227 @@
...
+    Enable a readiness check of leadership status.  If not the leader, don't come ready.  Use https://stackoverflow.com/questions/74644647/how-to-send-traffic-to-only-one-pod-in-a-deployment as a starting point.
+
+- **longhorn-manager**
+    
+    Changes to:
+    - engine_controller
+        - Check and revise engine cleanup.
+    - setting_controller
+        - Take SM pods out of restart lists for updates, since update of Deployment will restart pods.
+    - share_manager_controller
```
shuo-wu, line 68: Need to mention how these controllers (ShareManageController, VolumeAttachmentController, VolumeController, EngineController) manually migrates the volume attachment & mount from the unavailable pod to the new leader

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,227 @@
...
+
+Implement a fast failover mechanism that is not gated by Kubernetes' relatively slow detection and handling of node failure.  Recovery should be a matter of seconds rather than minutes.  Ideally, it would not require a restart of workload pods.
+
+### Non-goals
+
+Duplicate NFS servers providing Active/Active redundancy.
+
+## Proposal
+
+  1. Change the share-manager NFS server from a single pod to a deployment of two pods.  Only one can be serving data at any given time, but the other can have have the volume mounted and be ready to take over on short notice.
```
shuo-wu, line 29: In the worst case, if the nodes the both pods reside on crash, how can we handle the recovery? Do we need to fall back to the previous version of the NFS HA solution?

enhancements/20240325-nfs-server-ha.md
james-munson: Superseded by changes to the LEP.

james-munson:
Revised and re-pushed.

PhanLe1010 (commented):
In general LGTM.

There might be some question about the lease renewal period (currently set to 3s) but we can discuss later in the implementation detail phase

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,238 @@
...
+## Proposal
+
+Fast failover needs several pieces.
+
+1. Quick detection of failure
+    - Use the Kubernetes `Lease` mechanism.  When an RWX volume is created, create the lease at the same time as the share-manager pod and service.  It should have the same lifetime as the service.
+    - Update the lease periodically from ShareManager Run().
+2. Quick removal of the failed/failing server
+    - Check the lease from share-manager controller.  If it is expired, mark the share-manager CR as error and let it delete the pod.
+    - May require force delete, if kubelet not responsive.
```
ejweber, line 36: If we force delete the share-manager pod, it may be removed from the Kubernetes API, but still be running. Is this dangerous?

ejweber: I think no, assuming Kubernetes won't route traffic to it after it's force deleted?

ejweber (commented):
There is still something I'm struggling to understand. Let's assume the node running share-manager is actually down. This improvement allows us to detect it quite quickly and delete (potentially with force) the share-manager pod. Then, I guess the attach/detach controller needs to detach the underlying Longhorn volume from the down node and attach it wherever the new share-manager is. But if the node is down, won't the detach be stuck? (And if the node is NOT down, it SHOULD be stuck.) Can you check whether we need to describe some additional action taken by the attach/detach controller to allow for the quick failover?

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,238 @@
...
+
+- **Possible delay due to image pull on the successor node**
+The user should avoid a pull policy of `Always`.  If there are multiple RWX volumes, the failover node may already have the image.  There is also a ticket in development to pre-pull a variety of system-managed pod images: [[IMPROVEMENT] Pre-pull images (share-manager image and instance-manager image) on each Longhorn node](https://github.com/longhorn/longhorn/issues/8376).
+
+- **Insufficient resources on any successor node**
+We rely on Kubernetes scheduling to pick an appropriate successor node that meets all the usual constraints, if one exists.  If the pod can't be scheduled, there is little to be done immediately but to wait for the failed node to return.  For the user, the solution might be to add resources or decrease the cluster load in order to allow space for failover scheduling.  This design does not schedule the successor in advance (see Active/Passive below).
```
innobead, line 103: It seems we still focus on minimizing the failover time. Is it possible to offer the following options to users instead?

- Managed Lease, to reduce failover time
    - The new share manager pod wouldn't be able to start due to resource constraints.
- Lease + standby share manager pod (ensure availability)
    - This will ensure availability, but the downside is that it requires extra resource allocation for the standby pod.
    - No secondary engine, as the pod will be actually running when the volume is reattached to the node where it's running. The main purpose is to ensure the pod can be created earlier to ensure availability.

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,238 @@
...
+
+- **False positives**
+The initial choice for timeout interval may prove to be too sensitive.  That is addressed by using a configurable setting.  Increasing the timeout will make it less likely to restart share-manager unnecessarily, but also increase the time to recovery.
```
innobead, line 106: What's the benefit to use K8s lease instead of our own managed configurable timeout? Need extra monitoring cost for timeout calculation?

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,238 @@
...
+        - Expect
+            - The clients continue the tasks after the server's failover without IO or stale handle errors.
+            - Lock reclaim process can be finished earlier than the 90-second configured grace period.
+            - Outage should be equivalent to simple pod failover, less than **TBD** seconds.
+
+### Upgrade Strategy
+
+The only impact on upgrade is the creation of a new setting to define and default:  `share-manager-stale-timeout` (default, 5).
+
+The logic for creation and checking of Lease records ensures that all parties know and implement the lease strategy.  If some component doesn't, the behavior defaults to the same as current code.
```
innobead, line 160: Does this mean the existing share manager pods will be lease-managed after upgrade w/o restarting?

innobead (commented):
I like the idea of lease, but still have a few comments about the story of HA.

Reducing failover time is still a goal for us, but still need to think of ensuring availbility if it's feasible.

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,238 @@
...
+2. Lease creation and update.
+3. Add setting for stale timeout.
+4. Lease expiration checking.
+
+
+### Risks and Mitigation
+
+- **Rescheduling on same node**
+The replacement for a terminated share-manager pod will need to avoid being scheduled on the same node.  The node itself might not yet have transitioned to NotReady.  The expectation is that Kubernetes would notice a failure to respond during creation, but if it does not try to talk to the target node when scheduling, relying only on recorded state, then this might happen.  The size of the delay and a way to get around it if needed are **TBD**.
+
```
james-munson, line 98: Without intervention, Kubernetes does use the same node.  Overcome by using a node anti-affinity against the former owner when building the replacement pod manifest.

enhancements/20240325-nfs-server-ha.md
james-munson: I'm not sure what is meant by a "managed lease."

A pre-created standby pod would make it less likely for failover to be held up by image pull or resource constraints, I agree.  I think we can prevent traffic to it on the service IP by ensuring it reports itself as not ready.  That would entail expanding the lease to be a full leadership election, and only the leader is allowed to respond to the readiness check.  Are you proposing that we would somehow arrange for the corresponding engine not to be started until leadership changes?

Bear in mind that either way, there will still be a backend recovery grace period either way.

enhancements/20240325-nfs-server-ha.md
james-munson: Without a K8s lease, what would the managed configurable timeout be on?  Can you expand on this a little?

enhancements/20240325-nfs-server-ha.md
james-munson: No, the pod would have to be restarted into the new share-manager image for lease management to work.  It requires both the new longhorn-manager and share-manager to cooperate.

Old longhorn-manager + new share-manager:  No lease is created, so the share-manager just skips the update code.

New longhorn-manager + old share-manager:  Lease is created, but never claimed by a holder or renewed to advance the expiration time.  Staleness check sees that there is no holder, so the lease is never expired.  Failover is only based on reported K8s node state.

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,238 @@
...
+    - Update the lease periodically from ShareManager Run().
+2. Quick removal of the failed/failing server
+    - Check the lease from share-manager controller.  If it is expired, mark the share-manager CR as error and let it delete the pod.
+    - May require force delete, if kubelet not responsive.
+3. Quick creation of a replacement
+    - This is already part of the Share Manager lifecycle.
+    - But avoid former host.  Kubelet will not yet report it as down.
+4. Avoid client workload pod restart, if possible.
+    - Longhorn currently restarts the workload pods in volume controller when the share-manager pod restarts.  That may be necessary if the NFS mount is "hard", because the client can hang indefinitely, but not if it is "soft".
+    - Longhorn defaults to "softerr", but the user can override NFS mount options, so it can't be assumed.  Volume controller can Look at the storage class parameters or PV volumeAttributes to decide whether a restart is necessary..
```
derekbit, line 42: Is `softerr` inappropriate for fast failover?

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,288 @@
...
+
+- **False positives**
+The initial choice for timeout interval may prove to be too sensitive.  That is addressed by using a configurable setting.  Increasing the timeout will make it less likely to restart share-manager unnecessarily, but also increase the time to recovery.  
+One open question is whether delays in lease renewal and false positives are more likely in a heavily loaded cluster.  That can probably only be resolved with copious testing, including high-load cases.
+
+- **Experimental feature**
+If desired, the timeout setting can be set at a high value, such as 600 seconds, to guarantee that normal Kubernetes node-failure handling will take place first.  The (slight) overhead for Lease creation and update will still remain.  If RWX HA is released as an "experimental" feature, the high value can be the default.
+
+- **Webhook availability**
```
james-munson, line 151: Added this section to "Risks".

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,288 @@
...
+        - Delete the pod.  
+        - If the node is down *OR* expired, do a force delete as well.  If it truly is down, this is the same as current code.  If it is not, then the first delete should have succeeded, but either way, even if the pod is still running, this will ensure that Kubernetes does not route service traffic to it.  Also, deletion allows re-use of the same standard pod name, avoiding one potential upgrade complication.
+    - Other resources need to change ownership similarly.  Modify the usual rules to capture our suspicion that the expired node is dead:
+        - Add a Condition named "Delinquent" to longhorn.node.io, set to "False" by default.  A node is set to Delinquent=true when a share manager lease for a pod on that node expires.  It is cleared by the node itself in node_controller when it starts operation again.
+        - Change controller_manager's `IsNodeDownOrDeleted` to check `IsNodeDownOrDeletedOrDelinquent` for other resources, as well as any resource-specific `isResponsibleFor` calls that don't use the base implementation.  (Note: this is probably too intrusive.  See [Risks and Mitigation](#risks-and-mitigation).)
```
derekbit, line 59: After checking your longhorn-manager implementation, longhorn-manager is still using `IsNodeDownOrDeleted`. Do you plan to use `IsNodeDownOrDeletedOrDelinquent`?

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,288 @@
...
+    - For resources that map one-one with RWX volumes, use the resource controller's `isResponsibleFor()` method to check first whether the related share manager ownership has moved, and if so, change to match it.  That can work for the volume, volume attachment, and engine controllers.
+    - For engine and replica instances, the handling depends on the instance manager's state, which in turn is controlled by its node's `DownOrDeleted` status.  If the instance manager is allowed to remain in "running" state, then every piece of code that checks instance manager state would need to have an added check of node delinquent condition, at least for instances that belong to RWX volumes.
+
+- **Possible delay due to image pull on the successor node**
+The user should avoid a pull policy of `Always`.  If there are multiple RWX volumes, the failover node may already have the image.  There is also a ticket in development to pre-pull a variety of system-managed pod images: [[IMPROVEMENT] Pre-pull images (share-manager image and instance-manager image) on each Longhorn node](https://github.com/longhorn/longhorn/issues/8376).
```
derekbit, line 139: We don't have pull policy anymore. Need to update this.
cc @mantissahz

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,288 @@
...
+
+    - node_controller  
+    Add code to set `Delinquent` condition to `false` at startup.
+
+    - datastore/longhorn  
+    Revise `IsNodeDownOrDeleted` to `IsNodeDownOrDeletedOrDelinquent` (or add a new function) and check the node's condition.
+    Also add the condition to `IsNodeSchedulable` (which is only called by share_manager_controller, and affects unmount/remount).
+
+    - controller_manager  
+    Revise `isControllerResponsibleFor` to use `IsNodeDownOrDeletedOrDelinquent`.  Also make that change in any controller's `isResponsibleFor` code that does more than use the base call.
```
derekbit, line 112: Same question and concern as above.

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,288 @@
...
+The main goal is to minimize downtime on RWX workloads.
+
+In the current code, failure of a container in the share-manager pod or the pod itself is quickly handled.  It takes about 20 seconds to restore write access, of which about half is spent restarting the workload pods and letting them remount.  Longhorn's recovery backend has the ability to shorten the grace period wait when all clients are accounted for.
+
+The big problem happens when the node with the share-manager pod fails or restarts.  Just the time it takes for the node to go `NotReady` - in Kubernetes, typically about 45 seconds - is already too long an outage for share-manager failover.  We intend to make the NFS client recovery time for a failure of the node running the share-manager pod to be just as quick as for simple pod failure.
+
+### Goals
+
+- Implement a fast failover mechanism that is not gated by Kubernetes' relatively slow detection and handling of node failure.  Recovery should be a matter of seconds rather than minutes.  
+- Depending on mount type, it should not require a restart of workload pods.
```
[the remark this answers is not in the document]
james-munson: Done. This refers to the NFS mount options.

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,288 @@
...
+- Operation of the recovery backend will be unchanged.
+- After the failure of the share-manager pod, the application on the client side’s IO operations will be stuck until the share-manager and NFS are recreated, but the interruption will be brief, within a specified limit **TBD** regardless of the cause.
```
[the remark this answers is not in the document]
james-munson: Done.  I've called out a goal of 20 seconds.

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,288 @@
...
+- There will be only a single interruption for a failure.
```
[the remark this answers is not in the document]
james-munson: Done.  This was more confusing than helpful, so I've removed it.

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,288 @@
...
+    - If the lease is not expired, move on to normal handling.  
+    - If it has, set the share-manager CR `Status.OwnerID` to this node, if another node hasn't already done so.  (We can tell by comparing the owner field to the lease holder.)  The new owner will act as an interim manager to do tear-down and recreation of the share manager pod.  It updates the CR and, if successful, then returns "true".
+    - The new owner is logged, and then the reconcile can continue.  Mark the share-manager CR as error state and let it clean up the pod.  
+    - In cleanup,  
+        - Remember the old lease holder.  
+        - Clear the lease.  
+        - Delete the pod.  
+        - If the node is down *OR* expired, do a force delete as well.  If it truly is down, this is the same as current code.  If it is not, then the first delete should have succeeded, but either way, even if the pod is still running, this will ensure that Kubernetes does not route service traffic to it.  Also, deletion allows re-use of the same standard pod name, avoiding one potential upgrade complication.
```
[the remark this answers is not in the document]
james-munson: Done.  Delete the share-manager pod.

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,288 @@
...
+    - Previous releases of Longhorn used "hard" NFS mounts as the default (and only) option.  That is likely why there is code to restart workloads unconditionally when the share manager pod is recreated.  Currently, Longhorn defaults to "softerr", but the user can override NFS mount options, so it can't be assumed.  Volume controller can Look at the storage class parameters or PV volumeAttributes to decide whether a restart is necessary.
+    - This would also be affected by the `node-down-pod-deletion-policy` handler in kubernetes_pod_controller.
+
+5. Shorten the recovery grace period  
+Speak for the dead node's client, if any.  Without this step, workers may still be blocked for an NFS grace period after re-connection is possible, waiting for other clients to re-establish state with the new server.  The grace period is usually configured to 90 seconds.  If all workload pods are alive, it does not take the full grace period, but if a workload pod was on the failed node, it will not be able to clear its state.  Knowing its node ID, Longhorn should be able to send a recovery message on its behalf, avoiding the grace period timeout.
+
```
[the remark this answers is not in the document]
james-munson: I would like it to, but that is not implemented yet.

enhancements/20240325-nfs-server-ha.md
james-munson: So I am pushing up a revision that uses IsNodeDownOrDeletedOrDelinquent everywhere that it can determine that the question is being asked on behalf of an RWX volume.
The problem is that there are some cases, such as for checking the state of the instance manager, where that might not be known.  Does that make sense?

enhancements/20240325-nfs-server-ha.md
james-munson: Done.

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,288 @@
...
+    Add logic to `isResponsibleFor` to check for staleness first and take over ownership.
+    Create or set the node's `Delinquent` condition to `true` and event it.
+    Add staleness check to `cleanupShareManagerPod()` when deciding to force-delete.
+    Add an anti-affinity to pod manifest creation to avoid the previous owner node.
+
+    - node_controller  
+    Add code to set `Delinquent` condition to `false` at startup.
+
+    - datastore/longhorn  
+    Revise `IsNodeDownOrDeleted` to `IsNodeDownOrDeletedOrDelinquent` (or add a new function) and check the node's condition.
```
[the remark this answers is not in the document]
james-munson: Done.

enhancements/20240325-nfs-server-ha.md
james-munson: Revised again to use the RWX volume itself, so that the check can be scoped to the volume, not all RWX volumes.

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,286 @@
...
+
+2. Quick removal of the failed/failing server  
+If the lease is expired, we infer that the pod's node is dead and another controller needs to take over.  All other nodes will detect that.  Longhorn needs to pick one, and establish that choice so that multiple controllers don't all attempt to drive the recovery.  Longhorn will use the controller's `isResponsibleFor` method:  
+    - If the lease is not expired, move on to normal handling.  
+    - If it has, set the share-manager CR `Status.OwnerID` to this node, if another node hasn't already done so.  (We can tell by comparing the owner field to the lease holder.)  The new owner will act as an interim manager to do tear-down and recreation of the share manager pod.  It updates the CR and, if successful, then returns "true".
+    - The new owner is logged, and then the reconcile can continue.  Mark the share-manager CR as error state and let it clean up the pod.  
+    - In cleanup,  
+        - Remember the old lease holder.  
+        - Clear the lease.  
+        - Delete the pod.
```
derekbit, line 55: Will the deletion of a SM pod become the bottleneck of the fast failover feature?

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,286 @@
...
+    - In cleanup,  
+        - Remember the old lease holder.  
+        - Clear the lease.  
+        - Delete the pod.  
+        - If the node is down *OR* expired, do a force delete of the share-manager pod as well.  If it truly is down, this is the same as current code.  If it is not, then the first delete should have succeeded, but either way, even if the pod is still running, this will ensure that Kubernetes does not route service traffic to it.  Also, deletion allows reuse of the same standard pod name, avoiding one potential upgrade complication.
+    - Other resources need to change ownership similarly.  Modify the usual rules to capture our suspicion that the expired node is dead:
+        - Keep track of the delinquent state in the volume's lease object.  The otherwise informational Spec.AcquireTime field can be used for that.
+        - Change `IsNodeDownOrDeleted` to check `IsNodeDownOrDeletedOrDelinquent` for other volume-related resources, as well as any resource-specific `isResponsibleFor` calls that don't use the base implementation.
```
derekbit, line 59: Is `IsNodeDownOrDeletedOrDelinquent` only applied for the RWX volume-related resources?

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,286 @@
...
+    - Other resources need to change ownership similarly.  Modify the usual rules to capture our suspicion that the expired node is dead:
+        - Keep track of the delinquent state in the volume's lease object.  The otherwise informational Spec.AcquireTime field can be used for that.
+        - Change `IsNodeDownOrDeleted` to check `IsNodeDownOrDeletedOrDelinquent` for other volume-related resources, as well as any resource-specific `isResponsibleFor` calls that don't use the base implementation.  
+
+3. Quick creation of a replacement  
+This is already a normal part of the Share Manager lifecycle.  But it does need an adjustment to avoid the former host, which Kubernetes will try to re-schedule to since Kubelet does not yet report it as down.  
+    - Use the cached node identity, if found, to add a node anti-affinity into the existing set of selectors, tolerations, and affinities defined for share-manager pods.  Note that the node on which the pod is scheduled might not be the interim controller node, so ownership might change again.
```
derekbit, line 63: What does cached node identity mean?

derekbit: When is the anti-affinity removed? Who is responsible for the removal?

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,286 @@
...
+
+4. Avoid client workload pod restart  
+Longhorn currently restarts the workload pods in volume controller when the share-manager pod restarts.  That may be necessary if the NFS mount is "hard", because the client can hang indefinitely if the server does not respond.  Clients will not hang if mounted as "soft" or "softerr".  (In either case, the client will retry I/O attempts to the unresponsive server and eventually return an error to the calling workload application.  The only difference is in the error returned: EIO or ETIMEDOUT.)  In that case, they will eventually reconnect to the new server and resume operation without having to restart the workload pod to kill the client process.  
+    - Previous releases of Longhorn used "hard" NFS mounts as the default (and only) option.  That is likely why there is code to restart workloads unconditionally when the share manager pod is recreated.  Currently, Longhorn defaults to "softerr", but the user can override NFS mount options, so it can't be assumed.  Volume controller can Look at the storage class parameters or PV volumeAttributes to decide whether a restart is necessary.
```
derekbit, line 68: >     - Previous releases of Longhorn used "hard" NFS mounts as the default (and only) option.  That is likely why there is code to restart workloads unconditionally when the share manager pod is recreated. 

Is restsarting a workload required if the NFS is hard mounted?

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,286 @@
...
+
+    *TODO - Put a time diagram of durations and state here.*
+
+## Design
+
+### Implementation Overview
+
+- **share-manager**
+    
+    Add a constant, `share-manager-renew-interval`.  It should be often enough to detect failure quickly, but not so often that it overwhelms the network and API server with traffic.  Hard-code this to 3 seconds.
```
derekbit, line 86: Can `share-manager-renew-interval` configurable? Or, do we need to make configurable?

enhancements/20240325-nfs-server-ha.md
```diff
@@ -0,0 +1,286 @@
...
+
+    Add a goroutine to keep the lease from expiring.  Every `share-manager-renew-interval`, get the lease for the RWX volume (name it the same as the volume name) and update its `lease.spec.renewTime` to "now".  If the lease cannot be found, take no action.
+    
+- **longhorn-manager**
+    
+    - setting
+    Add a Boolean setting, `enable-share-manager-fast-failover`.  Because this is an experimental feature to begin with, default it to false.
```
derekbit, line 93: I would suggeest making the feature enabled by default. If users encounter any issues, they can disable it.

derekbit:
Need to add a diagram for the state machine in the LEP.

derekbit:
@james-munson Could you update the LEP? Thank you.

enhancements/20240325-nfs-server-ha.md
james-munson: No, it's actually very quick.

enhancements/20240325-nfs-server-ha.md
james-munson: Revised to be clearer, I hope.  The delinquent state of the node/volume pair is stored in the lease record, so when the share-manager-controller creates the pod, it can add the anti-affinity.  It is only used for the creation of that pod, and will be cleared when the pod refreshes the lease when it begins running.

enhancements/20240325-nfs-server-ha.md
james-munson: Yes, it is.

enhancements/20240325-nfs-server-ha.md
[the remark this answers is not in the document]
james-munson: The consensus of reviewers was that it was preferable to shorten the grace period in the NFS server configuration, rather than artificially cutting it short by sending an artificial recovery message for any dead workload pod.

enhancements/20240325-nfs-server-ha.md
james-munson: This is the subject of https://github.com/longhorn/longhorn/issues/9062.

enhancements/20240325-nfs-server-ha.md
james-munson: The consensus among reviewers is that this could use some soak time, and as an experimental feature, it should be off by default and opted into by users, at least in the first release.

james-munson:
Hmm.  Something went awry with the rebase.  I will close this PR and create a new one with the new version:  https://github.com/longhorn/longhorn/pull/9069

Issue #7872, opened 2024-02-06
[IMPROVEMENT] RWX volume scheduling


mazocode:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

The RWX functionality is hardly usable for us because the share-manager does not honor the storage classes nodeSelector, topology, nor any other nodeSelector configured. This means that it gets scheduled on nodes that are not suitable for this kind of workload or do not match the requirements. In addition, placing it randomly might not be beneficial to the traffic flow.

## Describe the solution you'd like

Add a nodeSelector derived from storage class topolgy configuration or nodeSelector tag. 

## Describe alternatives you've considered

So far I could not find a workaround.

## Additional context

RWX storage class example

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
provisioner: driver.longhorn.io
allowVolumeExpansion: true
allowedTopologies:
- matchLabelExpressions:
  - key: storage-provisioning/longhorn
    values:
    - ""
metadata:
  name: rep2x-rwx-hdd-fra4-b1c
parameters:
  diskSelector: hdd
  nfsOptions: vers=4.1,noresvport,softerr,timeo=600,retrans=5
  nodeSelector: fra4-b1c
  numberOfReplicas: "2"
  staleReplicaTimeout: "2880"
volumeBindingMode: WaitForFirstConsumer
reclaimPolicy: Delete
```

Neither topology nor nodeSelector tag gets honored when scheduling the share-manager pod.

mazocode:
Can we please reopen this, so that I don't feel like I waste my time here?

derekbit:
Just realized `nodeSelector` in `StorageClass.Parameters` is a bit confusing. The field is for selecting the candidates for **replace placement** rather than **engine placement**.

https://longhorn.io/docs/1.6.1/references/storage-class-parameters/

cc @shuo-wu @innobead

derekbit:
@mazocode Thanks for raising the issue. I've implemented the feature. Feel free to check the design https://github.com/longhorn/longhorn/pull/8294 and welcome any feedback.

shuo-wu:
> Just realized nodeSelector in StorageClass.Parameters is a bit confusing. The field is for selecting the candidates for replace placement rather than engine placement.

For regular volumes, the engine scheduling is determined by the workload pods hence typically `nodeSelector` has no ambiguity. But for the RWX volume, YES it's confusing. I am fine with renaming it to `replicaNodeSelector`.

[...]

Issue #2335, opened 2021-03-12
[FEATURE] Advanced Scheduling for Share Manager


awelsman:
I have a cluster with a number of nodes (EC2 instances) which, under considerable load, can die and be cycled back into the cluster by RKE. I want to run Longhorn on both those nodes and a few other, much more reliable ones. I have three nodes dedicated to Longhorn, one in each availability zone, and am storing my RWX volume replicas there. I can attach my volume to one of these dedicated nodes, but the moment I mount the volume to a deployment by using a PVC, the volume attaches to any random node Longhorn is installed on, not necessarily the one I selected through the UI.

It would be ideal if I could reliably schedule the Share Manager pod to a given dedicated node. I realize the drawbacks with high availability and NFS, but would like to avoid running the pod on any of my more unreliable nodes.

PhanLe1010:
Original discussion at https://forums.rancher.com/t/advanced-scheduling-for-share-manager/19717

awelsman:
Any updates on this?

joshimoo:
@awelsman this is definitly a good use case, to have more control over the scheduling of the internal longhorn components (share-manager, etc). Thanks for opening the enhancement, no commitment yet, but it's something to keep in mind next time we do some refactoring around the share-manager #2096

james-munson:
It seems like this is asking much the same thing as #4863
I would be happy to take this and combine them.

ChanYiLin:
Yes I just found these two issues might be the same.
I also haven't started working on this one.
cc @innobead

ChanYiLin:
Hi @james-munson 
I have transferred the ticket to you.
You can combined all related issues together.

innobead:
@derekbit is this same as what you have done for SM location scheduling? If yes, let's close this with the duplicated label. 

cc @PhanLe1010

derekbit:
Yes. It is same as https://github.com/longhorn/longhorn/issues/7872.
Let's close the ticket.

PR #8294, opened 2024-04-02
LEP: add "Share Manager Scheduling"

derekbit:

Issue Longhorn/longhorn#7872

### DIFF
--- a/enhancements/20240402-share-manager-scheduling.md
+++ b/enhancements/20240402-share-manager-scheduling.md
@@ -1,0 +1,65 @@
+# Share Manager Scheduling
+
+## Summary
+
+In the Longhorn storage system, a share manager pod of an RWX volume is created on a random node, without the ability for users to specify a preferred locality. The purpose of this feature is to enhance the locality of an RWX volume and its share manager pod.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/7872
+https://github.com/longhorn/longhorn/issues/4863
+https://github.com/longhorn/longhorn/issues/8255
+https://github.com/longhorn/longhorn/issues/2335
+
+## Motivation
+
+### Goals
+
+- Share manager pod respects the node selector specified in `storageClass.parameters["shareManagerNodeSelector"]`
+- Share manager pod complies with the affinity rules defined in `storageClass.allowedTopologies`
+- Share manager pod respects the newly introduced `storageClass.Parameters["shareManagerTolerations"]`
+
+### Non-goals [optional]
+
+`None`
+
+## Proposal
+
+### User Stories
+
+A share manager pod of an RWX volume is unable to adhere to the specified node selector and affinity rules, leading to potential inefficiencies and performance issues. The feature aims to enhance this by ensuring share manager pods can be scheduled according to the given node selector and affinity rules.
+
+### User Experience In Detail
+
+The introduction of node selector and affinity rule capabilities in a storage class significantly enhances user control over the scheduling of share manager pods for RWX volumes.
+
+- Node Selector:
+  Users can define node selectors directly within the storage class configuration by setting `storageClass.parameters["shareManagerNodeSelector"]`. When a share manager pod is to be scheduled, Kubernetes evaluates the node selectors specified by the user. The scheduler then ensures that the pod is placed only on nodes that match all of the specified labels. This mechanism provides a straightforward way to guide pod placement towards nodes that meet certain criteria, such as hardware capabilities, geographical location, or any other user-defined characteristic.
+
+- Allowed Topologies:
+  Users can set specific rules for where share manager pods should go in the cluster by using `storageClass.allowedTopologies`. This setting is turned into affinity rules which are applied to a share manager pod. The affinity helps decide which nodes the pod can be placed on, based on the labels of those nodes.
+
+- Tolerations:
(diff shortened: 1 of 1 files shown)


enhancements/20240402-share-manager-scheduling.md
```diff
@@ -0,0 +1,62 @@
+# Share Manager Scheduling
+
+## Summary
+
+In the Longhorn storage system, a share manager pod of an RWX volume is created on a random node, without the ability for users to specify a preferred locality. The purpose of this feature is to enhance the locality of an RWX volume and its share manager pod.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/7872
+
```
james-munson, line 10: Also
https://github.com/longhorn/longhorn/issues/4863
https://github.com/longhorn/longhorn/discussions/8255
https://github.com/longhorn/longhorn/issues/2335

enhancements/20240402-share-manager-scheduling.md
```diff
@@ -0,0 +1,62 @@
...
+### API changes
+
+`None`
+
+## Design
+
+### Implementation Overview
+
+When the share manager controller is in the process of reconciling a share manager, it attempts to search the associated storage class from the `persistentVolume.spec.StorageClassName`. If the storage class is nonexistent, node selectors and allowed topologies will also be absent, leading to a neglect of the share manager pod's locality.
```
james-munson, line 50: Just a note:  even if the storage class can't be found, the parameters fields are all captured in the PV spec.csi.volumeAttributes, so some matching might still be possible.

The allowedTopologies would not be, though, and we might decide that it is less confusing to do no scheduling adjustment at all rather than some but not others.

james-munson (commented):
LGTM.  Just a couple of small items.

enhancements/20240402-share-manager-scheduling.md
derekbit: You are right!
We can find parameters in the PV `spec.csi.volumeAttributes`, but `allowedTopologies` is not found in `spec.csi.volumeAttributes`.
To make the behavior unified, I currently read the settings from its storageClass.

I think we can handle them together in https://github.com/longhorn/longhorn/issues/8361. WDYT?

derekbit:
@james-munson updated the LEP and ready to review. Thank you.

james-munson (approved):
LGTM

enhancements/20240402-share-manager-scheduling.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Proposal
+
+### User Stories
+
+A share manager pod of an RWX volume is unable to adhere to the specified node selector and affinity rules, leading to potential inefficiencies and performance issues. The feature aims to enhance this by ensuring share manager pods can be scheduled according to the given node selector and affinity rules.
+
+### User Experience In Detail
+
+The introduction of node selector and affinity rule capabilities in a storage class significantly enhances user control over the scheduling of share manager pods for RWX volumes.
```
innobead, line 34: If no share manager locality parameters are specified, will we respect the corresponding parameters for replica scheduling? Such as node selector, taint, and topology as well? This can have another benefit to have the SM pod, engine and one replica together on the same node. 

Respect: setting in storage class for share manager -> settings in storage class for replica scheduling -> system managed locality settings

innobead (commented):
LGTM, just one question about locatliy setting overrding.

enhancements/20240402-share-manager-scheduling.md
```diff
@@ -0,0 +1,65 @@
...
+
+If the associated storage class is present, the system reads `storageClass.parameters["shareManagerNodeSelector"]` and `storageClass.allowedTopologies`. The node selectors specified in `storageClass.parameters["shareManagerNodeSelector"]` are combined with the global selectors from system-managed-components-node-selector and are applied to the share manager pod. Additionally, the system translates the `storageClass.allowedTopologies` into affinity rules, which are then applied to the configuration of the share manager pod as well.
+
+For tolerations, users are able to to specify these in the storage class through `storageClass.parameters["shareManagerTolerations"]`. These specified tolerations are combined with global tolerations defined under the global setting `taint-toleration`, enabling share manager pods to be allocated to nodes that have compatible taints.
+
+Once the share manager pod is allocated to a suitable node, the node's name is set to the volume attachment ticket.
+
+### Test plan
+
+1. Test RWX volumes from a storage class with `parameters["shareManagerNodeSelector"]`.
```
innobead, line 63: What's the difference between 1 and 3?

enhancements/20240402-share-manager-scheduling.md
derekbit: > If no share manager locality parameters are specified, will we respect the corresponding parameters for replica scheduling?

Yes. The share manager locality has nothing to do with the replica scheduling. It controls the share-manager pod for a RWX volume. Thus, users can apply the two localities simutaneously for put the all components on the same node.

enhancements/20240402-share-manager-scheduling.md
derekbit: `shareManagerNodeSelector` doesn't handle taint. If a node is tainted, users can use the toleration `storageClass.Parameters["shareManagerTolerations"]` to make the share-manager pod on the tainted node.

innobead (approved):

Issue #7927, opened 2024-02-15
[FEATURE] Create Longhorn CLI focusing on non CR resource operations


innobead:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

Create a Longhorn CLI focusing on non-custom-resource operations, as users can operate resources via Kubectl or similar tools, K8s user experience. However, some operations like install prerequisites, preflight check before installation/upgrade, runtime periodical check, headless support bundle download, check known issue mentioned in the knowledge pages, etc. are only partially operated through UI. 

We should provide a CLI to help with these headless operations for general functions, debugging, benchmarking, troubleshooting, etc. 

Change https://github.com/longhorn/longhorn-preflight to longhorn/cli and implement the corresponding functions.

shuo-wu:
We can leverage this CLI to guide users understand & config Longhorn systems on demands:
1. Use an interactive way to help config Longhorn nodes, disks, settings, backups, and so on.
2. Step by Step. In each step, we will explain what's for and list all options.

For example:
```
$ longhorn-cli guide

Do you need to launch v1 volumes in your clusters? v1 volumes are low-spec friendly but have limited performance.
yes/no: yes

How many CPUs you wish to reserve for the component (longhorn-instance-manger pod) that runs v1 volumes on each node?
Please input the reserved CPU percentage:

......

======

Do you need to add extra disks for the below nodes?
  1. Node 1: default disk
  2. Node 2: default disk
  3. Node 2: default disk
Please input the nodes we want to add disks: 1, 2, 3

The available disk type is:
  1. Block device (for v2 volumes)
  2. Filesystem (for v1 volumes)
Please select the disk type: 2

Please input the disk path: /var/lib/longhorn-extra

......
======

Do you have a backupstore?
  Currently, Longhorn supports:
  1. NFS. nfs://xxxxxx
  2. S3. xxx
  3. CIFS. xxx
If YES, please input the backup target: s3://xxxxx

What's the credential of the backup target?
Please input the secret name: minio-secret 
......

```

------

Besides, this CLI can handle some complicated operations such as recurring-job creation and configuration. Without this, users have to get a template YAML, check all fields and the corresponding options, modify it, create CRs, and bind it to some volumes. It's too trivial. An interactive CLI definitely helps.

derekbit:
Some functionalities of the CLI
- Installation
  - Environment check (https://github.com/longhorn/longhorn/issues/6657)
- Filesystem
  - Manual trim
[...]

PR #8309, opened 2024-04-05
Allow save-images to optionally use podman

wgnathanael:
We use podman locally instead of docker, this change allows us to run the save-images.sh script using podman instead of docker without editing the script each time.


innobead (approved):
LGTM

innobead:
@mantissahz Please help review this.

mantissahz (approved):
LGTM (Verified)
And @wgnathanael could we have a new version of `load-images.sh` as this one? That will be perfect.

Only need to modify the config file `/etc/containers/registries.conf` with `unqualified-search-registries = ["docker.io"]` for using the `podman`

wgnathanael:
Ok, Added the load-images.sh as well.

scripts/save-images.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 list="longhorn-images.txt"
+DOCKER_OR_PODMAN=${DOCKER_COMMAND:-docker}
```
derekbit, line 3: nit: Rename `DOCKER_OR_PODMAN` to `CONTAINER_CLI`

wgnathanael: applied in my latest force push

innobead:
@c3y1huang can consider to add these scripts to our CLI.

c3y1huang:
https://github.com/longhorn/longhorn/issues/7927

Issue #8209, opened 2024-03-19
[RELEASE] 1.6.1


innobead:
**What's the task? Please describe.**
Action items for releasing v<x.y.z>

**Describe the sub-tasks.**
  - Pre-Release
    - [x] Regression test plan (manual) - @roger-ryao (release captain)
    - [x] Run security testing of container images for pre-GA milestones - @roger-ryao  
    - [x] Verify longhorn chart PR to ensure all artifacts are ready for GA (`install`, `upgrade`)  @mantissahz  
    - [x] Run core testing (install, upgrade) for the GA build from the previous patch and the last patch of the previous feature release (1.5.4). - @chriscchien  
  - Release
    - [x] Release longhorn/chart from the release branch to publish to ArtifactHub 
    - [x] Release note @jillian-maroket 
	     - [x] Deprecation note
	     - [x] Upgrade notes including highlighted notes, deprecation, compatible changes, and others impacting the current users
  - Post-Release
    - [x] Update https://github.com/longhorn/longhorn/blob/master/deploy/upgrade_responder_server/chart-values.yaml @PhanLe1010 
    - [x] Add another request for the rancher charts for the next patch release (`1.6.2`) @rebeccazzzz  
  - Rancher charts: verify the chart can be installed & upgraded - @khushboo-rancher 
    - [x] rancher/image-mirrors update @mantissahz @PhanLe1010 
    - [x] rancher/charts 2.7 & 2.8 branches for rancher marketplace @mantissahz @PhanLe1010 
    
cc @longhorn/qa @longhorn/dev

innobead:
@roger-ryao The QA release captain needs to ensure the QA items in the description have been updated eventually. 

cc @longhorn/qa

roger-ryao:
Run security testing of container images for pre-GA milestones:

[v1.6.1-rc1](https://ci.longhorn.io/job/private/job/image-security-scan/1356/): 2 HIGH, 0 CRITICAL
[v1.6.1-rc2](https://ci.longhorn.io/job/private/job/image-security-scan/1362/): 2 HIGH, 0 CRITICAL

derekbit:
cc @c3y1huang

c3y1huang:
> Run security testing of container images for pre-GA milestones:
> 
[...]

PR #8326, opened 2024-04-09
Fix a typo and release longhorn v1.6.1 to upgrade responder server

PhanLe1010:
We also reduced the CPU and memory request/limit because we discovered the root cause of OOMKill is not because of these values being low

longhorn/longhorn#8209

### DIFF
--- a/dev/upgrade-responder/install.sh
+++ b/dev/upgrade-responder/install.sh
@@ -4,7 +4,7 @@ UPGRADE_RESPONDER_REPO="https://github.com/longhorn/upgrade-responder.git"
 UPGRADE_RESPONDER_REPO_BRANCH="master"
 UPGRADE_RESPONDER_VALUE_YAML="upgrade-responder-value.yaml"
 UPGRADE_RESPONDER_IMAGE_REPO="longhornio/upgrade-responder"
-UPGRADE_RESPONDER_IMAGE_TAG="master-head"
+UPGRADE_RESPONDER_IMAGE_TAG="longhorn-head"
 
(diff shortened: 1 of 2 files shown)


dev/upgrade-responder/install.sh
```diff
@@ -4,7 +4,7 @@ UPGRADE_RESPONDER_REPO="https://github.com/longhorn/upgrade-responder.git"
 UPGRADE_RESPONDER_REPO_BRANCH="master"
 UPGRADE_RESPONDER_VALUE_YAML="upgrade-responder-value.yaml"
 UPGRADE_RESPONDER_IMAGE_REPO="longhornio/upgrade-responder"
-UPGRADE_RESPONDER_IMAGE_TAG="master-head"
+UPGRADE_RESPONDER_IMAGE_TAG="longhorn-head"
```
shuo-wu, line 7: The latest ci still uses tag `master-head`: https://drone-publish.longhorn.io/longhorn/upgrade-responder/70/1/3
Do we need this change?

PhanLe1010: This script is just for creating a local instance of upgrade responder server. Since this repo is holding manifest for Longhorn version of the upgrade responder, I think this script should deploy a `longhorn-head` local instance. 

The CI is currently does:
1. If we merge PR on master-head, it will push a master-head image
2. If we merge PR on longhorn-head, it will push a longhorn-head image

So, it is unrelated to the script above

shuo-wu (approved):

innobead (approved):

Issue #8070, opened 2024-02-29
[IMPROVEMENT] Improve the UX of updating danger zone settings


derekbit:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

Starting from version 1.6.0, users can update danger zone settings and delay their activation until after all volumes are detached. However, there are user experiences that require improvement:

- It is difficult to determine if the settings have taken effect.
  - The Longhorn system can offer a clearer way to help users understand the current status.

- Following the detachment of all volumes, certain actions such as restarting instance manager pods do not occur quickly, might requiring users to wait for several minutes. To expedite the application of the settings, users need to unset and then set them again.

The improvement is inspired when working https://github.com/longhorn/longhorn/issues/8065 with @cometta.
The two UXs make users confused and not easy to know what is happening.

derekbit:
cc @mantissahz

innobead:
The current setting does not have any status. Could consider adding it to enrich the info or at least have an annotation for the status. `applied` or `satisfied` could be a good name for that status.

[/k8s/pkg/apis/longhorn/v1beta2/setting.go#L14-L20](https://github.com/longhorn/longhorn-manager/blob/5c5566b7c52e447bfb03b45ee2d0554a06375768/k8s/pkg/apis/longhorn/v1beta2/setting.go#L14-L20)
```go
type Setting struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	// The value of the setting.
	Value string `json:"value"`
}
```

innobead:
This involves a CRD update and isn't a critical improvement, so it's no need to backport.

cc @derekbit

derekbit:
@mantissahz Where are we with this issue?

mantissahz:
>@mantissahz Where are we with this issue?

@derekbit, it should be on `ready for testing` for CLI and UI.

yangchiu:
Testing on master-head (longhorn-manager 2d5d202) with the following steps:

1. Create a deployment to create and attach a volume
2. Create a priority class:
```
# Source: longhorn/templates/priorityclass.yaml
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: "longhorn-critical-1"
  labels:
    app.kubernetes.io/name: longhorn
    app.kubernetes.io/instance: longhorn
    app.kubernetes.io/version: v1.6.0-dev
[...]

PR #8424, opened 2024-04-23
crd: introduce the field `applied` in setting

mantissahz:

Issue # longhorn/longhorn#8070


chart/templates/clusterrole.yaml
```diff
@@ -35,7 +35,7 @@ rules:
   resources: ["volumesnapshotclasses", "volumesnapshots", "volumesnapshotcontents", "volumesnapshotcontents/status"]
   verbs: ["*"]
 - apiGroups: ["longhorn.io"]
-  resources: ["volumes", "volumes/status", "engines", "engines/status", "replicas", "replicas/status", "settings",
+  resources: ["volumes", "volumes/status", "engines", "engines/status", "replicas", "replicas/status", "settings", "settings/status",
```
innobead, line 38: See if we need to update the uninstall.yaml. I thought we don't.

innobead (approved):

PR #8428, opened 2024-04-24
Feat(LEP): Longhorn toolbox CLI

c3y1huang:

Issue longhorn/longhorn#7927

This LEP introduces the Longhorn Toolbox CLI, laying the foundation for upcoming development. The CLI aims to streamline manual Longhorn operations like troubleshooting and non-custom-resource operations.

`None`

`None`

### DIFF
--- a/enhancements/20240423-longhorn-toolbox-cli.md
+++ b/enhancements/20240423-longhorn-toolbox-cli.md
@@ -1,0 +1,543 @@
+# Longhorn Toolbox Commandline Interface (CLI)
+
+## Summary
+
+Navigating Longhorn's troubleshooting and manual operations documents can be challenging. This proposal introduces the Longhorn Toolbox Command-Line Interface (CLI) to enhance user experience. This CLI steamlines operations, making Longhorn's manual operations simpler and more intuitive for users.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/7927
+
+## Motivation
+
+### Challenges With Current Operations
+
+Currently, Longhorn users often face challenges when performing manual operations such as troubleshooting and non-custom-resource operations. These challenges include:
+
+- The need to refer to extensive documentation for each operations.
+- Complex and time-consuming steps for one-time operations.
+- Difficulty in gathering information for troubleshooting.
+- Manual processes for retrieving data and information during Longhorn failures.
+
+### Goals Of The Longhorn Toolbox CLI
+
+This proposal's primary goal is to introduce the Longhorn Toolbox CLI, laying the foundation for simplifying manual Longhorn operations. The specific objective include:
+
+- Providing a single, unified interface for common operations.
+- Reducing the time and effort required for one-time operations.
+- Streamlining troubleshooting by automating the information collection and error identification.
+- Simplifying data retrieval from Longhorn volumes during failures.
+
+#### Initial Implementation
+
+To demonstrate the capabilities of this CLI, the initial Proof of Concept (PoC) will focus on the following operations:
+
+- Preflight installation/uninstallation
+- Preflight checking
+- Volume trimming
+- Replica exporting
+
+Additional operations will be developed on a case-by-case basis to further enhance the Longhorn Toolbox CLI.
+
+### Non-goals [optional]
+
+While this proposal aims to address fundamental manual operations, it does not encompass every existing possible troubleshooting scenario or operation. The focus is on establishing a structured CLI for common use cases.
+
+## Proposal
+
+### User Stories
+
+#### Story 1: Preparation for Longhorn
+
+As a Longhorn user, I want to prepare my environment using the Toolbox CLI to easily set up my environment without the need for manual installations or searching through documents.
+
+- Before this enhancement: Users had to refer to documents to either prepare the environment using the Longhorn provided manifest or follow a detailed document to manually set up the environment for each cluster hosts.
+- After this enhancement: Users can simplify Longhorn prerequisites setup with a single command.
+
+#### Story 2: One-time operation
+
+As a Longhorn user, I want to perform one-time operations using the Toolbox CLI to execute tasks without navigating the Longhorn UI or remembering complex commands.
+
+- Before this enhancement: User need to remember or lookup steps for one-time operations, a process that could be complex and time-consuming.
+- After this enhancement: One-time operations are consolidated into a uniflied CLI, removing the need to remember and following complex steps. This improves efficiency and reduce the likelihood of errors.
+
+#### Story 3: Troubleshoot
+
+As a Longhorn user, I want to troubleshoot Longhorn using the Toolbox CLI to quickly identify and resolve issues.
+
+- Before this enhancement: Troubleshooting Longhorn requires manual information gathering from various sources, a time-consuming process.
+- After this enhancement: The troubleshooting process is streamlined, allowing user to quickly collect information and pinpoint issues. This enhancement improves efficiency and reduces the time needed to resolve Longhorn issues.
+
+#### Story 4: Data-retrival
+
+As a Longhorn user, I want to retrieve data from Longhorn volumes using the Toolbox CLI when Longhorn encounters failures.
+
+- Before this enhancement: User had to navigate through complex steps to manually mount Longhorn volumes for data access, a time-consuming and error-prone process.
+- After this enhancement: User an easily export data to a specified targeting path using the Toolbox CLI, simplifying the process and enhancing the user experience during the Longhorn failures.
+
+
+### User Experience In Detail
+
+When a user execute the `longhorn-toolbox` CLI, they should encounter clear and informative output to enhance their understanding and usage. This includes:
+- Detailed command descriptions: Each command is accompanied by a detailed description, providing users with a comprehensive overview of its purpose and usage.
+- Help menu at each command layer: User can find help menu at each layer of the command and subcommands that allows user to get detailed information without navigating through external documentation.
+- Clear messages and log control: User have the ability to change the log level, allowing them to control the verbolity of output. This ensures that user receive clear logs and results, making it easier to understand the CLI's action and responses.
+
+Examples:
+- Showing help at root command:
+    ```shell
+    > ./bin/longhorn-toolbox --help
+    Longhorn toolbox CLI for managing Longhorn
+
+    Install And Uninstall Commands:
+      install          Install Longhorn extensions
+      uninstall        Uninstall Longhorn extensions
+
+    Operation Commands:
+      trim             Longhorn trim commands
+      export           Export Longhorn
+
+    Troubleshoot Commands:
+      check            Check Longhorn
+      get              Get Longhorn resources
+
+    Other Commands:
+      global-options   Print global options inherited by all scommands
+
+    Use "longhorn-toolbox <command> --help" for more information about a given
+    command.
+    ```
+- Showing help at subcommand:
+    ```shell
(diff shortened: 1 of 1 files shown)


enhancements/20240423-longhorn-toolbox-cli.md
```diff
@@ -0,0 +1,543 @@
...
+- PostRun Function:
+    - Invoke receiver `Cleanup()` to clean up resources used by the `replicaGetter` after command execution.
+- Global Options (`SetGlobalOptionsRemote`):
+    - Sets global options for the subcommand, ensuring consistency and ease of use across all subcommands.
+- Command Flags:
+    - Defines flag for the command, allowing users to provide additional options and assign it to the `replicaGetter` instance.
+
+#### Remote (longhorn-toolbox)
+
+The `longhorn-toolbox` command is responsible for managing the required resources within the Kubernetes cluster for in-cluster operations, such as ConfigMap, DaemonSet, etc.
```
james-munson, line 298: Does the remote flavor use the KUBECONFIG envariable to find and connect to the target cluster?

james-munson (commented):
This looks really good.  Just one small question.

enhancements/20240423-longhorn-toolbox-cli.md
c3y1huang: By default, the remote uses the path specified in the `KUBECONFIG` environment variable. However, we will keep this customizable by providing it as a global option:
```go
cmd.PersistentFlags().StringVar(&globalOpts.KubeConfigPath, consts.CmdOptKubeConfigPath, os.Getenv("KUBECONFIG"), "Kubernetes config (kubeconfig) path")
```

james-munson (dismissed):
Looks good.

enhancements/20240423-longhorn-toolbox-cli.md
```diff
@@ -0,0 +1,543 @@
...
+    - A shared volume mount for the result output to `replica.json`.
+    - An init-container executing `toolbox get replica`.
+    - A container running `cat "replica.json"`.
+- When the DaemonSet is running successfully, retrieve the log of the container outputting the `replica.json`, then unmarshal JSON to YAML and output to the users.
+- Post-command cleanup:
+    - Upon command completion, cleanup the DaemonSet.
+
+##### Local (`toolbox get replica`)
+
+- Implement logic to collect replica information from host directories and `volume.meta` files:
```
ejweber, line 475: So the goal of this command is primarily to get information that is not available via the replica CR? Namely, on disk data structures, etc.?

For a broken volume, one thing we often want is to see the snapshot chain as it exists on disk. Your description calls out `volume.meta` files, but does the `Metadata` field also include snapshot meta files?

If I understand the goal correctly, another thing we want is the nominal and actual size of img files. This can help us understand if a snapshot is (unexpectedly) empty, etc.

ejweber (commented):
This is really in-depth and well-written. Thank you!

I have a few questions to discuss. Please find them in my review.

My major concern with implementing a CLI like this is that I want us to try to avoid duplicate effort wherever possible. (For example, if we can already get information easily with a kubectl command, we don't need a CLI command to do the same.) I was a bit worried when I first saw the `get replica` command, but I think it generally aims to retrieve additional information that we don't (and probably won't) put in the replica CRD. Overall, I think the LEP provides good examples of extra functionality that we can't achieve with it.

No real concerns from an implementation standpoint. It looks like we are using industry standard command line libraries, etc.

enhancements/20240423-longhorn-toolbox-cli.md
```diff
@@ -0,0 +1,543 @@
...
+    INFO[2024-04-23T15:19:16+08:00] Run 'longhorn-toolbox export replica stop' to stop exporting the replica
+    ```
+
+### API changes
+
+Not applicable in this context.
+
+## Design
+
+The Longhorn Toolbox CLI will be based on the existing https://github.com/longhorn/cli, originally designed for Longhorn preflight operations.
```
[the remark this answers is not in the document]
c3y1huang: As far as I know, the repo was still in PoC phase, originally named longhorn/preflight, and the image used for it was never released. So I didn't think it's necessary to mention the deprecation.

Please correct me if I am wrong, @derekbit .

enhancements/20240423-longhorn-toolbox-cli.md
```diff
@@ -0,0 +1,543 @@
...
+    - An init-container executing `toolbox install preflight`.
+    - A container running a pause image as an indicator to signal the completion of the `toolbox` command in the init-container.
+- When operating system is specified to `cos`, create:
+    - a ConfigMap include:
+        - The entrypoint.sh script.
+    - a DaemonSet include:
+        - Container running the `entrypoint.sh` script. Ref: [longhorn-gke-cos-node-agent.yaml
+        ](https://github.com/longhorn/longhorn/blob/2cbedf3902e72e347eed1eb8a45768462a4dd76b/deploy/prerequisite/longhorn-gke-cos-node-agent.yaml)
+        - Readiness probe for successful installation indication.
+        - Liveness probe for error detection.
```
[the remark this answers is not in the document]
c3y1huang: > The other commands mention a post-command cleanup, but this one does not. Is it intentional that we do NOT clean up the DeamonSet?
> 
> * If yes, is it because we prefer for it to run indefinitely in the cluster (perhaps executing actions on each host reboot)?

Its intentional for this DaemonSet to coexist with the Longhorn system on Container-Optimized OS due to the operating system's limitations regarding external service, directory permissions, and I/O operations.

Kubernetes work around this limitation by having its own rootfs with the necessary binaries, services, and executing tasks within it. Hense, Longhorn provides and requires this DaemonSet to keep the Longhorn data directory mounted to the Kubernetes rootfs and manually executing iSCSId within its container, as COS prohibits directly execution or having it managed by the systemd.

> * If yes, is there a CLI command to eventually remove it?

Users can either directly delete the DaemonSet or run `longhorn-toolbox uninstall preflight` to remove it.

> * If yes, what happens if we run the install command again while the DaemonSet is already running in the cluster?

If the DaemonSet is already running, executing the command will trigger the installation process again, attempting to create the ConfigMap and DaemonSet. However, as they already exist, the Kubernetes will throw an error, which the CLI can log to the user. Then, the CLI will proceed to validate the DaemonSet container status. If the container is running as expected, the CLI shows complete, or error otherwise.

enhancements/20240423-longhorn-toolbox-cli.md
```diff
@@ -0,0 +1,543 @@
...
+- Introduce command to prepare and create a `longhorn-volume-trimmer` DaemonSet.
+- The DaemonSet include:
+    - An init-container executing `toolbox trim volume`.
+    - A container running a pause image as an indicator to signal the completion of the `toolbox` command in the init-container.
+- Post-command cleanup:
+    - Upon command completion, cleanup the DaemonSet.
+
+##### Local (`toolbox trim volume`)
+
+- Implement logic to detect and execute `fstrim` for `RWO` and `RWX` volumes.
```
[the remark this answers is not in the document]
c3y1huang: This is a valid point. Utilizing the Longhorn REST API could indeed handle the task for this one-time operation, however, an additional benefit of using the local toolbox command provides an alternative option for users, especially in situations where Longhorn isn't working as expected (or we suspect it isn't). Plus, I think errors returned directly from the CLI could be more straightforward compared to the process of going through Longhorn's log.

enhancements/20240423-longhorn-toolbox-cli.md
c3y1huang: Indeed, the goal of this command is to retrieve information directly from the host, where the information may not be available or up-to-date via the replica CR.

For the initial phase of the CLI project, the implementation serves to gather the data required for exporting a replica to the host. And I anticipate that additional information will be added to the info collection over time. Please help create an improvement ticket for any other useful data you have in mind. Thank you!

enhancements/20240423-longhorn-toolbox-cli.md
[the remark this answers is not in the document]
ejweber: Got it. Thanks for the explanation!

ejweber (approved):
Thanks for the answers to my questions. LGTM.

enhancements/20240423-longhorn-toolbox-cli.md
james-munson: Completely agree with Eric's suggestion.  A way to see what the parentage chain looks like, whether there are any gaps in it, and a notion of the actual data content amount in snapshot .img files would be huge.  Worth an enhancement ticket on top of this already valuable tool.

james-munson (dismissed):

enhancements/20240423-longhorn-toolbox-cli.md
```diff
@@ -0,0 +1,543 @@
...
+```
+
+### Feature
+
+#### Install Preflight
+
+##### Remote Command (`longhorn-toolbox install preflight`)
+
+- Introduce command to prepare and create a `longhorn-preflight-installer` DaemonSet.
+- Include `--operating-system` command option to accommodate non-package-manager operating systems, such as container-optimized OS (COS).
```
derekbit, line 409: What is non-package-manager?

c3y1huang: I'm referring to operating systems that do not include a package manager. Reworded, thank you.

innobead:
cc @Vicente-Cheng @bk201 @WebberHuang1118 can check this from a support perspective

enhancements/20240423-longhorn-toolbox-cli.md
```diff
@@ -0,0 +1,543 @@
...
+- Resource Management: This command handles the creation and deletion of the resources like ConfigMap and DaemonSet within the Kubernetes cluster. For instance, the DaemonSet that is responsible for running operations related to the command actions.
+- Monitoring: The `longhorn-toolbox` command monitors the status of the DaemonSet containers. This proactive monitoring block ensures that actions are completed and allows for error handling.
+- Error Handling and Log Retrieval: In case of an action failure, the command retrieves the logs of the specific container involved. These logs shall provide users with clear understandings of the failure.
+- Results Output: For actions that yields a collection of information, the command formats the presents the result in a user-friendly YAML formal. These ensures that users can easily interpret the output.
+
+#### Local (toolbox)
+
+The `toolbox` command is designed to run within a DaemonSet pod inside the Kubernetes cluster, focusing on managing interactions within the cluster environment, including both in-cluster and host operations.
```
innobead, line 306: Can we rename `longhorn-toolbox-local` instead? Suggest still having `longhorn` in the artifact name to make it self-explanatory.

innobead (approved):
LGTM. 

The only suggestion is to rename the binary name as below. This can make the purpose of CLI extensible in the future. WDYT

`longhorn-toolbox` -> `longhornctl`
`toolbox` -> `longhornctl-local`

Vicente-Cheng (commented):
just some nit discussion, overall LGTM.

This part
>  Results Output: For actions that yields a collection of information, the command formats the presents the result in a user-friendly YAML formal. These ensures that users can easily interpret the output.

Should be very good for support! Thanks for this!

innobead:
> LGTM.
> 
> The only suggestion is to rename the binary name as below. This can make the purpose of CLI extensible in the future. WDYT
> 
> `longhorn-toolbox` -> `longhornctl` `toolbox` -> `longhornctl-local`

I will merge this PR, but please consider whether the naming is better during implementation.

Issue #2187, opened 2021-01-15
[IMPROVEMENT] Use `fsfreeze` instead of `sync` before snapshot


yasker:
**Is your feature request related to a problem? Please describe.**
Currently before snapshot operation, we will run a `sync` to flush the cache into the disk. It will be even safer if we also run `fsfreeze` against the filesystem, to stop the filesystem from written data to the disk before we take the snapshot. This will ensure the snapshot we took contains the stable filesystem image.

**Describe the solution you'd like**
Snapshot process:
1. `fsfreeze --freeze <fs_mount_point>`, which should include the cache flush. See https://linux.die.net/man/8/fsfreeze
2. Take a snapshot
3. `fsfreeze --unfreeze <fs_mount_point>`

Notice `fsfreeze` only works for certain filesystems. For the filesystems that doesn't support `fsfreeze`, `sync` will be used instead.

**Describe alternatives you've considered**

**Additional context**
Since we have to keep the freeze period as short as possible, it's better to do it inside the engine. One additional complexity is how can we know where is block device is mounted inside the engine.

innobead:
ref: https://github.com/longhorn/longhorn/issues/2029#issuecomment-791931309

khushboo-rancher:
Verified with Longhorn-master `03/19/2021`

Validation - **Pass**

The test steps from https://github.com/longhorn/longhorn/issues/2187#issuecomment-793703267 works fine.

@joshimoo The snapshot and backup creation worked fine while the writing was in process, but this works fine with v1.1.0 also, how can I make sure that `fsfreeze` is being validated with these test steps?

Regarding the test cases, we generally do these kind of testing with our pre release testing like basic operations with volume while writing is in the process, but I can't locate exact test case, I'll add this in our pre-release test.

khushboo-rancher:
Logs for `Froze filesystem`
```
[instance-manager-e-424ff07d] [pvc-43f5d029-511e-48ff-84bd-13545d3a7ef9-e-93cf8583] time="2021-03-19T15:12:59Z" level=info msg="Starting snapshot" snapshot= volume=pvc-43f5d029-511e-48ff-84bd-13545d3a7ef9 
[...]

PR #8447, opened 2024-04-25
Add a global FreezeFS setting and a volume FreezeFS field

ejweber:

longhorn/longhorn#2187

Adds the global and volume specific FreezeFSForSnapshot setting and field.


PhanLe1010 (dismissed):
LGTM

Can we double check if the question.yaml and values.yaml are valid yamls by `yq e . yourfile.yaml`

ejweber:
> Can we double check if the question.yaml and values.yaml are valid yamls by yq e . yourfile.yaml

No issues.

Issue #2856, opened 2021-08-10
[FEATURE] HA backing image


shuo-wu:
**Is your feature request related to a problem? Please describe.**
After download/upload/exporting from volumes, backing images need to be stored across different nodes immediately in case the node stores data unavailable.

**Additional context**
This is a feature that hasn't been done in #2530

bk201:
Found some behavior in Harvester, hope this will be improved together.

For a backing image that has no volumes derived from it:
- Evicting the Longhorn disk containing it doesn't evict the backing image to other disks.
- Draining the node containing the backing image doesn't move the image data either.

The backing image seems not to consider scheduling:
- Disable scheduling for all disks.
- Create a new backing image; the backing image is saved to one of them.

innobead:
> Found some behavior in Harvester, hope this will be improved together.
> 
> For a backing image that has no volumes derived from it:
> 
> * Evicting the Longhorn disk containing it doesn't evict the backing image to other disks.
> * Draining the node containing the backing image doesn't move the image data either.
> 
> The backing image seems not to consider scheduling:
> 
> * Disable scheduling for all disks.
> * Create a new backing image; the backing image is saved to one of them.

:point_up: this is good info. Basically, we should treat backing image as volume replica, so the behavior related to resilience and availability should be similar.

yangchiu:
@chriscchien Please help check if [test case `test_setting_backing_image_auto_cleanup`](https://ci.longhorn.io/job/public/job/master/job/sles/job/arm64/job/longhorn-tests-sles-arm64/896/testReport/tests/test_settings/test_setting_backing_image_auto_cleanup/) can pass after https://github.com/longhorn/longhorn-tests/pull/1877 merged. Many thanks.

chriscchien:
Verified **pass** on longhorn master(longhorn-manager `0a1d3a`) with test [steps](https://github.com/longhorn/longhorn/issues/2856#issuecomment-2089737102)

Waiting https://github.com/longhorn/longhorn/issues/2856#issuecomment-2208181883 to beresolved first, if no new issues are found, I will close this ticket, thanks.

chriscchien:
Found issue when multiple `nodeSelector`  or `diskSelector` were set to backing image, need to verify again after https://github.com/longhorn/longhorn/issues/6526#issuecomment-2209904516

ChanYiLin:
Hi @chriscchien 
[...]

PR #8452, opened 2024-04-26
feat(backingimage): add backing image enhancement LEP

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/2856

Add LEP for BackingImage enhancement

- HA
- nodeSelector, diskSelector
- Eviction Request


innobead:
cc @WebberHuang1118 @bk201

enhancements/20240426-backing-image-enhancement.md
```diff
@@ -0,0 +1,204 @@
...
+
+With this feature, Longhorn maintains the number of copy in the cluster to reduce the possibility of the data loss.
+
+#### Node Selector and Disk Selector
+
+Before this feature, Longhorn stores the BackingImage copy on a random node and a random disk. When a replica needs the BackingImage, Longhorn needs to copy the BackingImage to the disk where replica is.
+
+Similar to Volume Replicas scheduling, with adding `nodeSelector` and `diskSelector` to the BackingImage, Longhorn will place the BackingImage copies to the specific nodes or disks. BackingImages will be stored in the same set of nodes and disks as Replicas if they have the same `nodeSelector` and `diskSelector`. It improves space efficiency.
+
+Noted that, if the settings are not the same between BackingImages and Replicas. When a Replica is placed on the node where BackingImage cannot be stored, the Replica will be stuck waiting for BackingImage. User can diagnose the issue by checking the condition `WaitForBackingImage` in Volume and Replica.
```
shuo-wu, line 55: I am wondering if we can mutate the volume/replica selector during the volume creation with a backing image in some cases.

Actually, there are 2 cases:
1. There are overlappings between the node/disk selectors of replicas and the corresponding backing image. To avoid replicas being stuck on the node that cannot contain the backing image, the mutator should choose the overlapping part for the volume replicas. For example, when a user adds selector `node1` and `node2` for the volume creation while its backing image has selector `node1` and `node3`, the mutator should change the volume/replica selector to `node1` only.
2. There is no overlapping. [Test case 3 below] Then we reject the volume creation. For example, when a user adds selector `node1` for the volume creation while its backing image has selector `node2`, then the volume creation should be rejected.

ChanYiLin: I am afraid this would 
1. confuse the users since the change is hidden underneath
2. the logic becomes quite complicated but seems doesn't bring too much benefit. 
    - For overlapping case, in the original design, volume can still start but be degraded. Users can check the condition to see whats wrong and do the modification themselves. (comparing to doing the job for them but they may be not aware to the selectors change.)
    - For non overlapping case, no replicas will start, users can check the condition and do the modification to reschedule the replicas. (comparing to directly reject the creation)

WDYT?
cc @innobead @derekbit

shuo-wu: For the non-overlapping case, I think directly refusing the request would be more straightforward. Users would know what's wrong with the parameters immediately. 
For the overlapping part, I am fine with making the volume degraded if you think the silent mutation is unfriendly to users.

enhancements/20240426-backing-image-enhancement.md
```diff
@@ -0,0 +1,204 @@
...
+1. When a user disables the scheduling and set eviction request on the node or disk, Longhorn will move the BackingImage copies to other nodes or disks.
+
+
+### Non-goals
+
+#### Eviction Handling When Cordoning or Draining Node.
+
+We only support evicting BackingImage copy when manually set eviction request to the node or disk. That is, if a user want to cordon or drain the node. Longhorn won't evict the BackingImage copies automatically like what it does for the Replicas. In this case, the user needs to set eviction request to the node or disk first before draining the node.
+
+The reason is that we will need to set PDB to protect BackingImageManager from deleting during draining process. It increases the complexity of the process and adding risk of stucking in the process.
```
WebberHuang1118, line 38: Just a question, so the backing image file and backing image manager will both exist after node drain?

enhancements/20240426-backing-image-enhancement.md
WebberHuang1118, line 55: The replica will keep waiting? Or the Longhorn manager will try to schedule replica to another disk/node?

enhancements/20240426-backing-image-enhancement.md
ChanYiLin: Hi @WebberHuang1118  Yes,
We don't trigger any eviction for BackingImage when doing node drain
Users have to manually set eviction request to the node before node drain.

enhancements/20240426-backing-image-enhancement.md
ChanYiLin: it will keep waiting
because form replica's perspective, it doesn't know if the BackingImage is still downloading/syncing or unable to be scheduled. 

PV/PVC is not under Longhorn's control, and we can't set additional fields in PV/PVC
So users have to check Longhorn's Volume and Replicas condition

innobead: When a replica is decided to schedule to a node, don't we consider if the schedule node is eligible for the backing image? I thought this should be part of replica scheduling?

A good question from Webber. For any event of longhorn resources around volume, it will be nice to consider at least bubble an event to the PVC level, so users can describe the corresponding PVC to know about the status.

enhancements/20240426-backing-image-enhancement.md
```diff
@@ -0,0 +1,204 @@
...
+
+### Implementation Overview
+
+#### CRD
+
+- Add `numOfCopies` to BackingImage
+
+#### BackingImage Controller
+
+- After first BackingImage is ready, the controller starts maintaining the number of copies in the clusters.
```
innobead, line 77: How many replications will be happened concurrently? If we have 100 backing images are ready to do replicate another one? Do we have a concurrent mechanism? 

Note: A side topic. We have several concurrent situation like replica rebuilding, live upgrade, etc. They seem to do the concurrency independently. We Should think of managing these kinds of concurrent tasks in global to ensure resource usage efficient.

enhancements/20240426-backing-image-enhancement.md
```diff
@@ -0,0 +1,204 @@
...
+
+- Add `diskSelector` and `nodeSelector` to the BackingImage
+
+#### BackingImage Controller
+
+- When selecting node/disk for the BackingImage, Longhorn needs to follow the `diskSelector` and `nodeSelector` settings.
+
+#### Replica Controller
+
+- When the Replica starts on a node and request the BackingImage to be placed on that node, it checks the BackingImage's `diskSelector` and `nodeSelector` first before requesting.
```
innobead, line 95: Can we filter nodes which are eligible for backing image replication before scheduling a replica to one of them?

enhancements/20240426-backing-image-enhancement.md
```diff
@@ -0,0 +1,204 @@
...
+    - The evicted copy will be deleted
+
+5. Eviction - 3 (Negative Test)
+    - Set nodeTag: [node1], diskTag:[disk1]
+    - Create a BackingImage with following settings to place the copy on node1 
+        - `minNumberOfCopies = 1`
+        - `nodeSelector = [node1]`
+        - `diskSelector = [disk1]`
+    - Evict node1
+    - The copy would not be deleted because it is the only copy
```
innobead, line 194: Suggest having an info to let users know why this BI is still kept via condition or event. For example, check the backing image.

innobead (commented):
In general, LGTM. Have a few comments needed to resolve.

enhancements/20240426-backing-image-enhancement.md
ChanYiLin: Hi @innobead 
BackingImage scheduling is introduced in this feature
Before this, BackingImage can be placed on any nodes
So, we don't have this consideration in our replicas.
Besides, replica has its own `nodeSelector` and `diskSelector` 

Consider the following situation.
```
Volume ReplicaCounts: 2
Replicas:
- nodeSelector: [1, 3]

BackingImage:
- nodeSelector: [3]
```
It either 
1. creates replica on node1, node3 and shows that replica on node1 is waiting for BackingImage (current implementation)
2. creates replica on node3. and shows that 1 of the replica cannot be scheduled.

I prefer option 1 which user can check that the condition and find out that the BackingImage setting is not align with the Replicas

ChanYiLin: > A good question from Webber. For any event of longhorn resources around volume, it will be nice to consider at least bubble an event to the PVC level, so users can describe the corresponding PVC to know about the status.

Sure let me see if we already have similar function to reuse. If no, we should create a ticket for all this kind of events including replicas scheduling, engine issue, ... etc.

enhancements/20240426-backing-image-enhancement.md
ChanYiLin: We don't have concurrent limit to limit the BackingImage replication.
The mechanism is just like replenish Replicas when the number of Replica is smaller than the setting number
We also don't have concurrent limit for that situation, should we?
Image we have 100 Volumes having 1 Replica but they should have 2, so Longhorn  tried to replenish another Replica for all this 100 Volumes?

enhancements/20240426-backing-image-enhancement.md
ChanYiLin: Sure we can add a event for that.

enhancements/20240426-backing-image-enhancement.md
ChanYiLin: So
your suggestion is a prevention, and we need to update the replica scheduling in this feature
The current design is let the replica to be created, but tell the user that the replica is now waiting for the BackingImage
I am good with the prevention, the question is "should we do it in this same feature?"

enhancements/20240426-backing-image-enhancement.md
WebberHuang1118: Thanks @ChanYiLin

innobead:
@ChanYiLin and I discussed the feedback. He will update the PR with our planned additions and future plans for some improvements. 

Others, LGTM.

enhancements/20240426-backing-image-enhancement.md
ChanYiLin: sure, I can reject  the creation if they are not the same

ChanYiLin: Hi, I can adjust the replica scheduling to prevent replica to be scheduled on the node where BackingImage cannot be placed.

enhancements/20240426-backing-image-enhancement.md
ChanYiLin: I will follow the `Concurrent Replica Rebuild Per Node Limit` to set the concurrent limit to the BackingImage replenishment

enhancements/20240426-backing-image-enhancement.md
ChanYiLin: I will adjust the replica scheduling to prevent replica to be scheduled on the node where BackingImage cannot be placed.

enhancements/20240426-backing-image-enhancement.md
ChanYiLin: I have discussed with @derekbit 
for this kind of resources(backing image or replicas) left on node after eviction
We should add a status fields on node CRD to hint the users about the situation and we should also show the status on UI for better user experience.
We already create the ticket.
https://github.com/longhorn/longhorn/issues/8486

enhancements/20240426-backing-image-enhancement.md
```diff
@@ -0,0 +1,206 @@
...
+
+- Add `evictionRequested` to the BackingImage file status
+
+```go
+type BackingImageDiskFileStatus struct {
+	State BackingImageState `json:"state"`
+	Progress int `json:"progress"`
+	Message string `json:"message"`
+	LastStateTransitionTime string `json:"lastStateTransitionTime"`
+	EvictionRequested bool `json:"evictionRequested"`
```
innobead, line 115: `EvictionRequested` means it will be requested to evict, or it has been evicted? If the first, it should belong to the spec instead of status. (I will review this when reviewing the PR)

enhancements/20240426-backing-image-enhancement.md
```diff
@@ -0,0 +1,206 @@
...
+    
+3. Eviction - 1
+    - Create BackingImage with one copy
+    - Evict the node where copy is on
+    - The BackingImage will first create another copy in another node
+    - The evicted copy will be deleted 
+
+4. Eviction - 2
+    - Create BackingImage with two copies
+    - Evict the node where one of the copy is on
```
innobead, line 186: The old one will be evicted but will be created in another node.

innobead (commented):
Just two questions left.

enhancements/20240426-backing-image-enhancement.md
ChanYiLin: yes, but that is not because the eviction
but because of the `minNumberOfCopies`
That is, if the `minNumberOfCopies =  1` but because there is one volumes has used this BackingImage so it is replicated to other node where we are going to evict
We won't create another one if it meets the requirement of `minNumberOfCopies =  1`

enhancements/20240426-backing-image-enhancement.md
WebberHuang1118: > Yes it will be requested to evict As you said, it is supposed to belong to spec like the modification we did in volume
> 
> But the situation here is that Backingiamge, we don't have place for this field in the Spec So I place the field in status
> 
> ```
> // BackingImageSpec defines the desired state of the Longhorn backing image
> type BackingImageSpec struct {
> 	// +optional
> 	Disks map[string]string `json:"disks"`
> 	// +optional
> 	Checksum string `json:"checksum"`
> 	// +optional
> 	SourceType BackingImageDataSourceType `json:"sourceType"`
> 	// +optional
> 	SourceParameters map[string]string `json:"sourceParameters"`
> }
> 
> // BackingImageStatus defines the observed state of the Longhorn backing image status
> type BackingImageStatus struct {
> 	// +optional
> 	OwnerID string `json:"ownerID"`
> 	// +optional
> 	UUID string `json:"uuid"`
> 	// +optional
> 	Size int64 `json:"size"`
> 	// Virtual size of image, which may be larger than physical size. Will be zero until known (e.g. while a backing image is uploading)
> 	// +optional
> 	VirtualSize int64 `json:"virtualSize"`
> 	// +optional
> 	Checksum string `json:"checksum"`
> 	// +optional
> 	// +nullable
> 	DiskFileStatusMap map[string]*BackingImageDiskFileStatus `json:"diskFileStatusMap"`
> 	// +optional
> 	// +nullable
> 	DiskLastRefAtMap map[string]string `json:"diskLastRefAtMap"`
> }
> ```

`we don't have place for this field in the Spec` means this should refer to a per-disk structure? Only status has disk maps, neither does spec.

enhancements/20240426-backing-image-enhancement.md
innobead: That makes sense, but it's better to describe it with this condition `minNumberOfCopies = 1` if we don't have that in this test case.

enhancements/20240426-backing-image-enhancement.md
ChanYiLin: Yes, because we don't have much disk specific spec before.
so as you can see, in spec it is `Disks map[string]string `json:"disks"`` not something like `DiskFileStatusMap map[string]*BackingImageDiskFileSpec`

enhancements/20240426-backing-image-enhancement.md
ChanYiLin: got it

enhancements/20240426-backing-image-enhancement.md
```diff
@@ -0,0 +1,206 @@
...
+#### Node Selector and Disk Selector
+
+Before this feature, Longhorn stores the BackingImage copy on a random node and a random disk. When a replica needs the BackingImage, Longhorn needs to copy the BackingImage to the disk where replica is.
+
+Similar to Volume Replicas scheduling, with adding `nodeSelector` and `diskSelector` to the BackingImage, Longhorn will place the BackingImage copies to the specific nodes or disks. BackingImages will be stored in the same set of nodes and disks as Replicas if they have the same `nodeSelector` and `diskSelector`. It improves space efficiency.
+
+
+#### Replica Scheduling
+
+Since replica won't be able to start on the node where BackingImage can not be stored. Thus, we will not schedule the replica to the node and disk where the BackingImage can not be stored because of the `nodeSelector` and `diskSelector`.
```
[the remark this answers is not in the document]
ChanYiLin: No, a replica can't be scheduled can be due to many reasons, like space, engine-image and backingimage for example.
We can only look into the log to see why it can't be scheduled.

Creating the replica and let it hang for the BackingImage we can at least use condition to tell.
But now, with preventing the scheduling, there is no way to tell.

It is quite a big topic and I think it is useful
May I ask Harvester team to open a ticket to collect all the information we want through CSI instead of getting the Volume directly.
Then we can do the improvement there for all the cases at once.
Thanks

enhancements/20240426-backing-image-enhancement.md
ChanYiLin: I can change to spec and update the structure.
PTAL and see through the documents to see if any other things need to change, thanks.

enhancements/20240426-backing-image-enhancement.md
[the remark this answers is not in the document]
WebberHuang1118: > May I ask Harvester team to open a ticket to collect all the information we want through CSI instead of getting L > the Volume directly?
> Then we can do the improvement there for all the cases at once.

I will discuss this with @bk201, thanks.

enhancements/20240426-backing-image-enhancement.md
```diff
@@ -0,0 +1,206 @@
...
+```
+
+##### Node Controller 
+- When the node or disk is set to `evictionRequested = true`, node controller will update EvictionRequested for all the BackingImage copy on the node.
+
+
+#### BackingImage
+- `replenishBackingImageCopies()`
+    - if `nonFailedCopies >= MinNumberOfCopies`, we check if we need to replenish one copy for eviction
+        - replenish one if `NonEvictingCount < MinNumberOfCopies`
```
WebberHuang1118, line 136: What does `NonEvictingCoun` stand for? Can you explain more about this condition? Thanks.

enhancements/20240426-backing-image-enhancement.md
```diff
@@ -0,0 +1,206 @@
...
+    - if `nonFailedCopies < MinNumberOfCopies`
+        - replenish one copy to meet the MinNumberOfCopies requirement.
+- `cleanupEvictionRequestedBackingImageCopies()`
+    - If there is no non evicted healthy copy, don't delete the copy.
+    - Otherwise, delete the evicted copy.
```
WebberHuang1118, line 141: If all backing image copies are going to be deleted to satisfy eviction conditions, just ignore the eviction request. Do I understand correctly? Thanks.

enhancements/20240426-backing-image-enhancement.md
ChanYiLin: some copies maybe under eviction
we first replenish one of the copy then delete the evicted one.

for example
```
MinNumberOfCopies = 2
node1: bi
node2: bi (being evicted)
node3:

=> replenish one copy => synced to node3
node1: bi
node2: bi (being evicted)
node3:bi

then the bi on the node2 will then be deleted.
```

enhancements/20240426-backing-image-enhancement.md
ChanYiLin: Yes the last one will be left even it is evicted.
https://github.com/longhorn/longhorn-manager/pull/2742/files#diff-9cdffb3a94c7a2350738c352077504be988843d61387bd93d86ef4643d194a47R403-R416

enhancements/20240426-backing-image-enhancement.md
WebberHuang1118: thanks

enhancements/20240426-backing-image-enhancement.md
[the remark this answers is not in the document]
ChanYiLin: Thanks !

enhancements/20240426-backing-image-enhancement.md
WebberHuang1118: thanks

innobead (approved):

Issue #7571, opened 2024-01-07
[IMPROVEMENT] Fix outdated environment check in longhorn manager


innobead:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

Besides fixing the environment check in the bootstrap of the Longhorn manager, an environment check should also be added to the pre-upgrade job when installing Longhorn via Helm.

Consider consolidating this part together with the upgrade version check. 

[/app/daemon.go#L236-L244](https://github.com/longhorn/longhorn-manager/blob/a601b9b14da2ef3bb72f225dc3bc6b94515f0d33/app/daemon.go#L236-L244)
```go
func environmentCheck() error {
	namespaces := []lhtypes.Namespace{lhtypes.NamespaceMnt, lhtypes.NamespaceNet}
	nsexec, err := lhns.NewNamespaceExecutor(iscsiutil.ISCSIdProcess, lhtypes.HostProcDirectory, namespaces)
	if err != nil {
		return err
	}

	return iscsi.CheckForInitiatorExistence(nsexec)
}
```

roger-ryao:
**Verified on master-head 20240711**

- longhorn master-head https://github.com/longhorn/longhorn/commit/69739d8a63231934c5f7bb0c70619130be25fe23


**The test steps**
https://github.com/longhorn/longhorn/issues/7571#issuecomment-2081844335

**Result Passed**
I set up a **Debian** cluster. After installing Longhorn, the status of `multipathd` and required packages will be recorded in the node's `Status.Conditions`.

```console
> k -n longhorn-system describe nodes.longhorn.io ryao-test-w2-d0ff0ce9-4lw25
Status:
  Auto Evicting:  false
  Conditions:
    Last Probe Time:       
    Last Transition Time:  2024-07-12T03:33:36Z
    Message:               Node ryao-test-w2-d0ff0ce9-4lw25 is ready
    Reason:                
    Status:                True
    Type:                  Ready
    Last Probe Time:       
    Last Transition Time:  2024-07-12T03:33:25Z
    Message:               
    Reason:                
    Status:                True
    Type:                  Schedulable
    Last Probe Time:       
    Last Transition Time:  2024-07-12T03:33:56Z
    Message:               
    Reason:                
    Status:                True
    Type:                  MountPropagation
    Last Probe Time:       
    Last Transition Time:  2024-07-12T03:33:56Z
    Message:               Missing packages: [nfs-common open-iscsi cryptsetup dmsetup]
[...]

PR #8463, opened 2024-04-29
fix: check environment when doing pre-upgrade with Helm

mantissahz:

Issue # longhorn/longhorn#7571

Check if the environment settings are achieved when doing pre-upgrade job with Helm, ref: [environment_check.sh](https://github.com/longhorn/longhorn/blob/master/scripts/environment_check.sh)

### DIFF
--- a/chart/templates/preupgrade-job.yaml
+++ b/chart/templates/preupgrade-job.yaml
@@ -20,14 +20,23 @@ spec:
       - name: longhorn-pre-upgrade
         image: {{ template "registry_url" . }}{{ .Values.image.longhorn.manager.repository }}:{{ .Values.image.longhorn.manager.tag }}
         imagePullPolicy: {{ .Values.image.pullPolicy }}
+        securityContext:
+          privileged: true
         command:
         - longhorn-manager
         - pre-upgrade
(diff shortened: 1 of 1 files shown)


chart/templates/preupgrade-job.yaml
```diff
@@ -20,14 +20,27 @@ spec:
...
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+      volumes:
+      - name: proc
+        hostPath:
```
derekbit, line 42: Adding  a description for the `/proc` and `privileged: true` in the commit message  is appreciated.

mantissahz: Added in the commit message. Thanks

mantissahz:
Moved the check to the node controller.

mantissahz:
For checking iscsiadm/open-iscsi installed, we need this PR to be reopened.

related: https://github.com/longhorn/longhorn/issues/8974

derekbit (approved):
LGTM

PR #8616, opened 2024-05-22
feat(LEP): storage network for RWX volumes

c3y1huang:

Issue longhorn/longhorn#8184

This LEP proposes to extend Longhorn's storage network feature to support RWX volumes.

None.

None.

### DIFF
--- a/enhancements/20240522-storage-network-for-rwx-volumes.md
+++ b/enhancements/20240522-storage-network-for-rwx-volumes.md
@@ -1,0 +1,153 @@
+# Storage Network For Read-Write-Many (RWX) Volume
+
+## Summary
+
+This proposal outlines extending Longhorn's storage network feature to support Read-Write-Many (RWX) volumes. Having graduated from experimental status, RWX volumes should now be able to use storage network for data traffic, similar to Read-Write-Once (RWO) volumes. The enhancement enables data network isolation and potentially improve performance in specific network configuration that utilize storage network.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/8184
+
+## Motivation
+
+### Goals
+
+Enable users to utilize a dedicated network interface for RWX volume data traffic with Multus, similar to Read-Write-Once (RWO) volumes. This provides data network isolation and possibly can enhance performance in specific configurations.
+
+### Non-goals [optional]
+
+- This proposal does not involve performance evaluation of RWX volumes between storage networks and cluster networks.
+
+## Proposal
+
+### User Stories
+
+#### Story 1: RWX Volume Data Network Segregation
+
+As a Longhorn user, I want to use a dedicated network interface for RWX volume data traffic.
+
+- Before: Data network segregation was unavailable for RWX volumes. The data network segregation only works in the RWO volumes, and the data traffic between the mount point (client) and the share manager pod still goes through the cluster network.
+- After: User can use the storage network for RWX volumes, enabling network segregation.
+
+#### Story 2: Pre-existing and New RWX Volume Workload
+
+As a Longhorn user, I have storage network configured, and existing RWX volume workload before the v1.7.0 upgrade.
+
+- Pre-existing Workloads: There will be no disruption to pre-existing RWX volume workload pods since NFS mount point remains accessible in the host network namespace.
+- New Workloads: Newley created RWX volume workloads will be applied with the storage network.
+
+### User Experience In Detail
+
+Users will be able to create RWX volumes that uses the storage network for data traffic. Key user-facing changes include:
+
+- **CSI Annotations:** Share manager pods and CSI plugin pods will be annotated with the storage network for NFS client mounting. This annotation will only be applied once all volumes have been detached.
+- **Headless Service:** RWX volume will use a headless service and a custom endpoint, removing dependency on ClusterIP, which is not suitable for Multus networking.
+- **Custom Mounter Replacement:** The custom mounter uses `hostPID` and host `proc` namespace will be replaced with running the CSI plugin with `hostNetwork: true`.
+- **Workload Pod Restart:** When a CSI plugin pod restarts, the NFS share mountpoint becomes unavailable, rendering any attempt to access it unresponsive. User can manually restart the workload pod, or they can enable the `auto-delete-pod-when-volume-detached-unexpectedly` setting. This allows Longhorn to delete the associated workload pod when it is using the storage-network. If the pod is managed by a controller, the kubelet will create a new pod. This allows the CSI controller to handle volume remounting through the CSI node server.
+- **Upgrade to Longhorn v1.7.0:** If the storage network is configured before the upgrade, the CSI plugin pod will be annotated accordingly. The **Workload Pod Restart** does not apply to pre-existing RWX volume workload pods after upgrading to Longhorn v1.7.0, as the NFS share mountpoint remains accessible in the host network namespace.
+
+### API changes
+
+No API changes are required for this proposal.
+
+## Design
+
+### CNI Annotation
+
+Introduce CSI annotation for RWX volume pod stack to enable NFS client mounting over the storage network:
+	- Share manager pod
+	- CSI plugin pod
+
+The setting controller will apply these annotations once all volumes are detached.
+
+### Service
+
+Currently, Longhorn creates a share manager service auto-assigned with a cluster IP by the selector. The cluster IP is used for the share manager endpoint.
+
+However, the service selector is unaware of the Multus networking when the storage network is configured. Hence, for the RWX volumes, Longhorn will create a headless service and custom endpoint for the RWX volume resource stack.
+
+```golang
+// Service
+service.Spec.ClusterIP = "None"
+```
+```golang
+// Endpoint
+	newObj := &corev1.Endpoints{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            sm.Name,
+			Namespace:       sm.Namespace,
+			OwnerReferences: datastore.GetOwnerReferencesForShareManager(sm, false),
+			Labels:          labels,
+		},
+		Subsets: []corev1.EndpointSubset{},
+	}
+```
+
+The share manager will use the service FQDN for its endpoint instead of the value from service `ClusterIP` field.
+
+```yaml
+endpoint: nfs://pvc-117d3553-1a2c-4b42-aa49-2b221dfb5cd9.longhorn-system.svc.cluster.local/pvc-117d3553-1a2c-4b42-aa49-2b221dfb5cd9
+```
+
+### Kubernetes Endpoint Controller
+
+Introduce a new Kubernetes Endpoint controller to watch over share manager pod resources during *Add* and *Update* operations to sync the share manager details with the endpoint subset.
+
+```golang
+		desireSubset := corev1.EndpointSubset{
+			Addresses: []corev1.EndpointAddress{
+				{
+					IP:       storageIP,
+					NodeName: &pod.Spec.NodeName,
+					TargetRef: &corev1.ObjectReference{
+						Kind:      types.KubernetesKindPod,
+						Name:      pod.Name,
+						Namespace: pod.Namespace,
+						UID:       pod.UID,
+					},
+				},
+			},
+			Ports: []corev1.EndpointPort{
+				{
+					Name:     "nfs",
+					Port:     2049,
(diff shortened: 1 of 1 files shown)


enhancements/20240522-storage-network-for-rwx-volumes.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Proposal
+
+### User Stories
+
+#### Story 1: RWX Volume Data Network Segregation
+
+As a Longhorn user, I want to use a dedicated network interface for RWX volume data traffic.
+
+- Before: Data network segregation was unavailable for RWX volumes.
```
innobead, line 29: This is not clear. In fact, it works for RWO volume inside the share manager. It's better to clarify this.

`Data network segregation only works in the underlying RWO volume, but the data traffic between the mount point (client) and the share manager pod still goes through the cluster network.`

enhancements/20240522-storage-network-for-rwx-volumes.md
```diff
@@ -0,0 +1,139 @@
...
+- After: User can use the storage network for RWX volumes, enabling network segregation.
+
+### User Experience In Detail
+
+Users will be able to create RWX volumes that uses the storage network for data traffic. Key user-facing changes include:
+
+- **CSI Annotations:** Share manager pods and CSI plugin pods will be annotated with the storage network for NFS client mounting.
```
innobead, line 36: CSI Plugin is not system managed but a daemonset. When will we add the annotation? We will not wait for all volumes detached first?

enhancements/20240522-storage-network-for-rwx-volumes.md
```diff
@@ -0,0 +1,139 @@
...
+- **Headless Service:** RWX volume will use a headless service and a custom endpoint, removing dependency on ClusterIP, which is not suitable for Multus networking.
+- **Custom Mounter Replacement:** The custom mounter uses `hostPID` and host `proc` namespace will be replaced with running the CSI plugin with `hostNetwork: true`.
+- **Workload Pod Restart:** When a CSI plugin pod restarts, Longhorn will delete the associated workload pod when it is using the storage-network. If the pod is managed by a controller, the kubelet will create a new pod. This allows the CSI controller to handle volume remounting through the CSI node server.
```
innobead, line 39: Will this interrupt user running workloads? If yes, should we get the confirmation from users before doing this?

innobead (changes requested):
I have a few feedback and questions that need to be clarified first. 

BTW, the idea of using headless service for share manager pods is impressive.

innobead:
Added @mantissahz to the review, as this is involved in danger zone settings, which we are allowing to take effect asynchronously.

enhancements/20240522-storage-network-for-rwx-volumes.md
c3y1huang: > Will this interrupt user running workloads?

This will interrupt the user's running workloads.

> If yes, should we get the confirmation from users before doing this?

We can leverage on `auto-delete-pod-when-volume-detached-unexpectedly` setting for handling this.

enhancements/20240522-storage-network-for-rwx-volumes.md
c3y1huang: We will apply the annotation in the same way as the RWO volume using the setting controller, once all volumes have been detached.

innobead: So that means we will not apply the setting immediately but will eventually do so once all volumes are detached.

enhancements/20240522-storage-network-for-rwx-volumes.md
innobead: If we can ensure the behavior when setting the storage network is the same as other danger zone settings, then we don't need to leverage this setting or create a new one.

enhancements/20240522-storage-network-for-rwx-volumes.md
```diff
@@ -0,0 +1,141 @@
...
+
+- **CSI Annotations:** Share manager pods and CSI plugin pods will be annotated with the storage network for NFS client mounting. This annotation will only be applied once all volumes have been detached.
+- **Headless Service:** RWX volume will use a headless service and a custom endpoint, removing dependency on ClusterIP, which is not suitable for Multus networking.
+- **Custom Mounter Replacement:** The custom mounter uses `hostPID` and host `proc` namespace will be replaced with running the CSI plugin with `hostNetwork: true`.
+- **Workload Pod Restart:** When a CSI plugin pod restarts, the NFS share mountpoint becomes unavailable, rendering any attempt to access it unresponsive. User can manually restart the workload pod, or they can enable the `auto-delete-pod-when-volume-detached-unexpectedly` setting. This allows Longhorn to delete the associated workload pod when it is using the storage-network. If the pod is managed by a controller, the kubelet will create a new pod. This allows the CSI controller to handle volume remounting through the CSI node server.
```
innobead, line 39: I assume the annotation will only be applied once all volumes are detached, meaning there are no workloads using any active volumes. Do we still need this `workload pod restart` behavior?

innobead (commented):
In general, LGTM. 

Just one last question about if `workload pod restart` behavior is  necessary or not.

enhancements/20240522-storage-network-for-rwx-volumes.md
Vicente-Cheng, line 39: I know we need extra operations with workload pod if the csi-plugin (with multus) respawns.
The reason is described in the following section, `Share Manager Process Namespaces`, which is very clear.

But it still impacts the user. The workload pod restart means the application failed. Even the csi-plugin pod restart might be a corner case.

Vicente-Cheng (commented):
General is very clear about the current status. Thanks for it!
There is only one corner point: restart the workload pod. (with the above comment)

I checked some other solutions that might be facing the same situation.
Found an interesting article below: https://github.com/rook/rook/blob/master/design/ceph/multus-network.md#new-holder-daemonset

I am still checking what is the final solution with the multus-network on their implementation because this design says `not finalized`.

I will update again after I dig deeper into their code.

enhancements/20240522-storage-network-for-rwx-volumes.md
c3y1huang: During the storage-network setting application, there will be no associated workload pod to restart because all the volumes should be detached.

A workload pod will only restart if all the following conditions are met:
1. The pod is using RWX volumes + the cluster has storage networking configured.
1. The `auto-delete-pod-when-volume-detached-unexpectedly` setting is enabled. [This setting doesn't apply to the workload pods that don't have a controller. Longhorn never deletes them.](https://github.com/longhorn/longhorn-manager/blob/aebb5a8a1c5d5b9463c0c39b7ad4ae0766795477/types/setting.go#L691)
1. The workload pod is located on the same node as the restarted CSI plugin pod.

enhancements/20240522-storage-network-for-rwx-volumes.md
c3y1huang: When configuring the storage network, it is necessary to detach the volume, ensuring that no workload pods are using it.

However, Longhorn should handle the situation when the CSI plugin pod restarts unexpectedly to resolve dangling mount issues of already RWX volume attached workload pods.

Since Longhorn cannot control how users configure the NAD (maybe providing with own CNI plugin) and the host environment, this setting provides an option for users who managed to use the host network namespace with the storage network. In this scenario, the user does not need to restart the workload pods if the CSI plugin pod restarts unexpectedly.

enhancements/20240522-storage-network-for-rwx-volumes.md
c3y1huang: This behavior remains the same as for the other dangerous zone settings.

c3y1huang:
> I checked some other solutions that might be facing the same situation. Found an interesting article below: https://github.com/rook/rook/blob/master/design/ceph/multus-network.md#new-holder-daemonset
> 
> I am still checking what is the final solution with the multus-network on their implementation because this design says `not finalized`.

The holder pod idea seems quite similar what @ejweber had in mind. However, there is a downside: we cannot guarantee that the holder pod won't be restarted, and managing this just adds extra complexity.

It looks like Rook Ceph decided to leave this responsibility to the user when using Multus.
- [Multus-enabled clusters to begin "holder" pod deprecation.](https://github.com/rook/rook/blob/53be5ce1e928674a5e1866dcb68370cdfba17d61/ROADMAP.md?plain=1#L23)
- [Multus Prerequisites](https://rook.io/docs/rook/latest/CRDs/Cluster/network-providers/#multus-prerequisites)

innobead:
Discussed with @c3y1huang. Need to update the test cases for the upgrade. The rule is that we should not interrupt users' workloads during the upgrade if the storage network has been set up before the upgrade.

enhancements/20240522-storage-network-for-rwx-volumes.md
```diff
@@ -0,0 +1,153 @@
...
+1. **Regression Testing:** Cover both cluster network and storage network scenarios to ensure all functionalities work as expected.
+1. **Resilience Testing:** Verify data accessibility after CSI plugin pod restarts for both cluster network and storage network scenarios.
+1. **Cross-platform Testing:** Ensure RWX volume functionality on non-traditional operating systems such as Talos Linux and Container-Optimized OS for compatibility.
+1. **Upgrade Testing:** Validate that existing RWX volumes function correctly after an upgrade.
+
+### Upgrade strategy
+
+- **CSI Plugin Pod:** If the storage network is configured before upgrading, the CSI plugin pod will be annotated accordingly after the upgrade.
+
+- **Existing RWX Volumes:** Upgrading from previous version with existing volumes will be unaffected since the NFS client mount was established in the host network namespace that is not affected by the CSI plugin pod restart. When the workload pod restarts, the Kubelet will automatically reuse the existing client connection.
```
innobead, line 145: If the existing RWX volumes restart, will it be restarted with the new storage network applied? If yes, the original consumer workloads will retry to connect with an invalid connection, correct?

What's the expected behavior here? I assumed the workloads would be somehow killed (by respecting `SettingNameAutoDeletePodWhenVolumeDetachedUnexpectedly`) to restart so that they could connect the new share manager through the storage network. However, the connection resuming of the existing client connection will not work in this special case.

enhancements/20240522-storage-network-for-rwx-volumes.md
```diff
@@ -0,0 +1,153 @@
...
+
+- **New RWX Volumes:** When the storage network is configured before upgrading, New RWX volumes created after the upgrade will be applied with the storage network.
+
+- **Future Version Upgrades:** The Longhorn Kubernetes pod controller will trigger the deletion of workload pods, causing Kubelet to restart the workload pod. This process will remount NFS entries when the workload uses RWX volumes with the storage network.
```
innobead, line 149: Can you explain when `Future Version Upgrades` will happen? If the storage network is enabled before upgrade, after upgrade, the workloads should continue working w/o interruption.

enhancements/20240522-storage-network-for-rwx-volumes.md
c3y1huang: Here is the current behavior: https://github.com/longhorn/longhorn-tests/blob/114b39b9a607ad6058687b70ccdf3d07b145eb80/docs/content/manual/release-specific/v1.7.0/test-storage-network-support-for-rwx-volume.md

enhancements/20240522-storage-network-for-rwx-volumes.md
c3y1huang: For the upgrade after v1.7.0 (1.7.1, 1.7.2 ...), when an upgrade involves change to the CSI plugin DaemonSet, existing RWX volume workload pods that that using NFS share mounts on the storage-network will be restarted as the CSI plugin pods are restarted.

innobead: Clear. Thanks.

innobead (approved):

enhancements/20240522-storage-network-for-rwx-volumes.md
innobead: @c3y1huang needs to double check the behavior. If needed, we need to update this LEP.

c3y1huang: https://github.com/longhorn/longhorn/pull/8679/files

Issue #8376, opened 2024-04-17
[IMPROVEMENT] Pre-pull images (share-manager image and instance-manager image) on each Longhorn node


derekbit:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

Pre-pull on each Longhorn node 
- share-manager image for speeding up RWX volume failover
- instance-manager image 
## Additional context

https://suse.slack.com/archives/C02DR3N5T24/p1713318153020089?thread_ts=1713265309.479039&cid=C02DR3N5T24

derekbit:
cc @james-munson

derekbit:
Do we need a CRD for share-manager image like `engineImage`?
Further, we can generalize the `engineImage` CRD to `image` CRD that can manage the pre-pull images.

shuo-wu:
> Further, we can generalize the engineImage CRD to image CRD that can manage the pre-pull images.

Agree with this. We can launch a general image-pull daemonset for engine, instance-manager, and share-manager.

ejweber:
I was filing this as a bug, but we can probably just handle it here since the issue isn't closed yet:

## Describe the bug

After https://github.com/longhorn/longhorn/pull/8778 merged, there are now two containers in the longhorn-manager DaemonSet spec instead of one. This always causes the longhorn-post-upgrade job to fail.

## To Reproduce

1. `helm install -n longhorn-system longhorn ./chart`
2. `helm upgrade -n longhorn-system longhorn ./chart`

## Expected behavior

The Helm upgrade completes successfully.

## Additional context

```
eweber@laptop:~/longhorn> kl logs longhorn-post-upgrade-n67xg
warning: GOCOVERDIR not set, no coverage data emitted
time="2024-07-01T19:12:30Z" level=info msg="Running post-upgrade..." func=main.main.PostUpgradeCmd.func6 file="post_upgrade.go:41"
W0701 19:12:30.507888       1 client_config.go:659] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
time="2024-07-01T19:12:30Z" level=warning msg="found 2 containers in manager spec" func="app.(*postUpgrader).waitManagerUpgradeComplete" file="post_upgrade.go:99"
time="2024-07-01T19:12:30Z" level=warning msg="found 2 containers in manager spec" func="app.(*postUpgrader).waitManagerUpgradeComplete" file="post_upgrade.go:99"
[...]

PR #8648, opened 2024-05-27
feat(LEP): pre-pull share-manager, instance-manager images

mantissahz:

### DIFF
--- a/enhancements/20240516-pre-pull-images.md
+++ b/enhancements/20240516-pre-pull-images.md
@@ -1,0 +1,162 @@
+# Pre-Pull Images
+
+## Summary
+
+We will pre-pull the share-manager and instance-manager images on every worker node to speed up the startup of share-managers and instance-managers.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/8376
+
+## Motivation
+
+### Goals
+
+- Pre-pull the share-manager image on all worker nodes.
+- Pre-pull the instance-manager image on all worker nodes.
+
+### Non-goals [optional]
+
+- Not convert the `EngineImage` to the `Image`.
+
+## Proposal
+
+- A DaemonSet to pre-pull images
+- A global setting `start-pre-pull-manager-images`. Users can determine whether or not pre-pulling images. Default is `true`.
+
+### User Stories
+
+With the pre-pull images mechanism, users can deploy RWX volume and start to use volumes faster. It's also useful for [Share manager HA mechanism](https://github.com/longhorn/longhorn/issues/6205).
+
+There is the [garbage collection of unused containers and images feature](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#containers-images) of kubernetes. The `kubelet` performs garbage collection on unused images every two minutes and on unused containers every minute so the image controller will keep the DaemonSet to prevent the pre-pull images from being deleted by the garbage collection of unused containers and images.
+
+#### Story 1: Users don not want pre-pull images
+
+If users do not use RWX volumes, it will be unnecessary to pull the share manager image and occupy the disk usage.
+
+- Users can set the global setting `start-pre-pull-manager-images` false to stop pre-pulling share/instance manager images. The pre-pulling images DaemonSets will then be deleted.
+
+### User Experience In Detail
+
+Users can edit this global setting by Longhorn UI or the `kubectl edit` command.
+Example:
+```bash
+~# kubectl -n longhorn-system edit setting start-pre-pull-manager-images
+apiVersion: longhorn.io/v1beta2
+kind: Setting
+metadata:
+  ...
+  name: start-pre-pull-manager-images
+  namespace: longhorn-system
+...
+value: "true"
+```
+
+### API changes
+
+`None`
+
+## Design
+
+### Implementation Overview
+
+- New custom resource definition `Image`:
+
+  ```golang
+  type ImageState string
+
+  ImageStateDeploying = ImageState("deploying")
+  ImageStateDeployed  = ImageState("deployed")
+  ImageStateError     = ImageState("error")
+  ImageStateUnknown   = ImageState("unknown")
+
+  // ImageSpec defines the desired state of the Longhorn image
+  type ImageSpec struct {
+    // ImageURL indicates the image URL.
+    ImageURL string `json:"instanceManagerImage"`
+  }
+
+  // ImageStatus defines the observed state of the Longhorn image
+  type ImageStatus struct {
+    // OwnerID indicates the controller node that handles the image.
+    OwnerID string `json:"ownerID"`
+    // State indicates the Image state.
+    State ImageState `json:"state"`
+    // NodeDeploymentMap indicates the nodes that the image has been deployed on.
+    NodeDeploymentMap map[string]bool `json:"nodeDeploymentMap"`
+  }
+
+  // Image is where Longhorn stores image objects.
+  type Image struct {
+    Spec   ImageSpec   `json:"spec,omitempty"`
+    Status ImageStatus `json:"status,omitempty"`
+  }
+  ```
+
+  - `ImageState`:
+    - `ImageState` will be `ImageStateDeploying: deloying` when creating the DaemonSet. It will become `ImageStateDeployed: deployed` until the image controller detects that all worker nodes have the containers of the pod ready.
+    - `ImageStateUnknown` will be used when users set the global setting `start-pre-pull-manager-images` false.
+
+  - `NodeDeploymentMap` only synchronizes the containers status of the DaemonSet pod of the image; therefore, it can not detect if the share/instance manager images are deleted for some reason.
+    - Add `KubeNodeInformer.AddEventHandlerWithResyncPeriod` in the image controller to handle cases where share/instance manager images deleted on nodes.
+    - Add `SettingInformer.AddEventHandlerWithResyncPeriod` in the image controller to handle cases where the setting `start-pre-pull-manager-images` is changed.
+
+- New controller that handles the resource `Image`:
+
+  ```golang
+  NewImageController(...) (*ImageController, error) { // others image URLs
+    ic := &ImageController{
+      baseController: newBaseController("longhorn-image", logger),
+
+      namespace:      namespace,
+      controllerID:   controllerID,
+      serviceAccount: serviceAccount,
+
+      ...,
+    }
+
+    // add ImageInformer.AddEventHandlerWithResyncPeriod
+    // add SettingInformer.AddEventHandlerWithResyncPeriod
+    // add KubeNodeInformer.AddEventHandlerWithResyncPeriod
+    ...
+
+    return ic, nil
+  }
+
+  syncImage(imageName string) error {
+    // Check image owner
+    ...
+    // Check the global setting `start-pre-pull-manager-images`
+    ...
+    // Get the image DaemonSet and create the DaemonSet if not found or update the DaemonSet if the instance/share manager image URLs are upgraded.
+      // A pod with containers in the DaemonSet to pull the instance and share manager images.
+    ...
+    // Synchronize the nodes map with images deployed on the nodes.
+    ...
+    // Keep the image DaemonSet around after the image state is `deployed`.
+  }
+  ```
+
+### Test plan
+
+1. Fresh install:
+   1. The `Image` resources are created.
+   2. An image DaemonSet for pulling share and instance manager images is created.
+   3. The `Image` resources states are `deploying`.
+   4. All pods of the image DaemonSet for share and instance managers are created and ready on worker nodes.
+   5. The `Image` resources states are `deployed`.
+2. Upgrade Longhorn:
+   1. The old `Image` resources are deleted and the `Image` resources are created.
+   2. An image DaemonSet for pulling share and instance manager images is re-created.
+   3. The `Image` resources states are `deploying`.
+   4. All pods of the image DaemonSet for new share and instance managers are created and ready on worker nodes.
+   5. The `Image` resources states are `deployed`.
+
+### Upgrade strategy
+
+1. Delete old `Image` resources.
+2. Create new `Image` resources with new instance and share manager image URLs and the image controller will update the DaemonSet to start to pre-pull the images.
+
+## Note [optional]
+
+`None`


innobead (commented):
There are a few questions below.

- If the image is deleted manually on a node, what's the status of the image? how will this be handled? I assume users need to delete the pod on the node to make it repull. Also, how can we ensure NodeDeploymentMap up to date if the downloaded image is deleted on one node and the status can be updated accordingly to NodeDeploymentMap?
- Can we delete the daemonset if all images are ready? This can save unnecessary pods running but users can still check the image readyness status.
- I feel this is almost the same as the current engine image. Can we just migrate the engine image to this new CR if this is a minor feature.

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,131 @@
...
+  ImageStateDeploying = ImageState("deploying")
+  ImageStateDeployed  = ImageState("deployed")
+  ImageStateError     = ImageState("error")
+
+  ImageTypeInstanceManager = ImageType("instance-manager")
+  ImageTypeShareManager    = ImageType("share-manager")
+
+  ImageConditionTypeReady = "ready"
+  ImageConditionTypeReadyReasonDaemonSet = "daemonSet"
+  ImageConditionTypeReadyReasonBinary    = "binary"
```
shuo-wu, line 51: How does Longhorn utilize this condition?

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  // ImageStatus defines the observed state of the Longhorn image
+  type ImageStatus struct {
+    // OwnerID indicates the controller node that handles the image.
+    OwnerID string `json:"ownerID"`
+    // State indicates the Image state.
+    State ImageState `json:"state"`
+    // Incompatible indicates whether the Image can deploy on the Longhorn.
+    Incompatible bool `json:"incompatible"`
```
shuo-wu, line 68: I am not sure if the image pre-pull feature needs to understand this... 
Could you talk more about the goal and non-goal in the LEP?

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,131 @@
...
+  // ImageSpec defines the desired state of the Longhorn image
+  type ImageSpec struct {
+    // Image indicates the image URL.
+    Image string `json:"image"`
+    // Type indicates the image type, such as the share manager or instance manager image.
+    Type ImageType `json:"type"`
+  }
+
+  // ImageStatus defines the observed state of the Longhorn image
+  type ImageStatus struct {
```
derekbit, line 62: As the discussion in today's sync-up meeting, we can add all `EngineImage`'s status fields here.

derekbit: `status.refCount` might be useful for automatically deleting old share-manager and instance-manager images. That means, we need a auto-deletion mechanism as well.

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,131 @@
...
+    // Get the DaemonSet by the image Name and create the DaemonSet if not found.
+    ...
+    // Synchronize the nodes map with images deployed on the nodes.
+    ...
+  }
+  ```
+
+### Test plan
+
+1. Fresh install or upgrade Longhorn
```
derekbit, line 118: Will new `Image` resources for share-manager image and so on be automatically while upgrading longhorn-system?

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,131 @@
...
+2. Share and instance manager `Image` resources are created.
+3. Image Daemonsets of share and instance manager are created.
+4. `Image` resources state are `deploying`.
+5. All Image Daemonsets of share and instance manager pods are created.
+6. `Image` resources state are `deployed`.
+
```
derekbit, line 124: Need to handle the outdated images. For example, an `Image` resource for an old share-manager image is not referenced by any share-manager pods, the resource should be automatically deleted.

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,131 @@
...
+    // Synchronize the nodes map with images deployed on the nodes.
+    ...
+  }
+  ```
+
+### Test plan
+
+1. Fresh install or upgrade Longhorn
+2. Share and instance manager `Image` resources are created.
+3. Image Daemonsets of share and instance manager are created.
```
ejweber, line 120: TBH, this seems like a lot of overhead in the cluster for the somewhat small benefit of having the images pre-pulled and ready to go. What if we need multiple instance-manager and engine-images simultaneously? That's quite a few DaemonSets (and pods).

Would it be possible to maintain only a single DaemonSet with multiple containers? There could be one container for each image we still want deployed on the cluster. When a new Image resource is created, we could just modify the spec of the one DaemonSet, causing a rolling restart. It is fine if the DaemonSet restarts, since it won't really do anything but exist.

enhancements/20240516-pre-pull-images.md
PhanLe1010: Agree. This will be very similar to the engine-image CR.

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,131 @@
...
+  ImageConditionTypeReady = "ready"
+  ImageConditionTypeReadyReasonDaemonSet = "daemonSet"
+  ImageConditionTypeReadyReasonBinary    = "binary"
+
+  // ImageSpec defines the desired state of the Longhorn image
+  type ImageSpec struct {
+    // Image indicates the image URL.
+    Image string `json:"image"`
+    // Type indicates the image type, such as the share manager or instance manager image.
+    Type ImageType `json:"type"`
```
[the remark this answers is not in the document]
mantissahz: blindly pull the image for everything worker node via a DaemonSet

enhancements/20240516-pre-pull-images.md
mantissahz: Add the descriptions for the condition. Thanks.

enhancements/20240516-pre-pull-images.md
mantissahz: Removed the field. Thanks.

enhancements/20240516-pre-pull-images.md
mantissahz: Do we still need `refCount`?

enhancements/20240516-pre-pull-images.md
mantissahz: Separate the test case:
For upgrade, it will create new images according to the system provides and update the corresponding DaemonSets. And delete the old image reosurces.

enhancements/20240516-pre-pull-images.md
mantissahz: Separated the test case. Thanks
Yes, old image resources should be deleted automatically.

enhancements/20240516-pre-pull-images.md
mantissahz: Modified the part of the image controller to update the DaemonSet when the share/instance images are changed.

innobead (changes requested):
Can we just delete the daemsonset after all nodes already have pulled the image at the first time? 

After pulling images, the DS is still around like the current engine image. I feel we should change the behavior to make the pulling task dynamically instead of still occupying the system resources after the goal is achieved.

ejweber:
> Can we just delete the daemsonset after all nodes already have pulled the image at the first time?
>
> After pulling images, the DS is still around like the current engine image. I feel we should change the behavior to make the pulling task dynamically instead of still occupying the system resources after the goal is achieved.

This was discussed on Thursday before you were able to join. If we do as you suggest, there are two possibilities:

1. Our controller watches for the images to eventually be garbage collected by Kubernetes or the container-runtime on any cluster node (or for new nodes in the cluster that don't have the images). When it detects this condition, it recreates the DaemonSet. We can either reconcile on all `nodes.kubernetes.io` changes or reconcile `nodes.kubernetes.io` periodically to see what images are available on what nodes. The LEP as it is currently written has the following:

```
- Add `KubeNodeInformer.AddEventHandlerWithResyncPeriod` in the image controller to handle cases where share/instance manager images deleted on nodes.
```

2. Our controller doesn't do that. When our images are garbage collected or new nodes join the cluster, some nodes don't have the images we want.

I preferred maintaining the DaemonSet to offload all "thinking" to the Kubernetes DaemonSet controller (and probably not monitoring `nodes.kubernetes.io` at all). For users that didn't like the extra DaemonSet, they could just disable the setting. IIRC, that was the consensus at the end of the meeting, but I don't think opinions were strong either way.

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,157 @@
...
+    Status ImageStatus `json:"status,omitempty"`
+  }
+  ```
+
+  - `ImageCondition` and `ImageState`:
+
+    - `ImageCondition` will be `ImageStateDeploying: deloying`  and `ImageConditionTypeReady: false` with the DaemonSet , and the reason will be `ImageConditionTypeReadyReasonDaemonSet`. It will become `ImageStateDeployed: deployed` and ``ImageConditionTypeReady: true` until the image controller detects that all worker nodes have the containers of the pod ready.
+    - `ImageStateUnknown` will be used when users set the global setting `start-pre-pull-manager-images` false.
+
+  - `NodeDeploymentMap` only synchronizes the containers status of the DaemonSet pod of the image; therefore, it can not detect if the share/instance manager images are deleted for some reason.
```
ejweber, line 94: I don't think it is possible to delete an image if it is in use by a running container. Can you check it?

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,157 @@
...
+    // Synchronize the nodes map with images deployed on the nodes.
+    ...
+  }
+  ```
+
+### Test plan
+
+1. Fresh install:
+   1. Share and instance manager `Image` resources are created.
+   2. Image Daemonsets of share and instance manager are created.
```
ejweber, line 139: Are we still planning on having a DaemonSet for each type of image? Wouldn't one that covers both suffice?

mantissahz: Yes, we could have containers for images in a DaemonSet. I will test it and update the LEP. Thanks.

enhancements/20240516-pre-pull-images.md
mantissahz: Yes, I could delete the share-manager image manually when the pod is running.

```bash
master:~# root@rancher60-master:~# kl get no rancher60-worker1 -oyaml|grep share
    - docker.io/longhornio/longhorn-share-manager@sha256:ca0956077ab9a83e93704d3bc28ea664d2b154064e45d0145cee2bd58857ead6
    - docker.io/longhornio/longhorn-share-manager:master-head

master:~# # A couple of minutes after deleting the share-manager image.
master:~# root@rancher60-master:~# kl get no rancher60-worker1 -oyaml|grep share
master:~# kl get pod|grep mi
image-smi-5ceacedd-jl9q2                            1/1     Running   0          4m50s
image-smi-5ceacedd-5ndgs                            1/1     Running   0          4m50s
image-smi-5ceacedd-8mnb9                            1/1     Running   0          4m50s
image-imi-2ef43be6-58d7d                            1/1     Running   0          4m50s
image-imi-2ef43be6-njdcm                            1/1     Running   0          4m50s
image-imi-2ef43be6-8j98v                            1/1     Running   0          4m50s
master:~# kl get pod image-smi-5ceacedd-5ndgs -oyaml
apiVersion: v1
kind: Pod
metadata:
  ...
spec:
  ...
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchFields:
          - key: metadata.name
            operator: In
            values:
            - rancher60-worker1
  ...
status:
  ...
  containerStatuses:
  - containerID: containerd://bd2feadb18eae93ae22e9049a5c4bcd3c77fa1984e3c59d84443fbe6279ad84b
    image: docker.io/longhornio/longhorn-share-manager:master-head
    imageID: docker.io/longhornio/longhorn-share-manager@sha256:ca0956077ab9a83e93704d3bc28ea664d2b154064e45d0145cee2bd58857ead6
    lastState: {}
    name: image-smi-5ceacedd
    ready: true
    restartCount: 0
    started: true
    state:
      running:
        startedAt: "2024-06-04T01:14:25Z"
  ...

worker1:~# # delete the share image on the worker node
worker1:~# k3s crictl rmi docker.io/longhornio/longhorn-share-manager:master-head
Deleted: docker.io/longhornio/longhorn-share-manager:master-head
worker1:~#
worker1:~# k3s crictl images |grep  longhorn-share-manager:master-head
```

enhancements/20240516-pre-pull-images.md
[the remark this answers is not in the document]
shuo-wu: There is no need to know the image type as Longhorn is blindly pulling all required images and periodically checking them if necessary.

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,157 @@
...
+  ImageConditionTypeReadyReasonDaemonSet = "daemonSet"
+
+  // ImageSpec defines the desired state of the Longhorn image
+  type ImageSpec struct {
+    // Image indicates the image URL.
+    Image string `json:"image"`
+    // Type indicates the image type, such as the share manager or instance manager image.
+    Type ImageType `json:"type"`
+    // NeedCreateDaemonSet indicates whether the image controller needs to create the corresponding DaemonSet.
+    NeedCreateDaemonSet bool `json:"needCreateDaemonSet"`
```
shuo-wu, line 67: As we discussed, shouldn't Longhorn pull images by creating the corresponding daemonset? Is there an exception?

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  // ImageStatus defines the observed state of the Longhorn image
+  type ImageStatus struct {
+    // OwnerID indicates the controller node that handles the image.
+    OwnerID string `json:"ownerID"`
+    // State indicates the Image state.
+    State ImageState `json:"state"`
+    // Conditions indicate the state and message if the errors occurred.
+    Conditions []Condition `json:"conditions"`
```
shuo-wu, line 77: Seems redundant.

enhancements/20240516-pre-pull-images.md
shuo-wu: Then keeping the DaemonSet running seems meaningless. Longhorn has to periodically verify if an image exists on a node by checking the corresponding `kubernetesNode.status`

enhancements/20240516-pre-pull-images.md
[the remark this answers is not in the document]
mantissahz: Removed the `Type`. Thanks

enhancements/20240516-pre-pull-images.md
mantissahz: The controller will delete the DaemonSet after the image state is `deployed`.
Thanks for all the comments.

enhancements/20240516-pre-pull-images.md
mantissahz: Yes, removed the `Conditions`. Thanks.

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,143 @@
...
+  ImageStateDeployed  = ImageState("deployed")
+  ImageStateError     = ImageState("error")
+  ImageStateUnknown   = ImageState("unknown")
+
+  // ImageSpec defines the desired state of the Longhorn image
+  type ImageSpec struct {
+    // InstanceManagerImage indicates the instance manager image URL.
+    InstanceManagerImage string `json:"instanceManagerImage"`
+    // ShareManagerImage indicates the share manager image URL.
+    ShareManagerImage string `json:"shareManagerImage"`
```
shuo-wu, line 57: Can we blindly use one field `ImageURL` instead?

enhancements/20240516-pre-pull-images.md
mantissahz: Removed this field. Thanks

enhancements/20240516-pre-pull-images.md
mantissahz: Could you give an example? Would it be like this?

```golang
ImageURL := "longhornio/longhorn-instance-manager:master-head,longhornio/longhorn-share-manager:master-head"
```

enhancements/20240516-pre-pull-images.md
mantissahz: Updated. Thanks

enhancements/20240516-pre-pull-images.md
shuo-wu: I am fine with using either multiple CRs or a `ImageURLMap` with a single CR.

mantissahz: Choose a DaemonSet with multiple CRs. Thanks

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Goals
+
+- Pre-pull the share-manager image on all worker nodes.
+- Pre-pull the instance-manager image on all worker nodes.
+
+### Non-goals [optional]
+
+- Not convert the `EngineImage` to the `Image`.
+- Not delete the DaemonSet after images are deployed.
```
shuo-wu, line 21: This statement is confusing... Is this what we plan to do but not for now?  I prefer to directly remove this line

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,165 @@
...
+- A DaemonSet to pre-pull images
+- A global setting `start-pre-pull-manager-images`. Users can determine whether or not pre-pulling images. Default is `true`.
+
+### User Stories
+
+With the pre-pull images mechanism, users can deploy RWX volume and start to use volumes faster. It's also useful for [Share manager HA mechanism](https://github.com/longhorn/longhorn/issues/6205).
+
+There is the [garbage collection of unused containers and images feature](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#containers-images) of kubernetes. The `kubelet` performs garbage collection on unused images every two minutes and on unused containers every minute.
+
+- Keep the DaemonSet around with the global setting `start-pre-pull-manager-images` enabled to prevent the DamonSet from toggling on and off by detecting images deleted on worker nodes.
```
shuo-wu, line 34: > Keep the DaemonSet around ......

This statement is a little bit mieleading as well.

enhancements/20240516-pre-pull-images.md
mantissahz: You are right. Removed. Thanks.

enhancements/20240516-pre-pull-images.md
mantissahz: Modified to describe why we should keep the DaemonSet because of the garbage collection of unused container images. Thanks.

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,162 @@
...
+- Pre-pull the instance-manager image on all worker nodes.
+
+### Non-goals [optional]
+
+- Not convert the `EngineImage` to the `Image`.
+
+## Proposal
+
+- A DaemonSet to pre-pull images
+- A global setting `start-pre-pull-manager-images`. Users can determine whether or not pre-pulling images. Default is `true`.
```
derekbit, line 25: Can we use `pre-pull-manager-images` for the simplicity?

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,162 @@
...
+
+### User Stories
+
+With the pre-pull images mechanism, users can deploy RWX volume and start to use volumes faster. It's also useful for [Share manager HA mechanism](https://github.com/longhorn/longhorn/issues/6205).
+
+There is the [garbage collection of unused containers and images feature](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#containers-images) of kubernetes. The `kubelet` performs garbage collection on unused images every two minutes and on unused containers every minute so the image controller will keep the DaemonSet to prevent the pre-pull images from being deleted by the garbage collection of unused containers and images.
```
derekbit, line 31: The kubelet's garbage collection period is set to just two minutes. Does this cause the engine image deployed on each node to be repeatedly pulled and deleted?

If not, can we prevent the unnecessary garbage collection for share-manager images and pin the images?

enhancements/20240516-pre-pull-images.md
innobead: How about pre-pull-images?

derekbit:
After rethinking the whole picture of the feature, the objectives of the feature include
- Reduce Startup Time: Decrease the startup time for instance-manager and share-manager pods.
- Image Management: Pin the instance-manager and share-manager images on worker nodes. If disk usage reaches `HighThresholdPercent`, kubelet's image garbage collection will remove unused images. When the pre-pull feature is enabled, the image controller will repull the image. This cleanup and repull cycle will repeat continuously. Then, the potential issues are
  - Resource Waste: Unnecessary cleanup and repulling consume computing and networking resources.
  - Failover Delays: If a share-manager pod fails over to a node where the share-manager image has just been cleaned up, the failover time may increase significantly.

Therefore, it might be better to pin the share-manager and instance-manager images for the current Longhorn version, similar to the engine image.

mantissahz:
As discussed with @derekbit, we would use a simple way to pre-pull the share-manager image.

1. Pull the share-manager image in the `initContainer` of the longhorn-manger DaemonSet.
2. Use the command `ctr -n k8s.io images label [share-manager-image-url] io.cri-containerd.pinned=pinned` or APIs(if it has) to set the attribute `pinned` of the image true.
3. Unpin the old share-manager image by the command or APIs.
4. When uninstallation, unpin all share-manager images.

Are there any concerns if we use this way to pre-pull the images?
cc @longhorn/dev

ejweber:
> As discussed with @derekbit, we would use a simple way to pre-pull the share-manager image.
> 
> 1. Pull the share-manager image in the `initContainer` of the longhorn-manger DaemonSet.
> 2. Use the command `ctr -n k8s.io images label [share-manager-image-url] io.cri-containerd.pinned=pinned longhornio=true` or APIs(if it has) to set the attribute `pinned` of the image true.
> 3. Unpin the old share-manager image with the label `longhornio=true` by the command or APIs.
> 4. When uninstallation, unpin all share-manager images with the label `longhornio=true`.
> 
> Therefore we don't need to add a new CRD and a new controller to handle it.
> 
> Are there any concerns if we follow this way to pre-pull the images? cc @longhorn/dev

In general, it sounds like pinning is exactly the behavior we want. However, I'm not 100% sure I understand how this would work in practice:

1. Do we expect to be able to run `ctr ...` commands that take effect on the host from the longhorn-manager (or its init) container? Will we need some special configuration on the container to do this? Will we need to be able to handle `containerd` sockets in arbitrary locations? (IIRC, RKE2 uses `/run/k3s/containerd/containerd.sock` and it's generally required to [pass that address](https://gist.github.com/superseb/3b78f47989e0dbc1295486c186e944bf#ctr) when using `ctr` commands.
2. Is this approach container runtime agnostic? (I'm really not sure here, but [many different runtimes](https://kubernetes.io/docs/setup/production-environment/container-runtimes/) are supported for use with Kubernetes, and it seems this approach might only work with containerd.)

PhanLe1010:
Adding to the feedback from @ejweber , is there any downside if we just create 1 container for share-manager image and 1 container for instance-manager image inside longhorn-manager pod and let them go to eternally sleep instead of the init container ? That way we will automatically achieve pin and unpin effect without having doing anything, won't we?

ejweber:
> Adding to the feedback from @ejweber , is there any downside if we just create 1 container for share-manager image and 1 container for instance-manager image inside longhorn-manager pod and let them go to eternally sleep instead of the init container ? That way we will automatically achieve pin and unpin effect without having doing anything, won't we?

Interesting! This wouldn't have been possible with the original design (which folded in EngineImage), but with the current direction of this LEP, it might be a good solution.

Do the extra containers use resources?

Probably no. Their image layers have to be on the node anyway (that's the point of the LEP), and we can theoretically run them with a different entry point (e.g. `sleep infinity`) so we don't have to load the processes we would otherwise run into memory.

james-munson:
The snags with container and pod lifetime are in this blog post describing the "native sidecar" feature:  https://kubernetes.io/blog/2023/08/25/native-sidecar-containers/#how-did-users-get-sidecar-behavior-before-1-28

mantissahz:
Track on this PR https://github.com/longhorn/longhorn/pull/8775.

PR #8652, opened 2024-05-28
fix(image): introduce a new CRD `Image`

mantissahz:


mantissahz:
we do not need this CRD now.

Issue #8563, opened 2024-05-14
[IMPROVEMENT] Disable revision counter by default


derekbit:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

The purpose of revision counter is to help choose one replica containing the latest data. However, the mechanism introduces a significant performance drop in Longhorn volumes.

After revisiting the design, why we think disabling RC doesn't matter is
- Without `sync`, filesystem and block layers don't guarantee data integrity. Therefore, choosing any replica for rebuilding should be fine
- After `sync`, all IOs are flushed. All replicas should have identical data.

## Additional context

https://suse.slack.com/archives/C02EDCA93RA/p1715712709278379?thread_ts=1715659059.436039&cid=C02EDCA93RA

derekbit:
@shuo-wu @PhanLe1010 @WebberHuang1118 @Vicente-Cheng Any thoughts for the improvement?

PhanLe1010:
We discussed this internally. The discussion so far is that the reason we can safely disable RC are:
1. Without sync, filesystem and block layers don't guarantee data integrity. Therefore, choosing any replica for rebuilding should be fine
1. After sync, all IOs are flushed. All replica should have identical data.

There is one concern we need to solve which is: 

> Even though the filesystem doesn't guarantee data integrity, I think Longhorn must guarantee the data consistency across different replicas. The current revision counter try (best-effort) to compare the RC of replicas when engine starts and rebuild the replica data if there are mismatch. That mechanism is not perfect but at least it is working maybe 70-80% of the time? Now, if we want to remove RC, should we come up with a new mechanism to detect replicas' data mismatch during engine starts?

And the idea for this one is:
> we need to improve the auto-salvage part so that there will be only one replica salvaged when all replicas/engine crash. This should eliminate the mismatching issue.

Btw, we will have this in 1.7.0 instead of stable release 1.6.x

Thanks @innobead  @derekbit @shuo-wu for the discussion!

shuo-wu:
[The auto salvage feature should pick up one reusable replica only](https://github.com/longhorn/longhorn-manager/blob/41ab9a5e853097188e39c219ca2e3dc8b759f98c/controller/volume_controller.go#L1402-L1413). Considering the concurrent replica write out-of-order issue, this is required even if the revision counter is enabled.

WebberHuang1118:
Hi @PhanLe1010 
[...]

PR #8664, opened 2024-05-29
Disable revision counter by default

PhanLe1010:
longhorn/longhorn#8563


derekbit (approved):
LGTM

ejweber (approved):
I was briefly confused by the addition of another thing called `disableRevisionCounter`. I see now that we have for a long time had a `disableRevisionCounter` setting that only applied to UI volumes and a StorageClass field that fed into the volume spec field. Now, we are putting that field into the default StorageClass.

I think the StorageClass change here is not strictly necessary because we switched the default (if it is not specified) in https://github.com/longhorn/longhorn-manager/blob/e37452f10f886d853068a9395970d04a9c0b165b/csi/util.go#L176? I'm not opposed to it at all. I'm just trying to increase my understanding.

(As an aside, this seems to be the only such StorageClass parameter named differently than its corresponding volume spec field. It's a little weird IMO, but we can't really do anything about it now.)

LGTM.

PhanLe1010:
> I think the StorageClass change here is not strictly necessary because we switched the default (if it is not specified) in https://github.com/longhorn/longhorn-manager/blob/e37452f10f886d853068a9395970d04a9c0b165b/csi/util.go#L176? I'm not opposed to it at all. I'm just trying to increase my understanding.

I think it might be needed because otherwise user cannot enable RC for the default SC

PR #8679, opened 2024-05-31
feat(lep): storage network for RWX volumes: upgrade scenario

c3y1huang:

Issue longhorn/longhorn#8184

The current LEP doesn't fully address how pre-existing RWX volume workloads will behave during and after the upgrade to Longhorn v1.7.0 when the cluster has the storage network configured before the upgrade. This PR proposes to complete the upgrade scenarios.

- Behavior: https://github.com/longhorn/longhorn-tests/pull/1921

### DIFF
--- a/enhancements/20240522-storage-network-for-rwx-volumes.md
+++ b/enhancements/20240522-storage-network-for-rwx-volumes.md
@@ -44,7 +44,7 @@ Users will be able to create RWX volumes that uses the storage network for data 
 - **Headless Service:** RWX volume will use a headless service and a custom endpoint, removing dependency on ClusterIP, which is not suitable for Multus networking.
 - **Custom Mounter Replacement:** The custom mounter uses `hostPID` and host `proc` namespace will be replaced with running the CSI plugin with `hostNetwork: true`.
 - **Workload Pod Restart:** When a CSI plugin pod restarts, the NFS share mountpoint becomes unavailable, rendering any attempt to access it unresponsive. User can manually restart the workload pod, or they can enable the `auto-delete-pod-when-volume-detached-unexpectedly` setting. This allows Longhorn to delete the associated workload pod when it is using the storage-network. If the pod is managed by a controller, the kubelet will create a new pod. This allows the CSI controller to handle volume remounting through the CSI node server.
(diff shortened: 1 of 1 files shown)


enhancements/20240522-storage-network-for-rwx-volumes.md
```diff
@@ -142,7 +144,13 @@ If the RWX volume does not use the storage network, or the healess service, work
...
+
+- **Node Reboot:** Rebooting a node after the upgrade can impact pre-existing workloads. Reboots remove existing NFS mount entries. When the workload pod restarts, the CSI plugin recreates the mount within its own container network namespace. This means the NFS client connection becomes tied to the CSI plugin pod's lifecycle from this point forward, and Longhorn or the user becomes responsible for handling remounting the volume is the CSI plugin pod namespace is gone.
+
+- **Handling Remounts By Setting:** Two scenarios can cause the `auto-delete-pod-when-volume-detached-unexpectedly` setting to apply to a pre-existing workload, even though they are using the cluster network:
+	1. Node reboot: As mentioned above, a node reboot removed existing NFS client mount entries. The CSI plugin creates new connection in its own namespace when the workload pod restarts.
```
innobead, line 152: When the node reboots, the pod should be recreated eventually, why the case will respect the setting?

innobead:
LGTM except for one question.

We need to have a n important note to mention the upgrade scenario to  let users know when their workloads would be disrupted after the setting is applied.

enhancements/20240522-storage-network-for-rwx-volumes.md
c3y1huang: 1. In Longhorn v1.6.2, an RWX volume is created and mounted using the ClusterIP in the host network namespace.
1. Upgrade to v1.7.0, the workload is still using the NFS client connection in the host network.
1. Reboot the node.
1. The workload pod restarted after the node reboot. This means the RWX volume is now mounted with ClusterIP in the CSI plugin container network namespace.
1. Cluster plugin pod restart.
1. RWX volume workload client mount entry is unresponsive.
1. Restart the RWX volume workload pod.
1. RWX volume workload client mount entry is remounted and becomes responsive.

The setting addresses the dangling mount issue (step 6) after the CSI plugin pod restart (step 5) following a node reboot (step 3).

innobead (approved):

Issue #7051, opened 2023-11-06
[FEATURE] Support volume encryption for (encrypted) backing image volumes


innobead:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

Currently, Longhorn supports non-backing image volume encryption, and it is impossible to enable volume encryption for an unencrypted backing image. The goal here is to support volume encryption regardless of whether the volume is based on a backing image or not, with the potential condition being that the backing image is also encrypted using the same crypto key.

ref: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSEncryption.html#how-ebs-encryption-works

> Encrypt unencrypted resources
> You cannot directly encrypt existing unencrypted volumes or snapshots. However, you can create encrypted volumes or snapshots from unencrypted volumes or snapshots. If you enable encryption by default, Amazon EBS automatically encrypts new volumes and snapshots using your default KMS key for EBS encryption. 

ref: https://gitlab.com/cryptsetup/cryptsetup/-/wikis/FrequentlyAskedQuestions

> * **2.5 Can I encrypt an existing, non-empty partition to use LUKS?**
> 
> There is no converter, and it is not really needed. The way to do this is to make a backup of the device in question, securely wipe the device (as LUKS device initialization does not clear away old data), do a luksFormat, optionally overwrite the encrypted device, create a new filesystem and restore your backup on the now encrypted device. Also refer to sections "Security Aspects" and "Backup and Data Recovery".
> 

## Describe the solution you'd like

When creating a first snapshot from the backing image, should be encrypted first.

ChanYiLin:
## Experiments

### Create Encrypted Volume with BackingImage
1. Create StorageClass with BackingImage and Encryption Info
    ```
    kind: StorageClass
    apiVersion: storage.k8s.io/v1
    metadata:
      name: longhorn-crypto-global
    provisioner: driver.longhorn.io
    allowVolumeExpansion: true
    parameters:
      numberOfReplicas: "3"
      staleReplicaTimeout: "2880" # 48 hours in minutes
      fromBackup: ""
      encrypted: "true"
      backingImage: "parrot"
      backingImageDataSourceType: "download"
      backingImageDataSourceParameters: '{"url": "https://longhorn-backing-image.s3-us-west-1.amazonaws.com/parrot.raw"}'
      csi.storage.k8s.io/provisioner-secret-name: "longhorn-crypto"
[...]

PR #8740, opened 2024-06-12
feat(lep): backing image clone and encryption support

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7051

- Add LEP for backing image encryption support

### DIFF
--- a/enhancements/20240612-backing-image-encryption-and-clone-supprt.md
+++ b/enhancements/20240612-backing-image-encryption-and-clone-supprt.md
@@ -1,0 +1,267 @@
+# BackingImage Encryption Support
+
+## Summary
+
+Longhorn supports encrypted volumes by utilizing a Linux kernel-based disk encryption solution (LUKS, Linux Unified Key Setup). If users want to use BackingImage with encrypted Volume, the BackingImage needs to be encrypted as well. This feature allows users to clone the existing BackingImage and encrypt the content during cloning.
+
+### Related Issues
+
+- https://github.com/longhorn/longhorn/issues/7051
+
+## Motivation
+
+### Goals
+
+- Users can create a Encrypted Volume with the Encrypted BackingImage
+- Users can clone a BackingImage
+    - Users can encrypt a BackingImage during the cloning process
+    - Users can decrypt an encrypted BackingImage during the cloning process
+- For qcow2 image, if we want to encrypt it
+    - We will temporarily create raw image from the qcow2 first and then encrypt it.
+    - The result will be an encrypted raw image
+
+### Non-goals [optional]
+
+- Users can change the encryption key of the encrypted BackingImage
+- Encrypt qcow2 image to a encrypted qcow2 image
+
+
+## Proposal
+
+### User Stories
+
+#### Utilize encrypted Volume with BackingImage
+
+Before this feature, users can not utilize an encrypted volume with an unencrypted BackingImage. Users will get `unsupported disk encryption format unknown data, probably partitions` error during mounting the device.
+The content inside the BackingImage needs to be encrypted using the same key as the Volume before mounting to the workload.
+
+Thus, we need to support users to encrypt the BackingImage in Longhorn.
+
+#### Clone the BackingImage
+
+With this feature, users can create an identical BackingImage by cloning another BackingImage.
+
+#### Encryption and Decryption
+With introducing cloning mechanism, we can encrypt and decrypt the content inside the BackingImage during the cloning process.
+
+Noted that, it means it will cost additional space if users want to encrypt or decrypte the BackingImage. The encrypted or decrypted BackingImage will be a new BackingImager in the system.
+
+
+## Design
+
+### Implementation Overview
+
+#### CRD
+
+Add a field to the BackingImage
+- Secret/SecretNamespace: 
+    - Pointing to the k8s secret containing the key used for encryption and decryption
+    - We use `Secret` to check if the BackingIamge is encrypted. If the BackingImage has `Secret` then it means it is encrypted.
+
+```
+type BackingImageSpec struct {
+	Disks map[string]string `json:"disks"`
+	Checksum string `json:"checksum"`
+	SourceType BackingImageDataSourceType `json:"sourceType"`
+	SourceParameters map[string]string `json:"sourceParameters"`
+    
+	Secret string `json:"secret"`
+    SecretNamespace string `json:"secretNamespace"`
+}
+```
+
+#### Backing Image Data Source
+
+- Add a new source type: `clone`
+    - parameters:
+        - `source`: the source BackingImage
+        - `encryption`: `encrypt`, `decrypt` and `ignore`
+        - `secret`: the key used in the encryption and decryption.
+- Encryption
+    - `encrypt`: Longhorn copy the BackingImage and apply the encryption
+    - `decrypt`: Longhorn copy the BackingImage and decrypt the content
+    - `ignoe`: Longhorn simply copy the BackingImage
+- If we are encrypting the BackingImage, copy the `Secret` and `SecretNamespace` from parameters to `BackingImage.Spec.Secret` and  `BackingImage.Spec.SecretNamespace`
+- When choosing the node for the data source pod, we always choose one of the node and disk from the source BackingImage
+
+#### Webhook
+- If Source BackingImage has `Secret`, we don't allow to perform `"encrypt"`. (Forbid encrypted -> encrypted)
+- If Source BackingImage doesn't have `Secret`, we don't allow to perform `"decrypt"`. (Forbid unencrypted -> decrypted)
+- If the operation is to `"encrypt"` or `"decrypt"`, the `Secret` and `SecretNamespace` are needed.
+- User cannot change the `Secret` and `SecretNamespace`.
+
+#### Backing Image Manager - Data Source
+
+- When init the service, if the type is clone, then clone from source BackingImage by requesting sync service in the same pod.
+    ```golang
+        requestURL := fmt.Sprintf("http://%s/v1/files", client.Remote)
+        // credential contains the crypto secret
+        req, err := http.NewRequest("POST", requestURL, bytes.NewReader(encodedCredential))
+        if err != nil {
+            return err
+        }
+        req.Header.Set("Content-Type", "application/json")
+        q := req.URL.Query()
+        q.Add("action", "cloneFromBackingImage")
+        // the source backing image
+        q.Add("backing-image", sourceBackingImage)
+        // we need uuid to get the path
+        q.Add("backing-image-uuid", sourceBackingImageUUID)
+        // can be ignore, encrypt or decrypt
+        q.Add("encryption", encryption)
+        q.Add("file-path", filePath)
+        q.Add("uuid", uuid)
+        q.Add("disk-uuid", diskUUID)
+        q.Add("expected-checksum", expectedChecksum)
+    ```
+- When doing the cloning
+    - The source file is the source BackingImage file on the same disk
+    - If the operation is `"encrypt"` and the source file is `qcow2`
+        - We create a temp raw image from the qcow2 image and use it as source file
+    - Truncate the target file with the source file size.
+        - If the operation is `"encrypt"`, we `+16MB` for LUKS meta size
+        - If the operation is `"decrypt"`, we `-16MB`
+    - If the operation is `"encrypt"`
+        - Setup the loop device with the target file
+        - `cryptsetup luksFormat` and `luksOpen` the device
+        - set the target to the LUKS device
+    - If the operation is `"decrypt"`
+        - Setup the loop device with the source file
+        - `cryptsetup luksOpen` the device
+        - set the source to the LUKS device
+    - copy all the data from source to the target
+        - Need to write zeros since zeros also need to be encrypted.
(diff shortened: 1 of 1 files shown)


enhancements/20240612-backing-image-encryption-and-clone-supprt.md
```diff
@@ -0,0 +1,242 @@
...
+    - If the operation is `"encrypt"`
+        - Setup the loop device with the target file
+        - `cryptsetup luksFormat` and `luksOpen` the device
+        - set the target to the LUKS device
+    - If the operation is `"decrypt"`
+        - Setup the loop device with the source file
+        - `cryptsetup luksOpen` the device
+        - set the source to the LUKS device
+    - copy all the data from source to the target
+        - Need to write zeros since zeros also need to be encrypted.
```
james-munson, line 132: Presumably, if the operation is `"ignore"`, the zeros do not need to be expanded?  Or would a clone always be inflated to its full nominal size?

ChanYiLin: Yes you are right, we can conclude the operation to the following table

| - | unencrypted img | encrypted img |
| -------- | -------- | -------- |
| encrypt    |allow (write zero)|  x  |
| decrypt    | x  |     allow(skip zero)     |
| ignore     |   allow(skip zero)   | allow(skip zero)  |

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
```diff
@@ -0,0 +1,242 @@
...
+        q.Add("backing-image-uuid", sourceBackingImageUUID)
+        // can be ignore, encrypt or decrypt
+        q.Add("encryption", encryption)
+        q.Add("file-path", filePath)
+        q.Add("uuid", uuid)
+        q.Add("disk-uuid", diskUUID)
+        q.Add("expected-checksum", expectedChecksum)
+    ```
+- When doing the cloning
+    - The source file is the source BackingImage file on the same disk
```
shuo-wu, line 117: What if the cloning BI cannot be scheduled to the disk the source BI resides on? This is possible when the following conditions are satisfied:
1. All disks of the source BI files are unschedulable
2. The source BI and the clone BI have different node/disk selectors

shuo-wu (commented):
In general LGTM

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
ChanYiLin: I can come up with only one solution.
Like download the data from the URL, start the server on the source file, target file act as client to ask for the source data.
After, client can ask the server to close the server

I think we can improve this part in the next version v1.8.0.
Currently, we hint the user with the image that Longhorn can't find a ready node, disk for the image.
WDYT?

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
```diff
@@ -0,0 +1,251 @@
...
+
+## Design
+
+### Implementation Overview
+
+#### CRD
+
+Add a field to the BackingImage
+- Secret/SecretNamespace: 
+    - Pointing to the k8s secret containing The key used for encryption and decryption
```
derekbit, line 58:
```suggestion
    - Pointing to the k8s secret containing the key used for encryption and decryption
```

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
```diff
@@ -0,0 +1,251 @@
...
+
+- Add a new source type: `clone`
+    - parameters:
+        - `source`: the source BackingImage
+        - `encryption`: `encrypt`, `decrypt` and `ignore`
+        - `secret`: the key used in the encryption and decryption.
+- Encryption
+    - `encrypt`: Longhorn copy the BackingImage and apply the encryption
+    - `decrypt`: Longhorn copy the BackingImage and decrypt the content
+    - `ignoe`: Longhorn simply copy the BackingImage
```
derekbit, line 83: What's the use case of `ignore`?

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
```diff
@@ -0,0 +1,251 @@
...
+- If we are encrypting the BackingImage, copy the `Secret` and `SecretNamespace` from parameters to `BackingImage.Spec.Secret` and  `BackingImage.Spec.SecretNamespace`
+- When choosing the node for the data source pod, we always choose one of the node and disk from the source BackingImage
```
derekbit, line 85: Will we check if the disk space is enough before encryption or decryption?

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
```diff
@@ -0,0 +1,251 @@
...
+
+#### Webhook
```
derekbit, line 87: Changing `Secret` or `SecretNamespace` is not allowable, correct?

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
```diff
@@ -0,0 +1,251 @@
...
+        q.Add("encryption", encryption)
+        q.Add("file-path", filePath)
+        q.Add("uuid", uuid)
+        q.Add("disk-uuid", diskUUID)
+        q.Add("expected-checksum", expectedChecksum)
+    ```
+- When doing the cloning
+    - The source file is the source BackingImage file on the same disk
+    - If the operation is `"encrypt"` and the source file is `qcow2`
+        - We create a temp raw image from the qcow2 image and use it as source file
```
derekbit, line 119: What's the maximum space usage while encrypting a qcow2 image?

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
```diff
@@ -0,0 +1,251 @@
...
+    - Truncate the target file with the source file size.
+        - If the operation is `"encrypt"`, we `+16MB` for LUKS meta size
+        - If the operation is `"decrypt"`, we `-16MB`
```
derekbit, line 122: Add a reference doc for the 16 MB truncation.

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
```diff
@@ -0,0 +1,251 @@
...
+    - copy all the data from source to the target
+        - Need to write zeros since zeros also need to be encrypted.
+        - For other cases we can skip zero (ref to the below operation table)
+
+#### operation table
+
+| - | unencrypted img | encrypted img |
+| -------- | -------- | -------- |
+| encrypt    |allow (write zero)|  x  |
+| decrypt    | x  |     allow (skip zero)     |
```
derekbit, line 140:
```suggestion
| decrypt    | x  |     allow (skip zero)     |
```

How do you know if a block is zeroed and can be skipped? Sum the bytes in the block?

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
```diff
@@ -0,0 +1,251 @@
...
+        encryption: decrypt
+        secret: longhorn-crypto
+        secret-namespace: longhorn-system
+    ```
+- The checksum of the `parrot-cloned-decrypt` should be the same as `parrot`
+
+#### Qcow2 Image Encryption
+- Follow the same process, but change the BackingImage to qcow2
+    - `https://longhorn-backing-image.s3-us-west-1.amazonaws.com/parrot.qcow2`
+
```
derekbit, line 244: Add a UI change section.

derekbit (commented):
In general, LGTM.

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
shuo-wu: I am fine with it. If the disk the original BI resides on has no available space, Longhorn can directly error out or mark the clone BI as ERROR

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
ChanYiLin: fixed.

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
ChanYiLin: just simply clone the backing image

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
ChanYiLin: All the backing image creation currently has no disk space checking because we won't know the size of the backing image before creation.

But of course, cloning is different, we know the size beforehand. But currently we need to put the target file on the same disk of the source file. So we also need to consider the scheduling.

We can create another issue to improve the disk utilization when creating BackingImage including all the creation type and improve the node/disk selection logic to consider the space.

For now, I think we can skip the checking and make it failed due to the space issue and let user recreate it like other BackingImage type.

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
ChanYiLin: Yes

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
ChanYiLin: extra raw image size of the image.

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
ChanYiLin: It is implemented already
https://github.com/longhorn/backing-image-manager/blob/master/pkg/sync/handler.go#L154
```
				// Skip writing zero data
				if bytes.Equal(buf[0:nr], zeroByteArray[0:nr]) {
					_, handleErr = dst.Seek(int64(nr), io.SeekCurrent)
					nws = int64(nr)
				} else {
					nw, handleErr = dst.Write(buf[0:nr])
					nws = int64(nw)
				}
```

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
ChanYiLin: sure

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
ChanYiLin: added

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
ChanYiLin: > We can create another issue to improve the disk utilization when creating BackingImage including all the creation type and improve the node/disk selection logic to consider the space.
https://github.com/longhorn/longhorn/issues/8755

derekbit:
Will the progress of the encryption or decryption be displayed?

ChanYiLin:
- concurrent creating Backing Image number


> Will the progress of the encryption or decryption be displayed?

Hi @derekbit Yes, it is using the same mechanism as `upload` and `download` to copy the data from the sourceReader to targetWriter 4MB at a time.
The progress will be updated.

The PR is ready
You can check the implementation of this function  `IdleTimeoutCopy()`
https://github.com/longhorn/backing-image-manager/pull/251/files#diff-66b1ec3193d0ca8bd7fa4ee54caab93c7e0ed80edc090504487134b579bd2a71R684
https://github.com/longhorn/backing-image-manager/pull/251/files

ChanYiLin:
- Concurrent BackingImage creation limit
https://github.com/longhorn/longhorn/issues/8756

derekbit (approved):
LGTM.

TODO:
1. [Cross-disk and cross-node backing image encryption and decryption](https://github.com/longhorn/longhorn/issues/8758)
2. [Collect and display disk space usage for the backing images](https://github.com/longhorn/longhorn/issues/8757)
3. [Check the disk usage before encrypting or decryption a backing image](https://github.com/longhorn/longhorn/issues/8760)
4. [Re-encryption of an encrypted backing image](https://github.com/longhorn/longhorn/issues/8761)

cc @ChanYiLin @shuo-wu @innobead

innobead (approved):

enhancements/20240612-backing-image-encryption-and-clone-supprt.md
```diff
@@ -0,0 +1,267 @@
...
+      name: longhorn-crypto-global
+    provisioner: driver.longhorn.io
+    allowVolumeExpansion: true
+    parameters:
+      numberOfReplicas: "2"
+      staleReplicaTimeout: "2880" # 48 hours in minutes
+      fromBackup: ""
+      encrypted: "true"
+      backingImage: "parrot-cloned-encrypted"
+      backingImageDataSourceType: "clone"
```
Yu-Jack, line 191: Hi @ChanYiLin, I have a question. 

If I decided to use same secret, do I need to create a new storage class when I try to encrypt different backingimages each time? Or I could just use one storage class to encrypt all? 

The reason I asked that is I saw this example storage class has the `backingimage` and `backingImageDataSourceType`.

ChanYiLin: discussed with @Yu-Jack 
each backing image will have its own storageclass

PR #8775, opened 2024-06-18
feat(lep): pre-pull share-manager/instance-manager images

mantissahz:

### DIFF
--- a/enhancements/20240516-pre-pull-images.md
+++ b/enhancements/20240516-pre-pull-images.md
@@ -1,0 +1,63 @@
+# Pre-Pull Images
+
+## Summary
+
+We will pre-pull the share-manager image on every worker node to speed up the startup of share-managers.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/8376
+
+## Motivation
+
+### Goals
+
+- Pre-pull the share-manager image on all worker nodes.
+
+### Non-goals [optional]
+
+`None`
+
+## Proposal
+
+- Sidecar containers in the `longhorn-manager` DaemonSet to pre-pull images
+
+### User Stories
+
+With the pre-pull images mechanism, users can deploy RWX volume and start to use volumes faster. It's also useful for [Share manager HA mechanism](https://github.com/longhorn/longhorn/issues/6205).
+
+There is the [garbage collection of unused containers and images feature](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#containers-images) of kubernetes. The `kubelet` performs garbage collection on unused images every 2~5 minutes and on unused containers every minute so we have sidecar containers in the `longhorn-manager` DaemonSet to prevent the pre-pull images from being deleted by the garbage collection of unused containers and images.
+
+### API changes
+
+`None`
+
+## Design
+
+### Implementation Overview
+
+- Add containers in the `longhorn-manager` DaemonSet:
+
+  ```yaml
+  spec:
+    containers:
+    - name: longhorn-manager
+      ...
+    - name: pre-pull-share-manager-image
+        imagePullPolicy: IfNotPresent
+        image: longhornio/longhorn-share-manager:master-head
+        command: ["sh", "-c", "echo longhorn-share-manager image pulled && sleep infinity"]
+  ```
+
+### Test plan
+
+- Fresh install/Upgrade:
+  - After install or upgrade, the share-manager image are pulled on the each worker node.
+
+### Upgrade strategy
+
+`None`
+
+## Note [optional]
+
+`None`


enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Design
+
+### Implementation Overview
+
+- Add an initial container in the `longhorn-manager` DaemonSet:
+
+  ```yaml
+  spec:
+    initContainers:
```
innobead, line 44: This example is not really sidecar but a traditional init container instead, which will start and stop before the application container(s) starts. That means they could be garbage collected as usual. 

Please check https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/#pod-sidecar-containers to use `restartPolicy: Always` to ensure they run as sidecar.

innobead (changes requested):

enhancements/20240516-pre-pull-images.md
mantissahz: This field `restartPolicy` in `initContainers` is only implemented from v1.28 (native sidecar container alpha).
Moved the initial containers to containers and use the `sleep infinity` command.

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,66 @@
...
+  spec:
+    containers:
+    - name: longhorn-manager
+      ...
+    - name: pre-pull-share-manager-image
+        image: longhornio/longhorn-share-manager:master-head
+        command: ["sh", "-c", "echo longhorn-share-manager image pulled && sleep infinity"]
+    - name: pre-pull-instance-manager-image
+      image: longhornio/longhorn-instance-manager:master-head
+      command: ["sh", "-c", "echo longhorn-instance-manager image pulled && sleep infinity"]
```
innobead, line 52: Don't we need `ImagePullPolicy: IfNotPresent`? BTW, why do we need longhorn instance manager? I thought IM pods should be immediately run when Longhorn starts running.

enhancements/20240516-pre-pull-images.md
innobead: I am curious if this only takes effect in newer versions, but can still be applied in older Kubernetes versions without taking effect. This question is about the backward compatibility of Kubernetes.

enhancements/20240516-pre-pull-images.md
mantissahz: Added for clarity. If it is omitted, the default of ImagePullPolicy will be IfNotPresent.
And Removed the pre-pulling instance-manager container.

enhancements/20240516-pre-pull-images.md
mantissahz: >but can still be applied in older Kubernetes versions without taking effect.

No, it can not be applied in the older Kubernetes versions (I tested it in the v1.26.3+k3s1). There is no `restartPolicy` field in the `initContainer`.

innobead: Then Kubernetes resource API is not backward compatible.

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,64 @@
...
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/8376
+
+## Motivation
+
+### Goals
+
+- Pre-pull the share-manager image on all worker nodes.
+- Pre-pull the instance-manager image on all worker nodes.
```
innobead, line 16: Remove this as we don't need.

mantissahz: Removed the instance manager part. Thanks

derekbit:
cc @james-munson  for review

mantissahz:
The resource usage of the pre-pull image container `pre-pull-share-manager-image`:
```txt
root@rancher60-worker1:~# crictl stats
CONTAINER           NAME                           CPU %               MEM                 DISK                INODES
...
4493f3cdca904       longhorn-manager               0.25                48.65MB             77.82kB             18
4ba3c011b587e       csi-provisioner                0.07                12.3MB              36.86kB             11
5d6a790970abe       instance-manager               0.04                14.57MB             69.63kB             20
6a4a1c8533f36       longhorn-csi-plugin            0.00                15.89MB             77.82kB             19
73268e5ee9882       pre-pull-share-manager-image   0.00                1.61MB              36.86kB             9
...
```

It remains at 1.6MB for memory usage.
And it will not start the nfs-ganesha server.

enhancements/20240516-pre-pull-images.md
ejweber: Can we create an issue to move to the native sidecar pattern when we raise the minimum Kubernetes version to at least `v1.29` (when it is enabled by default), or potentially whatever version it goes GA in? It looks like this approach will work in the interim, but the native approach will likely be more "correct".

ejweber (approved):
I tried running a longhorn-manager like the latest PR describes looking for the issues described in https://kubernetes.io/blog/2023/08/25/native-sidecar-containers/#how-did-users-get-sidecar-behavior-before-1-28. We don't really care about startup order, but it will be bad if the pre-pull container somehow blocks deletion. Neither of the following seemed to be problematic:

- `kubectl delete longhorn-manager...` - Maybe it took longer than usual? However, the longhorn-manager pod was eventually deleted and recreated.
- `ssh <node> kill -9 <longhorn-manager PID>` - Looks like the readiness probe failed and the longhorn-manager container restarted. No effect from the pre-pull container.

I'll be curious to hear from @james-munson if he thinks there are other scenarios we should worry about, but LGTM.

innobead (approved):

enhancements/20240516-pre-pull-images.md
```diff
@@ -0,0 +1,63 @@
...
+
+### API changes
+
+`None`
+
+## Design
+
+### Implementation Overview
+
+- Add containers in the `longhorn-manager` DaemonSet:
```
shuo-wu, line 39: Which one is responsible for adding this container? And how?

innobead: Is it pre-defined in the longhorn manger daemonset manifest?

enhancements/20240516-pre-pull-images.md
mantissahz: Sure, here is the improvement ticket https://github.com/longhorn/longhorn/issues/8811

enhancements/20240516-pre-pull-images.md
mantissahz: Yes, it is pre-defined in the manifest as this PR https://github.com/longhorn/longhorn/pull/8778

derekbit:
> I tried running a longhorn-manager like the latest PR describes looking for the issues described in https://kubernetes.io/blog/2023/08/25/native-sidecar-containers/#how-did-users-get-sidecar-behavior-before-1-28. We don't really care about startup order, but it will be bad if the pre-pull container somehow blocks deletion. Neither of the following seemed to be problematic:
> 
> * `kubectl delete longhorn-manager...` - Maybe it took longer than usual? However, the longhorn-manager pod was eventually deleted and recreated.

@mantissahz Have you measured how long will a longhorn-manager pod be deleted? Will 

> * `ssh <node> kill -9 <longhorn-manager PID>` - Looks like the readiness probe failed and the longhorn-manager container restarted. No effect from the pre-pull container.
> 
> I'll be curious to hear from @james-munson if he thinks there are other scenarios we should worry about, but LGTM.

PR #8778, opened 2024-06-18
fix(images): pre-pull / pin share-manager image

mantissahz:

Add initial containers in the longhorn-manager DaemonSet to pre-pull images.


ejweber (changes requested):
Looks like we're missing the instance-manager part. Is that coming in a followup PR, or should it be here?

innobead:
Why do we need to have instance manager? I thought it should run just after the longhorn manager starts and should not be possible to be garbage collected?

derekbit:
> Why do we need to have instance manager? I thought it should run just after the longhorn manager starts and should not be possible to be garbage collected?

Users can only enable v2 data engine. After upgrading Longhorn system, the new instance-manager pods won't be created if there are v2 engines or replicas running on the old instance-manager pods.
Therefore, we need to pin the default instance-manager image for speeding up the v2 live upgrade process supported in v1.8.0.

innobead:
live upgrade is not like failover (share manager) which has latency concerns, but it should eventually happen w/o any interruption, even though there would be some delays for pulling images first time. Do we still need to have a dedicated sidecar for it?

derekbit:
> live upgrade is not like failover (share manager) which has latency concerns, but it should eventually happen w/o any interruption, even though there would be some delays for pulling images first time. Do we still need to have a dedicated sidecar for it?

I prefer to have a dedicated sidecar for it to avoid unexpected IM image pulling failure. Unlike a share-manager image, we can remove the sidecar for an IM image after a live upgrade. WDYT?

innobead:
Sounds good. Maybe let's just keep it simple initially. Don't need to dynamically remove it until it becomes a resource usage issue.

At least I know the initial reason.

derekbit:
> Sounds good. Maybe let's just keep it simple initially. Don't need to dynamically remove it until it becomes a resource usage issue.
> 
> At least I know the initial reason.

For IM image, we can defer to v1.8.0, and we can think about it more. Let's focus on share-manager image here.

innobead (approved):

derekbit (approved):

PR #8790, opened 2024-06-20
feat(backingimage): add clone source type and secret to the backing image

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7051

Check the LEP here: https://github.com/longhorn/longhorn/pull/8740


derekbit (approved):
LGTM

innobead (approved):

Issue #2765, opened 2021-07-02
[IMPROVEMENT] Improve rebuilding is canceled if it takes longer than 24 hours


PhanLe1010:
We have a limit for rebuilding. It is capped at 24 hours [here](https://github.com/longhorn/longhorn-engine/blob/c101e7a2388ca41e84bfc0cad33366f9e22b279c/pkg/replica/client/client.go#L20).

Some ideas from @yasker :

Those long timeout are not good, as we should use progress instead of timeout to decide if the operation is still going on. like Backup or other operations. Though here we're only dealing with a single GRPC call. A few ways to solve this:
* Increase the timeout and document it. Warning user in the worst case scenario, a rebuild lasting more than xx hours might fail. (Not ideal due to we still going to timeout)
* Remove the timeout on this level and depends on GRPC for network detection, and depends on the upper layer or lower layer timeout (I vaguely remembered there is one? or it's after the GRPC call) for whole process timeout (Need to check if the upper/lower layer can catch it correctly)
* Divide one GRPC into multiple GRPC calls with ability to determine the progress (Can be complex to implement)

**Additional context:**
The rebuilding time is a function of: the size of snapshot + the number of holes in the snapshot

mattmattox:
@PhanLe1010 Could we make this a tunable setting under the "Danger Zone" section? Because there could be events where rebuilds are very slow.

For example, a cluster that has a large amount of data and has a power outage event. This event triggers a rebuild of most of its data which can put a heavy load on the longhorn storage servers. In this event, larger volumes could timeout cause additional load as another rebuild has to be started IE rebuilding the same data multiple times. But if we had a tunable setting, this timeout could be increased temporarily until all data has been rebuilt then changed back to default.

innobead:
@derekbit 

Should we make the timeout configurable?

Some causes make the rebuilding slow
disk starvation: many applications are consuming disk ios, slow disks
network issue: starvation, small bandwidth, high resend rate/packet loss...
cpu bound

Two ideas
Make the timeout configuration
Or, if the rebuilding is in progress, can we make it  adaptive and automatically increase the timeout?

@shuo-wu 
Replica rebuilding is a sync call hence the timeout & disk performance determines the max capacity Longhorn can rebuilding, which then determines the max volume size

Maybe we need to make the call async first, which is already mentioned in the SPDK rebuilding PR

@PhanLe1010 
[...]

PR #8812, opened 2024-06-24
Make long gRPC timeout configurable

PhanLe1010:
longhorn/longhorn#2765


chart/README.md
```diff
@@ -293,6 +293,7 @@ During installation, you can either allow Longhorn to use the default system set
 | defaultSettings.guaranteedInstanceManagerCPU | Percentage of the total allocatable CPU resources on each node to be reserved for each instance manager pod when the V1 Data Engine is enabled. The default value is "12". |
 | defaultSettings.kubernetesClusterAutoscalerEnabled | Setting that notifies Longhorn that the cluster is using the Kubernetes Cluster Autoscaler. |
 | defaultSettings.logLevel | Log levels that indicate the type and severity of logs in Longhorn Manager. The default value is "Info". (Options: "Panic", "Fatal", "Error", "Warn", "Info", "Debug", "Trace") |
+| defaultSettings.longGRPCTimeOut | Number of seconds that Longhorn allows for the completion of replica rebuilding and snapshot cloning operations. |
```
innobead, line 296: I know the manager PR was merged, but suggest changing the setting name to something more explicit like `replicaSyncTimeout`. What do you think?

innobead (changes requested):

chart/README.md
PhanLe1010: We were discussing a similar topic at https://github.com/longhorn/longhorn-engine/pull/1125/files/93a14578a17d45abacd050539a71360b3c600bf2#diff-a13a6464fb17f3625bc0b3508f2313ed8f091abe05a1b370258321b517f0974e 

Since the setting is not only use for replica building but also file syncing and snapshot syncing, I think it might make more sense to have a more generic name. WDYT?

innobead: Didn't notice this had been discussed. It's fine for me if all agree.

innobead (approved):

Issue #6001, opened 2023-05-25
[FEATURE] v2 volume supports live upgrade for data plane


derekbit:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

The data path for SPDK volumes is established through spdk_tgt. To upgrade `spdk_tgt` without downtime, the bdev raids of attached volumes must be live migrated to the new `spdk_tgt`.

Subtasks
- [x] https://github.com/longhorn/longhorn/issues/6647
- [x] https://github.com/longhorn/longhorn/issues/8456
- [x] https://github.com/longhorn/longhorn/issues/8458
- [x] https://github.com/longhorn/longhorn/issues/8464

derekbit:
Current design
![image](https://github.com/longhorn/longhorn/assets/12527233/cfd1e845-d744-4bac-8103-2973cc6ffada)

chriscchien:
Verified **pass** on longhorn master(longhorn-instance-manager `84ba8e`)

- V2 volume attach/detach wokerd well
- Below V2 volume restore path worked well
   - Restore V2 volume to RWO V1 volume
   - Restore V2 volume to RWX V1 volume
   - Restore V2 volume to RWO V2 volume

derekbit:
@chriscchien
Can you help verify the data plane APIs by the following steps?

1. Deploy a Longhorn v1.7.0-rc1 2-node cluster (node1: `rancher60-worker1`, node2:  `rancher60-worker2`)
2. Enable v2 data engine
3. Create workloads with v2 volumes
```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: longhorn-v2-data-engine
provisioner: driver.longhorn.io
allowVolumeExpansion: true
reclaimPolicy: Delete
volumeBindingMode: Immediate
parameters:
  # backup, backingImage and snapshot related parameters are not supported in SPDK preview stage
  numberOfReplicas: "2"
  staleReplicaTimeout: "2880"
  fsType: "ext4"
  dataEngine: "v2"
  ---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: longhorn-volv-pvc-1
  namespace: default
spec:
  accessModes:
    - ReadWriteOnce
  storageClassName: longhorn-v2-data-engine
  resources:
    requests:
      storage: 5Gi
---
apiVersion: v1
kind: Pod
metadata:
  name: volume-test-1
  namespace: default
spec:
  restartPolicy: Always
  containers:
[...]

PR #8814, opened 2024-06-24
Add LEP for Live Upgrade For Data Engine of V2 Volumes

derekbit:

Issue Longhorn/longhorn#6001


derekbit:
@DamiaSan @shuo-wu @innobead Could you take a look at the data plane design? Thank you.

derekbit:
@DamiaSan @shuo-wu @innobead Can you take a look at the data plane design? Thank you.

enhancements/20240624-live-upgrade-for-data-engine-of-v2-volumes.md
```diff
@@ -0,0 +1,225 @@
...
+      ```
+    - EngineCreateRequest
+      During an upgrade, the initiator and target reside on different nodes. `initiator_address` and `target_address` specify the respective addresses of the initiator and target.
+      ```
+      message EngineCreateRequest {
+        ...
+        bool upgrade_required = 7;
+        string initiator_address = 8;
+	    string target_address = 9;
+      }
```
DamiaSan, line 117: Don't we need here also target and initiator ports?

enhancements/20240624-live-upgrade-for-data-engine-of-v2-volumes.md
```diff
@@ -0,0 +1,225 @@
...
+      ```
+    - InstanceDeleteTargetRequest
+      ```
+      message InstanceDeleteTargetRequest {
+      	DataEngine data_engine = 1;
+      	string name = 2;
+      	string type = 3;
+      }
+      ```
```
DamiaSan, line 204: Could you make a brief description of the fields of these last 4 messages?

enhancements/20240624-live-upgrade-for-data-engine-of-v2-volumes.md
```diff
@@ -0,0 +1,225 @@
...
+  - Replica Rebuilding
+      - Rebuilding failed replicas
+
+  Then, the upgrade process is complete once all volumes on the upgrading node are healthy.
+
+
+  Here is an example for upgrading instance-manager for v2 engine on the upgrading node `node1`:
+
+  ![An example for upgrading instance-manager for v2 engine on the upgrading node `node1`](image/v2-volume-upgrade.png)
+
```
DamiaSan, line 86: Is the title of the fifth figure correct? It is the same of the previous one, "Switch over target to ..."

DamiaSan (commented):
Just a few questions, but the general concept is correct for me.

enhancements/20240624-live-upgrade-for-data-engine-of-v2-volumes.md
derekbit: Yes, we need. address is the join of an ip and a host.

enhancements/20240624-live-upgrade-for-data-engine-of-v2-volumes.md
derekbit: Good catch! Updated. Thank you.

enhancements/20240624-live-upgrade-for-data-engine-of-v2-volumes.md
derekbit: Added. Thank you.

enhancements/20240624-live-upgrade-for-data-engine-of-v2-volumes.md
```diff
@@ -0,0 +1,229 @@
...
+      - Resume the linear device mapper and continue IO processing.
+        ```
+        InstanceResume()
+        |-> EngineResume()
+        ```
+  - Instance Manager Upgrade
+      - If the existing instance-manager pod does not have any running engines with targets, the instance-manager and its pod will be deleted by the node controller.
+      - Replicas managed by the deleted instance-manager are marked as ERROR, causing any volume with replicas on the upgrading node to become degraded.
+      - A new instance-manager is then created and starts running.
+  - Switch Over Target Back
```
shuo-wu, line 71: Similarly, will Longhorn revert to the temporary target if there is something wrong with the new target?

enhancements/20240624-live-upgrade-for-data-engine-of-v2-volumes.md
```diff
@@ -0,0 +1,225 @@
...
+## Motivation
+
+### Goals
+
+- Support both destructive and live upgrades of the data engine for a v2 volume with multiple replicas.
+- Upgrade volumes one node at a time.
+
+### Non-goals [optional]
+
+- Support live upgrades of the data engine for a v2 volume with a single replica.
```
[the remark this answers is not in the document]
derekbit: No. It is a hard limit.

enhancements/20240624-live-upgrade-for-data-engine-of-v2-volumes.md
```diff
@@ -0,0 +1,229 @@
...
+        ```
+        InstanceSwitchOverTarget()
+        |-> EngineSwitchOverTarget()
+        ```
+      - Delete the old target for the volume on the upgrading node.
+        ```
+        InstanceDeleteTarget()
+        |-> EngineDeleteTarget()
+        ```
+      - Resume the linear device mapper and continue IO processing.
```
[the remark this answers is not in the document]
derekbit: > due to this case, it's better to resume the dm device IO before deleting the old target.

Yes, but I found the nvme will somehow crash if deleting the old target after resuming IO. I'm still investigating the root cause.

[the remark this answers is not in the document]
derekbit: > due to this case, it's better to resume the dm device IO before deleting the old target.

Ideally, yes. But we need to overcome https://github.com/longhorn/longhorn/pull/8814#discussion_r1657944110 first.

enhancements/20240624-live-upgrade-for-data-engine-of-v2-volumes.md
derekbit: Same issue for now https://github.com/longhorn/longhorn/pull/8814#discussion_r1657944110

enhancements/20240624-live-upgrade-for-data-engine-of-v2-volumes.md
```diff
@@ -0,0 +1,225 @@
...
+
+### User Stories
+
+After upgrading the Longhorn system, users may want to upgrade the data engine of active v2 volumes to access new features or bug fixes without service interruption.
+
+### API changes
+
+## Design
+
+Unlike v1 volumes, upgrading the data engine for individual v2 volumes is not possible. This limitation exists because each node disk can only be managed by a single SPDK target daemon (spdk_tgt) within an instance-manager pod at any given time. The engine and replicas of a volume cannot be moved to a newly created instance-manager pod using a new (default) instance-manager image. This means the existing instance-manager pod must be removed before deploying a new one with the new image.
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

derekbit:
Replaced by https://github.com/longhorn/longhorn/pull/9807

Issue #8844, opened 2024-06-27
[BUG] Orphan longhorn-engine-manager and longhorn-replica-manager services 


PhanLe1010:
## Describe the bug

When trying StackState, it detects that Longhorn system has 2 orphan longhorn-engine-manager and longhorn-replica-manager services with no available endpoint

![image](https://github.com/longhorn/longhorn/assets/22139961/a576cf33-9297-448e-944e-648398e3d729)

## To Reproduce

Just install a fresh Longhorn 1.5.x or 1.6.x version and see the orphan services 

## Expected behavior

In the past, we had instance manager engine and instance manager replica separately so these services are having selectors to select the correct IM:
```yaml
apiVersion: v1
kind: Service
metadata: 
  name: longhorn-engine-manager
  namespace: longhorn-system
spec: 
  selector: 
    longhorn.io/component: instance-manager
    longhorn.io/instance-manager-type: engine
---
apiVersion: v1
kind: Service
metadata: 
  name: longhorn-replica-manager
  namespace: longhorn-system
spec: 
  selector: 
    longhorn.io/component: instance-manager
    longhorn.io/instance-manager-type: replica
```

However, since 1.5.x, we combined the IM into instance manager AIO but forget to update the selector of the services. I think there are 2 action items:
1.  Investigate whether we need the 2 services. If not perhaps we should remove them
2. If they are still needed, we need to update the selector to point to the instance manager AIO

c3y1huang:
@innobead @chriscchien , do you have any insights on why the headless services were introduced in v1.2.3? 
- https://github.com/longhorn/longhorn/pull/3345
- https://github.com/longhorn/longhorn/issues/3344

It seems that Longhorn hasn't been using these two services since the integration of the instance manager e/r. I suggest we remove them. What do you think @PhanLe1010 , @derekbit ?

PhanLe1010:
@c3y1huang Looks like the creation of `longhorn-engine-manager` and `longhorn-replica-manager` headless service is to fix this security issue https://github.com/longhorn/security/issues/2 . It is a hotfix while the better fix is using mtls. 

I am reading up to see if we still need that hotfix

PhanLe1010:
## Test Plan

1. Install Longhorn v1.6.2 using Helm
1. See that there are longhorn-engine-manager and longhorn-replica-manager services
1. Upgrade Longhorn to master-head
1. Verify that there are no longhorn-engine-manager and longhorn-replica-manager services
[...]

PR #8854, opened 2024-06-29
Remove headless services for instance manager pods

PhanLe1010:
See explanation at https://github.com/longhorn/longhorn/issues/8844#issuecomment-2197769331

longhorn/longhorn#8844


shuo-wu (approved):

PhanLe1010:

Issue #8855, opened 2024-06-29
[IMPROVEMENT] iscsi and nfs installation daemonsets should be in longhorn-system namespace


samtudor0112:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

The iscsi and nfs installation daemonsets used in the deployment instructions are in the default namespaces. Seems to make sense to put them in the longhorn-system namespace.

## Describe the solution you'd like

Add longhorn-system namespace to the daemonsets and update docs.

derekbit:
@samtudor0112 
The two daemonsets are just used for creating the two utilities. After the installation, they can be deleted. Does they have any negative impact for now?

samtudor0112:
I thought the idea of a daemonset was that if additional nodes are added to the cluster, the dependencies will also be installed there?

derekbit:
> I thought the idea of a daemonset was that if additional nodes are added to the cluster, the dependencies will also be installed there?

The dependencies will be installed on the host side.

samtudor0112:
> > I thought the idea of a daemonset was that if additional nodes are added to the cluster, the dependencies will also be installed there?
> 
> The dependencies will be installed on the host side.

According to the docs, these are node requirements? https://longhorn.io/docs/1.6.2/deploy/install/#installation-requirements

> Each node in the Kubernetes cluster where Longhorn is installed must fulfill the following requirements

derekbit:
> > > I thought the idea of a daemonset was that if additional nodes are added to the cluster, the dependencies will also be installed there?
> > 
> > 
> > The dependencies will be installed on the host side.
> 
> According to the docs, these are node requirements? https://longhorn.io/docs/1.6.2/deploy/install/#installation-requirements

Yes.

> 
> > Each node in the Kubernetes cluster where Longhorn is installed must fulfill the following requirements

samtudor0112:
Any update on this? The change is ready and just pending review 
https://github.com/longhorn/longhorn/pull/8856
https://github.com/longhorn/website/pull/932

PR #8856, opened 2024-06-29
Add longhorn-system namespace to installer daemonsets

samtudor0112:
Issue https://github.com/longhorn/longhorn/issues/8855

Add longhorn-system namespace to installer daemonsets

### DIFF
--- a/deploy/prerequisite/longhorn-cifs-installation.yaml
+++ b/deploy/prerequisite/longhorn-cifs-installation.yaml
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   name: longhorn-cifs-installation
+  namespace: longhorn-system
   labels:
     app: longhorn-cifs-installation
   annotations:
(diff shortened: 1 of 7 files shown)


samtudor0112:
Works
image

derekbit:
@PhanLe1010 @ejweber @shuo-wu Do you think if we need the change?

derekbit:
@PhanLe1010 @ejweber @shuo-wu Do  @innobead  you think if we need the change?

ejweber:
> @PhanLe1010 @ejweber @shuo-wu Do @innobead you think if we need the change?

I don't have much of a preference. Either `longhorn-system` or `default` seem fine to me. I lean a slight bit toward @samtudor0112's point of view because:

1. I think if I was creating these from scratch today, I would put them in the `longhorn-system` namespace.
2. Whether or not is is necessarily intended, I think the DaemonSets can be used in the way described in https://github.com/longhorn/longhorn/issues/8855#issuecomment-2198439492. (They can be left active in the cluster to ensure newly added nodes get dependencies installed.) In that case, it is better for them not to be in the `default` namespace.

I'll leave it to @derekbit or @innobead to make the call.

derekbit (approved):

samtudor0112:
Anything further needs to be done before this is merged?

samtudor0112:
Can this be merged yet?

ejweber (approved):

derekbit:
@samtudor0112 
Thanks for your contribution.

PR #8942, opened 2024-07-07
feat(v2 upgrade/chart): update for the changes in the data plane

derekbit:

Issue longhorn/longhorn#6001


derekbit:
Merge it first because longhorn-manager is already merged.

Issue #8300, opened 2024-04-03
[BUG] OpenShift 4.15.3 - Lonhorn 1.6.1 - longhorn-ui nginx (13: Permission denied)


n00bsi:
## Question

How to run 1.6.1 on OpenShift 4.15.3 ?
1.4.2 worked on OpenShift 4.15.3

## Environment

 - Longhorn version: 1.6.1
 - Kubernetes version:  v1.28.7+6e2789b
 - Node config
   - OS type and version: Red Hat Enterprise Linux CoreOS 415.92.202403080220-0 (Plow)
   - Kernel version: 5.14.0-284.55.1.el9_2.x86_64
   - CPU per node: 4
   - Memory per node: 8 GB worker / 16 GB controlplane
   - Disk type: Virtual Disk
   - Network bandwidth and latency between the nodes:
 - Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal):  VMWare

## Additional context

```
[core@osdemo1 ~]$ oc get nodes -o wide
NAME                 STATUS     ROLES                  AGE   VERSION           INTERNAL-IP   EXTERNAL-IP   OS-IMAGE                                                       KERNEL-VERSION                 CONTAINER-RUNTIME
osdemo1.ourdomain.tld   Ready      control-plane,master   15d   v1.28.7+6e2789b   192.168.7.221   <none>        Red Hat Enterprise Linux CoreOS 415.92.202403080220-0 (Plow)   5.14.0-284.55.1.el9_2.x86_64   cri-o://1.28.4-4.rhaos4.15.git92d1839.el9
osdemo2.ourdomain.tld   Ready      control-plane,master   15d   v1.28.7+6e2789b   192.168.7.222   <none>        Red Hat Enterprise Linux CoreOS 415.92.202403080220-0 (Plow)   5.14.0-284.55.1.el9_2.x86_64   cri-o://1.28.4-4.rhaos4.15.git92d1839.el9
osdemo3.ourdomain.tld   Ready      control-plane,master   15d   v1.28.7+6e2789b   192.168.7.223   <none>        Red Hat Enterprise Linux CoreOS 415.92.202403080220-0 (Plow)   5.14.0-284.55.1.el9_2.x86_64   cri-o://1.28.4-4.rhaos4.15.git92d1839.el9
[...]

PR #8950, opened 2024-07-08
fix(deploy): add a deploying manifest file with OpenShift enabled.

mantissahz:

Issue # longhorn/longhorn#8300

Longhorn can be installed by the deploy manifest file on the OpenShift platform.

### DIFF
--- a/deploy/longhorn-okd.yaml
+++ b/deploy/longhorn-okd.yaml
@@ -1,0 +1,5134 @@
+---
+# Builtin: "helm template" does not respect --create-namespace
+apiVersion: v1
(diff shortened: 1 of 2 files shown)


derekbit:
@mantissahz 
longhorn.yaml is generated by [script/generate-longhorn-yaml.sh](https://github.com/longhorn/longhorn/blob/master/scripts/generate-longhorn-yaml.sh), so you need to change the script. The script can generate both longhorn.yaml and longhorn-okd.yaml.

derekbit (approved):

Issue #8142, opened 2024-03-08
[FEATURE] Add additional monitoring settings to ServiceMonitor resource.


kejrak:
## Is your feature request related to a problem?

I was thinking to add multiple settings to ServiceMonitor resource. I have been doing multiple upgrades across my infrastructure applications with deploying prometheus-operator. I have my own "source" label for scraping. An option to change a source label is missing from Longhorn ServiceMonitor.

## Describe the solution you'd like

I would like to add additionalLabels option to serviceMonitor template as well as multiple another settings for prometheus monitoring (etc. annotations, interval, scrapeTimeout, tlsConfig, relabeling options...)

ejweber:
I don't think we need to go so far as to verify that particular ServiceMonitor configurations are WORKING the way a user might hope to test this issue. We can leave it to Prometheus to work correctly when configured correctly. We just need to verify that setting the new variables correctly results in a valid ServiceMonitor object (and setting them incorrectly does not).

1. Leaving the new variables to their default values results in the same ServiceMonitor as before:

```
# Helm install Longhorn before this commit with `serviceMonitor.enabled = true`.
metrics:
  serviceMonitor:
    # -- Setting that allows the creation of a Prometheus ServiceMonitor resource for Longhorn Manager components.
    enabled: true

eweber@laptop:~/longhorn> helm upgrade -n longhorn-system longhorn ./chart
Release "longhorn" has been upgraded. Happy Helming!
NAME: longhorn
LAST DEPLOYED: Thu Jul 11 14:51:19 2024
NAMESPACE: longhorn-system
STATUS: deployed
REVISION: 136
TEST SUITE: None
NOTES:
Longhorn is now installed on the cluster!

Please wait a few minutes for other Longhorn components such as CSI deployments, Engine Images, and Instance Managers to be initialized.

Visit our documentation at https://longhorn.io/docs/

eweber@laptop:~/longhorn> kl get servicemonitor -oyaml > before

# Helm install Longhorn after this commit while ONLY setting `serviceMonitor.enabled = true`.

metrics:
  serviceMonitor:
    # -- Setting that allows the creation of a Prometheus ServiceMonitor resource for Longhorn Manager components.
    enabled: true
    # -- Setting that adds additional labels to the Prometheus ServiceMonitor.
    additionalLabels: {}
    # -- Setting that adds annotations to the Prometheus ServiceMonitor.
    annotations: {}
    # -- Interval at which Prometheus scrapes the metrics from the target.
[...]

PR #8985, opened 2024-07-12
docs(metrics): move Helm values to a separate section in README.md

ejweber:

This is an addon for longhorn/longhorn#8142 that doesn't affect functionality.

It's weird to have so many metrics/serviceMonitor settings in `Other Settings` instead of in their own group.


ejweber:
@mergify backport v1.7.x v1.6.x

No need to backport to v1.5.x, since we don't have a `chart/README.md` there.

PhanLe1010 (approved):

Issue #9067, opened 2024-07-23
[BUG] Wrong NetworkPolicy labels


lahwaacz:
The `instance-manager` NetworkPolicy tries to select pods with `longhorn.io/component: instance-manager`:

https://github.com/longhorn/longhorn/blob/091256437632f5a0532eafef6462d26381cd2807/chart/templates/network-policies/instance-manager-networking.yaml#L8-L10

But the DaemonSet applies only `app: longhorn-manager` label:

https://github.com/longhorn/longhorn/blob/7e5b8047ed4acbd8c05264da6a721b4543f539de/chart/templates/daemonset-sa.yaml#L13-L15

In general, the `longhorn.io/component` label seems to be used only in the network policies, which should all be verified.

Another problem I noticed is that this `ports:` section should be obviously after the following `{{- end }}` tag (it does not make sense to gate it by the condition):

https://github.com/longhorn/longhorn/blob/091256437632f5a0532eafef6462d26381cd2807/chart/templates/network-policies/ui-frontend-network-policy.yaml#L40-L45

Btw. why are the network policies duplicated in the [examples directory](https://github.com/longhorn/longhorn/tree/master/examples/network-policy)?

ejweber:
Thanks for the report @lahwaacz. I am looking at it.

> The `instance-manager` NetworkPolicy tries to select pods with `longhorn.io/component: instance-manager`:
> But the DaemonSet applies only `app: longhorn-manager` label:

Is this one causing you problems in practice? The policy is meant to target instance-manager pods. These pods are deployed by the Longhorn system after installation, and their manifests do not appear in the Helm chart by design.

- https://github.com/longhorn/longhorn-manager/blob/62cd59c13f793147a33e72fb25f58458cfed785b/controller/instance_manager_controller.go#L1132
- https://github.com/longhorn/longhorn-manager/blob/62cd59c13f793147a33e72fb25f58458cfed785b/types/types.go#L317
- https://github.com/longhorn/longhorn-manager/blob/62cd59c13f793147a33e72fb25f58458cfed785b/types/types.go#L287
- https://github.com/longhorn/longhorn-manager/blob/62cd59c13f793147a33e72fb25f58458cfed785b/types/types.go#L279

```
eweber@laptop:~/longhorn> kl get pod -oyaml instance-manager-6f29b4b11e7b25924b3046d7da054dbd | grep labels -A 6
  labels:
    longhorn.io/component: instance-manager
[...]

PR #9077, opened 2024-07-24
Network policy

lahwaacz:

Issue #9067

Current network policies are too open, they should be allow only traffic that is specifically needed for Longhorn functionality and deny the rest.

The policies should also be in line with the documentation: https://longhorn.io/docs/1.6.2/references/networking/

#### Future work

- [ ] Review instance-manager policy
- [ ] Review backing-image-manager policy
- [ ] Review backing-image-data-source policy
- [ ] Add a configurable list of CIDRs to explicitly allow traffic to the webhooks and recovery backend from all cluster nodes (see https://github.com/longhorn/longhorn/issues/9067#issuecomment-2247049239)
- [ ] Add egress to all policies
- [ ] Add missing network policies (CSI plugin, CSI sidecar, driver deployer, etc.)

I will go offline in the following days and have a busy next week, but I will definitely revisit this later! Maintainers are invited to push to my branch or split the rest to a separate issue if doing it by parts is preferred.


lahwaacz:
The removed `examples/network-policies/` directory is linked from the documentation, which should be updated to refer to the policies installed by the chart: https://longhorn.io/docs/1.6.2/references/networking/

lahwaacz:
I have some questions about the network policies:

- In the [documentation](https://longhorn.io/docs/1.6.2/references/networking/), Instance Manager expects ingress from "Node in the Cluster" on TCP port 3260. Does this mean one specific node (e.g. the one where the Instance Manager runs) or can all nodes in the cluster connect to any Instance Manager?
- In the [documentation](https://longhorn.io/docs/1.6.2/references/networking/), Backing Image Manager specifies egress to Instance Manager, but Instance Manager does not specify ingress from Backing Image Manager. Which is correct?

innobead:
Assigned to @ChanYiLin to help with this review.

cc @derekbit

chart/templates/network-policies/ui-frontend-network-policy.yaml
M4t7e, line 1: These changes might be not sufficient especially when no ingress controller is used. See: https://github.com/longhorn/longhorn/issues/9067#issuecomment-2454536990

lahwaacz: You are right, this case is not addressed in this pull request. I was focusing on "simple" improvements that don't depend on any configuration, but the review is taking too long regardless... :disappointed:

chart/templates/network-policies/backing-image-data-source-network-policy.yaml
```diff
@@ -9,19 +9,23 @@ spec:
...
-    - podSelector:
-        matchLabels:
-          longhorn.io/component: instance-manager
-    - podSelector:
-        matchLabels:
-          longhorn.io/component: backing-image-manager
-    - podSelector:
-        matchLabels:
-          longhorn.io/component: backing-image-data-source
+    - from:
```
ChanYiLin, line 14: longhorn-manager will send a request to the backing-image-data-source on port 8000 to retrieve the status of the backing image.
See: https://longhorn.io/docs/1.7.2/references/networking/#backing-image-data-source

chart/templates/network-policies/manager-network-policy.yaml
```diff
@@ -9,27 +9,27 @@ spec:
...
+                  - longhorn-csi-plugin
+                  - longhorn-driver-deployer
+        - podSelector:
+            matchLabels:
+              longhorn.io/managed-by: longhorn-manager
+            matchExpressions:
+              - { key: recurring-job.longhorn.io, operator: Exists }
+        - podSelector:
+            matchExpressions:
+              - { key: longhorn.io/job-task, operator: Exists }
```
ChanYiLin, line 31: - Conversion Webhook Server, port `9501`
- Admission Webhook Server, port `9502`
- Recovery Backend Server, port `9503`
are now resides in longhorn-manager with the ports.

This changes restrict the port to only 9500, which seems not correct.
Other components may not be able to connect longhorn-manager:9501/9502/9503 for these 3 services

chart/templates/network-policies/ui-frontend-network-policy.yaml
ChanYiLin: Hi @lahwaacz 
Sorry for the late review and thanks for your contribution, I have left some comment, PTAL, thanks!

ChanYiLin (commented):
Thanks for your contribution, I have left some comments, PTAL, thanks!

chart/templates/network-policies/backing-image-data-source-network-policy.yaml
lahwaacz: What do you mean? This ingress rule applies to longhorn-manager, see 3 lines below.

ChanYiLin: Oh, sorry, don't know why I saw it wrong.
Resolved.

chart/templates/network-policies/manager-network-policy.yaml
lahwaacz: So what is the `podSelector` or other for these components?

ChanYiLin: These services are all in the longhor-manager pod with different port(9501, 9502 and 9503)

so the pod selector will be the same as longhorn-manager

lahwaacz: Done, but I can't test it right now

ChanYiLin: I can do the test next Monday
And we can then proceed this PR.

ChanYiLin:
Hi @lahwaacz 
The PR looks good to me
Can you fix the commit to meet the convention requirement
Then we can merge the PR
Thanks

ChanYiLin:
@c3y1huang @derekbit 
Can you also help to take a look, thanks

chart/templates/network-policies/instance-manager-networking.yaml
```diff
@@ -9,19 +9,36 @@ spec:
     matchLabels:
       longhorn.io/component: instance-manager
   policyTypes:
-  - Ingress
+    - Ingress
```
c3y1huang, line 12: Should we also include port `3260`? 
https://github.com/longhorn/website/blob/master/content/docs/1.9.0/references/networking.md?plain=1#L56

cc @PhanLe1010

chart/templates/network-policies/manager-network-policy.yaml
ChanYiLin: Tested. I can create a BackingImage and a Volume, then mount it to the pod.

chart/templates/network-policies/instance-manager-networking.yaml
lahwaacz: With what selector? What other components access this port?

PhanLe1010: The port 3260 is used by iSCSI initiators to establish connections with iSCSI targets (the tgtd target inside instance-manager pod)

ChanYiLin: https://longhorn.io/docs/1.8.0/nodes-and-volumes/volumes/iscsi/

We actually allow users to connect Longhorn volume through iscsi
That means this port should be always open
https://longhorn.io/docs/1.8.0/references/networking/#instance-manager
> Node in the Cluster	3260

ChanYiLin: Hi @lahwaacz 
can you help to add the port 3260 to always accessible?
Thanks

Issue #9113, opened 2024-07-29
[TASK] Replace the checking codes in environement_check.sh with `longhornctl` commands


derekbit:
## What's the task? Please describe

https://github.com/longhorn/longhorn/issues/9075#issuecomment-2255073613

derekbit:
@c3y1huang has some comments and discussion in https://github.com/longhorn/longhorn/pull/9123#discussion_r1697759066.

derekbit:
We will deprecate environment_check.sh in v1.8.0 instead.
https://github.com/longhorn/longhorn/pull/9123#issuecomment-2261762174

derekbit:
environement_check.sh will be removed in v1.9.0.

PR #9123, opened 2024-07-30
fix(script): modify environment_check.sh

mantissahz:

Issue # longhorn/longhorn#9113

`environment_check.sh` will only download the Longhorn CLI tool `longhornctl` and execute the commands `longhornctl check preflight` or `longhornctl install preflight` with/without the `--enable_spdk` flag.

### DIFF
--- a/scripts/environment_check.sh
+++ b/scripts/environment_check.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 NVME_CLI_VERSION="1.12"
+LONGHORN_CLI_VERSION="v1.7.0-rc2"
 
 ######################################################
 # Log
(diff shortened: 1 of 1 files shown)


scripts/environment_check.sh
```diff
@@ -492,29 +184,41 @@ Usage: $0 [OPTIONS]
 
 Options:
     -s, --enable-spdk           Enable checking SPDK prerequisites
-    -p, --expected-nr-hugepages Expected number of 2 MiB hugepages for SPDK. Default: 1024
+    -i, --install               Install and set up the preflight dependencies before installing Longhorn
```
c3y1huang, line 187: I am a bit confused about the purpose of this script. Initially, it was meant to check preflight dependencies because there is no `longhornctl`. Now that Longhorn provides a tool for dependency installation and check, so what is the benefit of wrapping `longhornctl` in a bash script with its own menu? Is it intended to include the preflight installation in this script? If so, how will the flags be handled (e.g. --operating-system=cos)? And are we planning to include other subcommands in the future?

derekbit: @innobead wants to retain the `environemnt_check.sh` for now. The benefit is users can check the environment in one single command. I think we can remove `environemnt_check.sh` in the future.

c3y1huang: @derekbit should we include the preflight installation in this script?

derekbit: > If so, how will the flags be handled (e.g. --operating-system=cos)

You're right. IIRC, we don't handle `cos` in the original environemnt_check.sh. Can we add a paragraph for the operating system and similar cases? WDYT @c3y1huang @mantissahz @innobead 

Or, I'm fine with retaining both `environment_check.sh` with the old codes (marked it as deprecated) and `longhornctl` (perferred way), but we need to remove `environment_check.sh` in the end.

derekbit: > @derekbit should we include the preflight installation in this script?

I don't think so. WDYT?

c3y1huang: > > @derekbit should we include the preflight installation in this script?
> 
> I don't think so. WDYT?

I don't think we should include it. It complicates the script and obscures the main purpose.

c3y1huang: > You're right. IIRC, we don't handle `cos` in the original environemnt_check.sh. Can we add a paragraph for the operating system and similar cases? WDYT @c3y1huang @mantissahz @innobead
> 

I don't think we should include the preflight install.

> Or, I'm fine with retaining both `environment_check.sh` with the old codes (marked it as deprecated) and `longhornctl` (perferred way), but we need to remove `environment_check.sh` in the end.

I am in favor of deprecating the `environment_check.sh`.

derekbit:
@mantissahz Can you update the script according to the discussion? Thank you.

scripts/environment_check.sh
yangchiu: > You're right. IIRC, we don't handle cos in the original environemnt_check.sh. Can we add a paragraph for the operating system and similar cases?

For your reference, `longhornctl` doesn't work on `amazon linux` and `sle-micro` as well:

```
msg="failed to get package manager" error="operating systems (amzn, sl-micro) are not supported"
```

derekbit: > > You're right. IIRC, we don't handle cos in the original environemnt_check.sh. Can we add a paragraph for the operating system and similar cases?
> 
> For your reference, `longhornctl` doesn't work on `amazon linux` and `sle-micro` as well:
> 
> ```
> msg="failed to get package manager" error="operating systems (amzn, sl-micro) are not supported"
> ```

@yangchiu Can you create a BUG ticket? We can resolve it in v1.8.0.

mantissahz:
>@mantissahz Can you update the script according to the discussion? Thank you.

Hi @derekbit, 

According to the discussion, there are two ways to go:

1. If we retain both environment_check.sh with the old codes and mark it as deprecated, we will close this PR and have another PR that will update the document.
2. Remove the `longhornctl` installation part of this PR and still we will need to update the document.

I will vote for deprecating the environment_check.sh

scripts/environment_check.sh
yangchiu: Created https://github.com/longhorn/longhorn/issues/9139 for it. Thank you!

derekbit:
> > @mantissahz Can you update the script according to the discussion? Thank you.
> 
> Hi @derekbit,
> 
> According to the discussion, there are two ways to go (which is the final choice?):
> 
> 1. If we retain both environment_check.sh with the old codes and mark it as deprecated, we will close this PR and have another PR that will update the document.
> 2. Remove the `longhornctl` installation part of this PR and still we will need to update the document.
> 
> I will vote for deprecating the environment_check.sh

We can deprecate the environment_check.sh in v.18.0, because there are some OS supportability issues. such as https://github.com/longhorn/longhorn/issues/9139.

IMHO, dowloading and then executing `longhornctl` is not a good UX. We can figure out the better way later.
One example is 
```
curl -sfL https://get.k3s.io | K3S_URL=https://myserver:6443 K3S_TOKEN=mynodetoken sh -
```

cc @innobead @c3y1huang

mantissahz:
>IMHO, dowloading and then executing longhornctl is not a good UX. We can figure out a better way later.

I agree with you too.

Close this PR because 
```txt
For v1.7.0, we can retain the two methods and no need to update envrionment_check.sh for now.
```

Thanks.

Issue #8746, opened 2024-06-13
[IMPROVEMENT] Make backup deletion async and force backup creation wait until there is no backup being delete


ChanYiLin:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

- Make backup deletion async
    - Currently, backup deletion is blocking call by using binary to delete the backup in the backupstore
    - This is time consuming and not follow the eventual consistency convention in K8s
    - By making it async, we can add `"Deleting"` status to postpone other backups creation to prevent deletion lock issue. 
- Wait for backup deletion to be finished
    - Backup creation will fail when there is another backup being delete. We can improve it by making backup wait until there is no backup being deleted before triggering the backup process.

~~- Monitor progress time
    - We can help users to record the progress time of the backup. This can help users to adjust the interval time of Recurring Job to prevent two jobs running at the same time.~~ => we disable recurring job overlap, so two jobs won't run at the same tmie.

## Describe the solution you'd like

.

## Describe alternatives you've considered
.

## Additional context
.

cc @derekbit

derekbit:
Can we move it to v1.8.0? WDYT? @ChanYiLin @innobead

derekbit:
@ChanYiLin Let's add it into v1.8.0 important note.

roger-ryao:
**Verified on v1.8.0-dev-20241103 20241113**

- longhorn v1.8.0-dev-20241103 https://github.com/longhorn/longhorn/commit/8265d91c74e0d4d83608d6210628215315a6fab9

**The test steps**
https://github.com/longhorn/longhorn/pull/9152

Cases:
1. Normal Case
2. Error Case (use nfs)
3. Controller Crashes Case (use nfs)


**Result Passed**

- [x] 1. Normal Case
- [x] 2. Error Case (use nfs)
- [x] 3. Controller Crashes Case (use nfs)
- [ ] 4. require/manual-test-plan ❌


Hi @ChanYiLin 
I suggest we reference the following documents for manual test planning in v1.8.0:
- https://github.com/longhorn/longhorn-tests/blob/master/docs/content/manual/release-specific/v1.7.0/test-storage-network-support-for-rwx-volume.md
- https://longhorn.github.io/longhorn-tests/manual/release-specific/v1.7.0/test-rwx-fast-failover/

derekbit:
@ChanYiLin What's the status of the ticket?

derekbit:
@ChanYiLin Let's move this forward.

[...]

PR #9152, opened 2024-08-01
feat(LEP): delete backup in the backupstore asynchronously

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/8746

Add LEP for backup deletion refactor.
To prevent deletion lock issue from happening, we should disallow backup creation when there is a backup being created.
We refactor the backup deletion flow to make it asynchronous and add a new state `Deleting`

### DIFF
--- a/enhancements/20240801-delete-backup-in-the-backupstore-asynchronously.md
+++ b/enhancements/20240801-delete-backup-in-the-backupstore-asynchronously.md
@@ -1,0 +1,171 @@
+# Delete backup in the backupstore asynchronously
+
+## Summary
+
+When deleting the backup, Longhorn deletes the backup in the backupstore using binary command synchrounosly. During the backup deletion, Longhorn creates a deletion lock in the backupstore to prevent other backups' creation. Backup creation fails immediately without retrying when encountering the deletion lock. This confuses and annoys users.
+
+This feature refactors the code to execute the backup deletion command asynchronously and introduces a new `Deleting` state to notify other backups to wait until the deletion process is complete.
+
+This feature aims to prevent creation failures caused by deletion lock issues.
+
+### Related Issues
+
+- https://github.com/longhorn/longhorn/issues/8742
+
+## Motivation
+
+### Goals
+
+- Delete the backup in the backupstore asynchronously.
+- When the Backup is being deleted, the state would be `Deleting`
+- Creation of other backups should be delayed until the backup deletion is complete.
+
+## Design
+
+### Implementation Overview
+
+#### Normal case
+
+To make deletion asynchronously, we can simply run the command in another go routine.
+
+Then we periodically check if the backup still exists in the backupstore before removing the finalizer.
+
+However, since the command runs in another go routine, we need to handle the scenario of command failure.
+
+Thus, we introduce an in-memory map as share memory for the go routine to notify the controller.
+
+### Error handling
+
+```
+type DeletingStatus struct {
+	State        longhorn.BackupState
+	ErrorMessage string
+}
+
+type BackupController struct {
+    ...
+    // Use to track the result of the deletion command.
+    // Also used to track if controller crashes after the deletion command is triggered.
+    deletingMapLock       *sync.Mutex
+    inProgressDeletingMap map[string]*DeletingStatus
+    ...
+}
+```
+
+Go routine can add the error message to the `inProgressDeletingMap` if the command fails. Controller then can retry the command in the next reconciliation.
+
+### Deleting/Error State with BackOff
+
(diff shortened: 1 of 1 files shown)


enhancements/20240801-delete-backup-in-the-backupstore-asynchronously.md
```diff
@@ -0,0 +1,171 @@
...
+```
+# Normal Case
+Completed => Deleting 
+=> finalizer removed (CR isgone)
+
+# Command failure Case
+Completed => Deleting 
+=> Error (found the error message in the map) 
+=> Deleting (Retry the command) => finalizer removed (CR isgone)
+```
```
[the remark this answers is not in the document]
ChanYiLin: I think extending the max delay to 1 time per day makes sense
Does it sound good you?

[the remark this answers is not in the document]
ChanYiLin: Because if we have limited retry, we may need another mechanism for users to trigger the backup deletion again even it's CR already has deletionTimeStamp.

ChanYiLin:
### Conclusion from the discussion
1. We can extend the retry delay to 1 time per day, so the error message can be more easily caught by the users. It also decrease the burden of the retry.

enhancements/20240801-delete-backup-in-the-backupstore-asynchronously.md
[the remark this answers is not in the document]
derekbit: Sounds good. Can we move the error message to a condition such as `Deletion`?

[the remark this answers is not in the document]
ChanYiLin: After discussion, since we have many kind of error message to show
- Deletion error
- Backup creation error
We can summarize organize them in the future.
Currently we will only show them in the .Status.Messages for users to read.

enhancements/20240801-delete-backup-in-the-backupstore-asynchronously.md
```diff
@@ -0,0 +1,171 @@
...
+1. Create a Volume
+2. Write some data and then create a BackupA
+3. Write some data and then create a sSnapshot
+4. Exec into the backupstore pod and make the backup.cfg immutable
+    ```
+    $ chattr +i backups/backup_backup-5640dfd33a054f98.cfg`
+    ```
+5. Delete the BackupA, at the same time, create a BackupB from the Snapshot(you can click from UI)
+6. BackupA will be in `Deleting` and `Error` state repeatedly to retry the deletion. When in `Error` state, it shows error message related to permission 
+7. BackupB will be `InProgress` when BackupA is in `Deleting` state. BackupB should be complete after awhile.
```
shuo-wu, line 148: Do you mean that `BackupB will be InProgress when BackupA is in Error state`?

enhancements/20240801-delete-backup-in-the-backupstore-asynchronously.md
```diff
@@ -0,0 +1,171 @@
...
+
+1. Create a Volume
+2. Write some data and then create a BackupA
+3. Exec into the backupstore pod and make the backup.cfg immutable, example
+    ```
+    $ chattr +i backups/backup_backup-5640dfd33a054f98.cfg`
+    ```
+5. Delete the BackupA
+6. BackupA will be in `Deleting` and `Error` state repeatedly to retry the deletion. When in `Error` state, it shows error message related to permission
+7. When the BackupA is in `Deleting` state, delete the longhorn manager pod directly. (you can find the one doing the deleting with `Backup.Status.OwnerID`)
```
shuo-wu, line 161: Will there be an ownership transfer?

shuo-wu (approved):
In general LGTM

enhancements/20240801-delete-backup-in-the-backupstore-asynchronously.md
derekbit: cc @ChanYiLin

enhancements/20240801-delete-backup-in-the-backupstore-asynchronously.md
ChanYiLin: Yes you are right

enhancements/20240801-delete-backup-in-the-backupstore-asynchronously.md
ChanYiLin: It is okay that it gets ownership transfer
because for another manager, it will also find out that the Backup is in `Deleting` state, but there is no record in the in-memory map
Thus update the Backup to error state to retry the deletion command.

derekbit (approved):

innobead (approved):

Issue #9155, opened 2024-08-01
[TASK] Migrate ZenHub Board to GitHub Project


derekbit:

derekbit:
Completed.

PR #9156, opened 2024-08-01
ci(workflow): add "Migrate ZenHub Board to GitHub Project" GitHub Action

derekbit:

Issue Longhorn/longhorn#9155


innobead (approved):
LGTM. 

A side note for iteration. Just make sure that after the current sprint finishes, the unresolved issues in the sprint should be removed from the current sprint. However , members can add them back to the current sprint if they want to continue working on that.

innobead (approved):

Issue #8954, opened 2024-07-08
[BUG] Successful backups to an S3 compatible Backend are not displayed in the "Backups" section of the UI in Longhorn v1.6.2


conall88:
## Describe the bug

Successful backups to an S3 compatible Backend are not displayed in the "Backups" section of the UI.
![image](https://github.com/longhorn/longhorn/assets/10930485/63fedaf4-2c2f-415a-9ea8-60382c60e2f2)

Note that the following API calls when navigating to this section return HTTP200 OK:

```
/nodes
/settings
/backupimages
/backupvolumes
/backuptargets 
```

## To Reproduce

Create secret (base64 decoded values):
```
apiVersion: v1
data:
  AWS_ACCESS_KEY_ID: 60c8d7...
  AWS_ENDPOINTS: https://contabostorage.com
  AWS_REGION: default
  AWS_SECRET_ACCESS_KEY: 9bfe...
  VIRTUAL_HOSTED_STYLE: true
kind: Secret
metadata:
  annotations:
    longhorn.io/backup-target: s3://eu2@contabo/conall-s3-cos/longhorn-backup
  creationTimestamp: "2024-06-29T19:59:54Z"
  name: contabo-s3-secret
  namespace: longhorn-system
  resourceVersion: "3038189"
  uid: bd090cd3-ce4f-43c6-9515-c06899e67daf
type: Opaque
```

Define backup target and credential secret in General configuration:
![image](https://github.com/longhorn/longhorn/assets/10930485/66c19319-9074-4b4c-ad9a-5089d3dcbf35)

Define a backup schedule for the default group:

![image](https://github.com/longhorn/longhorn/assets/10930485/8b39e007-18c7-4939-b86f-1e606a32016b)

Successfully take  snapshots and backups:
![image](https://github.com/longhorn/longhorn/assets/10930485/ef24ec19-e098-4630-8683-ebf9e87a8719)

Confirm scheduled backups have succeeded, and backup data is present in the S3 Bucket:

![image](https://github.com/longhorn/longhorn/assets/10930485/00e68c7d-8392-4487-bf61-f4e8a9e0e7d7)

## Expected behavior

Backups should be listed and selectable to restore via the UI.

## Support bundle for troubleshooting
[...]

Issue #8319, opened 2024-04-08
[BUG] Old backups are not cleaned up after timeout


gothka:
## Describe the bug
We've about 20,000+ old backups that are stale and using up resources on the cluster and could potentially blow up the cluster anytime due to this. We believe there's a few reasons on why this is happening: 

the default timeout for backup is hardcoded to 1 minute here https://github.com/longhorn/go-common-libs/blob/3993c59084471a20ccd863282a0af1757b7a9fbf/types/exec.go#L14 due to this when backup is failed during backing up to s3 in our case, all the stale/old backups are left on the cluster without LH engine cleaning them. 

https://github.com/longhorn/longhorn-manager/blob/91bc9cb701ac3116d37311fb8c732ac049a560f4/engineapi/backups.go#L169

We've to clean this up manually, the command below shows all the old backups and these object don't have `BackupVolume` set to them so they can't be restored unless manually created on the cluster

`k get backups.longhorn.io -n longhorn-system -o json | jq '.items[]|select(.status.state!="")|select(.metadata.creationTimestamp | fromdateiso8601 < 1711554025)|.metadata.name'` 

This command roughly outputs 26,000+ old backups that are stale and worth mentioning that deleting is easy but it'll incur s3 api costs which isn't quite cheap given num of api calls the client has to make to the s3 bucket.

This is happening on one of the cluster where manage more than 700 volumes in the cluster.

## To Reproduce

Run a cluster with 700+, and set the backup target with s3 and let the backups fail and eventually you'll old backups piling up without being deleted.

## Expected behavior
All old backups whether they're backed up or not should be deleted from the cluster.

## Support bundle for troubleshooting

Sent via email

## Environment

 - Longhorn version: v1.4.3
 - Impacted volume (PV): 
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl): Rancher catalog
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: EKS
   - Number of control plane nodes in the cluster: EKS managed
   - Number of worker nodes in the cluster: 60+
 - Node config
   - OS type and version: Amazon linux 2
   - Kernel version:
   - CPU per node: 2 to 4
   - Memory per node: 8+ Gi
   - Disk type (e.g. SSD/NVMe/HDD): NVMe/SSD
   - Network bandwidth between the nodes (Gbps): 
[...]

PR #9157, opened 2024-08-02
fix(setting): add a default setting `backupExecutionTimeout`

mantissahz:

Issue # longhorn/longhorn#8954 longhorn/longhorn#8319

### DIFF
--- a/chart/README.md
+++ b/chart/README.md
@@ -284,6 +284,7 @@ During installation, you can either allow Longhorn to use the default system set
 | defaultSettings.backingImageRecoveryWaitInterval | Number of seconds that Longhorn waits before downloading a backing image file again when the status of all image disk files changes to "failed" or "unknown". |
 | defaultSettings.backupCompressionMethod | Setting that allows you to specify a backup compression method. |
 | defaultSettings.backupConcurrentLimit | Maximum number of worker threads that can concurrently run for each backup. |
+| defaultSettings.backupExecutionTimeout | Number of minutes that Longhorn allows for the backup execution. The default value is "1". |
 | defaultSettings.backupTarget | Endpoint used to access the backupstore. (Options: "NFS", "CIFS", "AWS", "GCP", "AZURE") |
 | defaultSettings.backupTargetCredentialSecret | Name of the Kubernetes secret associated with the backup target. |
 | defaultSettings.backupstorePollInterval | Number of seconds that Longhorn waits before checking the backupstore for new backups. The default value is "300". When the value is "0", polling is disabled. |
@@ -300,7 +301,6 @@ During installation, you can either allow Longhorn to use the default system set
 | defaultSettings.disableRevisionCounter | Setting that disables the revision counter and thereby prevents Longhorn from tracking all write operations to a volume. When salvaging a volume, Longhorn uses properties of the "volume-head-xxx.img" file (the last file size and the last time the file was modified) to select the replica to be used for volume recovery. This setting applies only to volumes created using the Longhorn UI. |
 | defaultSettings.disableSchedulingOnCordonedNode | Setting that prevents Longhorn Manager from scheduling replicas on a cordoned Kubernetes node. This setting is enabled by default. |
 | defaultSettings.disableSnapshotPurge | Setting that temporarily prevents all attempts to purge volume snapshots. |
-| defaultSettings.rwxVolumeFastFailover | Turn on logic to detect and move RWX volumes quickly on node failure. |
 | defaultSettings.engineReplicaTimeout | Timeout between the Longhorn Engine and replicas. Specify a value between "8" and "30" seconds. The default value is "8". |
 | defaultSettings.failedBackupTTL | Number of minutes that Longhorn keeps a failed backup resource. When the value is "0", automatic deletion is disabled. |
 | defaultSettings.fastReplicaRebuildEnabled | Setting that allows fast rebuilding of replicas using the checksum of snapshot disk files. Before enabling this setting, you must set the snapshot-data-integrity value to "enable" or "fast-check". |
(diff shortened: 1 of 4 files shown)


chart/README.md
```diff
@@ -284,6 +284,7 @@ During installation, you can either allow Longhorn to use the default system set
 | defaultSettings.backingImageRecoveryWaitInterval | Number of seconds that Longhorn waits before downloading a backing image file again when the status of all image disk files changes to "failed" or "unknown". |
 | defaultSettings.backupCompressionMethod | Setting that allows you to specify a backup compression method. |
 | defaultSettings.backupConcurrentLimit | Maximum number of worker threads that can concurrently run for each backup. |
+| defaultSettings.backupExecuteTimeout | Number of minutes that Longhorn sets the timeout for the backup engine execution. When the value is "0", automatic deletion is disabled. |
```
derekbit, line 287: How about `backupExecutionTimeout`?
What does `automatic deletion is disabled` mean?

mantissahz: Renamed.

>What does automatic deletion is disabled mean?

Updated for disabling the timeout.

chart/questions.yaml
```diff
@@ -374,6 +374,13 @@ questions:
     type: int
     min: 0
     default: 1440
+  - variable: defaultSettings.backupExecutionTimeout
+    label: Backup Execute Timeout
+    description: "Number of minutes that Longhorn sets the timeout for the backup engine execution. When the value is \"0\", disable the timeout."
```
derekbit, line 379: What does `disable the timeout` mean?

mantissahz: The request/operation (binary command) will not be interrupted by a timeout (there is no timeout).

derekbit: Got it. Can we make it clearer?
For example, `If timeout is -1, it will wait infinitely until .....`.

mantissahz: Modified. Thanks

chart/questions.yaml
```diff
@@ -374,6 +374,13 @@ questions:
     type: int
     min: 0
     default: 1440
+  - variable: defaultSettings.backupExecutionTimeout
+    label: Backup Execute Timeout
+    description: "Number of minutes that Longhorn sets the timeout for the backup engine execution. If timeout is set to 0, it will wait indefinitely until the execution returns."
```
derekbit, line 379: I'm thinking whether 0 is reasonable. WDYT? @c3y1huang @ChanYiLin @shuo-wu

shuo-wu: Although backup creation is an async call, I don't think an infinite wait is the best default setting... But on the other hand, do we support interrupting the progress when the timeout reaches?

mantissahz: >I don't think an infinite wait is the best default setting

@shuo-wu, sorry for misleading you, the default setting is 1 minute.

>But on the other hand, do we support interrupting the progress when the timeout reaches?

I think we only return an error when the timeout reaches.

shuo-wu: > > But on the other hand, do we support interrupting the progress when the timeout reaches?
>
> I think we only return an error when the timeout reaches.

I mean, how can Longhorn stop it if the backup creation actually starts after the timeout?

mantissahz: If the backup creation actually starts after the timeout (the Error `DeadlineExceeded`), the backup state will be marked as `pending` and wait for the next reconciling.

https://github.com/longhorn/longhorn-manager/blob/abb8f260796dcc5ee1ab84c5a20385217aeb7243/engineapi/backup_monitor.go#L77-L119

Longhorn would not stop the backup creation when the timeout reaches.

But the setting `backupExecutionTimeout` only handles the backup binary operations like `engines/longhorn backup ls --volume-only`, `backup inspect-volume`, and so on.

mantissahz: And I think disabling the timeout here is not a good idea and we have [ExecuteEngineBinaryWithoutTimeout](https://github.com/longhorn/longhorn-manager/blob/abb8f260796dcc5ee1ab84c5a20385217aeb7243/engineapi/backups.go#L144) already if the operations need to wait until they complete.

mantissahz: Yes, WDYT?
Some operations need a timeout.

derekbit: Sounds good to me.

mantissahz: Modified the default value to 1, and the value can not be smaller than 1.

chart/questions.yaml
```diff
@@ -374,6 +374,13 @@ questions:
     type: int
     min: 0
     default: 1440
+  - variable: defaultSettings.backupExecutionTimeout
+    label: Backup Execute Timeout
+    description: "Number of minutes that Longhorn sets the timeout for the backup engine execution. The default value is \"1\"."
+    group: "Longhorn Default Settings"
+    type: int
+    min: 0
```
derekbit, line 382: Does it mean `0` is allowed?

derekbit: In addition, why not use seconds?

mantissahz: >Does it mean 0 is allowed?

Yes, modified to 1

>In addition, why not use seconds?

According to the user's response, they usually set a long timeout (10 minutes).
Either one (second or minute) is fine with me. WDYT?

derekbit: > And the origin time unit from the library go-common-libs is minute.

Can you guide me the link? Thank you.

mantissahz: here it is https://github.com/longhorn/go-common-libs/blob/907f57efd48f43e1ca32f4ffa634d3c6b55b5a54/types/exec.go#L14

derekbit: Thank you.

chart/questions.yaml
```diff
@@ -374,6 +374,13 @@ questions:
     type: int
     min: 0
     default: 1440
+  - variable: defaultSettings.backupExecutionTimeout
+    label: Backup Execute Timeout
+    description: "Number of minutes that Longhorn sets the timeout for the backup engine execution. The default value is \"1\"."
```
derekbit, line 379: @jillian-maroket @jhkrug Can you review the wording? Thank you.

chart/README.md
```diff
@@ -325,6 +325,7 @@ During installation, you can either allow Longhorn to use the default system set
 | defaultSettings.replicaZoneSoftAntiAffinity | Setting that allows Longhorn to schedule new replicas of a volume to nodes in the same zone as existing healthy replicas. Nodes that do not belong to any zone are treated as existing in the zone that contains healthy replicas. When identifying zones, Longhorn relies on the label "topology.kubernetes.io/zone=<Zone name of the node>" in the Kubernetes node object. |
 | defaultSettings.restoreConcurrentLimit | Maximum number of worker threads that can concurrently run for each restore operation. |
 | defaultSettings.restoreVolumeRecurringJobs | Setting that restores recurring jobs from a backup volume on a backup target and creates recurring jobs if none exist during backup restoration. |
+| defaultSettings.rwxVolumeFastFailover | Turn on logic to detect and move RWX volumes quickly on node failure. |
```
jillian-maroket, line 328: The first part is unclear so here are my best guesses:

Option 1: Setting that allows Longhorn to detect node failure and immediately migrate affected RWX volumes.
Option 2: Setting that allows Longhorn to quickly migrate RWX volumes affected by node failure.

chart/questions.yaml
jillian-maroket: I did not find any mentions of a "backup engine" in the doc. The draft is unclear so here are my best guesses:

Option 1: Number of minutes that Longhorn allows for backup creation. The default value is \"1\"
Option 2: Number of minutes the backup engine is allowed to run. The default value is \"1\".

jillian-maroket (commented):
I provided a few options instead of definite suggestions because the drafts are unclear.

chart/questions.yaml
derekbit: @mantissahz Should be `backup execution`?

mantissahz: >@mantissahz Should be backup execution?

Yes, and modified like `Option 1`
```txt
Number of minutes that Longhorn allows for backup execution. The default value is "1"
```

chart/README.md
mantissahz: Modified to `Option 1`. Thank you.

jillian-maroket (approved):

derekbit (approved):
LGTM

mantissahz:
@Mergifyio backport v1.7.x

Issue #9237, opened 2024-08-15
[TASK] Longhorn sprint release 


derekbit:
## What's the task? Please describe

Development, Release, and Verification Workflow:

- During the Current Sprint:
  - Developers address assigned issues and move them to `Ready for Testing` once merged.
  - At the end of the sprint, a sprint release will be automatically published.
- In the Following Sprint:
  - QA teams will verify the `Ready for Testing` items from the previous sprint using the published sprint release.
  - Update the sprint period to the current sprint

## Additional context

cc @innobead @khushboo-rancher @longhorn/dev @longhorn/qa

PR #9249, opened 2024-08-18
feat(ci): automatically scan and notify testing items every Sunday

derekbit:

Issue Longhorn/longhorn#9237

### DIFF
--- a/.github/workflows/scan_and_notify_testing_items.py
+++ b/.github/workflows/scan_and_notify_testing_items.py
@@ -1,0 +1,307 @@
+import requests
+import os
+import sys
+import time
+import json
+from datetime import datetime, timedelta
+
+
+GITHUB_GRAPHQL_URL = "https://api.github.com/graphql"
+
+
+def get_github_project_info(github_token, github_org, github_project):
+    headers = {
+        "Authorization": f"Bearer {github_token}",
+        "Content-Type": "application/json"
+    }
+    query = '''
+    {
+      organization(login: "%s") {
+        projectsV2(first: 20) {
+          nodes {
+            id
+            title
+            number
+          }
+        }
+      }
+    }
+    ''' % (github_org)
+    payload = {
+        "query": query
+    }
+
+    response = requests.post(GITHUB_GRAPHQL_URL, headers=headers, json=payload)
+    if response.status_code == 200:
+        # fine project by title
+        nodes = response.json().get("data").get("organization").get("projectsV2").get("nodes")
+        for node in nodes:
+            if node.get("title") == github_project:
+                return node
+    else:
+        response.raise_for_status()
+
+
+def get_current_sprint(github_token, project_id):
+    headers = {
+        "Authorization": f"Bearer {github_token}",
+        "Content-Type": "application/json"
+    }
+    query = '''
+    query {
+      node(id: "%s") {
+        ... on ProjectV2 {
+          fields(first: 20) {
+            nodes {
+              ... on ProjectV2IterationField {
+                configuration {
+                  iterations {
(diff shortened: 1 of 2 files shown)


.github/workflows/scan_and_notify_testing_items.py
```diff
@@ -0,0 +1,163 @@
...
+
+def scan_and_notify(github_org, github_repo, github_project):
+    github_token = os.getenv("GITHUB_TOKEN")
+    webhook_url = os.getenv("SLACK_WEBHOOK_URL")
+
+    project = get_github_project_info(github_token, github_org, github_project)
+    print(f"GitHub Project Details: {project}")
+
+    project_id = project.get("id")
+    issues = list_issues_in_project(github_token,
```
innobead, line 150: Can we filter the issues belonging to the current sprint only? For other non-current sprints, QA should track them on another tab of the project or use a different workflow if need an automatic notification.

This workflow is for letting the team know how many issues are ready for testing by the end of the current sprint.

innobead (changes requested):
In general, LGTM. Just one feedback.

.github/workflows/scan_and_notify_testing_items.yml
```diff
@@ -0,0 +1,63 @@
+name: Scan and Notify Testing Items
+
+on:
+  schedule:
+    # Trigger every Sunday at 12:00
+    - cron: '0 12 * * 0'
```
innobead, line 6: Ideally, we should trigger this every two weeks before the end of the sprint. However, since the cron syntax may not support that frequency, a weekly trigger works for me

.github/workflows/scan_and_notify_testing_items.py
derekbit: > Can we filter the issues belonging to the current sprint only? For other non-current sprints, QA should track them on another tab of the project or use a different workflow if need an automatic notification.
> 
> This workflow is for letting the team know how many issues are ready for testing by the end of the current sprint.

Can we show two lists at the same time? One is for current sprint and the other one is for non-current sprints.

```
List of current sprints:
1.
2.
3.
...

List of non-current sprints:
1.
2.
3.
...
```

.github/workflows/scan_and_notify_testing_items.yml
derekbit: Just figured out a better method. We can run the workflow every Sunday and compare the date with the Project's current print. Then, scan and notify only when it is the end of the sprint. I will improve the PR.

.github/workflows/scan_and_notify_testing_items.py
innobead: What's the reason for showing non current sprint issues?

BTW, can we add the issue owners next to each issue? I think this is helpful for members quickly know their status.

derekbit: > What's the reason for showing non current sprint issues?

To avoid missing the non-current sprint issues that should be tested before.

> BTW, can we add the issue owners next to each issue? I think this is helpful for members quickly know their status.

Sure.

innobead: Yeah. That was why I said we can have a separate workflow for this purpose. But, let's make it simple  to have both in this workflow fine for me!

derekbit:
The colleague of IT team for slack app is on vacation. Waiting for his approval for incoming webook.
https://suse.slack.com/archives/C02CWABQMEX/p1723887838447749

derekbit:
Ready for review.

yangchiu (approved):
LGTM

Issue #9247, opened 2024-08-16
Improvement of OpenSSF Scorecard Score


harshitasao:
Hi, I'm Harshita. I’m working with [CNCF and the Google Open Source Security Team for the GSoC 2024 term](https://github.com/cncf/mentoring/issues/1196). We are collaborating to enhance security practices across various CNCF projects. The goal is to improve security for all CNCF projects by both using OpenSSF Scorecards and implementing its security improvements.

As this project already has scorecard action, I'm here to increase the final score by going over each check. I've listed all of the checks where work needs to be done, in order of its criticality. I plan to submit each PR for each fix. Please let me know what you think and for which ones a PR is welcome that I will submit it ASAP.

**Current Score:** 6.5

**Scorecard report:** https://scorecard.dev/viewer/?uri=github.com/longhorn/longhorn

Here's a few checks we can work on to improve the project's security posture:

- [ ] [Signed-Releases](https://github.com/ossf/scorecard/blob/main/docs/checks.md#signed-releases): Score = 0
    - The releases need to be signed; we can use [cosign using github actions ](https://github.com/avisi-cloud/cosign-tutorial#using-github-actions) for that, and it will increase the score to 8. Additionally, [SLSA provenance](https://slsa.dev/spec/v0.1/index) also needs to be added, which can be done using [this](https://github.com/slsa-framework/slsa-github-generator), increasing the overall score to 10.

- [X] [Token-Permissions](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions): Score = 0
    - The issue here is that many workflows doesn’t have a top-level read-only permissions block like present in the other workflows. Scorecard is quite severe in this check: a single workflow without top-level permissions gets a 0/10 for the check.
    - PR: https://github.com/longhorn/longhorn/pull/9250

- [ ] [Fuzzing](https://github.com/ossf/scorecard/blob/main/docs/checks.md#fuzzing): Score = 0
    - Integrating the project with OSS-Fuzz by following the instructions [here](https://google.github.io/oss-fuzz/). The most difficult one on the list, maintainers help, is highly appreciated. For example, helping in identifying the components where fuzz testing will be added.

- [X] [Pinned-Dependencies](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies): Score = 0
    - Github actions are not pinned by hash, which results in a low score.
    - Pinning dependencies to a specific hash rather than allowing mutable versions or ranges of versions improves supply chain security.
[...]

PR #9250, opened 2024-08-18
 fix: fixed token permission and pinned dependencies issue

harshitasao:

Part of #9247 

fixed token permission and pinned dependencies issue reported by [scorecard](https://scorecard.dev/viewer/?uri=github.com/longhorn/longhorn).

### DIFF
--- a/.github/workflows/add-to-projects.yml
+++ b/.github/workflows/add-to-projects.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Is Longhorn Member
-      uses: tspascoal/get-user-teams-membership@v1.0.4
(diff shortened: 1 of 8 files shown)


.github/workflows/add-to-projects.yml
```diff
@@ -7,15 +7,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Is Longhorn Member
-      uses: tspascoal/get-user-teams-membership@v1.0.4
+      uses: tspascoal/get-user-teams-membership@39b5264024b7c3bd7480de2f2c8d3076eed49ec5 # v1.0.4
```
innobead, line 10: Why do we specify the version by a commit hash instead of a tag? Suggest still using a semantic version instead of commit raw string.

innobead (changes requested):
Thanks for the contribution. One feedback needs to resolve.

derekbit:
@harshitasao Can you help resolve the comment from @innobead? Thanks.

Issue #9347, opened 2024-08-29
[BUG] RWX fast failover setting not included in upgrade responder metrics.


james-munson:
## Describe the bug

It would be good to know how much the setting rwx-volume-fast-failover is used, but it is currently omitted from the upgrade responder settings metrics.

## Expected behavior

Upgrade responder shows this setting.

## Environment

 - Longhorn version: 
 - Impacted volume (PV): 
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl):
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version:
   - Number of control plane nodes in the cluster:
   - Number of worker nodes in the cluster:
 - Node config
   - OS type and version:
   - Kernel version:
   - CPU per node:
   - Memory per node:
   - Disk type (e.g. SSD/NVMe/HDD):
   - Network bandwidth between the nodes (Gbps):
 - Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal):
 - Number of Longhorn volumes in the cluster:

innobead:
Do we know why we need to backport this to 1.7.x? Upgrade responder should be independent from Longhorn releases.

PR #9376, opened 2024-09-03
fix(metrics): add rwx-volume-fast-failover to upgrade responder chart

james-munson:

https://github.com/longhorn/longhorn/issues/9347

Called out as necessary by @PhanLe1010 in https://github.com/longhorn/longhorn-manager/pull/3121#issuecomment-2327136164

Also moved entries for `FreezeFilesystemForSnapshot` to keep the list alphabetized, and added some missing `{}`.
Made my best guess as to the capitalization of the setting tag.


derekbit (approved):
LGTM

PhanLe1010 (approved):

Issue #8884, opened 2024-07-03
[IMPROVEMENT] backup backing image should store the secret and secret namespace if it is encrypted


ChanYiLin:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

If users backup an encrypted backing image 
The secret/secretnamespace information is not stored in the backupstore backup backing image config 
So the `BackupBackingImage` CR won't have the information

If users want to restore it back, we still need to add the secret/secretnamespace to the restored `BackingImage` CR
Thus, in the v1.7.0, we allow users to fill in the secret/secretnamespace when restoring the backing image

## Describe the solution you'd like

However, 
we should store the information in the backup store
so we can sync the information back and stored in `BackupBackingImage` CR
When users want to restore the backing image, the information will be in the `BackingImage` CR automatically.

## Describe alternatives you've considered
.
## Additional context
cc @derekbit @a110605

ChanYiLin:
PRs:

- longhron-type: https://github.com/longhorn/types/pull/22

yangchiu:
Verification passed on master-head (longhorn-manager 4d77907) following [the test steps](https://github.com/longhorn/longhorn/issues/8884#issuecomment-2235236505).

Created an encrypted backing image following the steps in https://longhorn.io/docs/1.7.1/advanced-resources/backing-image/backing-image-encryption/. After Backing up the encrypted backing image, checked the corresponding fields of the encrypted backing image backup, confirmed that `secret` and `secretNamespace` were present as expected.

```
$ kubectl get backupbackingimages.longhorn.io -n longhorn-system parrot-cloned-encrypted -oyaml
apiVersion: longhorn.io/v1beta2
kind: BackupBackingImage
metadata:
  creationTimestamp: "2024-09-30T06:50:50Z"
  finalizers:
  - longhorn.io
  generation: 1
  name: parrot-cloned-encrypted
  namespace: longhorn-system
  resourceVersion: "20711"
  uid: 052742d6-6677-47c9-848c-1e678e0358ff
spec:
  syncRequestedAt: null
  userCreated: true
status:
  backingImage: parrot-cloned-encrypted
  backupCreatedAt: "2024-09-30T06:50:53Z"
  checksum: 4d0eee6c81f1282988ffd86e08366bef368b423b3919ad8256f8891c4a18645ab7d2af4a55c540df6938046d5ad2a08d316ae0fe6711ac8be0ad015fd26149dc
  compressionMethod: lz4
  labels: null
[...]

PR #9434, opened 2024-09-10
feat(crd): add secret and secretnamespace in backupbackingimage

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/8884

update
- `backupBackingImage` status CRD to add `secret` and `secretNamespace`


derekbit (approved):
LGTM

PR #9437, opened 2024-09-10
fix(metrics): add rwx-volume-fast-failover to upgrade responder insta…

james-munson:
…ll script.

https://github.com/longhorn/longhorn/issues/9347

It appears that dev/upgrade-responder/install.sh is meant to track the upgrade-responder chart with respect to the `requestSchema`, so catch it up.


PhanLe1010 (approved):
LGTM

james-munson:
@mergify backport v1.7.x

Issue #8430, opened 2024-04-24
[FEATURE] v2 volume supports auto salvage 


derekbit:
## Additional context

Some discussions
https://suse.slack.com/archives/C02DR3N5T24/p1713926457976349

DamiaSan:
See `data consistency` in https://github.com/longhorn/longhorn/pull/7031

chriscchien:
Verified **pass** on longhorn v1.8.0-dev(https://github.com/longhorn/longhorn-manager/commit/a503949c751f8bb07d9835f3680b93d5bc807e30)

Test steps:

1. Dynamic provision a v2 volume, having data in the mount point.
2. Set `auto-salvage` to `false`
3. Force delete all v2 volume instance-manager pod.
4. V2 volume become faulted and remains in that state.
5. Set `auto-salvage` to `true`
6. The faulted v2 volume become healthy and data in the workload is correct.
7. Force delete all v2 volume instance-manager pod again.
8. V2 olume become faulted then healthy again.
9. After volume healthy, data in workload correct.

In v1.7.2, set `auto-salvage` to `true` and force delete all v2 volume instance-manager pod, the v2 volume will loop from faulted -> attaching -> faulted.

PR #9486, opened 2024-09-19
feat(lep): auto-salvage support for v2 volumes

c3y1huang:

Issue longhorn/longhorn#8430

Proposes extending Longhorn's auto-salvage feature to support v2 volumes

`None`

https://suse.slack.com/archives/C02DR3N5T24/p1713926457976349

### DIFF
--- a/enhancements/20241028-auto-salvage-support-for-v2-volumes.md
+++ b/enhancements/20241028-auto-salvage-support-for-v2-volumes.md
@@ -1,0 +1,101 @@
+# Auto-salvage Support For V2 Volumes
+
+## Summary
+
+This document proposes extending Longhorn's auto-salvage feature to support v2 volumes. Currently, auto-salvage automatically recovers v1 volumes when all its replicas fail. This proposal aims to provide the same functionality for v2 volumes, improving data availability and reducing operational overhead.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/8430
+
+## Motivation
+
+### Goals
+
+- **Improve Data Availability For V2 Volumes:** Auto-salvage ensures v2 volumes automatically recovers from replica failures, minimize downtime.
+- **Reduce Operational Overhead:** Automating recovery, freeing user from manual intervention during failures.
+
+### Non-goals [optional]
+
+- `None`
+
+## Proposal
+
+### User Stories
+
+#### Story 1: Auto-salvage V2 Volumes When All Replicas Failed
+
+As a Longhorn user, I want Longhorn to automatically salvage a v2 volume when all its replicas fail, similar to the existing functionality for v1 volumes.
+
+- **Before:** Manual intervention was required to recover a faulted v2 volume when all replicas failed.
+- **After:** With auto-salvage enabled, Longhorn will attempt to salvage usable replicas and bring the v2 volume back online automatically.
+
+### User Experience In Detail
+
+- **Auto-salvage:** When the setting is enabled, if all replicas for a volume fail, Longhorn attempts salvage usable replicas to recover the volume state.
+- **Volume Trim Operation Blocking:** For degraded v2 volumes, filesystem trim operation will be blocked to preserve a reliable volume head size for identifying usable replica candidates.
+
+### API changes
+
+#### SPDK RPC Protobuf
+
+- Introduce `salvage_requested` boolean field in the `SpdkInstanceSpec` message. which is passed to the instance GRPC server during engine instance creation.
+  ```go
+  message SpdkInstanceSpec {
+  	map<string, string> replica_address_map = 1;
+  	string disk_name = 2;
+  	string disk_uuid = 3;
+  	uint64 size = 4;
+  	bool expose_required = 5;
+  	string frontend = 6;
+  	bool salvage_requested = 7;
+  }
+  ```
+
+- Introduce `salvage_requested` boolean field in the `EngineCreateRequest` message to pass to the SPDK server during engine creation.
+  ```go
+  message EngineCreateRequest {
+      string name = 1;
+      string volume_name = 2;
+      uint64 spec_size = 3;
+      map<string, string> replica_address_map = 4;
+      string frontend = 5;
+      int32 port_count = 6;
+      bool upgrade_required = 7;
+      string initiator_address = 8;
+      string target_address = 9;
+      bool salvage_requested = 10;
+  }
+  ```
+
+## Design
+
+> **Note:**
+> The design applies only to v2 volumes. The v1 volume candidate selection remains unchange.
+
+### Failed Usable Replica Filtering
+
+When `EngineCreate()` is called with `salvage_requested` set to `true`, the SPDK server retrieves `Replica` from SPDK server cache. We assume that all remaining replicas' lvols are identical and their lvol heads contain that latest data. The lvol head with the largest size is assumed to hold the most recent, valid data since larger size should indicates that it includes the latest writes or updates. Therefore, the head sizes of all replicas are sorted, and replicas with head sizes different from the largest in the sorted list are excluded from the candidate list.
+
+### Engine Creation
+
+During SPDK replica creation, if the replica already exists, the lvol bdev creation process is skipped. Instead, `Replica.construct()` is called to build the `Replica` object with existing lvol.
+
+### Volume Filesystem Trim Operation Blocking
+
+Block filesystem trim operation on degraded v2 volumes to preserve the lvol head size, ensuring reliable salvage replica candidate selection.
+
+### Test plan
+
+1. **Feature Testing:**
+   1. Use existing robot/integration test cases (cluster reboot, node powerdown, network disconnection, etc.) to verify auto-salvage functionality for v2 volumes after replica failures.
+   1. Introduce new robot test cases to verify that trim operations are blocked for degraded v2 volumes.
+1. **Regression Testing:** Ensure v1 volume auto-salvage and volume trim operation remains unaffected by the this extended feature implementation.
+
+### Upgrade strategy
+
+No specific upgrade strategy is required for this feature extension.
+
+## Note [optional]
+
+None


enhancements/20240919-auto-salvage-support-for-v2-volumes.md
```diff
@@ -0,0 +1,111 @@
+# Auto-salvage Support For V2 Volumes
+
+## Summary
+
+This document proposes extending Longhorn's auto-salvage feature to support v2 volumes. Currently, auto-salvage automatically recovers v1 volumes when all its replicas fail. This proposal aims to provide the same functionality for v2 volumes, improves data availability and reduing operational overhead.
```
DamiaSan, line 5: improving data availability and reducing operational overhead.

enhancements/20240919-auto-salvage-support-for-v2-volumes.md
```diff
@@ -0,0 +1,111 @@
...
+  	// +optional
+  	// +nullable
+  	Snapshots map[string]*ReplicaSnapshotInfo `json:"snapshots"`
+
+  	// Deprecated: Replaced by field `spec.evictionRequested`.
+  	// +optional
+  	EvictionRequested bool `json:"evictionRequested"`
+  }
+
+  type ReplicaSnapshotInfo struct {
```
DamiaSan, line 83: If this type is related to the message `ReplicaSnapshotDiskInfo`, I suggest to use the same name to better understand their relationship.

DamiaSan (commented):
This functionality has some overlaps with the [delta rebuilding](https://github.com/longhorn/longhorn/pull/7031), for example in the case when the node where the raid resides restart (we must elect an healty replica and compare all others replica with this one to ensure they are aligned). So I think we need an unique implementation for these 2 functionalities. WDYT?
cc @innobead

c3y1huang:
> This functionality has some overlaps with the [delta rebuilding](https://github.com/longhorn/longhorn/pull/7031), for example in the case when the node where the raid resides restart (we must elect an healty replica and compare all others replica with this one to ensure they are aligned). So I think we need an unique implementation for these 2 functionalities. WDYT?

Based on the Slack discussion: `There is no way to guarantee that there won't be any data block missing in the head when the volume suddenly crashes`. As @shuo-wu suggested, our best approach, for now, is to ensure that the replica candidates have the complete snapshot set. I don't see the overlap between this proposal and the future integration with the delta rebuilding. Could you elaborate on the potential overlap? IMO we can enhance the handling of the scenario when the RAID node goes offline, possibly by introducing a new field in `Replica.Status.Snapshots`. What do you think?

enhancements/20240919-auto-salvage-support-for-v2-volumes.md
c3y1huang: Thank you for the suggestion. I prefer to keep this more generic, as it's present in the Replica CR, which should allow us to add other details in the future.

However, if the team agrees to align with `ReplicaSnapshotDiskInfo`, I am open to that as well.

DamiaSan:
> > This functionality has some overlaps with the [delta rebuilding](https://github.com/longhorn/longhorn/pull/7031), for example in the case when the node where the raid resides restart (we must elect an healty replica and compare all others replica with this one to ensure they are aligned). So I think we need an unique implementation for these 2 functionalities. WDYT?
> 
> Based on the Slack discussion: `There is no way to guarantee that there won't be any data block missing in the head when the volume suddenly crashes`. As @shuo-wu suggested, our best approach, for now, is to ensure that the replica candidates have the complete snapshot set. I don't see the overlap between this proposal and the future integration with the delta rebuilding as this proposal leans more towards the control plan selection of the reusable candidate. Could you elaborate on the potential overlap?
> 
> IMO we can enhance the handling of the scenario when the RAID node goes offline with [delta rebuilding](https://github.com/longhorn/longhorn/pull/7031). An approach is to address this in the engine instance, similar to how v1 volume handles it, or incorporate additional information in `Replica.Status.Snapshots`? What do you think?
> 
> @derekbit @shuo-wu , any thoughts about this?

The overlap is the selection method and the check that replicas have the complete snapshot set.

c3y1huang:
Summary from LEP discussion:
1. Remove `Replica.Status.Snapshots` from this proposal. There shouldn't be a scenario for inconsistent snapshot sets between failed reusable replicas.
1. Filter replica candidates based on the size of the head volume.
1. Block trim operations when the volume is degraded to maintain reliable volume head size.

enhancements/20240919-auto-salvage-support-for-v2-volumes.md
c3y1huang: The Snapshots field will be removed as discussed in the LEP meeting.

enhancements/20240919-auto-salvage-support-for-v2-volumes.md
```diff
@@ -0,0 +1,95 @@
...
+      map<string, string> replica_address_map = 4;
+      string frontend = 5;
+      int32 port_count = 6;
+      bool upgrade_required = 7;
+      string initiator_address = 8;
+      string target_address = 9;
+      bool salvage_requested = 10;
+  }
+  ```
+- When `salvage_requested` is set to `True` in the requests, the SPDK server sorts the volume head sizes of the replicas. A replica mode is marked as `ERR` if its head size is not the same as the largest size in the sorted list.
```
DamiaSan, line 76: What do you mean exactly with head size? Because in SPDK we create thinly provisioned volumes, which means that no cluster is allocated for the volume at creation time. Clusters are allocated only when needed, for example when you want to write over it. So what we should check is which clusters are allocated for every head volume.

c3y1huang: Referring to the [Replica.Head](https://github.com/longhorn/longhorn-spdk-engine/blob/20be505a19b72474e50d1703b1fbda66b28e5e18/pkg/api/types.go#L21) lvol here.

c3y1huang: After discussing with @shuo-wu , the concern is that `Replica.Head` relies on the SPDK server monitoring for updates, which can lead to potential cases where `Replica.Head` contains outdated information.

After discussing with @derekbit , a possible approach is to introduce a new field, such as `requestRuntime`, in the `ReplicaGetRequest`. This field would request the server to fetch run-time information directly from the bdev lvol during a `ReplicaGet` request.

innobead (commented):
In general, LGTM.

DamiaSan (dismissed):

c3y1huang:
The CI codespell check failed to build. I'm not sure if it's a temporary thing.

enhancements/20240919-auto-salvage-support-for-v2-volumes.md
```diff
@@ -0,0 +1,122 @@
...
+
+When `EngineCreate()` is called with `salvage_requested` set to `true`, the SPDK server retrieves runtime `Replica` object using `ReplicaGet()` with `runtime_requested` set to `true`. All replicas lvol head size are then sorted. Replicas with a head size different from the largest size in the sorted list are marked as `ERR`.
+
+### Engine Creation
+
+During SPDK replica creation, if the replica already exists, the lvol bdev creation process is skipped. Instead, `Replica.construct()` is called to build the `Replica` object with existing lvol.
+
+### Snapshot Sync During Engine Creation
+
+Currently, the SPDK server updates the `Engine` object as regular intervals in the monitoring go routine. During salvage scenarios, the `Snapshot` custom resource (CR) could get deleted before the `Engine` object is updated, leading to the deletion of the actual snapshot chain data. To address this, `EngineGet()` will use `ReplicaGet()` with `runtime_requested` set to `true` to obtain runtime `Replica` objects. This information is then used to update the `Engine` CR with the latest replica details.
```
shuo-wu, line 103: > During salvage scenarios, the `Snapshot` custom resource (CR) could get deleted before the `Engine` object is updated, leading to the deletion of the actual snapshot chain data.

Can you talk a little bit more about the above? Before salvage, the engine and all replicas should be state ERROR, will there be any snapshot CR deletion? Or will longhorn manager pods send actual requests to SPDK for snapshot CR deletion of errored engines?

c3y1huang: Sorry, I cannot reproduce this issue anymore. So I've removed the runtime request from the design.

enhancements/20241028-auto-salvage-support-for-v2-volumes.md
```diff
@@ -0,0 +1,101 @@
...
+  ```
+
+## Design
+
+> **Note:**
+> The design applies only to v2 volumes. The v1 volume candidate selection remains unchange.
+
+### Failed Usable Replica Filtering
+
+When `EngineCreate()` is called with `salvage_requested` set to `true`, the SPDK server retrieves `Replica` from SPDK server cache. All replicas lvol head size are then sorted. Replicas with a head size different from the largest size in the sorted list are marked as `ERR`.
```
derekbit, line 78: > Replicas with a head size different from the largest size in the sorted list are marked as `ERR`.

Can you add purposes of `Replicas with a head size different from the largest size in the sorted list are marked as `ERR``?

c3y1huang: Sorry, I forgot to update after https://github.com/longhorn/longhorn-spdk-engine/pull/222#issuecomment-2458428694. It's now updated. Thank you!

enhancements/20241028-auto-salvage-support-for-v2-volumes.md
```diff
@@ -0,0 +1,101 @@
...
+  ```
+
+## Design
+
+> **Note:**
+> The design applies only to v2 volumes. The v1 volume candidate selection remains unchange.
+
+### Failed Usable Replica Filtering
+
+When `EngineCreate()` is called with `salvage_requested` set to `true`, the SPDK server retrieves `Replica` from SPDK server cache. All replicas lvol head size are then sorted. Replicas with a head size different from the largest size in the sorted list are excluded from the candidate list.
```
derekbit, line 78: > Replicas with a head size different from the largest size in the sorted list are excluded from the candidate list.

Can you briefly introduce why we choose the largest size here? It will help other developers understand the context.

c3y1huang: Updated. Thank you!

derekbit (approved):
LGTM. The LEP is quite straightforward.

Issue #9411, opened 2024-09-06
[TASK] Improve Longhorn project boards and WIKI page


derekbit:
## What's the task? Please describe

- Improve the boards
  - [ ] Longhorn Sprint
    - [ ] Make the sprint empty if the issues are not moved to ready for testing at end of sprint. 
  - [ ] Longhorn Community Sprint
  - [ ] Longhorn QA Sprint

- [WIKI page](https://github.com/longhorn/longhorn/wiki/Issue-Management)
  - [ ] Change the issue management page to Project Sprint Management
  - [ ] Update the WIKI page

## Additional context

cc @innobead

innobead:
cc @khushboo-rancher

PR #9510, opened 2024-09-21
chore(workflow): fix false alarm in update-projects.yaml

derekbit:

Issue Longhorn/longhorn#9411

### DIFF
--- a/.github/workflows/update-projects.yaml
+++ b/.github/workflows/update-projects.yaml
@@ -31,6 +31,7 @@ jobs:
 
     - uses: EndBug/project-fields@v2
       id: fields
+      continue-on-error: true
       with:
         operation: get
         fields: Status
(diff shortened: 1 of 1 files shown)


.github/workflows/add-to-projects.yml
```diff
@@ -43,7 +43,7 @@ jobs:
         github-token: ${{ secrets.CUSTOM_GITHUB_TOKEN }}
         item-id: ${{ steps.add-project.outputs.itemId }}
         field-keys: Status,Sprint
-        field-values: "New,[0]"
+        field-values: "In Progress,[0]"
```
innobead, line 46: Added back the missing `new` column, so we don't need this change.

innobead (changes requested):
LGTM, just one part needs to be corrected.

innobead (approved):

Issue #8757, opened 2024-06-14
[IMPROVEMENT] Collect and display disk space usage for the backing images


derekbit:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

Longhorn currently does not collect or display disk space usage for backing images. To help users better manage host storage, we need to improve this functionality as the volumes and replicas.

## Additional context

- UI ticket: https://github.com/longhorn/longhorn/issues/9325

ChanYiLin:
## Test Steps

- Create a sparse file `test` (`$ truncate -s 512M backing.img`)
- Create a BackingImage `test` with this sparse file
- Create a BackingImage with a ubuntu image
- Check the node disk status (`$ kubectl get lhn -n longhorn-system`)
    - a new field `scheduledBackingImage` record the backing image on the disk and its real size
        - `test` real size is smaller than 512M
    - `storageScheduled` includes the size of BackingImage
```
        scheduledBackingImage:
          test: 53248
          ubuntu: 3860852736
        scheduledReplica: {}
        storageAvailable: 32820428800
        storageMaximum: 66257006592
        storageScheduled: 3860905984
```

roger-ryao:
**Verified on master-head / v1.8.0-dev-20241020 20231021**

- longhorn v1.8.0-dev-20241020 https://github.com/longhorn/longhorn/commit/c0235113894a5baa6dbcf871caa9d54f1ea5d311

**The test steps**
https://github.com/longhorn/longhorn/issues/8757#issuecomment-2311897006

**Result**
    - [x] a new field `scheduledBackingImage` record the backing image on the disk and its real size
            - `test` real size is smaller than 512M
    - [x] `storageScheduled` includes the size of BackingImage

**Output**
```bash
> kl get lhn -o yaml |grep -H2 scheduledBackingImage
(standard input)-        filesystemType: ext2/ext3
(standard input)-        instanceManagerName: instance-manager-4ace7950fb5c5e9aa349ca00cd158a10
(standard input):        scheduledBackingImage:
(standard input)-          "512": 46559232
(standard input)-        scheduledReplica: {}
--
(standard input)-        filesystemType: ext2/ext3
(standard input)-        instanceManagerName: instance-manager-da3f44ea0ac6694896cf9be03125f04f
(standard input):        scheduledBackingImage:
(standard input)-          "512": 46559232
(standard input)-        scheduledReplica: {}
--
(standard input)-        filesystemType: ext2/ext3
[...]

PR #9582, opened 2024-10-07
feat(crd): add realSize field to backingimage and scheduledBackingImage field to node

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/8757

- store realsize of backingimage in status
- add the realsize to storageScheduled in diskStatus
- add new field scheduledBackingImage in diskStatus

### DIFF
--- a/chart/templates/crds.yaml
+++ b/chart/templates/crds.yaml
@@ -342,6 +342,9 @@ spec:
                       type: string
                     progress:
                       type: integer
+                    realSize:
+                      format: int64
+                      type: integer
                     senderManagerAddress:
                       type: string
                     sendingReference:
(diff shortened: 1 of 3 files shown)


chart/templates/crds.yaml
```diff
@@ -559,6 +562,11 @@ spec:
                 type: object
               ownerID:
                 type: string
+              realSize:
+                description: Real size of image, which may be smaller than the size
```
innobead, line 566: Can we add the size unit to the description?

ChanYiLin: Fixed. thanks

innobead (approved):

PR #9583, opened 2024-10-07
chore: set data engine explicitly for the default storageclass

shuo-wu:

https://github.com/longhorn/longhorn/issues/9584

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -172,6 +172,8 @@ persistence:
     selector: ""
   # -- Setting that allows you to enable automatic snapshot removal during filesystem trim for a Longhorn StorageClass. (Options: "ignored", "enabled", "disabled")
   removeSnapshotsDuringFilesystemTrim: ignored
+  # -- Setting that allows you to specify the data engine version for the default Longhorn StorageClass. (Options: "v1", "v2")
+  dataEngine: v1
 
 preUpgradeChecker:
   # -- Setting that allows Longhorn to perform pre-upgrade checks. Disable this setting when installing Longhorn using Argo CD or other GitOps solutions.
--- a/chart/templates/storageclass.yaml
+++ b/chart/templates/storageclass.yaml
@@ -53,4 +53,5 @@ data:
       {{- end }}
       {{- if .Values.persistence.disableRevisionCounter }}
       disableRevisionCounter: "{{ .Values.persistence.disableRevisionCounter }}"
+      dataEngine: "{{ .Values.persistence.dataEngine }}"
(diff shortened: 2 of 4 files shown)


innobead:
> #### Which issue(s) this PR fixes:

@shuo-wu let's make sure an issue for this PR, so QA can help validate it.

shuo-wu:
The ticket is created with a manual test attached.

derekbit:
@shuo-wu 
Can you execute [generate-longhorn-yaml.sh](https://github.com/longhorn/longhorn/blob/master/scripts/generate-longhorn-yaml.sh)?

shuo-wu:
@derekbit 

Done. Thank you for the reminder

chart/values.yaml
```diff
@@ -172,6 +172,8 @@ persistence:
     selector: ""
   # -- Setting that allows you to enable automatic snapshot removal during filesystem trim for a Longhorn StorageClass. (Options: "ignored", "enabled", "disabled")
   removeSnapshotsDuringFilesystemTrim: ignored
+  # -- Setting that allows you to specify the data engine version for the default Longhorn StorageClass. (Options: "v1", "v2", "all")
```
innobead, line 175: Just noticed we have `all` for data engine. What does that mean?

shuo-wu: Just re-check the code. Option `all` is more like a special flag that [explicitly makes Longhorn list all 2 versions of instance managers/volumes](https://github.com/longhorn/longhorn-manager/blob/56903baa3890d0cac86d5a477f4863271b9043f9/types/types.go#L481-L483). But it seems that using an empty `dataEngine` could achieve the same result.

I am thinking if we should forbid users to use it.

shuo-wu: OK. In Volume/Engine/Replica CRD, it allows `v1` and `v2` only. Let me remove option `all` for storageclass

innobead: > Just re-check the code. Option `all` is more like a special flag that [explicitly makes Longhorn list all 2 versions of instance managers/volumes](https://github.com/longhorn/longhorn-manager/blob/56903baa3890d0cac86d5a477f4863271b9043f9/types/types.go#L481-L483). But it seems that using an empty `dataEngine` could achieve the same result.
> 
> I am thinking if we should forbid users to use it.

Yeah. I remembered that context discussed with @derekbit before. Just need a doc for that intended internal usage.

innobead (approved):

shuo-wu:
@mergify backport to v1.7.x

PR #9587, opened 2024-10-07
fix(cve): update CSI images for fixing CVE-2024-24790

derekbit:

Issue Longhorn/longhorn#9563

### DIFF
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -35,7 +35,7 @@ questions:
     label: Longhorn Engine Image Tag
     group: "Longhorn Images Settings"
   - variable: image.longhorn.ui.repository
-    default:  longhornio/longhorn-ui
+    default: longhornio/longhorn-ui
     description: "Repository for the Longhorn UI image."
     type: string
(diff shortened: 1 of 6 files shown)


chart/questions.yaml
```diff
@@ -611,9 +599,7 @@ questions:
   default: false
 - variable: defaultSettings.storageNetwork
   label: Storage Network
-  description: "Longhorn uses the storage network for in-cluster data traffic. Leave this blank to use the Kubernetes cluster network.
-  WARNING:
```
innobead, line 615: Do we use any different formatting by IDE?

innobead (commented):
LGTM, just one comment about the file format for question yaml.

chart/questions.yaml
innobead: Will this become repeated changes as contributors might use different formats?

Would this format affect presentation issues in rancher UI?

innobead: Sounds good

innobead (approved):

PR #9615, opened 2024-10-11
chore(scripts): add a script for updating chart/README.md

derekbit:

Issue #


shuo-wu (approved):
In general LGTM

innobead (approved):

innobead (commented):
After merging this PR, just noticed that as this is for the head branch, can we make the image tags for our components as branch-based?

For example, for 1.7.x, longhorn-manager should be `longhorn-manager:v1.7.x-head`? Also, we need to check if this change might affect our QA regression testing. 

cc @derekbit @longhorn/qa

derekbit:
> After merging this PR, just noticed that as this is for the head branch, can we make the image tags for our components as branch-based?
> 
> For example, for 1.7.x, longhorn-manager should be `longhorn-manager:v1.7.x-head`? Also, we need to check if this change might affect our QA regression testing.
> 
> cc @derekbit @longhorn/qa

The script only updates the chart/README.md. I think we should only execute this script in the CI flow to make the longhorn-images.txt and README.md consistent. The update won't affect the QA regression testing.

innobead:
yeah, true. only readme.

PR #9678, opened 2024-10-18
chore(template): update release.md

derekbit:

Issue #

### DIFF
--- a/.github/ISSUE_TEMPLATE/release.md
+++ b/.github/ISSUE_TEMPLATE/release.md
@@ -29,12 +29,12 @@ Action items for releasing v<x.y.z>
   - longhorn-share-manager
   - backing-image-manager
   - longhorn-spdk-engine (needed after GA)
-- [ ] Update `Best Practices>Operating System` in Longhorn official document.
 - [ ] Update image versions in https://github.com/longhorn/longhorn/tree/v<x.y>.x/chart/README.md.
-- [ ] Trigger the GA release build by [longhorn/longhorn Actions](https://github.com/longhorn/longhorn)
+- [ ] Trigger the GA release build by [longhorn/release Actions](https://github.com/longhorn/release/actions/workflows/release.yml)
 
 **The QA captain needs to coordinate the following items before the GA release.**
 
+- [ ] Update `Best Practices>Operating System` in Longhorn official document.
 - [ ] Regression test plan (manual)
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/release.md
```diff
@@ -49,17 +48,21 @@ Action items for releasing v<x.y.z>
 **The Release Captain needs to finish the following items.**
 
 - [ ] Release note - Release Captain
+  - [ ] Update `Best Practices>Operating System` in Longhorn official document.
```
innobead, line 51: Should it be updated earlier, right after QA testing, instead of waiting until the release is wrapped up?

innobead (approved):
LGTM

.github/ISSUE_TEMPLATE/release.md
derekbit: In practice, we prepare the release note before QA finishes testing, so we can update it while updating the new version of doc. Any concerns?

innobead: I assumed below instead?

> we prepare the release note after QA finishes testing

It's okay if we can make sure version updates won't be missed. Ideally, QA should have an issue to verify OS version bumps for each feature release. They will finish the test and update the [best practices] doc accordingly by them.

Probably we can just say that checking OS versions bump in the best practices doc only for feature release. (However, we have a version policy indicating that we will still update SLES or Micro versions in patches as well)

derekbit: > They will finish the test and update the [best practices] doc accordingly by them. I thought this is what we do currently.

Got it. It sounds like we are not on the same page. I thought the **[best practices] doc** is updated by the release captain rather than QA captain. Let's move back it the Pre-Release stage and explicitly assign it QA captain.

Thank you.

innobead (approved):

.github/ISSUE_TEMPLATE/release.md
derekbit: > You just reminded me that probably we need to have an issue template for regular tasks (OS versions, K8s libs, CSI, [Wiki: Version Update Policy](https://github.com/longhorn/longhorn/wiki/Version-Update-Policy)) for each feature release. WDYT?

Sounds good. Will draft ones for the tasks.

Issue #9736, opened 2024-10-31
[FEATURE]: Automatic online filesystem resize for RWX volumes.


PhanLe1010:
## Describe the bug

Fail to resize RWX PVC at filesystem resizing step 

## To Reproduce

1. Create a RWX PVC
    ```
    apiVersion: v1
    kind: PersistentVolumeClaim
    metadata:
      name: rwx-test
      namespace: default
    spec:
      accessModes:
        - ReadWriteMany
      storageClassName: longhorn
      resources:
        requests:
          storage: 10Gi
    ```
1. Create `pod-1` running on `node-1` and use the `rwx-test` PVC
    ```
    apiVersion: v1
    kind: Pod
    metadata:
      name: pod-1
      namespace: default
    spec:
      nodeName: node-1 # Replace by your node name
      restartPolicy: Always
      containers:
        - name: volume-test
          image: nginx:stable-alpine
          imagePullPolicy: IfNotPresent
          livenessProbe:
            exec:
              command:
                - ls
                - /data/lost+found
            initialDelaySeconds: 5
            periodSeconds: 5
          volumeMounts:
            - name: volv
              mountPath: /data
          ports:
            - containerPort: 80
      volumes:
        - name: volv
          persistentVolumeClaim:
            claimName: rwx-test
    ```
1. Share manager pod and the volume will be attached to `node-1` as `node-1` has the only workload pod so Longhorn prefer `node-1` at this point
1. Create `pod-2` running on `node-2` and use the `rwx-test` PVC
    ```
    apiVersion: v1
    kind: Pod
    metadata:
      name: pod-2
      namespace: default
    spec:
      nodeName: node-2 # Replace by your node name
      restartPolicy: Always
      containers:
        - name: volume-test
          image: nginx:stable-alpine
          imagePullPolicy: IfNotPresent
          livenessProbe:
            exec:
              command:
                - ls
                - /data/lost+found
            initialDelaySeconds: 5
            periodSeconds: 5
          volumeMounts:
            - name: volv
              mountPath: /data
          ports:
            - containerPort: 80
      volumes:
        - name: volv
          persistentVolumeClaim:
            claimName: rwx-test
    ```
1. Delete `pod-1`
1. Note that share manager pod and the volume are still be attached to `node-1`. They cannot move
1. Edit `rwx-test` PVC to have 15Gi size 
[...]

PR #9758, opened 2024-11-06
docs(lep): rwx resize improvement

james-munson:

Issue https://github.com/longhorn/longhorn/issues/9736

LEP for forwarding RWX resize requests to be handled on the node with the share manager on it.

### DIFF
--- a/enhancements/20241111-rwx-resize.md
+++ b/enhancements/20241111-rwx-resize.md
@@ -1,0 +1,118 @@
+# Title
+
+RWX Resize
+
+## Summary
+
+Expansion of volumes requires resizing the filesystem inside the block device /dev/longhorn/\<volume\> after the block device is grown.  In the case of an RWX volume, that is on the node where the share-manager pod has it mounted.  But the CSI layer sends the `NodeExpandVolume` RPC request to the CSI plugin on the node(s) where the volume is attached, which might not include the mount node.  If it doesn't, the resize does not take effect.
+
+This enhancement is meant to ensure that the CSI request or some equivalent is forwarded to the share-manager node.
+
+### Related Issues
+
+[BUG] RWX expansion fails. Fail to resize RWX PVC at filesystem resizing step [#9736](https://github.com/longhorn/longhorn/issues/9736)
+
+## Motivation
+
+### Goals
+
+Resize of RWX volume should succeed regardless of location of share-manager and workload pods.
+
+This should also take care of the manual resize step documented in https://longhorn.io/docs/1.7.2/nodes-and-volumes/volumes/expansion/#rwx-volume, and allow fully automated expansion of RWX volumes without scale-down, resolving `[FEATURE] Automatic online RWX volume expansion` [#8118](https://github.com/longhorn/longhorn/issues/8118)
+
+### Non-goals [optional]
+
+## Proposal
+
+A csi-plugin can detect that the device path does not exist locally.  But it doesn't really have to, because it can tell that the volume is RWX.  It can delegate the resize to the owning share manager in all cases, even on the local node.
+
+Implement a "FilesystemResize" RPC call in the share-manager, much like its "FilesystemTrim", and let the csi-plugin refer the resize action to it.
+
+### User Stories
+
+There's no extra work for the user.  In fact, this should make an existing feature easier to use.  However, the user must have restarted the share-manager pod after upgrade to the release with this feature, to get an image that supports the RPC call.
+
+### API changes
+
+- Add a FilesystemResize method to share-manager RPC protocol
+
+### Alternatives
+
+It might be tempting to find a way to forward the request to the longhorn-csi-plugin on the correct node.
+
+If we handle the expansion in share manager we must:
+ 1. re-implement the logic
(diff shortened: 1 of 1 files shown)


enhancements/20241111-rwx-resize.md
```diff
@@ -0,0 +1,102 @@
+# Title
+
+RWX Resize
+
+## Summary
+
+Expansion of volumes requires resizing the filesystem inside the block device /dev/longhorn/\<volume\> after the block device is grown.  In the case of an RWX volume, that is on the node where the share-manager pod has it mounted.  But the CSI layer sends the `NodeExpandVolume` RPC request to the CSI plugin on the node(s) where the volume is attached, which might not include the mount node.  If it doesn't, the resize does not take effect.
```
PhanLe1010, line 7: WDYT about this?
```suggestion
Expansion of volumes requires resizing the filesystem inside the block device /dev/longhorn/\<volume\> after the block device is grown.  In the case of an RWX volume, that is on the node where the share-manager pod has it attached.  But the CSI layer sends the `NodeExpandVolume` RPC request to the CSI plugin on the node(s) where the volume is mounted, which might not include the share-manager pod's node.  If it doesn't, the resize does not take effect.
```

enhancements/20241111-rwx-resize.md
```diff
@@ -0,0 +1,102 @@
...
+Unfortunately, that won't work.  The csi-plugin has no idea of the actual filesystem type.  The fields of the `NodeExpandVolume` request are derived from the mount it knows about.  On the csi-plugin pod, `mount -l` shows
+```
+10.43.248.207:/pvc-8d1d3f46-76b6-484e-9a3a-93cc821ffd5b on /var/lib/kubelet/plugins/kubernetes.io/csi/driver.longhorn.io/46e93d5a9308ccb4b10edb9f7d2e362f5e6f45f125927196166405a669c36d2b/globalmount type nfs4 (rw,relatime,vers=4.1,rsize=1048576,wsize=1048576,namlen=255,softerr,softreval,noresvport,proto=tcp,timeo=600,retrans=5,sec=sys,clientaddr=24.144.89.5,local_lock=none,addr=10.43.248.207)
+10.43.248.207:/pvc-8d1d3f46-76b6-484e-9a3a-93cc821ffd5b on /var/lib/kubelet/pods/44f0a9cc-6995-4fa9-a914-ca0748793315/volumes/kubernetes.io~csi/pvc-8d1d3f46-76b6-484e-9a3a-93cc821ffd5b/mount type nfs4 (rw,relatime,vers=4.1,rsize=1048576,wsize=1048576,namlen=255,softerr,softreval,noresvport,proto=tcp,timeo=600,retrans=5,sec=sys,clientaddr=24.144.89.5,local_lock=none,addr=10.43.248.207)
+```
+
+If asked to do a `xfs_growfs` on `/var/lib/kubelet/...` it will fail with 
+```
+could not find block size of device /var/lib/kubelet/plugins...
+```
```
PhanLe1010, line 62: Could you elaborate more about why the csi-plugin need to find the mount point of the volume? 
I am thinking it just need to know the longhorn block device at /dev/longhorn/volume name and run `resize2fs` against it. Futher more, csi-plugin can figure out which node has the block device by looking at the volume/engine resource.

PhanLe1010 (commented):
In general, LGTM

enhancements/20241111-rwx-resize.md
james-munson: We don't run `resize2fs` directly, we prefer to call `ResizeFs.Resize` which takes both the device path and the mount path:  https://github.com/kubernetes/mount-utils/blob/e448c96afa03f6e2556117ed7ce7dee0462fa3ca/resizefs_linux.go#L46-L68

Specifically, that's because it knows that while `resize2fs` can operate on either one, `xfs_resizefs` requires the mount path:
```go
	klog.V(3).Infof("ResizeFS.Resize - Expanding mounted volume %s", devicePath)
	switch format {
	case "ext3", "ext4":
		return resizefs.extResize(devicePath)
	case "xfs":
		return resizefs.xfsResize(deviceMountPath)
	case "btrfs":
		return resizefs.btrfsResize(deviceMountPath)
	}
	return false, fmt.Errorf("ResizeFS.Resize - resize of format %s is not supported for device %s mounted at %s", format, devicePath, deviceMountPath)"
```

james-munson:
```
xfs_growfs /dev/longhorn/pvc-8d1d3f46-76b6-484e-9a3a-93cc821ffd5b 
xfs_growfs: /dev/longhorn/pvc-8d1d3f46-76b6-484e-9a3a-93cc821ffd5b is not a mounted XFS filesystem
```

PhanLe1010: Got it. Thank you for the explaination

PhanLe1010 (dismissed):
LGTM

enhancements/20241111-rwx-resize.md
james-munson: I put this into the LEP just to document it.

PhanLe1010 (approved):

PR #9807, opened 2024-11-14
docs(lep): add "V2 Data Engine Live Upgrade"

derekbit:

Issue #


derekbit:
@yangchiu 
Found an error of linking backport issue
image

enhancements/20241114-v2-data-engine-live-upgrade.md
```diff
@@ -0,0 +1,232 @@
...
+
+2. According to `DataEngineUpgradeManager.Status.State`, `DataEngineUpgradeManager Controller` does
+    - `undefined`
+        1.  Update `status.instanceManagerImage` from `defaultInstanceManagerImage` setting
+		2.  Check whether the default instance manager image has be pulled on each Longhorn node. If not, update `status.State` to `error` and `status.ErrorMessage`
+        3.  Update `status.state` to `initializing`
+
+	- `initializing`  
+        4. Update `status.upgradeNodes`  
+           - If `spec.nodes` is empty, list all nodes and add them to `status.upgradeNodes`
```
PhanLe1010, line 112: Is `DataEngineUpgradeManager` controller modifying its CR spec? If yes, might be violating K8s design pattern?

enhancements/20241114-v2-data-engine-live-upgrade.md
```diff
@@ -0,0 +1,232 @@
...
+    State UpgradeState
+    ErrorMessage string
+```
+
+## Controllers
+
+### DataEngineUpgradeManager Controller
+
+- Create `DataEngineUpgradeManager` CR for each node upgrade
+- Monitor the `DataEngineUpgradeManager` CRs
```
PhanLe1010, line 88: 1 `DataEngineUpgradeManager` CR for each node or for the whole cluster?

enhancements/20241114-v2-data-engine-live-upgrade.md
```diff
@@ -0,0 +1,232 @@
...
+        4. Update `status.upgradeNodes`  
+           - If `spec.nodes` is empty, list all nodes and add them to `status.upgradeNodes`  
+           - If `spec.nodes` is not empty, list the nodes in the `spec.nodes` and add them to `status.upgradeNodes`  
+        5. Then, update the `status.currentState` to `upgrading`
+
+    -  `upgrading`  
+        6. Iterate `status.upgradeNodes`  
+        7. Create `upgradeNode` resource for next node if there is no active `nodeUpgrade` resource  
+        8. Set the next node to `status.upgradingNode`  
+        9. Add the finished node to `status.upgradedNodes`
```
PhanLe1010, line 120: NIT: looks like `status.upgradedNodes` is not defined in `DataEngineUpgradeManager`

enhancements/20241114-v2-data-engine-live-upgrade.md
```diff
@@ -0,0 +1,232 @@
...
+
+3.  According to `NodeDataEngineUpgrade.status.state`, `NodeDataEngineUpgrade controller` does
+	  
+    - `undefined`
+		1. Update `status.currentState` to `initializing`
+
+    - `initializing`
+		1. Set upgrading node to **unscheduable** by setting `node.spec.upgradeRequested` to `true`
+		2. List volumes on upgrading node
+		3. Snapshot all volumes
```
PhanLe1010, line 134: Could you elaborate more about why we need to take snapshot of all volumes?

enhancements/20241114-v2-data-engine-live-upgrade.md
derekbit: Typo. It is `status.upgradeNodes`.

enhancements/20241114-v2-data-engine-live-upgrade.md
derekbit: It is for the whole cluster. Fixed the mistake. Thank you.

enhancements/20241114-v2-data-engine-live-upgrade.md
derekbit: No, the controller reads the `spec.nodes` and determines and sets the `status.upgradeNodes`. It won't update spec fields.

enhancements/20241114-v2-data-engine-live-upgrade.md
derekbit: The purpose is to accelerate the replica rebuild. We can skip the snapshot lvol. WDYT?

PhanLe1010: Snapshot without checksum might not accelerate the replica rebuild? 
If we wait for snapshot to have checksum, it will turn out to be same time as rebuilding them directly?

enhancements/20241114-v2-data-engine-live-upgrade.md
PhanLe1010: Who is setting `spec.nodes`?

enhancements/20241114-v2-data-engine-live-upgrade.md
derekbit: Yes, I'm thinking if we should wait for the checksum. Can we skip the snapshot blindly?

enhancements/20241114-v2-data-engine-live-upgrade.md
```diff
@@ -0,0 +1,320 @@
...
+    
+    - `undefined`
+        1.  Update `status.instanceManagerImage` from `defaultInstanceManagerImage` setting
+        2.  Check whether the default instance manager image has be pulled on each Longhorn node. If not, update `status.State` to `error` and `status.ErrorMessage`
+        3.  Update `status.state` to `initializing`
+
+    - `initializing`  
+        1. Update `status.upgradeNodes`  
+          - If `spec.nodes` is empty, list all nodes and add them to `status.upgradeNodes`  
+          - If `spec.nodes` is not empty, list the nodes in the `spec.nodes` and add them to `status.upgradeNodes`
```
innobead, line 200: Does this mean the unknown nodes will be silently filtered? Throwing errors is recommended, as it informs users that the nodes should match those in the cluster.

enhancements/20241114-v2-data-engine-live-upgrade.md
```diff
@@ -0,0 +1,320 @@
...
+        3.  Update `status.state` to `initializing`
+
+    - `initializing`  
+        1. Update `status.upgradeNodes`  
+          - If `spec.nodes` is empty, list all nodes and add them to `status.upgradeNodes`  
+          - If `spec.nodes` is not empty, list the nodes in the `spec.nodes` and add them to `status.upgradeNodes`  
+        2. Then, update the `status.currentState` to `upgrading`
+        
+    - `upgrading`  
+        1. Iterate `status.upgradeNodes`
```
innobead, line 204: "Should `spec.nodes` be immutable for the node once the node upgrade has started, or should we allow mutations in case we support upgrade cancellation? It’s suggested to keep the scenario simple initially."

enhancements/20241114-v2-data-engine-live-upgrade.md
```diff
@@ -0,0 +1,320 @@
...
+        2.  Waiting for the replicas become `running`
+
+	- `finalizing`
+		1. Update the `volume.spec.image` of detached volumes to default instance manager image
+ 
+    - `completed`
+
+## Test Plan
+
+1. Single-node Longhorn cluster is not allowed to performance v2 data engine live upgrade.
```
innobead, line 253: An offline upgrade still needs to be supported. For example, users should be able to trigger the upgrade as usual, but it will only take effect after all volumes are detached, and when any volume is reattached.

enhancements/20241114-v2-data-engine-live-upgrade.md
```diff
@@ -0,0 +1,320 @@
...
+        
+    - `upgrading`  
+        1. Iterate `status.upgradeNodes`  
+        2. Find a node that is not upgraded and create a `nodeDataEngineUpgrade` resource
+        3. Set the node to `status.upgradingNode`
+        4. Update the `status.currentState` to `completed` if all nodes are finished
+        
+    - `completed`
+        
+3.  According to `NodeDataEngineUpgrade.status.state`, `NodeDataEngineUpgrade controller` does
```
innobead, line 211: Since the upgrade is node-based, the process for recovering from an error during the status transition (before the node upgrade reaches 'completed') needs to be clarified.

innobead (commented):
In general, LGTM for this initial version, but I believe some additional details need to be discovered or clarified eventually. 

There are also some minor feedback, mostly related to naming and clarifying the behavior.

enhancements/20241114-v2-data-engine-live-upgrade.md
derekbit: Yes, offline upgrade still works and only `v2 data engine live upgrade` is not allowed.

enhancements/20241114-v2-data-engine-live-upgrade.md
derekbit: `spec.nodes` is immutable. `upgrade cancellation` is not explicitly supported. For now, if an upgrade is stuck, user can directly delete the `dataEngineUpgradeManager` resource if it really needs to be caneclled.

enhancements/20241114-v2-data-engine-live-upgrade.md
derekbit: What does `unknown nodes` mean?

enhancements/20241114-v2-data-engine-live-upgrade.md
```diff
@@ -0,0 +1,320 @@
...
+### CRD Changes
+
+- node.longhorn.io
+
+```
+spec:
+	...
+	// Request to upgrade the instance manager for v2 volumes on the node.
+	// +optional
+	DataEngineUpgradeRequested bool `json:"dataEngineUpgradeRequested"`
```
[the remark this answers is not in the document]
derekbit: upgrade manager controller will watch the `DataEngineUpgrade` and set node's `spec. DataEngineUpgradeRequested` when the node is being lively upgraded, so user doesn't need to set the field manually. The field will be set to `false` after the live upgrade.

enhancements/20241114-v2-data-engine-live-upgrade.md
```diff
@@ -0,0 +1,320 @@
...
+
+1.  Upgrade the Longhorn system by helm upgrade, manifest or other method.
+    
+2.  After upgrading, an instance-manager resource with the default instance manager image is created on each node but remains in
+    
+
+- `stopped`: if the running instance-manager pod using an old instance manager image still contains running replicas or engines.
+- `running`: if the old instance-manager is deleted because it does not contains running replicas and engines.
+
+3.  Create a `DataEngineUpgradeManager` resource
```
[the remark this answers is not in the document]
derekbit: No, offline upgrade is still supported.
Even though there is no `DataEngineUpgradeManager` CR, an instance manager pod with default instance manager image will be created and running when the old instance manager pod doesn't have running instances.
Then, after creating a `DataEngineUpgradeManager` CR, the upgrade of the node will be skipped.

enhancements/20241114-v2-data-engine-live-upgrade.md
```diff
@@ -0,0 +1,320 @@
...
+    ```
+    apiVersion: longhorn.io/v1beta2
+    kind: DataEngineUpgradeManager
+    metadata:
+      name: upgrade
+      namespace: longhorn-system
+    spec:
+      dataEngine: v2
```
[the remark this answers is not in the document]
derekbit: I'm not sure if it can be extended to v1 data engine, so I prefer a general naming. WDYT?

enhancements/20241114-v2-data-engine-live-upgrade.md
derekbit: If a volume on the upgrading node is somehow stuck, the `nodeDataEngineUpgrade.status.state` will be stuck in `upgrading` state and won't be in an `error` state, so user needs to resolve the issue, and the node upgrade can continue. Currently, the error message will be set in the `status.Message`, so the status is

```
status:
  state: upgrade
  message: an error is found.....
```

If `nodeDataEngineUpgrade.status.state` is in `error` state, it won't be recovered from the error. User needs to recreate a `DataEngineUpgradeManager` for a new upgrade.

enhancements/20241114-v2-data-engine-live-upgrade.md
```diff
@@ -0,0 +1,320 @@
...
+
+![Architecture Diagram](image/v2-data-engine-live-upgrade/arch-diagram.svg)
+
+1.  Create a `DataEngineUpgradeManager` resource
+    
+2.  According to `DataEngineUpgradeManager.Status.State`, `DataEngineUpgradeManager Controller` does
+    
+    - `undefined`
+        1.  Update `status.instanceManagerImage` from `defaultInstanceManagerImage` setting
+        2.  Check whether the default instance manager image has be pulled on each Longhorn node. If not, update `status.State` to `error` and `status.ErrorMessage`
```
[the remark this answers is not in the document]
derekbit: We can change the status to 
```
status:
  state:
  conditions:
    - <condition name>:
      status:
      reason:
      message:
```

However, I tried to use condition and could not find a better name for the condition. Any suggestion? In addition, for the upgrade status, `condition` seems redundant because the message is already set to `status.Message`. That's why I don't use condition here.

enhancements/20241114-v2-data-engine-live-upgrade.md
```diff
@@ -0,0 +1,320 @@
...
+```
+
+- volume.longhorn.io
+
+```
+spec:
+	...
+	// TargetNodeID is specifies the node where the volume's initiator instance should be scheduled during data engine live upgrade.
+	// +optional
+	TargetNodeID string `json:"targetNodeID"`
```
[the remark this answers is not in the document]
derekbit: Great idea. Thanks.

enhancements/20241114-v2-data-engine-live-upgrade.md
```diff
@@ -0,0 +1,320 @@
...
+	InstanceStateSuspended = InstanceState("suspended")
+	...
+)
+```
+
+
+
+### Webhook Validation
+
+1.  When the v2 data engine of a node is being upgraded (i.e., `node.spec.dataEngineUpgradeRequested` is `true`), the volume attachment will be rejected, but detachment is allowed.
```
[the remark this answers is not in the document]
derekbit: > Images being created by users should be prevented with best efforts if the name does not follow the aforementioned naming convention.

Sorry, I don't get it. Could you give me an example?

enhancements/20241114-v2-data-engine-live-upgrade.md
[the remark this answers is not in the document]
innobead: Okay, so it's opposite. Longhorn updates the field instead of users. What users should do is to create `DataEngineUpgrade` CR to trigger the upgrade.

So, how can we prevent users from updating this field? Ideally, users should not update the field.

enhancements/20241114-v2-data-engine-live-upgrade.md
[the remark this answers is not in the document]
innobead: I messed up the review of the backing image with this PR. What I meant to say is that `DataEngineUpgrade.spec.name` should follow an expected naming convention, so it should probably be automatically generated. This way, users don’t have to worry about naming the CR.

enhancements/20241114-v2-data-engine-live-upgrade.md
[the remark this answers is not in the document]
innobead: Conditions can be used to represent mandatory requirements for the upgrade, rather than representing upgrade status.

How about using `DefaultImageOnAllNodesReady` as the condition for indicating the readiness of the default image?

enhancements/20241114-v2-data-engine-live-upgrade.md
innobead: What I’m asking is that users might add nodes to `spec.nodes` that don’t actually exist in the cluster. In this case, the upgrade should not be started and allowed.

innobead: However, you mentioned in another comment, `spec.nodes` is immutable, so we should decide the content of it. Then, why do need these two `if` conditions?

enhancements/20241114-v2-data-engine-live-upgrade.md
innobead: Then, this flow should be explained, as this is a valid case for us.

enhancements/20241114-v2-data-engine-live-upgrade.md
innobead: Sounds good. Let's also briefly explain this flow.

enhancements/20241114-v2-data-engine-live-upgrade.md
derekbit: > Then, why do need these two if conditions?

Ideally, we should upgrade all nodes cluster, but the design allows users to upgrade some nodes in the cluster according to users' upgrade plan.

innobead: Then, we might need to ensure the following items.

- All nodes should be eventually upgraded to the same IM version before the next upgrade to prevent IM version drift. 
- If some nodes added by users in `spec.nodes` are not recognized, the upgrade should be avoided until `spec.nodes` is corrected by the users.

derekbit: > All nodes should be eventually upgraded to the same IM version before the next upgrade to prevent IM version drift.

We need to add the pre-upgrade check.

enhancements/20241114-v2-data-engine-live-upgrade.md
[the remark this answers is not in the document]
derekbit: Update the implementation. The field is removed.

enhancements/20241114-v2-data-engine-live-upgrade.md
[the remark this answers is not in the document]
derekbit: > DataEngineUpgrade.spec.name should follow an expected naming convention, so it should probably be automatically generated

If users trigger the upgrade, UI can help generate a name, but how to achieve it via kubectl?

derekbit:
Close the PoC PR. @PhanLe1010 will address the live upgrade.

Issue #9114, opened 2024-07-29
[CI] Move reusabled github actions to longhorn/gh-configs repo


innobead:
## What's the CI task to develop? Please describe

To reduce duplicated GitHub action or app config manifests to create in each repo, they can be moved to https://github.com/longhorn/gh-configs and then reused in each repo by using https://docs.github.com/en/actions/using-workflows/reusing-workflows.

d-cryptic:
Hi, I want to contribute on this issue. Can you assign this to me please?

derekbit:
@d-cryptic Sure! Thank you.

PR #9812, opened 2024-11-14
chore(ci) - called add-to-projects workflow and referenced here

d-cryptic:
Issue #9114 

This PR contains the workflow files which calls the workflows from gh-config repo

NA

NA

### DIFF
--- a/.github/workflows/add-to-projects.yml
+++ b/.github/workflows/add-to-projects.yml
@@ -5,5 +5,5 @@ on:
 
 jobs:
   call-add-to-projects-workflow:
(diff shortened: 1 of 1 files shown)


innobead:
Thanks for the contribution, but this action is more specific for longhorn, so we don't need to move this to the common repo. 

The original goal is to reduce duplicated configurations among repos including GitHub action or other bot configurations like stale, mergify, renovate, etc.

derekbit:
@d-cryptic 
After rethinking the PR, the workflow in this PR is not reused by other repos, so we don't need to move it the longhorn/gh-configs.

Issue #8425, opened 2024-04-24
[TEST] [ROBOT] implement manual test case `The node the DR volume attached to is rebooted`


roger-ryao:
## What's the test to develop? Please describe

Implement manual test case [The node the DR volume attached to is rebooted](https://longhorn.github.io/longhorn-tests/manual/pre-release/node-not-ready/node-restart/dr-volume-node-rebooted/) in robot

## Describe the tasks for the test

Implement manual test case `The node the DR volume attached to is rebooted` in robot

roger-ryao:
When executing the test case, I discovered the issue  https://github.com/longhorn/longhorn/issues/9752

roger-ryao:
According to https://github.com/longhorn/longhorn/issues/9752#issuecomment-2453923408, I was not successful in changing the NFS backup server on the control node through the file at https://raw.githubusercontent.com/longhorn/longhorn/master/deploy/backupstores/nfs-backupstore.yaml.

I need to spend some time today looking into this issue.

roger-ryao:
To successfully execute this case, there are two points that need to be addressed:
1. `CoreDNS` must be scaled up to 3 replicas
    - https://github.com/longhorn/longhorn/issues/9870
2. `NFS`/`MinIO` must be deployed to the control node
    - https://github.com/longhorn/longhorn/pull/9881

Issue #9752, opened 2024-11-04
[BUG] DR volume fails to reattach and faulted after node stop and start during incremental restore


roger-ryao:
## Describe the bug
When I was working on the [[TEST] [ROBOT] implementation of the manual test case for 'The node the DR volume is attached to is rebooted' (#8425)](https://github.com/longhorn/longhorn/issues/8425), I discovered the issue. When the node with the attached DR volume is stopped during the incremental restore, we expect the DR volume to reattach to another available node. However, the DR volume goes into a `Faulted` state. Even after restarting the previously stopped node, the DR volume remains in a `Faulted` state and does not recover as expected.

## To Reproduce

https://longhorn.github.io/longhorn-tests/manual/pre-release/node-not-ready/node-restart/dr-volume-node-rebooted/ 
- Scenario 2
1. Create a pod with a Longhorn volume.
2. Write data to the volume and get its MD5 checksum.
3. Create the first backup for the volume.
4. Create a DR (Disaster Recovery) volume from this backup.
5. Wait for the DR volume to complete the initial restore.
6. Write more data to the original volume and get the new MD5 checksum.
7. Create a second backup for the volume.
8. During the incremental restore of the DR volume, immediately `stop` the node to which the DR volume is attached.
9. Wait for the instance status change to `stopped` and then `start` the node
10. Wait for the DR volume to detach and then reattach itself to the node.
11. Wait for the DR volume to complete the restore after reattachment.
12. Activate the DR volume and check the MD5 checksum of the data.

## Expected behavior

The DR volume should reattach to an available node and continue the restore process without entering a faulted state, even after a node reboot.

## Support bundle for troubleshooting

[supportbundle_295c69fd-6ae1-4b69-a7be-8b0b023fbf84_2024-11-04T04-08-10Z.zip](https://github.com/user-attachments/files/17613810/supportbundle_295c69fd-6ae1-4b69-a7be-8b0b023fbf84_2024-11-04T04-08-10Z.zip)

## Environment

 - Longhorn version: `master-head`
 - Impacted volume (PV): `vol-1`
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl): Kubectl
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: `v1.31.1+k3s1`
   - Number of control plane nodes in the cluster: `1`
   - Number of worker nodes in the cluster: `3`
 - Node config
   - OS type and version:  SLES 15-sp5 
[...]

PR #9881, opened 2024-12-02
chore: deploy MinIO/NFS on ctrl node for stability

roger-ryao:
https://github.com/longhorn/longhorn/issues/8425

- Deploys the backup store on the control node to improve reliability. This is based on discussions with @yangchiu and @chriscchien.
- Previous tests showed that when the `DR volume` is attached to the same node as the `longhorn-test-minio` pod or an `NFS` pod, rebooting the node causes the `DR` volume to enter a Faulted state.

Note : **This change might affect existing backup and restore processes in environments without a dedicated control node.**

https://github.com/longhorn/longhorn/issues/8425
https://github.com/longhorn/longhorn/issues/9752#issuecomment-2453923408

`MinIO`/`NFS` on the control node regression test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7724
Screenshot_20241202_121227
Screenshot_20241202_121207

### DIFF
--- a/deploy/backupstores/minio-backupstore.yaml
+++ b/deploy/backupstores/minio-backupstore.yaml
@@ -41,6 +41,27 @@ spec:
       labels:
         app: longhorn-test-minio
     spec:
+      nodeSelector:
+        node-role.kubernetes.io/control-plane: "true"
+      tolerations:
+      - key: "node-role.kubernetes.io/master"
+        operator: "Exists"
+        effect: "NoSchedule"
+      - key: "node-role.kubernetes.io/master"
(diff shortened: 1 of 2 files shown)


roger-ryao:
> I don't believe we want to modify the backupstore yaml files here, as deploying backupstores on the control plane is only for testing purpose and doesn't need to impact customers or users. Keeping custom backupstore yaml files in longhorn-tests repo should be good enough.

Close this PR and create another one: https://github.com/longhorn/longhorn-tests/pull/2173 on the longhorn-tests repo.

Issue #9830, opened 2024-11-18
[IMPROVEMENT] Check NFS versions in /etc/nfsmount.conf instead


derekbit:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

```
https://longhorn.io/docs/1.5.4/deploy/install/#installing-nfsv4-client
only checks that the kernel is able to use nfs4, NOT that it is actually in use!
The documentation on that page should be updated to check (at laest) /etc/nfsmount.conf instead of just checking the kernel.
```

## Describe the solution you'd like

- [ ] Update the official document
- [ ] Update the longhorn-manager node condition check  @mantissahz 
- [ ] Update the longhorn cli cc @mantissahz

COLDTURNIP:
## PR
- longhorn/cli#129

c3y1huang:
We also need to ensure this works on a Talos Linux cluster. cc @longhorn/qa

COLDTURNIP:
PR for node controller: longhorn/longhorn-manager/#3335

mantissahz:
>PR for node controller: https://github.com/longhorn/longhorn-manager/pull/3335

@COLDTURNIP could you update the PR and necessary information like reproduce steps on the `Pre Ready-For-Testing Checklist`?

COLDTURNIP:
# Test Steps

- Step 1. a Longhorn ready cluster, ensure the following changes are applied:
  - latest char (longhorn/longhorn#9910)
  - the longhornctl is updated (longhorn/cli#129)
  - the longhorn-manager is updated (longhorn/longhorn-manager#3335)
  - Ensure `longhornctl check preflight` all green.
- Step 2. enter the shell of each `longhorn-manager` daemonset containers. The host's `/etc` should be mounted at container's `/host/etc`
- Step 3. backup host's `/host/etc/nfsmount.conf` if present
- Step 4. create/modify host's `/host/etc/nfsmount.conf`, change the `Defaultvers` to specific version (see below)
- Step 5. check the nodes using `longhornctl check preflight`. If the `Defaultvers` is not acceptable, a warning item will be attached: `NFS4 is supported, but default protocol version is not 4, 4.1, or 4.2. Please refer to the NFS mount configuration manual page for more information: man 5 nfsmount.conf`
[...]

PR #9910, opened 2024-12-05
feat(chart): mount host's /etc on Longhorn manager for NFS configuration check

COLDTURNIP:

Issue longhorn/longhorn#9830

LH manager's node controller checks the NFS configuration for LH node runtime status. It needs host's `/etc/` to read `/etc/nfsmount.conf`

### DIFF
--- a/deploy/longhorn.yaml
+++ b/deploy/longhorn.yaml
@@ -4896,6 +4896,8 @@ spec:
           mountPath: /host/dev/
         - name: proc
           mountPath: /host/proc/
+        - name: etc
+          mountPath: /host/etc/
         - name: longhorn
           mountPath: /var/lib/longhorn/
           mountPropagation: Bidirectional
@@ -4932,6 +4934,9 @@ spec:
       - name: proc
         hostPath:
           path: /proc/
+      - name: etc
+        hostPath:
+          path: /etc/
       - name: longhorn
         hostPath:
           path: /var/lib/longhorn/
--- a/chart/templates/daemonset-sa.yaml
+++ b/chart/templates/daemonset-sa.yaml
@@ -69,6 +69,8 @@ spec:
           mountPath: /host/dev/
         - name: proc
           mountPath: /host/proc/
+        - name: etc
+          mountPath: /host/etc/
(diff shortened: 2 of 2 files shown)


c3y1huang (dismissed):
The document on [Root and Privileged Permission - Longhorn Manager](https://github.com/longhorn/website/blob/c4fa470afbf0fc55968020e1417ad40e230ce110/content/docs/1.8.0/deploy/install/_index.md?plain=1#L158-L161) section needs to be updated.

COLDTURNIP:
Doc PR: longhorn/website#1019

deploy/longhorn.yaml
```diff
@@ -4896,6 +4896,8 @@ spec:
           mountPath: /host/dev/
         - name: proc
           mountPath: /host/proc/
+        - name: etc
+          mountPath: /host/etc/
```
innobead, line 4900: Suggest using a read-only mount to prevent any unexpected changes from impacting the user's environment, or consider mounting the exact path specifically as we only need that specific config file.

innobead (changes requested):

deploy/longhorn.yaml
COLDTURNIP: Will change to RO `/etc` since the target file `/etc/nfsmount.conf` is optional.

COLDTURNIP: We'll need `recursiveReadOnly: Enabled` (and `mountPropagation: None`) after Kubernetes v1.31

innobead: Can we just mount a specific file? as we don't need the whole /etc on host?

innobead: Clear. It might be a compromise for now. Please help create a follow-up ticket for `recursiveReadOnly: Enabled`.

innobead (approved):

derekbit (approved):

derekbit:
@mergify backport v1.7.x

The change doesn't need to backport to v1.6.x which doesn't support CLI and node.status.conditions.

Issue #6361, opened 2023-07-21
[FEATURE] SPDK volume supports live migration


derekbit:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

SPDK volume supports live migration as v1 volume

## Describe the solution you'd like

A clear and concise description of what you want to happen

shuo-wu:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Longhorn-Sprint-59d8105b2fbdf4155c4589ef/issues/longhorn/longhorn/6361?planning-poker) with Zenhub @DamiaSan @derekbit @innobead

DamiaSan:
For, volume live migration can we use the same procedure of [online volume rebuilding](https://github.com/longhorn/longhorn-spdk-engine/wiki/How-to-create-a-copy-of-a-volume-on-a-different-node-with-SPDK) right?

PhanLe1010:
e2e test case ticket is https://github.com/longhorn/longhorn/issues/9905

TODO: We should design a test skeleton for https://github.com/longhorn/longhorn/issues/9905 as part of this ticket

PhanLe1010:
## Test plan:

Case 1: Migration confirmation 

Case 2: Migration rollback

Case 3:Engine crash during migration 
1. Original engine crash
1. Migrating engine crash

Case 4: Replica crash during live migration:
1. Some replica crash
1. All replica crash: failed

Case 5: Run all e2e migration test cases with v2 volume

PR #9925, opened 2024-12-07
feat: v2 Volume Live Migration

PhanLe1010:
longhorn/longhorn#6361


enhancements/20241206-v2-volume-live-migration.md
```diff
@@ -0,0 +1,180 @@
...
+  volumes:
+    - name: block-vol
+      persistentVolumeClaim:
+        claimName: longhorn-block-vol
+```
+Longhorn will expose the volume to both node-1 and node-2.
+1. User then can do operation like flush/pausing IO on pod-1 and delete pod-1. 
+   This would mean migration confirmation. 
+   Longhorn would then detach the volume from node-1. 
+   The volume become sorely attached to node-2.
```
james-munson, line 106: "sorely" -> "solely"

enhancements/20241206-v2-volume-live-migration.md
```diff
@@ -0,0 +1,180 @@
...
+2. User can also delete the pod-2.
+   This would mean migration rollback
+   Longhorn would then detach the volume from node-2.
+   The volume become sorely attached to node-1.
```
james-munson, line 110: ditto.

enhancements/20241206-v2-volume-live-migration.md
```diff
@@ -0,0 +1,180 @@
...
+
+The reason, we have to create new inactive replicas r1' and r2' is because v1 replica is designed to be connected to a single engine at time.
+However, it seems that v2 replica can be connected by multiple engine at a time.
+More over, creating and matching new inactive replicas in v1 live migration seems a bit complicated especially when it comes to handle migration revert.
+Therefore, I decided to not creating new inactive replicas in v2 live migration. 
+The flow would be like this:
+
+![v2 volume live migration flow](./assets/images/v2-volume-live-migration/v2-volume-live-migration-flow.png)
+
+// TODO: describe the state transition by sudo code
```
james-munson, line 159: "sudo" should be "pseudo".  Is there a plan to fill in this part?

enhancements/20241206-v2-volume-live-migration.md
```diff
@@ -0,0 +1,180 @@
...
+1. Start engine e2 with replica r1' and r2'
+1. Now migration is ready. User can either do migration confirmation or migration rollback 
+1. If user does migration confirmation:
+   1. Longhorn delete e1
+   1. Active e2
+   1. Active r1' and r2'
+   1. Deactive r1 and r2
+1. If user does migration rollback:
+    1. Longhorn delete e2
+    1. Delete r1' and r2'
```
james-munson, line 149: I gather that only zero or one engine can be active at a time, so an attempt to activate e2 before e1 is deleted will be rejected?  That's how we protect against simultaneous writes from both?

james-munson (changes requested):
Some small spelling items and a couple of questions.

james-munson (commented):
Taking off the "request changes" tag in case I don't get a chance to later.

enhancements/20241206-v2-volume-live-migration.md
PhanLe1010: Thank you. Fixed

enhancements/20241206-v2-volume-live-migration.md
PhanLe1010: Thank you. Fixed

enhancements/20241206-v2-volume-live-migration.md
PhanLe1010: No, we don't stop user from writing to 2 engine at the same time. It is the responsibility of user to coordinate the write

But you are correct that there at most on active engine at a time

innobead: Have we explained users should manage exclusive IO operation during the migration in the current doc? If no, it might be helpful to improve the doc to  prevent users from having any wrong expectation.

innobead (approved):
LGTM.

enhancements/20241206-v2-volume-live-migration.md
PhanLe1010: Hi @innobead 
I think only Harvester uses this live migration feature and they are aware of this. 
I just looked up and see that we don't mention this live migration feature at all in our document. 
Maybe we don't need to document it?

innobead: Okay, that's fine.

PR #9952, opened 2024-12-11
feat(script/env_check): check overriden NFS version

COLDTURNIP:

Issue longhorn/longhorn#9830

Though the node supports NFSv4, the NFS client picks the protocol version by `Defaultvers` defined in `nfsmount.conf`. This may cause problem while using backup-related features.

This change is kind of a backport version of longhorn/cli#129. It is needed since Longhorn v1.6 is still maintained, while `longhornctl check preflight` is provided after v1.7.

### DIFF
--- a/scripts/environment_check.sh
+++ b/scripts/environment_check.sh
@@ -388,6 +388,7 @@ check_nfs_client() {
     return 1
   fi
 
+  local nfs_module_not_found=1
   for option in "${options[@]}"; do
     kubectl exec ${pod} -- nsenter --mount=/proc/1/ns/mnt -- bash -c "[ -f /boot/config-${kernel} ]" > /dev/null 2>&1
     if [ $? -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


scripts/environment_check.sh
```diff
@@ -396,13 +397,27 @@ check_nfs_client() {
...
 
-  error "NFS clients ${options[*]} not found. At least one should be enabled"
-  return 1
+  if [ $nfs_module_not_found -ne 0 ]; then
+    error "NFS clients ${options[*]} not found. At least one should be enabled"
+    return 1
+  fi
+
+  local nfs_ver_overriden=$(kubectl exec ${pod} -- nsenter --mount=/proc/1/ns/mnt -- bash -c "grep '^\s*Defaultvers=' /etc/nfsmount.conf" 2> /dev/null)
+  if [ "$nfs_ver_overriden" ]; then
```
COLDTURNIP, line 412: Doesn't use `[[ "..." =~ regex_ptn ]]` to ensure shell compatibility.

derekbit (approved):

COLDTURNIP:
@mergify backport v1.7.x

COLDTURNIP:
@mergify backport v1.6.x

derekbit:
My bad. We don't need to implement this in master and v1.7.x, because we've implemented it in the CLI.

PR #9982, opened 2024-12-16
feat(chart): mount host's /boot & /proc on Longhorn manager for env check

COLDTURNIP:

Issue longhorn/longhorn#9830

For Longhorn manager, besides `/etc`, the `/boot` and `/proc` hostpath mounts are needed for environment check, thus we only need RO mount for safety.

Previous change: #9910
Doc change: longhorn/website#1019

### DIFF
--- a/chart/templates/daemonset-sa.yaml
+++ b/chart/templates/daemonset-sa.yaml
@@ -65,10 +65,12 @@ spec:
         volumeMounts:
         - name: boot
           mountPath: /host/boot/
+          readOnly: true
         - name: dev
           mountPath: /host/dev/
         - name: proc
           mountPath: /host/proc/
+          readOnly: true
         - name: etc
           mountPath: /host/etc/
           readOnly: true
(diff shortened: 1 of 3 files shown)


innobead:
> For Longhorn manager, besides `/etc`, the `/boot` and `/proc` hostpath mounts are needed for environment check, thus we only need RO mount for safety.

Do we use these paths for other purposes that need write permission?

COLDTURNIP:
> > For Longhorn manager, besides `/etc`, the `/boot` and `/proc` hostpath mounts are needed for environment check, thus we only need RO mount for safety.
> 
> Do we use these paths for other purposes that need write permission?

`/boot` stores the kernel and ramdisk files that we're not expected to access. We read only the kernel config to ensure the necessary modules loaded.

And manager accesses `/proc` for the following purpose:

- Look up the PID of container daemon, then enter the namespace through this PID to collect node information, such as disk info.
- Read the compressed kernel config `/proc/config.gz` to ensure the necessary modules loaded.

These operations should be read-only.

derekbit (approved):
LGTM. Thanks @COLDTURNIP

COLDTURNIP:
@mergify backport v1.7.x

The change doesn't need to backport to v1.6.x which doesn't support CLI and node.status.conditions.
Ref: #9910

PR #10016, opened 2024-12-19
release: v1.8.0-rc1

derekbit:


chart/questions.yaml
```diff
@@ -53,7 +53,7 @@ questions:
     label: Longhorn Instance Manager Image Repository
     group: "Longhorn Images Settings"
   - variable: image.longhorn.instanceManager.tag
-    default: v2_20221123
```
innobead, line 56: The replaced value is quite old. It's v2_20221123 where we did not run branched-based release.

Issue #9764, opened 2024-11-06
[TASK] Verify Longhorn v1.8.0 on Talos v1.8.3


derekbit:
## What's the task? Please describe

The latest Talos is [v1.8.3](https://github.com/siderolabs/talos/releases), so we need to verify it on Longhorn v1.8.0 and update https://longhorn.io/docs/1.8.0/best-practices/#operating-system.

derekbit:
Ref: https://github.com/longhorn/longhorn/issues/9764

cc @c3y1huang @innobead

roger-ryao:
Currently, I have observed two issues:

1. Triggering our `Talos Jenkins job` with `v1.7.6` causes it to interrupt halfway:
- https://ci.longhorn.io/job/public/job/master/job/talos/job/amd64/job/longhorn-tests-talos-amd64/4/console
- https://ci.longhorn.io/job/public/job/master/job/talos/job/amd64/job/longhorn-tests-talos-amd64/2/console

2. When trying to use the `v1.8.2` or `v1.8.3` AMI provided by the `Talos team`, the worker node failed to join the cluster, and Terraform returned an error message:
```bash
Warning: Deprecated

  with data.talos_cluster_kubeconfig.this,
  on main.tf line 263, in data "talos_cluster_kubeconfig" "this":
 263: data "talos_cluster_kubeconfig" "this" {

Use `talos_cluster_kubeconfig` resource instead. This data source will be removed in the next minor version of the provider.

(and one more similar warning elsewhere)

Error: Error applying configuration

  with talos_machine_configuration_apply.worker[0],
  on main.tf line 235, in resource "talos_machine_configuration_apply" "worker":
 235: resource "talos_machine_configuration_apply" "worker" {

rpc error: code = InvalidArgument desc = configuration validation failed: 1 error occurred:
        * issuing CA key is required for controlplane nodes (.machine.ca.key)



Error: Error applying configuration

  with talos_machine_configuration_apply.worker[2],
  on main.tf line 235, in resource "talos_machine_configuration_apply" "worker":
 235: resource "talos_machine_configuration_apply" "worker" {

rpc error: code = InvalidArgument desc = configuration validation failed: 1 error occurred:
        * issuing CA key is required for controlplane nodes (.machine.ca.key)



Error: Error applying configuration

  with talos_machine_configuration_apply.worker[1],
  on main.tf line 235, in resource "talos_machine_configuration_apply" "worker":
[...]

PR #10119, opened 2025-01-02
fix(gh-action): reference PR to right backport issues

innobead:
Signed-off-by: davidko <dko@suse.com>

ref: https://github.com/longhorn/longhorn/issues/9764#issuecomment-2567254532


yangchiu:
@Mergify backport v1.8.x

PR #10245, opened 2025-01-18
chore(template): update bug, feature and improvement templates

derekbit:

Issue #

Improve the quality of issue reported.


.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -0,0 +1,78 @@
...
+      description: A clear and concise description of what you expected to happen.
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: Support Bundle for Troubleshooting
+      description: Please provide a support bundle when the issue happens. You can generate a support bundle using the link at the footer of the Longhorn UI. Check [here](https://longhorn.io/docs/latest/troubleshoot/support-bundle/). Then, attach to the issue or send to longhorn-support-bundle@suse.com.
+    validations:
+      required: false
```
innobead, line 41: True instead?

.github/ISSUE_TEMPLATE/feature.yaml
```diff
@@ -0,0 +1,41 @@
...
+      description: A clear and concise description of what the problem is. Ex. I'm always frustrated when [...].
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: Describe the solution you'd like
+      description: A clear and concise description of what you want to happen and share how this will benefit Longhorn and its users.
+    validations:
+      required: true
```
innobead, line 27: False instead? As the reporter might not have a concrete idea for the solution.

innobead (changes requested):
This is good to leverage the structure issue schema provided by GitHub.

.github/ISSUE_TEMPLATE/bug.yaml
derekbit: Might fail to create a support bundle.
But yes, setting it to `true` can make sure users aware of this information. If they cannot provide a support bundle, they should provide reasons.

derekbit: Updated. Thank you.

.github/ISSUE_TEMPLATE/feature.yaml
derekbit: Updated. Thank you.

innobead (approved):

Issue #5137, opened 2022-12-23
[TASK] Add resource missing check in PR check


innobead:
## What's the task? Please describe

When adding new resources, it's easy to miss to add resources authorization/RBAC to uninstall manifest. That will be a problem for users who use the deploy manifest to install/uninstall Longhorn. 

## Describe the items of the task (DoD, definition of done) you'd like
> Please use a task list for items on a separate line with a clickable checkbox https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists

- [ ] `item 1`

## Additional context

Add any other context or screenshots about the task request here.

innobead:
Moving this to ready for testing, as the code has been merged.

roger-ryao:
**Verified on v1.9.0-dev-20250406 20250408**

- longhorn `v1.9.0-dev-20250406` https://github.com/longhorn/longhorn/commit/903cc557b96da8a97684dccf4f67f74dc2ee687b

**The test steps**

https://github.com/longhorn/longhorn/issues/5137#issuecomment-2711041948

**Result Passed**

PR #10255, opened 2025-01-20
chore(script): regenerate uninstall.yaml from deploy/longhorn.yaml

derekbit:

Issue Longhorn/longhorn#5137

To prevent the update of CRDs, uninstall.yaml will be regenerated from deploy/longhorn.yaml.
The python file will be executed by the workflow in https://github.com/longhorn/longhorn-manager/pull/3486.

### DIFF
--- a/scripts/update-uninstall-manifest.py
+++ b/scripts/update-uninstall-manifest.py
@@ -1,0 +1,69 @@
+import yaml
+import os
+from typing import List, Dict, Iterator
+from pathlib import Path
+
+class YAMLResourceProcessor:
+    def __init__(self, longhorn_manifest: str, uninstall_file: str):
+        self.longhorn_manifest = Path(longhorn_manifest)
+        self.uninstall_file = Path(uninstall_file)
+        self.output_file = self.uninstall_file.parent / 'uninstall.tmp.yaml'
+
+    def load_yaml_documents(self, file_path: Path) -> Iterator[Dict]:
+        """Load YAML documents from a file, skipping None values."""
+        try:
+            with open(file_path, 'r') as file:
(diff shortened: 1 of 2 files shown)


scripts/update-uninstall-manifest.py
```diff
@@ -0,0 +1,69 @@
...
+                    resources.append(name.split(".")[0])
+        return resources
+
+    def update_cluster_role(self, doc: Dict, resources: List[str]) -> Dict:
+        """Update ClusterRole document with new resources."""
+        if doc['kind'] == "ClusterRole":
+            for rule in doc.get('rules', []):
+                if rule.get('apiGroups') == ["longhorn.io"]:
+                    rule['resources'] = resources
+        return doc
```
innobead, line 39: Just a notice. The change here is based on the existing uninstall.yaml, so if it has some changes, the change might affect the expected result here. 

[/uninstall/uninstall.yaml#L66-L70](https://github.com/longhorn/longhorn/blob/b265b3f9abd577d20759b5a5feb443bf6230b0f9/uninstall/uninstall.yaml#L66-L70)
```yaml
  - apiGroups: ["longhorn.io"]
    resources: ["volumes", "engines", "replicas", "settings", "engineimages", "nodes", "instancemanagers", "sharemanagers",
                "backingimages", "backingimagemanagers", "backingimagedatasources", "backuptargets", "backupvolumes", "backups",
                "recurringjobs", "orphans", "snapshots", "supportbundles", "systembackups", "systemrestores", "volumeattachments", "backupbackingimages"]
    verbs: ["*"]
```

innobead (approved):
LGTM

PR #10288, opened 2025-01-23
chore(crd): update crds.yaml and manifests (PR longhorn/longhorn-manager#)

derekbit:
It was triggered by longhorn/longhorn-manager#.


derekbit:
Closed because this is a testing PR.

PR #10289, opened 2025-01-23
chore(crd): update crds.yaml and manifests (PR longhorn/longhorn-manager#5)

derekbit:
It was triggered by longhorn/longhorn-manager#5.


derekbit:
Sorry. This is created by testing.

PR #10290, opened 2025-01-23
chore(crd): update crds.yaml and manifests (PR longhorn/longhorn-manager#3509)

derekbit:
It was triggered by longhorn/longhorn-manager#3509.

### DIFF
--- a/deploy/longhorn.yaml
+++ b/deploy/longhorn.yaml
@@ -113,6 +113,7 @@ data:
       dataEngine: "v1"
 ---
 # Source: longhorn/templates/crds.yaml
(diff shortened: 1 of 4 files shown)


deploy/longhorn.yaml
```diff
@@ -113,6 +113,7 @@ data:
       dataEngine: "v1"
 ---
 # Source: longhorn/templates/crds.yaml
+# Generating crds.yaml from github.com/longhorn/longhorn-manager/k8s/pkg/apis and the crds.yaml will be copied to longhorn/longhorn chart/templates and cannot be directly used by kubectl apply.
```
innobead, line 116: Just noticed the message should be "# Generated crds..."

innobead: Let's fix this minor typo in manager repo first. Thanks.

Issue #10323, opened 2025-02-03
[BUG] Backup Execution Timeout setting issue in Helm chart


vvanholl:
### Describe the Bug

Hi,

I attempted to use the backupExecutionTimeout setting in the Helm chart from version 1.8.0 but noticed that it was never applied.

Upon checking the logs of the manager pod, I found the following error:

```
{"file":"/go/src/github.com/longhorn/longhorn-manager/types/setting.go:1586","func":"github.com/longhorn/longhorn-manager/types.GetCustomizedDefaultSettings","level":"error","msg":"Customized settings are invalid, will give up using them: undefined setting backup-execute-timeout","time":"2025-02-03T20:51:22Z"}
```

The issue seems to originate from the longhorn-default-setting ConfigMap, where the parameter is incorrectly named backup-execute-timeout instead of backup-execution-timeout. Due to this mismatch, the manager does not recognize the setting and discards it.

```
apiVersion: v1  
data:  
  default-setting.yaml: |-  
    replica-auto-balance: best-effort  
    storage-minimal-available-percentage: 10  
    storage-reserved-percentage-for-default-disk: 0  
    upgrade-checker: false  
    default-replica-count: 2  
    default-data-locality: best-effort  
    backup-execute-timeout: 60  
    recurring-successful-jobs-history-limit: 10  
```

### To Reproduce

Deploy Longhorn v1.8.0 using the Helm chart.
Set backupExecutionTimeout in the Helm values file.
Check the longhorn-default-setting ConfigMap.
Observe that the setting appears as backup-execute-timeout instead of backup-execution-timeout.
Check the manager pod logs and observe the error.

### Expected Behavior

The manager should recognize and apply the backupExecutionTimeout setting correctly.

### Support Bundle for Troubleshooting

[supportbundle_cc181dab-7d25-423d-b33d-01e75efaf0ba_2025-02-03T21-00-28Z.zip](https://github.com/user-attachments/files/18648222/supportbundle_cc181dab-7d25-423d-b33d-01e75efaf0ba_2025-02-03T21-00-28Z.zip)

### Environment

- Longhorn version: 
- Impacted volume (PV): 
- Installation method (e.g. Rancher Catalog App/Helm/Kubectl):
- Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version:
  - Number of control plane nodes in the cluster:
  - Number of worker nodes in the cluster:
- Node config
  - OS type and version:
  - Kernel version:
[...]

PR #10324, opened 2025-02-03
fix: correct name in default settings configmap

james-munson:

Issue https://github.com/longhorn/longhorn/issues/10323

Make name in default settings configmap match name in settings.

### DIFF
--- a/chart/templates/default-setting.yaml
+++ b/chart/templates/default-setting.yaml
@@ -46,7 +46,7 @@ data:
     failed-backup-ttl: {{ .Values.defaultSettings.failedBackupTTL }}
     {{- end }}
     {{- if not (kindIs "invalid" .Values.defaultSettings.backupExecutionTimeout) }}
-    backup-execute-timeout: {{ .Values.defaultSettings.backupExecutionTimeout }}
+    backup-execution-timeout: {{ .Values.defaultSettings.backupExecutionTimeout }}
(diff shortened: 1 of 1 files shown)


chart/templates/default-setting.yaml
```diff
@@ -46,7 +46,7 @@ data:
     failed-backup-ttl: {{ .Values.defaultSettings.failedBackupTTL }}
     {{- end }}
     {{- if not (kindIs "invalid" .Values.defaultSettings.backupExecutionTimeout) }}
-    backup-execute-timeout: {{ .Values.defaultSettings.backupExecutionTimeout }}
+    backup-execution-timeout: {{ .Values.defaultSettings.backupExecutionTimeout }}
```
derekbit, line 49: @james-munson Can you add it into questions.yaml and official doc as well? Thanks.

derekbit:
@mergify backport v1.7.x v1.8.x

chart/templates/default-setting.yaml
james-munson: It is already in questions.yaml.  I'll make a separate PR to add it to the 1.8 and 1.9 docs.

derekbit: Good. Thanks @james-munson

derekbit (approved):

Issue #10379, opened 2025-02-10
[BUG] longhorn/longhorn `Scorecard supply-chain security` always failed


derekbit:
### Describe the Bug

longhorn/longhorn `Scorecard supply-chain security` always failed

### To Reproduce

Run GitHub Action `Scorecard supply-chain security` on longhorn/longhorn

### Expected Behavior

Should succeed

### Support Bundle for Troubleshooting

https://github.com/longhorn/longhorn/actions/workflows/scorecards.yml

### Environment

- Longhorn version: 
- Impacted volume (PV): 
- Installation method (e.g. Rancher Catalog App/Helm/Kubectl):
- Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version:
  - Number of control plane nodes in the cluster:
  - Number of worker nodes in the cluster:
- Node config
  - OS type and version:
  - Kernel version:
  - CPU per node:
  - Memory per node:
  - Disk type (e.g. SSD/NVMe/HDD):
  - Network bandwidth between the nodes (Gbps):
- Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal):
- Number of Longhorn volumes in the cluster:

derekbit:
Build passed. Completed.

PR #10380, opened 2025-02-10
fix(workflow): fix "Scorecard supply-chain security"

derekbit:

Issue Longhorn/longhorn#10379

### DIFF
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -55,7 +55,7 @@ jobs:
       # Upload the results as artifacts (optional). Commenting out will disable uploads of run results in SARIF
       # format to the repository Actions tab.
       - name: "Upload artifact"
-        uses: actions/upload-artifact@97a0fba1372883ab732affbe8f94b823f91727db # v3.pre.node20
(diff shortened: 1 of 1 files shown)


.github/workflows/scorecards.yml
```diff
@@ -55,7 +55,7 @@ jobs:
       # Upload the results as artifacts (optional). Commenting out will disable uploads of run results in SARIF
       # format to the repository Actions tab.
       - name: "Upload artifact"
-        uses: actions/upload-artifact@97a0fba1372883ab732affbe8f94b823f91727db # v3.pre.node20
+        uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
```
innobead, line 58: Why don't we use @v4? I assume the minor or patch version of GitHub Actions should generally be fine.

innobead (changes requested):

.github/workflows/scorecards.yml
derekbit: Updated. Thank you.

innobead (approved):

Issue #10410, opened 2025-02-12
[TASK] Automatically Add Sprint Tag When Changing Issue to "Ready for Testing"


roger-ryao:
## What's the task? Please describe

After discussing with @derekbit , we propose that when an issue's status is changed to `Ready for Testing`, the system should automatically add the corresponding `sprint` tag. This change will help reduce the manual effort required to confirm whether the issue fix is included in the `sprint release image`.

PR #10435, opened 2025-02-18
feat(workflow): set Ready For Testing ticket's sprint to current

derekbit:

Issue Longhorn/longhorn#10410

Set "Ready For Testing" ticket's sprint to current if it is not set.

### DIFF
--- a/.github/workflows/periodic-issue-sprint-update.yml
+++ b/.github/workflows/periodic-issue-sprint-update.yml
@@ -1,4 +1,4 @@
-name: Update-Issue-Sprint
+name: Periodic Issue Sprint Update
 
 on:
   schedule:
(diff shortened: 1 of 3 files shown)


.github/workflows/periodic-issue-sprint-update.yml
```diff
@@ -53,6 +53,18 @@ jobs:
         new-iteration: next
         statuses: "Review"
 
+    - name: Longhorn Sprint Issues - Update None to Current Sprint
+      if: steps.check_sprint_build_required.outcome == 'success'
+      uses: derekbit/move-to-next-iteration@master
```
c3y1huang, line 58: Will this be moved to Longhorn projects in the future?
https://github.com/derekbit/move-to-next-iteration

c3y1huang (approved):

.github/workflows/periodic-issue-sprint-update.yml
derekbit: Great question. Do you think it is better to move it into Longhorn? I'm fine with it.

Issue #9456, opened 2024-09-13
[FEATURE] v2 volume supports UBLK frontend


derekbit:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

According to preliminary results in https://github.com/longhorn/longhorn/issues/5159, replacing the iSCSI frontend with UBLK has shown significant performance gains. Currently, v2 uses nvmf-tcp as its frontend protocol, but SPDK also supports UBLK (see [UBLK documentation](https://spdk.io/doc/ublk.html)). The purpose of this ticket is to further compare nvme-tcp with UBLK to determine if UBLK could replace nvme-tcp, and to explore the associated error handling strategies.

## Describe the solution you'd like

- Compare the performance of nvme-tcp and UBLK in various scenarios, focusing on data transfer rates, latency, and resource utilization.
- If UBLK is identified as a potential candidate, investigate the feasibility of replacing nvme-tcp with UBLK as the frontend protocol.
- Develop error handling procedures to ensure stability and reliability when using UBLK.

## Describe alternatives you've considered

Retaining the current nvmf-tcp protocol and optimizing its performance.

## Additional context

 https://spdk.io/doc/ublk.html

derekbit:
Assigned to @PhanLe1010, who has a strong interest in UBLK. Mark it as `P1`; if we have enough bandwidth, we can include it in v1.8.0 or defer it to v1.9.0 later.

cc @innobead

derekbit:
cc @shuo-wu @DamiaSan

PhanLe1010:
Update:

Pushed the draft PR for the data plane:

1. [fet: implement ublk frontend for v2 data engine longhorn/go-spdk-helper#198](https://github.com/longhorn/go-spdk-helper/pull/198)
2. [fet: implement ublk frontend for v2 data engine longhorn/longhorn-spdk-engine#328](https://github.com/longhorn/longhorn-spdk-engine/pull/328)
3. [fet: implement ublk frontend for v2 data engine longhorn/longhorn-instance-manager#911](https://github.com/longhorn/longhorn-instance-manager/pull/911)

Next step:
* Write the LEP 
* Implement the control plan

cc @derekbit @shuo-wu @innobead

chriscchien:
[...]

PR #10587, opened 2025-03-13
feat: LEP for UBLK frontend for v2 data engine

PhanLe1010:
longhorn/longhorn#9456


enhancements/20250313-ublk-frontend-for-v2-engine.md
```diff
@@ -0,0 +1,393 @@
+# UBLK Frontend For V2 Engine
+
+## Summary
+
+Longhorn v2 data engine current only support NVME-TCP frontend. 
+We would like to support UBBLK frontend for v2 data engine.
+In high performance evn, UBLK frontend may have better performance compares to NVME-TCP.
```
liyimeng, line 7: evn --> env?

enhancements/20250313-ublk-frontend-for-v2-engine.md
```diff
@@ -0,0 +1,393 @@
...
+	SpecSize   uint64
+	ActualSize uint64
+	Frontend   string
+	Endpoint   string
+
+	ctrlrLossTimeout     int
+	fastIOFailTimeoutSec int
+	ReplicaStatusMap     map[string]*EngineReplicaStatus
+
+	NvmeTcpFrontend *NvmeTcpFrontend
```
liyimeng, line 257: Is it possible to have them convergence as a Frontend interface instead?

enhancements/20250313-ublk-frontend-for-v2-engine.md
PhanLe1010: Thank you. Fixed!

enhancements/20250313-ublk-frontend-for-v2-engine.md
PhanLe1010: Interesting idea!

We often need to reference to the inner field of NvmeTcpFrontend or UblkFrontend. They have different struct. Could you please collaborate more of how to combine them into a single interface? 

This is the implementation code on the engine repo https://github.com/longhorn/longhorn-spdk-engine/pull/328

liyimeng: hmm, when looking at the picture enhancements/assets/images/ublk-frontend-for-v2-engine/nvme-tcp-frontend-vs-ublk-frontend.png, I thought the frondend might share some common operations.  So I thought we can make definition like this 

```

type Frontend interface{
   getDevice() string
    ....
}

func NewFrontend(t FrontendType) Frontend{

   if  t== FrontendType.UBLK {
      return &UblkFrontEnd{}
   }else {
     return &NvmeFrontEnd{}
  }
}

```

But when checking the code you provided , it seem quite difference from what I thought. 

Sorry for making the noise.

enhancements/20250313-ublk-frontend-for-v2-engine.md
```diff
@@ -0,0 +1,397 @@
...
+
+### User Stories
+
+In high performance evn, UBLK frontend may have better performance compares to NVME-TCP.
+Ref https://github.com/longhorn/longhorn/wiki/Longhorn-Performance-Investigation
+
+### User Experience In Detail
+
+UBLK framework added in the Linux 6.0 kernel.
+User need to have kernel version >= 6.0 to use this feature.
```
c3y1huang, line 35: What happens if a user tries to use `ublk` as the frontend but their kernel version is `<6.0`? Should Longhorn block volume creation or issue a warning?

enhancements/20250313-ublk-frontend-for-v2-engine.md
```diff
@@ -0,0 +1,397 @@
...
+1. UBLK initiator creates new ublk device for the raid bdev by calling `UblkStartDisk`
+1. UBLK initiator set `UblkInfo.UblkID` to the avaiable ID
+1. UBLK initiator links that device to dm mapper
+1. UBLK initiator duplicates dm device to `e.Endpoint`
+
+#### The flow when deleting a new v2 engine with new UBLK frontend
+1. Longhorn manager send gRPC call to instance manager to delete engine with UBLK frontend
+1. UBLK initiator removes dm mapper device
+1. UBLK initiator removes `e.Endpoint`
+1. UBLK initiator delete ublk device associated with the raid1 bdev by calling `UblkStopDisk` with the recorded `UblkInfo.UblkID` info
```
c3y1huang, line 378: If the UBLK initiator crashes before calling `UblkStopDisk`, will the UBLK device be orphaned? How is cleanup or recovery handled in this scenario?

enhancements/20250313-ublk-frontend-for-v2-engine.md
```diff
@@ -0,0 +1,397 @@
...
+1. Longhorn manager send gRPC call to instance manager to create engine with UBLK frontend 
+1. Instance manager connect to the replicas
+1. Instance manager creates a raid1 bdev from these replicas' NVME bdevs
+1. Instance manager create ublk frontend by calling `handleUblkFrontend`
+1. Instance manager starts ublk target if it has not been started by calling `UblkCreateTarget`
+1. UBLK initiator cleanups the previous ublk device associated with the raid1 bdev by calling `UblkGetDisks` and `UblkStopDisk`
+1. UBLK initiator finds an available ublk id by calling `UblkGetDisks`
+1. UBLK initiator creates new ublk device for the raid bdev by calling `UblkStartDisk`
+1. UBLK initiator set `UblkInfo.UblkID` to the avaiable ID
+1. UBLK initiator links that device to dm mapper
```
c3y1huang, line 371: What happens if the UBLK device is partially set up? e.g.:
- The initiator crashes before linking the device to `dm mapper`
- `UblkStartDisk` succeeds, but linking to `dm mapper` fails 

Is there a retry/rollback mechanism to recover?

enhancements/20250313-ublk-frontend-for-v2-engine.md
shuo-wu: Probably Longhorn cannot block the volume creation for this case since the validation webhook can be on the node different from the attached node.

PhanLe1010: In this case when user attach the volume, it might just failed similar to attaching encrypted volume without having dm_crypt module loaded. WDYT?

enhancements/20250313-ublk-frontend-for-v2-engine.md
PhanLe1010: The next time user attach the volume, ublk device will be cleaned up (we lookup all ublk devices and check the path to see if it contains the volume name) before we start new ublk device.  WDYT? 
This is similar to how we handle tgt target at the moment

enhancements/20250313-ublk-frontend-for-v2-engine.md
PhanLe1010: The engine would crash.

The next time we start the engine it will:
* Stop the old ublk device
* Start a new one
* Try to link to dm mapper

enhancements/20250313-ublk-frontend-for-v2-engine.md
derekbit: Is there any potential risk if the volume is not reattached and the orphaned ublk device remains on the host?

enhancements/20250313-ublk-frontend-for-v2-engine.md
PhanLe1010: > In an attaching and detaching loop, right? I'm fine with the harmless behavior.

Yeah, you are right!

PhanLe1010: > BTW, can you add what kernel modules are required?

Surely. Added. Thank you

enhancements/20250313-ublk-frontend-for-v2-engine.md
PhanLe1010: If the IM does not restart, it is non issue. 

If the IM restarts, it might be orphan. This is an issue indeed. I don't know how to solve this issue yet. It is pretty difficult to remove orphan ublk of crashed IM. In some case, manually try to remove on the host or reloading the kernel module still doesn't help. I had to reboot the node. I suspect that ublk kernel module is still a bit new. We can continue to investigate this corner case in a new ticket. WDYT? cc @shuo-wu

enhancements/20250313-ublk-frontend-for-v2-engine.md
```diff
@@ -0,0 +1,397 @@
...
+### Non-goals
+
+In this ticket, supporting live engine upgrade is not a goal. 
+It will be handled in different ticket https://github.com/longhorn/longhorn/issues/9104
+
+## Proposal
+
+### User Stories
+
+In high performance evn, UBLK frontend may have better performance compares to NVME-TCP.
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. Fixed

enhancements/20250313-ublk-frontend-for-v2-engine.md
```diff
@@ -0,0 +1,397 @@
...
+Ref https://github.com/longhorn/longhorn/wiki/Longhorn-Performance-Investigation
+
+### User Experience In Detail
+
+UBLK framework added in the Linux 6.0 kernel.
+User need to have kernel version >= 6.0 to use this feature.
+
+
+In order to use UBLK frontend for v2 data engine, user just need to create a v2 volume with `volume.Spec.Frontend: ublk`.
```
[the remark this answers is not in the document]
PhanLe1010: I would prefer to keep `volume.Spec.Frontend: ublk` as:
* blockdev is legacy term. This is referring to block device with nvme-tcp protocol for v2 engine or iscsi protocol for v1 engine
* `nvmf`: the nvme-tcp target only
* `iscsi`: the tgtd target only
* `ublk`: block device with ublk protocol 

WDYT?

enhancements/20250313-ublk-frontend-for-v2-engine.md
```diff
@@ -0,0 +1,397 @@
...
+    size: "21474836480"
+    snapshotDataIntegrity: disabled
+    snapshotMaxCount: 250
+    snapshotMaxSize: "0"
+    staleReplicaTimeout: 20
+    unmapMarkSnapChainRemoved: disabled
+```
+
+### API changes
+We will add a new field to the engine CR's status: `UblkID` indicating the current ID of the ublk device on node corresponding to this engine.
```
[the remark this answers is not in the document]
PhanLe1010: I manually added an upper limit of `MaxUblkId                = 65535` because I believe a single node should have more than that number of device. WDYT?

enhancements/20250313-ublk-frontend-for-v2-engine.md
```diff
@@ -0,0 +1,397 @@
...
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/9456
+
+## Motivation
+
+### Goals
+
+Support UBLK frontend for v2 data engine
```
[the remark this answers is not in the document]
PhanLe1010: It will use the same CPU mask as SPDK. It might use a bit of huge page. 
Maybe we can do benchmark after the feature?

enhancements/20250313-ublk-frontend-for-v2-engine.md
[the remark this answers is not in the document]
derekbit: Got your point, but both nvmf and iscsi servers serve a similar purpose ([ref](https://longhorn.io/docs/1.8.1/nodes-and-volumes/volumes/iscsi/)), and they don't expose the device as a traditional block device like /dev/longhorn/xxx. Could this potentially cause confusion?

enhancements/20250313-ublk-frontend-for-v2-engine.md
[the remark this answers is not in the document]
derekbit: Sounds good to me. Thank you.

enhancements/20250313-ublk-frontend-for-v2-engine.md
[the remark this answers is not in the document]
derekbit: Sure. No rush. Thank you.

enhancements/20250313-ublk-frontend-for-v2-engine.md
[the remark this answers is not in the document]
PhanLe1010: I see your point. But I would say that adding ublk does not make it more confused than the current implementation. And in the ublk world, ublk already imply user space "block device". WDYT?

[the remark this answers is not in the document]
derekbit: Sure. Then, let's make it as it is. Thank you.

derekbit (dismissed):
In general, LGTM

enhancements/20250313-ublk-frontend-for-v2-engine.md
innobead: Can we add these potential limitations in this proposal as well?

enhancements/20250313-ublk-frontend-for-v2-engine.md
```diff
@@ -0,0 +1,397 @@
...
+1. Longhorn manager send gRPC call to instance manager to delete engine with UBLK frontend
+1. UBLK initiator removes dm mapper device
+1. UBLK initiator removes `e.Endpoint`
+1. UBLK initiator delete ublk device associated with the raid1 bdev by calling `UblkStopDisk` with the recorded `UblkInfo.UblkID` info
+2. Instance manager disconnects from the replicas
+1. Longhorn manager delete replica and stops expose them as NVME-TCP target
+
+### Test plan
+
+// TODO
```
innobead, line 384: Have we finished the test plan?

innobead (changes requested):
In general, LGTM. Just a few comments.

enhancements/20250313-ublk-frontend-for-v2-engine.md
PhanLe1010: Thank you. Added

derekbit (approved):
LGTM. Thank you Phan

Issue #8022, opened 2024-02-26
[FEATURE] v2 supports volume expansion


derekbit:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

v2 supports volume expansion as v1 volume

derekbit:
@AaronDewes 
Thanks for your contribution. I created the feature request ticket for tracking the status of the v2 volume expansion and resize. Feel free to discussion and update the status here. Thank you.

cc @innobead @shuo-wu @DamiaSan

AaronDewes:
Thanks!

As I've mentioned before, I have not yet tested this properly yet. I'm planning to run this locally today and make improvements or let you know if I have questions. Because I've only used Longhorn before, but not contributed yet, I'm not familiar with the internal architecture.

innobead:
@AaronDewes as this is an upcoming feature in 1.7.0, so I would expect you can actively work on the feature. Please work with @derekbit who is our maintainer as well to try to make the feature happen eventually. 

Assigned to you first, as you are working on that. However, if you can't make it, just let us know and we will arrange the assignee. Thanks for your contribution!

derekbit:
> Thanks!
> 
> As I've mentioned before, I have not yet tested this properly yet. I'm planning to run this locally today and make improvements or let you know if I have questions. Because I've only used Longhorn before, but not contributed yet, I'm not familiar with the internal architecture.

@AaronDewes
Feel free to ping me on the Slack channel or leave messages here if you have any questions.
BTW, please let us know what docs or materials we can improve for external contributors.
Thank you.

derekbit:
Just found the related PR in spdk side 
https://review.spdk.io/gerrit/c/spdk/spdk/+/22619

cc @AaronDewes @DamiaSan

DamiaSan:
> Just found the related PR in spdk side https://review.spdk.io/gerrit/c/spdk/spdk/+/22619

I think this patch will be merged soon on SPDK master branch.

derekbit:
@DamiaSan Could you collaborate with @AaronDewes to land the implementation in Longhorn? Thanks.

cc @innobead @shuo-wu

DamiaSan:
> @DamiaSan Could you collaborate with @AaronDewes to land the implementation in Longhorn? Thanks.
> 
> cc @innobead @shuo-wu

Yes, sure. @AaronDewes do you already have some questions or something I can do for you?

DamiaSan:
[...]

Issue #978, opened 2019-12-24
Rollback/Recover the volume if the expansion of all replicas failed


shuo-wu:
If all replica expansion failed and the volume became `Faulted` state, we don't know the salvage works fine and there is no way to roll back the expansion.

yasker:
Backend merged. Need instruction for UI.

shuo-wu:
Prerequisites of the API:
1. `volume.controllers[0].isExpanding == false`
2. `volume.controllers[0].size != 0 && volume.controllers[0].size != volume.size`

Need to add tips when users try to call this API:
If the in-progress expansion you want to cancel is triggered by the **PVC size update**, this API won't help revert the PVC. Since the PVC size can not be shrink, users need to clean up then recreate the PVC and PV after the expansion canceling success:
1. Update the field `spec.persistentVolumeReclaimPolicy` to `Retain` for the related PV. Otherwise, the volume will be removed when the PVC/PV is gone.
2. Delete PVC and PV.
3. Recreate PVC and PV via yaml file or Longhorn UI

yasker:
@smallteeths Can you help with the UI? Thanks.

shuo-wu:
The way to prevent the expansion from success so that the expansion will keep retry:
1. Create a volume and attach it.
2. Enter the replica data directory and run the following command for each replica:
```
echo "while true; do if [ ! -e \"volume-snap-expand-4294967296.img.meta.tmp\" ]; then mkdir volume-snap-expand-<PLEASE INPUT THE EXPANSION SIZE>.img.meta.tmp; fi; sleep 1; done;" > test.sh && chmod +x test.sh && ./test.sh &
```
e.g., If I want to make sure the volume`v1` fails to expand to 4Gi, and the only replica of the volume is in `instance-manager-r-894f8e3b`:
```
~ kubectl -n longhorn-system exec -it instance-manager-r-894f8e3b bash
root@instance-manager-r-894f8e3b:/# cd /host/var/lib/longhorn/replicas/v1-d482cb0c/
root@instance-manager-r-894f8e3b:/host/var/lib/longhorn/replicas/v1-d482cb0c# echo "while true; do if [ ! -e \"volume-snap-expand-4294967296.img.meta.tmp\" ]; then mkdir volume-snap-expand-4294967296.img.meta.tmp; fi; sleep 1; done;" > test.sh && chmod +x test.sh && ./test.sh &
[1] 25183
```
3. Detach the volume
[...]

PR #10612, opened 2025-03-19
feat(lep): v2 supports volume expansion

DamiaSan:
Issue https://github.com/longhorn/longhorn/issues/8022

### DIFF
--- a/enhancements/20250319-v2-volume-expansion.md
+++ b/enhancements/20250319-v2-volume-expansion.md
@@ -1,0 +1,111 @@
+# V2 Volume Expansion
+
+## Summary
+
+This enhancement enable the possibility to expand volume's size also for v2 engine.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/8022
+
+## Motivation
+
+### Goals
+
+User will be able to increase the size of a volume created with v2 engine, in the same way it is possible for v1 engine.
+
+### Non-goals
+
+Make an additional step in v2 feature parity with v1.
+
+## Proposal
+
+The work will involve 5 different Longhorn components:
+- `longhorn-instance-manager`, where `V2DataEngineProxyOps` have to implement the `VolumeExpand` proxy operation
+- `longhorn-spdk-engine`, where `Engine` and `Replica` have to implement the expand operation on both client and server side
+- `types`, where new RPCs for the expand operation have to be created
+- `go-spdk-helper`, where we have to change a json parameter's name, as requested by SPDK
+- `SPDK`, where we have to make some changes to the NVMe controller, in order to let the logical volume resize event to be correctly handled by nvmf and nvme layers in SPDK.
+
+What about `RWX` volumes, no change is needed because the `NFS` filesystem resize action is taken at upper level in `longhorn-manager`.
+
+### API changes
+
+In `go-spdk-helper`, the json's name of the `Size` parameter inside `BdevLvolResizeRequest` struct will be changed from `size` to `size_in_mib`.
+
+In `types`, the new proto message `ExpandRequest` will be created and also the following APIs into the `SPDKService` service:
+- `ReplicaVolumeExpand`
+- `EngineVolumeExpand`
+
+`ExpandRequest` contains the name and the (new) size of the volume, it is used as parameter in the new APIs `ReplicaVolumeExpand` and `EngineVolumeExpand`.
+
+In `longhorn-spdk-engine`, the following new APIs will be created:
+- `VolumeExpand` in the `Engine` struct
+- `VolumeExpand` in the `Replica` struct
+
+## Design
+
+### Implementation Overview
+
+Let's see in detail the implementations for `longhorn-spdk-engine` and `SPDK`, the other ones are quite trivial.
+
+#### SPDK
+
+In SPDK, we create a longhorn volume composing different logical volumes, which are the replicas, into a raid1 bdev. The raid1 bdev has the size equal to the minimum size of its base bdevs, so the raid1 can be composed also by base bdevs with different size. If a base bdev is resized, the raid size is increased when this minimum size increases, and so only when all its base bdevs has been correctly resized. When longhorn need to expand a volume, it makes a resize request over all logical volumes that are the replicas of the same longhorn volume. Blob layer, before to resize the blob, freeze I/O and resume it after having completed the operation.
+
+Inside Longhorn we use thinly provisioned logical volumes, i.e. logical volumes that will allocate new space only when it is needed. The resize operation of such a volume is so a simpe `realloc` of the array that contains all the LBA allocated to the volume, and as consequences of this we have:
+- lvol resize don't reserve any new space
+- the time to perform the resize is not affected by the size of the expand operation
+
+The RPC `bdev_lvol_resize` is a blocking call and the only one reason why it can fail is basically lack of memory; and if the SPDK process can't allocate new memory, its functionality is seriously compromised and so the process and maybe the entire node must be restarted.
+
+Actually, when creating a raid1 bdev, as base bdevs we use nvme bdevs created attaching to logical volume exported via nvmf; this is true also for local logical volume, i.e. volumes that reside on the same tgt of the raid1. The simplest solution would has been to add directly the logical volume to the raid, but we have to do otherwise principally to support the live upgrade but also to threat all base bdevs in the same way.
+
+Inside SPDK, the operations for the nvmf exported logical volume and for the attached nvme bdev are executed by the same poller; let's follow the flow when we resize a logical volume:
+- the `bdev_lvol_resize` RPC is called by `go-spdk-helper`
+- the RPC is received by SPDK and lvol/blob layers operate the resize
+- a notification about the bdev block count change is raised
+- nvmf layer handle the notification and send an event over nvmf to notify the attribute change in the namespace, waiting for the reply (the logical volume is added to the nvmf subsystem as a nvme namespace)
+- nvme controller receive the event and start to process it, executing the operations needed to update the attributes of the changed namespace; only after having completed these operations it will send back a response to the event caller
+- these operation send commands synchronously to the other side of the communication channel, but the nvmf layer is waiting for a response and so the resize operation get stuck
+
+The solution is to handle all the operations needed to handle the namespace attribute change in the nvme controller asynchronously.
+The 3 operations that actually are executed synchronously inside the function `nvme_ctrlr_process_async_event`, which handle the namespace attribute change event, are:
+- update the namespace list, by the function `nvme_ctrlr_clear_changed_ns_log`
+- identify active namespaces, by the function `nvme_ctrlr_identify_active_ns`
+- update every active namespace, by the function `nvme_ctrlr_update_namespaces`
+
+The development simply makes the execution of these operations asynchronous, i.e. one function is executed only after the previous one has completed. This completely solve the expand issue.
+
+#### longhorn-spdk-engine
+
+First of all, we must consider that a device mapper I/O suspend is not needed before to start with the expand operation. The reasons are:
+- as mentioned above, SPDK blob layer freezes I/O before to proceed with the resize operation 
+- the raid bdev will have the new size only after all base bdevs has this new size, and so the raid will accept I/O on the added space only at that time.
+
+Using thinly replicas, we also don't need any preliminary check to see if the volume can be expanded.
+
+Basically, when the engine receive a volume expand request, first of all it routes the request to all available replicas of the volume. In v1 engine, the expand operation is considered correctly executed if at least one replica is able to perform the expansion: replicas that return an error are marked as error and so they will be removed by the engine and later resynced with other healthy replicas. But in v2 engine this can't be done because, as mentioned above, raid1 is expanded only when all its base bdevs are correctly expanded. A temporary network interruption, or a memory outage, can make the replica expansion to fail but not to remove it from the raid, so in this case we can't threat the volume as expanded because we should have a different size between engine's size and raid1's size. Then for the v2 engine we must design a different flow: if one replica is not able to resize, the engine expansion stop, rollback all the replicas that have previously been expanded and come back to the caller with an error.
+
+After having correctly expanded all replicas, the engine will perform a device mapper reload to update the volume information inside device mapper volume table.
+
+### Test plan
+
+#### Success case
+
+- Create a storage class with v2 engine
+- Create a PVC that references the created storage class
+- Successfully resize the PVC with a bigger size
+- Check the PVC has the new increased size
+
+#### Failed case
+
+- Create a storage class with v2 engine
+- Create a PVC that references the created storage class
+- Try to resize the PVC with a bigger size and get an error over one replica
+	- If the error is due to a network issue, we can have 2 scenarios
+		- the expand request isn't received from the replica due to e temporary network issue. In this case the replica is still part of the engine, so we can proceed with further step
+		- the replica has gone offline during the expansion, so in this case we must wait for the replica to be removed from the engine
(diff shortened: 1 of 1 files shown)


enhancements/20250319-v2-volume-expansion.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Non-goals
+
+Make an additional step in v2 feature parity with v1.
+
+## Proposal
+
+The work will involve 5 different Longhorn components:
+- `longhorn-instance-manager`, where `V2DataEngineProxyOps` have to implement the `VolumeExpand` proxy operation
+- `longhorn-spdk-engine`, where `Engine` and `Replica` have to implement the expand operation on both client and server side
```
derekbit, line 25: Don't we need to reload the table of the dm device?

enhancements/20250319-v2-volume-expansion.md
```diff
@@ -0,0 +1,71 @@
...
+- `go-spdk-helper`, where we have to change a json parameter's name, as requested by SPDK
+- `SPDK`, where we have to make some changes to the NVMe controller, in order to let the logical volume resize event to be correctly handled by nvmf and nvme layers in SPDK.
+
+### API changes
+
+In `go-spdk-helper`, the json's name of the `Size` parameter inside `BdevLvolResizeRequest` struct will be changed from `size` to `size_in_mib`.
+
+In `types`, the new message `ExpandRequest` will be created and also the following APIs into the `SPDKService` service:
+- `ReplicaVolumeExpand`
+- `EngineVolumeExpand`
```
derekbit, line 36: Add the definition of `ExpandRequest`

enhancements/20250319-v2-volume-expansion.md
```diff
@@ -0,0 +1,71 @@
...
+
+In `longhorn-spdk-engine`, the following new APIs will be created:
+- `VolumeExpand` in the `Engine` struct
+- `VolumeExpand` in the `Replica` struct
+
+## Design
+
+### Implementation Overview
+
+Let's see in detail the work over SPDK, that is the more complex.
```
derekbit, line 46: What will happen if one replica failed to expand but other replicas succeeded?

enhancements/20250319-v2-volume-expansion.md
```diff
@@ -0,0 +1,71 @@
...
+
+Actually, when creating a raid1 bdev, as base bdevs we use nvme bdevs created attaching to nvmf exported logicval volume; this is true also for local logical volume, i.e. volumes that reside on the same tgt of the raid1. The simplest solution would has been to add directly the logical volume to the raid, but we have to do otherwise principally to support the live upgrade but also to threat all base bdevs in the same way.
```
derekbit, line 48: Will we check if all base bdevs can be expanded first? E.g. check the space capacity.

enhancements/20250319-v2-volume-expansion.md
```diff
@@ -0,0 +1,71 @@
...
+
+Inside SPDK, the operations for the nvmf exported logical volume and for the attached nvme bdev are executed by the same poller; let's follow the flow when we resize a logical volume:
+- the `bdev_lvol_resize` RPC is called by `go-spdk-helper`
```
derekbit, line 51: Is the time required to expand a volume affected by the new volume size?
For example, is expanding a volume from 10G to 1T the same in duration as expanding from 10G to 100G?

enhancements/20250319-v2-volume-expansion.md
```diff
@@ -0,0 +1,71 @@
...
+- the RPC is received by SPDK and lvol/blob layers operate the resize
+- a notification about the bdev block count change is raised
+- nvmf layer handle the notification and send an event over nvmf to notify the attribute change in the namespace, waiting for the reply (the logical volume is added to the nvmf subsystem as a nvme namespace)
```
derekbit, line 54: Is the RPC a blocking call?

enhancements/20250319-v2-volume-expansion.md
```diff
@@ -0,0 +1,71 @@
...
+- update every active namespace, by the function `nvme_ctrlr_update_namespaces`
+
+The development simply makes the execution of these operations asynchronous, i.e. one function is executed only after the previous one has completed. This completely solve the expand issue.
+
+### Test plan
+
+- Create a storage class with v2 engine
+- Create a PVC that references the created storage class
+- Resize the PVC with a bigger size
+- Check the PVC has the new increased size
```
derekbit, line 71: Need to add test plan for failed cased.

enhancements/20250319-v2-volume-expansion.md
```diff
@@ -0,0 +1,96 @@
...
+- Successfully resize the PVC with a bigger size
+- Check the PVC has the new increased size
+
+#### Failed case
+
+- Create a storage class with v2 engine
+- Create a PVC that references the created storage class
+- Try to resize the PVC with a bigger size and get an error over one replica (how do we simulate this?)
+- Resolve the problem that cause the error above
+- Retry to resize
```
derekbit, line 95: Can you elaborate how to resolve it in codes (this part should be mentioned in `Design`)? How to handle the unexpanded replica?

enhancements/20250319-v2-volume-expansion.md
derekbit: Unresolved the question. Not addressed in the LEP.

enhancements/20250319-v2-volume-expansion.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Implementation Overview
+
+Let's see in detail the implementations for `longhorn-spdk-engine` and `SPDK`, the other ones are quite trivial.
+
+#### SPDK
+
+In SPDK, we create a longhorn volume composing different logical volumes, which are the replicas, into a raid1 bdev. The raid1 bdev has the size equal to the minimum size of its base bdevs, so the raid1 can be composed also by base bdevs with different size. If a base bdev is resized, the raid size is increased when this minimum size increases, and so only when all its base bdevs has been correctly resized. When longhorn need to expand a volume, it makes a resize request over all logical volumes that are the replicas of the same longhorn volume.
+
+Inside longhorn we use thinly provisioned logical volumes, i.e. logical volumes that will allocate new space only when it is needed. Therefore, the resize operation of such a volume is simply a realloc of the array that contains all the LBA allocated to the volume, and so the time to perform the operation is not affected by the size of the expand operation, which is a blocking call.
```
derekbit, line 54: If it is a blocking call, then this question https://github.com/longhorn/longhorn/pull/10612/files#r2004912485 is important. Because there is a timeout between longhorn-manager and longhorn-instance-manager and between longhorn-instance-manager and longhorn-spdk-engine, how do we make sure the volume expansion can be finished within the timeouts?

derekbit (commented):
The LEP is incomplete for now and mainly for happy path. Should address the failed cases more.

enhancements/20250319-v2-volume-expansion.md
DamiaSan: I don't understand what you mean exactly: should we retry the expand request?
Basically, the operation can fail if we have network problem so that the RPC doesn't arrive to SPDK, and inside SPDK  the operation can fail only if another operation is in progress over the same lvol or in case of insufficent memory.

enhancements/20250319-v2-volume-expansion.md
DamiaSan: The operation simply operate a realloc of an array, it doesn't involve any space request, so we can threat it for example like the snapshot operation, where we don't care for timeouts and we don't have any retry.

enhancements/20250319-v2-volume-expansion.md
DamiaSan: In the LEP i have put the following statement:  _the time to perform the operation is not affected by the size of the expand operation_

enhancements/20250319-v2-volume-expansion.md
derekbit: I mean if the case (one replica is unable to expand because of the insufficient space) happens but other replicas succeed, how to address it?  You only mention `Resolve the problem that cause the error ...`, but how?

enhancements/20250319-v2-volume-expansion.md
derekbit: Okay, then please add this in the doc to explain `the time to perform the operation is not affected by the size of the expand operation`.

enhancements/20250319-v2-volume-expansion.md
```diff
@@ -0,0 +1,96 @@
...
+- `VolumeExpand` in the `Engine` struct
+- `VolumeExpand` in the `Replica` struct
+
+## Design
+
+### Implementation Overview
+
+Let's see in detail the implementations for `longhorn-spdk-engine` and `SPDK`, the other ones are quite trivial.
+
+#### SPDK
```
derekbit, line 50: Do we need to suspend IO while expanding volumes?

enhancements/20250319-v2-volume-expansion.md
DamiaSan: > I mean if the case (one replica is unable to expand because of the insufficient space) happens or somehow failed but other replicas succeed, how to address it? 
> 

I guess it depends on what is the problem, and the only 2 reasons for a volume expand failure is a network problem, so that the RPC doesn't arrive to SPDK, and a memory outage in SPDK's host. The expand operation doesn't allocate any new space on SPDK's host, it simply make a `realloc` on an array.
However I can update the LEP with these 2 cases.

> > should we retry the expand request?
> 
> Not really. Reverting to the original size is also an option, but the doc lacks this kind of error handling. Do you know how v1 volume handles this?

I've had a look at v1 expand operation, but it is more complex than v2 one. In v2 the expand operation is very similar to the snapshot operation, and it's  the snapshot flow that I have taken as example for the expand one.
It is not needed to revert to the original size in case of error, because on the replica where the operation succeed, no new space is allocated.

enhancements/20250319-v2-volume-expansion.md
DamiaSan: Blob layer freezes i/o before to proceed with the resize operation; the raid bdev will accept i/o on the new space only after all base bdevs has the new size.
So it is not needed to suspend I/O, I will update the doc with this info.

derekbit: One more question, should a linear device mapper need to be suspended?

enhancements/20250319-v2-volume-expansion.md
derekbit: > It is not needed to revert to the original size in case of error, because on the replica where the operation succeed, no new space is allocated.

Will you delete the replicas failing to be expanded from the engine and only retain the successful ones?

Can you elaborate more on the cases and what the flow will do?

For a 3-replica volume, it is being expanded
- Case 1: replica 1 and replica 2 succeeded, but replica 3 failed
- Case 2: all replicas failed

derekbit:
Also need to check the RWX volume's expansion. Please address RWX volume in the LEP as well.

enhancements/20250319-v2-volume-expansion.md
DamiaSan: I am going to describe better the flow in `longhorn-spdk-engine` Implementation Overview, so I think that all these doubts will be cleared.

enhancements/20250319-v2-volume-expansion.md
DamiaSan: I will specify this in the LEP.

DamiaSan:
> Also need to check the RWX volume's expansion. Please address RWX volume in the LEP as well.

Ok, thanks for the suggestion.

derekbit:
> > > Also need to check the RWX volume's expansion. Please address RWX volume in the LEP as well.
> > 
> > 
> > Ok, thanks for the suggestion. I will have to study better how RWX volumes works.
> 
> In `longhorn-instance-manager` we have `func (p *Proxy) VolumeExpand` which actually is implemented in `V1DataEngineProxyOps` and not in `V2DataEngineProxyOps`. Following all the flow of `func (ops V1DataEngineProxyOps) VolumeExpand`, I haven´t found any operation over NFS for the RWX volumes, so I guess that all needed operations are made at upper level and then no actions shoudl be taken in `func (ops V2DataEngineProxyOps) VolumeExpand` flow implementation. Am I missing something @derekbit ?

Please check this https://github.com/longhorn/longhorn/blob/master/enhancements/20241111-rwx-resize.md and see if need to extra handling.

BTW, [v1 volume supports offline expansion](https://longhorn.io/docs/1.8.1/nodes-and-volumes/volumes/expansion/). Will the implementation support it?

enhancements/20250319-v2-volume-expansion.md
```diff
@@ -0,0 +1,101 @@
...
+- Create a PVC that references the created storage class
+- Successfully resize the PVC with a bigger size
+- Check the PVC has the new increased size
+
+#### Failed case
+
+- Create a storage class with v2 engine
+- Create a PVC that references the created storage class
+- Try to resize the PVC with a bigger size and get an error over one or more replicas
+	- If the error is due to a network issue, i.e. the expand request isn't received from the replica, solve the issue and restore a regular network traffic
```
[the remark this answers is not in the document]
DamiaSan: No, v1 code doesn't have a specific error handling for this case.
Suppose that for a network issue the expand request for one replica get an error, and so this replica didn't receive the expand request. What should I put in the test plan as description to solve the issue?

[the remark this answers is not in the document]
derekbit: > What should I put in the test plan as description to solve the issue?

Cancellation and reverting to the original size are required. This is what v1 does.

For all replica failed case of  v1 volume, you can check https://github.com/longhorn/longhorn/issues/978.

enhancements/20250319-v2-volume-expansion.md
```diff
@@ -0,0 +1,101 @@
...
+
+#### longhorn-spdk-engine
+
+A device mapper I/O suspend is not needed before to start with the expand operation because:
+- as mentioned above, SPDK blob layer freezes I/O before to proceed with the resize operation 
+- the raid bdev will have the new size only after all base bdevs has this new size, and so the raid will accept I/O on the added space only at that time.
+
+Basically, when the engine receive a volume expand request, first of all it routes the request to all available replicas of the volume. If one replica is not able to perform the expansion and so it return with an error, the engine expansion stop and come back to the caller with an error. The replicas that have already been expanded don't need to be reverted, because thinly replicas don't reserve space; the volume will appear with the new size only when all replicas are correctly expanded. If a further expand request will be done, these replicas that have already been expanded will simply reply success without doing anything. Using thinly replicas, we don't need any preliminary check to see if the volume can be expanded.
```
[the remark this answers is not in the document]
DamiaSan: I don't see any possible issue for this case, however I can insert a rollback to handle partial failures.

Just wont to point out that also the rollback could fail, because there are only 3 possible causes for a replica expand to fail:
- network issue, the replica doesn't receive any expand request
- mismatch between engine's replica map and replica's one, so that when the replica receive the expand request it cannot find any replica with the name present in the expand request
- memory outage in SPDK, the replica can't complete correctly the operation

In all these cases, if the expand fail also the rollback probably will fail, so how should we behave?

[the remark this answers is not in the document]
derekbit: A rollback is necessary to align with v1 volume's behavior.

> In all these cases, if the expand fail also the rollback probably will fail, so how should we behave?

As I mentioned, please check how a v1 volume behaves first and make them consistent. :)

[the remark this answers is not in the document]
DamiaSan: I didn't know the existence of expand cancel and expand rollback operation, I will have a look. Thanks.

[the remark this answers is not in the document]
innobead: Please make the behavior the same as v1. This can ensure the same user experience, especially for feature parity features between v1 and v2.

derekbit:
Replaced by https://github.com/longhorn/longhorn/pull/11373

Issue #9213, opened 2024-08-08
[FEATURE] Allow auto deleting snapshot when a backup is created from that snapshot.


ChanYiLin:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

Currently, Backup can be created in 2 ways
1. On demand: Create Backup through UI (create a Backup CR with API) or create a Backup CR through kubectl
2. RecurringJob

For RecurringJob, we already have a settings to automatically clean up the snapshot which is `Automatically Cleanup Recurring Job Backup Snapshot` [link](https://longhorn.io/docs/1.6.2/references/settings/#automatically-clean-up-system-generated-snapshot)
Job deletes the snapshot before and after the backup createion
- before backup createion: https://github.com/longhorn/longhorn-manager/blob/master/app/recurring_job.go#L330
- after backup createion: https://github.com/longhorn/longhorn-manager/blob/master/app/recurring_job.go#L714 

For on demand backup creation, we could have a setting to automatically clean up the snapshot after the backup is created to improve the space efficiency.

## Describe the solution you'd like

Add a new Label `post-action-delete-snapshot` to Backup as a parameters
If the Label is set, controller will automatically delete the snapshot after the backup is created.

UI modification
1. Add a check box when creating Backup: `auto-delete-snapshot`
2. When it is checked, pass the parameters `auto-delete-snapshot=true` to create backup api

Longhorn will then add the label to the Backup and delete the snapshot after the backup is created.

## Additional context

cc @derekbit

mantissahz:
Hi @ChanYiLin,

what's the difference between this feature and this setting [Auto Cleanup Snapshot When Delete Backup](https://longhorn.io/docs/1.7.2/references/settings/#auto-cleanup-snapshot-when-delete-backup)

ChanYiLin:
Just saw there is a discussion related to this feature
https://github.com/longhorn/longhorn/discussions/5797

cc @mantissahz @derekbit

mantissahz:
Got it.
So it will be like `Auto Cleanup Snapshot After Backup Created`

yangchiu:
@mantissahz Is this feature applicable to v2 volumes?

mantissahz:
>Is this feature applicable to v2 volumes?

@yangchiu Yes, any problems occurred?

yangchiu:
[...]

PR #10649, opened 2025-03-27
feat(backup): introduce the setting 'auto-cleanup-snapshot-after-on-demand-backup-completed'

mantissahz:

Issue # longhorn/longhorn#9213


chart/questions.yaml
```diff
@@ -813,6 +813,13 @@ questions:
         group: Longhorn Default Settings
         type: boolean
         default: false
+      - variable: defaultSettings.autoCleanupSnapshotAfterBackupCompleted
+        label: Auto Cleanup Snapshot When Delete Backup
+        description: >-
+          Setting that automatically cleans up the snapshot after the backup is completed.
```
derekbit, line 819: Is this only for on-demand backup? If yes, we need to mention it in the description.

mantissahz: > Is this only for on-demand backup? If yes, we need to mention it in the description.

Did you mean is this setting used for an individual backup that needs to delete the snapshot when the backup is completed?
If yes, the answer is `No`. This setting is global for all backups created after the setting is modified.

mantissahz: Modified for on-demand backups.

derekbit (approved):
LGTM

Issue #4987, opened 2022-12-03
[IMPROVEMENT] Affinity in the longhorn-ui deployment within the helm chart


atropos112:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)
I am currently using the helm chart of longhorn in my cluster at home. If a node shuts down abruptly (for whatever reason) this sometimes results in the longhorn ui no longer being accessible (because it was running on that node). This is obviously far from ideal given that longhorn forms a critical part of my cluster as such I'd like it to work even if a node goes down. A seemingly obvious solution is to increase replica count of the longhorn ui this however does not solve the problem because there is no antiaffinity on the deployment-ui yaml and hence it is possible that all 3 replicas go on the same node (which for me they did, how lucky am I).

## Describe the solution you'd like
Add overridable affinity in the deployment-ui yaml file that can be overriden via values.yaml file to allow me to fill in podAntiAffinity to ensure no node has two longhorn-ui pods or more on it.

## Describe alternatives you've considered
I have considered copying the helm chart and modifying myself, but this will not update with whatever changes are done upstream so thats no good. I can also use the yaml instead of helm chart but that losses all the benefits of using the helm chart. I can also use kustomize on top of the helm chart but i consider this to be a final resort and do not classify this as a "good solution" rather a desperate one.

chriscchien:
Verified in longhorn master `06c418`
Result **Pass**

Precondition
- Deploy Longhorn by both chart and yaml file on 3 worker nodes cluster, can see 2 lonnghorn UI pod running.

Case 1
1. Deploy Longhorn by both chart and yaml file on 3 worker nodes cluster, can see 2 lonnghorn UI pod running.
2. Power off 1 of those worker node, verify that UI still can work.
3. Power on the node
4. Repeat step 2 ~ 3 on other nodes, verify that UI can be accessed.

Case 2
1. In setting page set `Pod Deletion Policy When Node is Down` to `delete-deployment-pod`
2. Power off the 2 nodes which longhorn UI pods located on.
3. Minutes later new UI pod created and UI can access again.

chriscchien:
Verified in longhorn v1.4.x `404956`, with all v1.4.x images and [steps](https://github.com/longhorn/longhorn/issues/4987#issuecomment-1354290487)
Result were **passed** on both install longhorn by chart and yaml file

PR #10652, opened 2025-03-27
chore(ui): allow specifying affinity in values

lgutter:

Issue #4987

Allow the user to override the affinity in the Longhorn UI deployment. 
In some case it is needed to tweak these settings, and currently the helm chart does not allow this.

The linked Issue is already closed, but the proposed solution in the issue was never addressed, which was to allow overriding of the affinity in the values.

The default output of the helm chart has not changed, which also means no changes to deploy.yaml are needed.

### DIFF
--- a/chart/templates/deployment-ui.yaml
+++ b/chart/templates/deployment-ui.yaml
@@ -55,6 +55,9 @@ spec:
     spec:
       serviceAccountName: longhorn-ui-service-account
       affinity:
+      {{- if .Values.longhornUI.affinity }}
+      {{- toYaml .Values.longhornUI.affinity | nindent 8 }}
+      {{- else }}
         podAntiAffinity:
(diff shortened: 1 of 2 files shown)


chart/templates/deployment-ui.yaml
```diff
@@ -55,6 +55,9 @@ spec:
     spec:
       serviceAccountName: longhorn-ui-service-account
       affinity:
+      {{- if .Values.longhornUI.affinity }}
```
c3y1huang, line 58: Should we have the `if` statement? Perhaps we could just define the default affinity to the value.yaml (it's currently commented out in this PR).

lgutter: Good feedback, I personally agree with defining the defaults in values.yaml. That also allows overwriting a single value without having to redefine the entire affinity block.

Let me know if you want this, I'd be happy to commit that change.

PR #10668, opened 2025-04-01
chore(crd): update crds.yaml and manifests (PR longhorn/longhorn-manager#3505)

derekbit:
It was triggered by longhorn/longhorn-manager#3505.


mantissahz (approved):
LGTM

Issue #10662, opened 2025-03-31
[IMPROVEMENT] Remove the upper bound of v2-data-engine-guaranteed-instance-manager-cpu


derekbit:
### Is your improvement request related to a feature? Please describe (👍 if you like this request)

The v2-data-engine-guaranteed-instance-manager-cpu has an upper limit of [8000 milliCPU](https://github.com/longhorn/longhorn-manager/blob/master/types/setting.go#L1388), but this limit is not essential.

### Describe the solution you'd like

Remove the upper bound.

roger-ryao:
**Verified on v1.9.0-dev-20250406 20250409**

- longhorn `v1.9.0-dev-20250406` https://github.com/longhorn/longhorn/commit/903cc557b96da8a97684dccf4f67f74dc2ee687b

**Test Steps**
https://github.com/longhorn/longhorn/issues/10662#issuecomment-2765340619

**Test Passed**

We can set the value of the setting `v2-data-engine-guaranteed-instance-manager-cpu` to a value larger than 8000.

PR #10672, opened 2025-04-01
feat(setting): remove ValueIntRangeMaximum of v2-data-engine-guaranteed-instance-manager-cpu

derekbit:

Issue Longhorn/longhorn#10662


mantissahz (approved):
LGTM

Issue #6764, opened 2023-09-22
[FEATURE] Cleanup orphaned volume runtime resources if the resources already deleted


innobead:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

https://github.com/harvester/harvester/issues/4517#issue-1892287382 it seems there would be a chance to have an orphaned volume runtime resources even though the corresponding resource has been deleted. 

Resources include: engine or replica process, orphaned tgt process and iscsi session. 

## Describe the solution you'd like

Leverage the orphan controller, and automatically delete these orphaned processes. 

## Additional context

cc @Vicente-Cheng @ejweber

PhanLe1010:
I think we also need to cleanup the orphan tgt and iscsi session as mentioned at https://github.com/longhorn/longhorn/issues/7386#issuecomment-2195984587 . WDYT?

derekbit:
> I think we also need to cleanup the orphan tgt and iscsi session as mentioned at [#7386 (comment)](https://github.com/longhorn/longhorn/issues/7386#issuecomment-2195984587) . WDYT?

Do you mean cleaning up iscsi sessions with wrong ip address while starting an instance-manager pod?

PhanLe1010:
> > I think we also need to cleanup the orphan tgt and iscsi session as mentioned at [#7386 (comment)](https://github.com/longhorn/longhorn/issues/7386#issuecomment-2195984587) . WDYT?
> 
> Do you mean cleaning up iscsi sessions with wrong ip address while starting an instance-manager pod?

I am thinking about a routinely cleaning like: longhorn-manager periodically sends the cleaning request to IM with the data about existing IM IP. IM can clean up the session with non-existing IP. Similar to how we would cleanup orphan engine/replica process. 

IM itself will not know whether an IP is no longer valid. WDYT?

derekbit:
> > > I think we also need to cleanup the orphan tgt and iscsi session as mentioned at [#7386 (comment)](https://github.com/longhorn/longhorn/issues/7386#issuecomment-2195984587) . WDYT?
> > 
> > 
> > Do you mean cleaning up iscsi sessions with wrong ip address while starting an instance-manager pod?
> 
> I am thinking about a routinely cleaning like: longhorn-manager periodically sends the cleaning request to IM with the data about existing IM IP. IM can clean up the session with non-existing IP. Similar to how we would cleanup orphan engine/replica process.
> 
> IM itself will not know whether an IP is no longer valid. WDYT?

[...]

PR #10675, opened 2025-04-02
docs(lep): orphaned engine and replica runtime instance cleanup

COLDTURNIP:

Issue #6764 

When a node disconnected or suspended then get back to the cluster, the existed living processes are orphaned. We need a way to identify and clean up these processes.

### DIFF
--- a/enhancements/20250331-orphaned-runtime-cleanup.md
+++ b/enhancements/20250331-orphaned-runtime-cleanup.md
@@ -1,0 +1,227 @@
+# Orphaned Engine And Replica Runtime Instance Cleanup
+
+## Summary
+
+Orphaned runtime instance cleanup identifies unmanaged engine and replica runtime instances on the nodes, and provides a list of the orphan instances on each node. Longhorn by default keeps the instance. Also, it provides a configurable way to clean up such resources automatically.
+
+### Related Issues
+
+[https://github.com/longhorn/longhorn/issues/6764](https://github.com/longhorn/longhorn/issues/6764)
+
+## Motivation
+
+### Goals
+
+- Identify the orphaned engine and replica processes
+- The scanning process should not stick to the reconciliation of the controller
+- Provide the user a way to select and trigger the deletion of the orphaned runtime instance
+- Support the global auto-deletion of orphaned runtime
+- Compatible with existing cluster, and is able to clean up the existing orphaned runtime in the acceptable previous versions.
+
+### Non-goals
+
+- Clean up orphaned data stores on the nodes
+- Support the per-node auto-deletion of orphaned runtime
+- Support the auto-deletion of orphaned runtime exceeding the TTL
+- Prevent the process deletion race between CRs (see the design note below)
+
+## Proposal
+
+1. Introduce new types of existing CRD `orphan` that represents and tracks the orphaned runtime instances. This new type of orphan CR is reconciled by existing orphan controller. The orphan controller deletes the runtime instance if receives a deletion request. The orphan CR reflects the existence of the runtime instance.
+
+2. The monitor on each instance manager controller keeps watching the runtime instances managed by instance manager, compares them with the scheduled engine and replica, and then finds the orphaned runtime instances. Once a runtime instance is considered to be orphaned, create a corresponding orphan CR, and keep update the instance state on this orphan CR. The monitor will delete the orphan CR when an instance disappears from the node, or, the corresponding engine/replica is scheduled back.
+
+3. The node controller keeps tracking the node status. When a node is evicted or disconnects from the cluster, node controller initiates the deletion on orphan CRs belongs to this node.
+
+### User Stories
+
+When a network outage occurs to some Longhorn node, it may contain multiple engine or replica runtime instances not tracked by the Longhorn System. The corresponding engine and replica CRs may be removed or relocated to another node during the outage. When the node comes back, the corresponding runtime resources are no longer tracked by the Longhorn system. These runtime resources, including the processes, are called orphaned.
+
+Orphaned runtime resources continue to consume CPU and memory. Users have no way to clean up such resources except forcibly restarting the entire instance manager pod on the node, which will unnecessarily increase the chance of a system outage.
+
+After the enhancement, Longhorn automatically finds out the orphaned runtime instances on Longhorn nodes. Users can visualize and manage the orphaned replica runtime instances via Longhorn GUI or command line tools. Additionally, Longhorn can delete the orphaned runtime resources automatically if users enable the global auto-deletion option.
+
+### User Experience In Detail
+
+- Via Longhorn GUI
+    - Users can check instance status and see if Longhorn already identifies orphaned engines/replicas.
+    - Users can choose the items in the orphaned replica directory list and clean up them.
+    - Users can enable the global auto-deletion on setting page. By default, the auto-deletion is disabled.
+
+- Via `kubectl`
+    - Users can list the orphaned runtime instances directories by `kubectl -n longhorn-system get orphans`.
+    - Users can delete the orphaned runtime instances directories by `kubectl -n longhorn-system delete orphan <name>`.
+    - Users can enable or disable the global auto-deletion by `kubectl -n longhorn-system edit settings orphan-engine-instance-auto-deletion`
+    - Users can enable or disable the global auto-deletion by `kubectl -n longhorn-system edit settings orphan-replica-instance-auto-deletion`
+
+## Design
+
+### Implementation Overview
+
+- Orphan CRs track the liveness of corresponding orphaned runtime instance
+- Orphan CRs are created by instance manager monitor
+- Orphan CRs are deleted by
+  - Longhorn node controller: when node is deleted or evicted
+  - Instance manager monitor:
+    - When runtime instance disappear, delete the corresponding orphan runtime instance CR.
+    - When auto deletion is enabled, delete all orphan runtime instance CRs on this node.
+    - When runtime instance is rescheduled back to the node, delete the corresponding orphan runtime instance CR since it is no longer orphaned.
+
+**Settings**
+
+  - Add setting `orphan-resource-auto-deletion`.
+    - This is a string of semicolon-seperated list. Possible items:
+      - `replicaData` to enable auto deletion on replica data store. This replaces the old setting `orphan-auto-deletion`.
+      - `engineInstance` to enable auto deletion on engine runtime instance.
+      - `replicaInstance` to enable auto deletion on replica runtime instance.
+    - Default value is empty, which means to disable auto deletion for all kinds of orphaned resources.
+    - While upgrade Longhorn to v1.9.0, default set to `replicaData` if old setting `orphan-auto-deletion` is enabled.
+  - Delete `orphan-auto-deletion`
+
+**Instance manager controller**
+
+  - Start the instance monitor during initialization.
+  - Reconciles the settings events and orphan events.
+    - When auto deletion enabled, delete the orphan runtime instance CRs on this node.
+    - When the instance manager is no longer running, delete the orphan runtime instance CRs on this node.
+    - Delete the exist orphan runtime instance CRs if the runtime instance disappeared from instance manager.
+    - Delete the exist orphan runtime instance CRs if the corresponding engine/replica is scheduled back to this node.
+
+**Instance manager monitor**
+
+  - Establishes a GRPC stream with instance manager to watch the status of runtime instances.
+  - Receives engine instance status update events from GRPC stream
+  - Update runtime instance status on the list in instance manager CR's `status.instanceEngines` and `status.instanceReplicas`
+  - Compare the known engine/replica runtime instances with corresponding engine/replica CRs:
+    - If there's no such CR, this instance is considered to be orphaned.
+    - If there's a corresponding CR:
+      - If the `status.currentState` is different from `spec.desireState`, ignore it because the state and ownership may change.
+      - If the corresponding CR is in running state:
+        - If the owner ID is not node ID, ignore it.
+        - The instance is considered to be orphaned if and only if the node ID is not the current controller ID.
+      - If the corresponding CR is in stopped state:
+        - The instance is considered to be orphaned if and only if the node ID is not the current controller ID.
+      - To other corresponding CR states, ignore it because of unstable state.
+  - Compare the orphaned engine/replica runtime instances with the exist orphan CRs
+    - Create an orphan CR for each orphaned instance if missing.
+
+  ```
+                     ┌────────────────────┐
+                     │                    │
+                     │ an engine/replica  │
+                     │ instance listed by │
+                     │  instance manager  │
+                     │                    │
+                     └──────────┬─────────┘
+                                │
+                      ┌─────────▼─────────┐
+                      │                   │
+                      │    compare the    │
+                      │   corresponding   │
+                      │ engine/replica CR │
+                      │                   │
+                      └─────────┬─────────┘
+                                │
+                          ┌─────▼─────┐
+                          │           │ not exist
+                          │ CR exist? ├──────────────────────────────┐
+                          │           │                              │
+                          └─────┬─────┘                              │
+                                │ exist                              │
+   no, CR state ┌───────────────▼─────────────────────────┐          │
+   will change  │                                         │          │
+      ┌─────────┤ status.currentState == spec.desireState │          │
+      │         │                                         │          │
+      │         └───────────────┬─────────────────────────┘          │
+      │                         │ yes, stable state                  │
+      │                 ┌───────▼────────┐                           │
+      │                 │                │                           │
+      │                 │ currentState ? │                           │
+      │                 │                │                           │
+      │                 └───────┬────────┘                           │
+      │                         │                                    │
+      │          ┌──────────────┼──────────────┐                     │
+      │          │              │              │                     │
+      │    ┌─────▼────┐    ┌────▼────┐    ┌────▼────┐                │
+      │    │  other   │    │         │    │         │                │
+      │    │(starting,│    │ running │    │ stopped │                │
+      │    │ stopping,│    │         │    │         │                │
+      │    │ error   )│    └────┬────┘    └────┬────┘                │
+      │    └─┬────────┘         │              │                     │
+      │      │                  │              │                     │
+  ┌───▼──────▼──┐       ┌───────▼────────┐     │                     │
+  │             │       │                │     │                     │
+  │ no decision │    no │  CR owner ==   │     │                     │
+  │             ◄───────┤ desired node ? │     │                     │
+  └──────▲──────┘       │                │     │                     │
+         │              └───────┬────────┘     │                     │
+         │                  yes │              │                     │
+         │                ┌─────▼──────────────▼─────┐    ┌──────────▼───────────┐
+         │                │                          │    │                      │
+         │                │ CR node == current node? │    │ instance is orphaned │
+         │                │                          │    │                      │
+         │                └─────┬──────────────┬─────┘    └──────────▲───────────┘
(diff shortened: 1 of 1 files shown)


enhancements/20250331-orphaned-runtime-cleanup.md
```diff
@@ -0,0 +1,215 @@
...
+  - Reconciles the settings events and orphan events.
+    - When auto deletion enabled, delete the orphan runtime instance objects on this node.
+
+**Instance manager monitor**
+
+  - Establishes a GRPC stream with instance manager to watch the status of runtime instances.
+  - Receives engine instance status update events from GRPC stream
+  - Update runtime instance status on the list in instance manager object's `status.instanceEngines` and `status.instanceReplicas`
+  - Compare the known engine/replica runtime instances with corresponding engine/replica objects:
+    - If there's no such object, mark this instance as orphaned.
```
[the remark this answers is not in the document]
COLDTURNIP: It is actually marked on Kubernetes object. In this step, we filter the orphaned instances. Will reword it:

> If there's no such object, this instance is considered to be orphaned.

enhancements/20250331-orphaned-runtime-cleanup.md
```diff
@@ -0,0 +1,215 @@
...
+            │              └───────┬──────────────┬───────┘  └──────────▲───────────┘
+            └──────────────────────┘              └─────────────────────┘            
+               object on this node                  object on other node             
+  ```
+
+**Orphan object**
+
+  - The object ID is calculated from engine/replica ID.
+    - To engine instance: `orphan-engine-${checksum}`
+    - To replica instance: `orphan-replica-${checksum}`
```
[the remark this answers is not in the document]
COLDTURNIP: I'd like to know the reason of appending a random string. Wouldn't it more convenient to a determined name that can be calculated from the name of instance?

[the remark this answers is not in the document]
derekbit: > Wouldn't it more convenient to a determined name that can be calculated from the name of instance?

The `random string` can be calculated from the name of the instance as well.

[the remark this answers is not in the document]
COLDTURNIP: How about 
```suggestion
    - To engine instance: `orphan-${engine-name}-${checksum}`
    - To replica instance: `orphan-${replica-name}-${checksum}`
```

enhancements/20250331-orphaned-runtime-cleanup.md
```diff
@@ -0,0 +1,215 @@
...
+      - If the current state is not the desired state, ignore it.
+      - If the object is in running state:
+        - If the owner ID is not node ID, ignore it.
+        - Mark the instance orphaned if and only if the node ID is not the current controller ID.
+      - If the object is in stopped state:
+        - Mark the instance orphaned if and only if the node ID is not the current controller ID.
+      - To other object states, ignore it because of unstable state.
+  - Compare the orphaned engine/replica runtime instances with the exist orphan objects
+    - Create a corresponding orphan object for each orphaned instance if missing.
+    - Delete the exist orphan runtime instance objects if the runtime instance disappeared from instance manager
```
c3y1huang, line 93: I find the use of the word `object` in this doc a bit confusing - are you referring to Kubernetes Custom Resources (CRs), or a cached in-memory object used by the controllers? Consider using CR when referring to Kubernetes Custom Resources.

enhancements/20250331-orphaned-runtime-cleanup.md
```diff
@@ -0,0 +1,215 @@
...
+  - Record the data engine type in `Parameters["dataEngine"]`.
+
+**Orphan controller**
+
+  Reconciles the orphan events.
+
+  - If the `deletionTimestamp` is non-zero, and the current controller is responsible for this orphan object, proceed with deletion request.
+    - If the orphan's node ID is the current controller ID:
+      - Check with the corresponding engine or replica object before deleting the runtime instance.
+        - if the engine / replica object not exists, ask instance manager to delete the runtime instance directly.
```
c3y1huang, line 183: > ask instance manager to delete the runtime instance directly

Could you clarify how this request is made or triggered?

enhancements/20250331-orphaned-runtime-cleanup.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Implementation Overview
+
+- Orphan objects track the liveness of corresponding orphaned runtime instance
+- Orphan objects are created by instance manager monitor
+- Orphan objects are deleted by
+  - instance manager controller: when auto deletion is enabled
+  - engine and replica controller: when runtime instance is rescheduled back to the node
+  - Longhorn node controller: when node is deleted or evicted
+  - Instance manager monitor: when runtime instance disappear
```
c3y1huang, line 63: I feel we should minimize the handling of orphaned engine/replica deletions across multiple controllers to avoid potential conflicts. Is it possible to limit this responsibility to the orphan controller only?

enhancements/20250331-orphaned-runtime-cleanup.md
```diff
@@ -0,0 +1,215 @@
...
+    - If there's no such object, mark this instance as orphaned.
+    - If there's an object:
+      - If the current state is not the desired state, ignore it.
+      - If the object is in running state:
+        - If the owner ID is not node ID, ignore it.
+        - Mark the instance orphaned if and only if the node ID is not the current controller ID.
+      - If the object is in stopped state:
+        - Mark the instance orphaned if and only if the node ID is not the current controller ID.
+      - To other object states, ignore it because of unstable state.
+  - Compare the orphaned engine/replica runtime instances with the exist orphan objects
```
c3y1huang, line 91: Is there a possibility of false identification during orphan detection when the network is not stable?

enhancements/20250331-orphaned-runtime-cleanup.md
COLDTURNIP: Orphan controller first fetch the instance manager CR, and try to create a client of instance manager (`engineapi.NewInstanceManagerClient`). With the information recorded in orphan CR, including the instance name and the data engine type, orphan controller can invoke `instanceManagerClient.InstanceDelete` with empty `diskUUID` and `cleanupRequired` disabled.

enhancements/20250331-orphaned-runtime-cleanup.md
COLDTURNIP: Since the instance monitor has both the watching stream and a 60sec polling period, it is possible to merge the setting and engine/replica reschedule trigger into instance manager monitor:

1. Orphan instance judgement, then
2. Create new orphan objects, then
3. Delete existed or created objects because of
    1. auto deletion enabled
    2. engine or replica rescheduled back to this node
    3. runtime instance disappear

And keep the deletion of node evicted/deleted in node controller, just the same place we delete orphaned data store. WDYT?

enhancements/20250331-orphaned-runtime-cleanup.md
COLDTURNIP: Confirmed that engine/replica instances will be recovered if it died or be deleted accidentally. So the problem is to prevent keeping deleting instances on orphan controller. Once the instance disappeared, the instance manager monitor should also update the instance state on corresponding orphan CR, hence the orphan controller is able to remove the finalizer directly.

enhancements/20250331-orphaned-runtime-cleanup.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Non-goals
+
+- Clean up orphaned data stores on the nodes
+- Support the per-node auto-deletion of orphaned runtime
+- Support the auto-deletion of orphaned runtime exceeding the TTL
+
+## Proposal
+
+1. Introduce some new type of existing CRD `orphan` and controller that represents and tracks the orphaned runtime instances. The controller deletes the runtime instance if receives a deletion request. The orphan CR reflects the existence of the runtime instance.
```
derekbit, line 28: Do we introduce a new controller in the implementation?

derekbit, line 28:
```suggestion
1. Introduce new types of existing CRD `orphan` and controller that represents and tracks the orphaned runtime instances. The controller deletes the runtime instance if receives a deletion request. The orphan CR reflects the existence of the runtime instance.
```

derekbit, line 28: > The controller deletes the runtime instance if receives a deletion request.

Which controller?

enhancements/20250331-orphaned-runtime-cleanup.md
```diff
@@ -0,0 +1,214 @@
...
+  - Longhorn node controller: when node is deleted or evicted
+  - Instance manager monitor:
+    - When runtime instance disappear, delete the corresponding orphan runtime instance CR.
+    - When auto deletion is enabled, delete all orphan runtime instance CRs on this node.
+    - When runtime instance is rescheduled back to the node, delete the corresponding orphan runtime instance CR since it is no longer orphaned.
+
+**Settings**
+
+  - Add setting `orphan-engine-instance-auto-deletion`. Default value is `false`.
+  - Add setting `orphan-replica-instance-auto-deletion`. Default value is `false`.
```
c3y1huang, line 69: NIT: Consider consolidating this into a single setting `orphan-instance-auto-deletion`, which accepts a list with possible values of `engine, replica`. If no values are provided, the behavior is no-op.

enhancements/20250331-orphaned-runtime-cleanup.md
COLDTURNIP: No, we're going to reuse existing orphan controller. Sorry for misleading and will refine it.

COLDTURNIP: The orphan controller will take the responsibility to ask instance manager to delete the instance process. Will reword this part.

enhancements/20250331-orphaned-runtime-cleanup.md
COLDTURNIP: > If no values are provided, the behavior is no-op.

Do you mean not perform auto deletion when the value is empty/invalid?

c3y1huang: If we decided to go with the list approach, input should be restricted to valid values, limiting the acceptable inputs to `engine` and `replica`. When the list is empty, Longhorn should behave as same as when both `orphan-engine-instance-auto-deletion` and `orphan-replica-instance-auto-deletion` are set to `false`.

derekbit: @c3y1huang Sorry, I'm a bit confused. What is the list approach?

enhancements/20250331-orphaned-runtime-cleanup.md
```diff
@@ -0,0 +1,214 @@
...
+
+[https://github.com/longhorn/longhorn/issues/6764](https://github.com/longhorn/longhorn/issues/6764)
+
+## Motivation
+
+### Goals
+
+- Identify the orphaned engine and replica processes
+- The scanning process should not stick to the reconciliation of the controller
+- Provide the user a way to select and trigger the deletion of the orphaned runtime
```
derekbit, line 17: For wording alignment. Please apply to other related text as well.
```suggestion
- Provide the user a way to select and trigger the deletion of the orphaned runtime instance
```

enhancements/20250331-orphaned-runtime-cleanup.md
```diff
@@ -0,0 +1,214 @@
...
+**Integration tests**
+
+- `orphan` CRs will be created correctly while the process instance created unexpectedly. And they can be cleaned up without touching the data store.
+- `orphan` CRs will be created correctly to a re-connected node. And they can be cleaned up without touching the data store.
+- `orphan` CRs will be removed when an engine / replica is relocated back to the node.
+- `orphan` CRs will be removed when the node is evicted or down.
+- Auto-deletion setting.
+
+
+## Note[optional]
```
derekbit, line 214: Add Orphan CR LEP here. Readers can understand the context more.

enhancements/20250331-orphaned-runtime-cleanup.md
c3y1huang: I am thinking if we should consider using a single setting `orphan-instance-auto-deletion`, which takes a list as value, instead of having 2 separate settings `orphan-engine-instance-auto-deletion` and `orphan-replica-instance-auto-deletion` that take boolean as value. WDYT?

enhancements/20250331-orphaned-runtime-cleanup.md
```diff
@@ -0,0 +1,214 @@
...
+      - Check with the corresponding engine or replica CR before deleting the runtime instance.
+        - If the engine / replica CR not exists, this runtime instance is deletable.
+        - If the engine / replica CR exists, and the node ID is not the current controller ID, then the runtime instance is deletable.
+        - Once the runtime instance is deletable, create an instance manager client from the instance manager CR, and delete the runtime instance.
+      - If the instance disappears from the instance manager, remove the finalizer to delete the orphan CR.
+    - Otherwise, the orphan runtime instance lives on another node, but the controller loses its ownership because of disconnection. Remove the finalizer to complete deletion.
+
+**Engine and replica controller**
+
+  - Once an engine or replica CR's node ID is assigned, delete the corresponding orphan runtime instance CR if exists on assigned node.
```
derekbit, line 189: Can you elaborate more on the purpose?

enhancements/20250331-orphaned-runtime-cleanup.md
derekbit: After discussing with @c3y1huang, we have a setting `orphan-auto-deletion` that takes boolean as a value and only handles the orphaned replica directories.

We can deprecate the `orphan-auto-deletion` setting and introduce a new setting, `orphan-resource-auto-deletion`, which supports values like `replica` (replica directory), `engine instance`, and `replica instance` (you can adjust the naming). This approach can avoid introducing additional settings in the future.

WDYT @COLDTURNIP

enhancements/20250331-orphaned-runtime-cleanup.md
derekbit: > We try to delete the engine on node-1
> Before we finished delete the engine, engine.Spec.NodeID quickly move back to node-1

When the engine moves back to node-1 and before a new engine process is created, will the old engine process be deleted? Or will both the new and old engine processes coexist in the instance manager pod?

enhancements/20250331-orphaned-runtime-cleanup.md
COLDTURNIP: A list here would be more flexible. To this approach, I suggest to use a comma-separate string, and accept the following three possible items:
- `replicaData` for replica data store
- `engineInstance` for engine runtime instance
- `replicaInstance` for replica runtime instance

derekbit: > A list here would be more flexible. To this approach, I suggest to use a comma-separate string, and accept the following three possible items:
> 
> * `replicaData` for replica data store
> * `engineInstance` for engine runtime instance
> * `replicaInstance` for replica runtime instance

LGTM.

enhancements/20250331-orphaned-runtime-cleanup.md
COLDTURNIP: Sure. Once an engine or replica CR is scheduled to a node, and there's an existing instance process on this node for this engine/replica CR, then this instance is no longer considered to be orphaned. If there's an existing orphan CR for this instance, the orphan CR should be deleted without removing the instance process. This part should be judged in the instance monitor instead of engine/replica controller. Will refine in the LEP.

enhancements/20250331-orphaned-runtime-cleanup.md
c3y1huang: > To this approach, I suggest to use a comma-separate string

Other settings use a semicolon as the separator; consider doing the same for consistency.

enhancements/20250331-orphaned-runtime-cleanup.md
COLDTURNIP: > When the engine moves back to node-1 and before a new engine process is created, will the old engine process be deleted? Or will both the new and old engine processes coexist in the instance manager pod?

To our current v1.8.1 implementation, when a engine is scheduled back to the node, the old engine process will be killed first, then launch another engine process.

enhancements/20250331-orphaned-runtime-cleanup.md
```diff
@@ -0,0 +1,228 @@
...
+### Implementation Overview
+
+- Orphan CRs track the liveness of corresponding orphaned runtime instance
+- Orphan CRs are created by instance manager monitor
+- Orphan CRs are deleted by
+  - Longhorn node controller: when node is deleted or evicted
+  - Instance manager monitor:
+    - When runtime instance disappear, delete the corresponding orphan runtime instance CR.
+    - When auto deletion is enabled, delete all orphan runtime instance CRs on this node.
+    - When runtime instance is rescheduled back to the node, delete the corresponding orphan runtime instance CR since it is no longer orphaned.
```
WebberHuang1118, line 68: The presence of an orphan engine instance can block node draining, which in turn causes the Harvester upgrade process to get stuck in [Phase 4](https://docs.harvesterhci.io/v1.5/upgrade/troubleshooting#phase-4-upgrade-nodes). We've [encountered this issue multiple times](https://docs.harvesterhci.io/v1.5/upgrade/v1-4-1-to-v1-4-2#1-upgrade-is-stuck-in-the-pre-drained-state).

With this enhancement, will the orphan engine be automatically cleaned up as part of the node draining process?

WebberHuang1118: No, Harvester does not perform node eviction. Instead, it relies on Longhorn volume live migration.

> When an engine is rescheduled and left the node, instance monitor must be able to identify & delete the orphaned engine process in this design.

The IM handles this in real time? For example, if node draining is blocked by an orphan engine, how long does it usually take for the Longhorn manager to detect and remove the orphan engine?

Also, Harvester performs node draining sequentially—one node at a time during the upgrade process.

enhancements/20250331-orphaned-runtime-cleanup.md
derekbit: Because the race condition caused by engine deletion and engine creation occurs on the same node, I'm thinking if we can introduce a per-node read-write lock for 
- read lock: for engine creation 
- write lock: for engine deletion

During the write lock protection, controller can check whether the engine process is really deletable.

enhancements/20250331-orphaned-runtime-cleanup.md
COLDTURNIP: Once the volume is attaching to the new node, the IM handles process deletion as soon as the engine controller receives the old engine deletion, and the IM monitor (mentioned in this LEP) will also receives the process status changing event immediately from the GRPC event stream. Also, IM monitor has a process status polling period every minute, hence the process exiting will be detected withing 60sec in worst case. In this LEP, the orphan judgement occurs each time when the IM monitor receives and polls the status of node processes.

WebberHuang1118 (approved):
LGTM, thanks.

enhancements/20250331-orphaned-runtime-cleanup.md
COLDTURNIP: It would be technically possible to implement a lock between engine/replica and orphan controllers. Let me clarify the detail.

* while initiating a deletion, engine or orphan controller acquire the write lock, and will release after instance manager confirmed deletion completed
* while initiating a creation, engine controller acquire the read lock, and will release after instance manager confirmed creation completed
* the lock and unlock events should be recorded
* if a controller failed to acquire the lock, raise an conflict error to re-enqueue the event again

derekbit (approved):

PR #10850, opened 2025-05-06
chore(workflow): update update-community-review-sprint-issue-status.yaml

derekbit:

Issue #


.github/workflows/update-issue.yaml
```diff
@@ -29,6 +29,51 @@ jobs:
...
+        env.field-values != ''
+        }}
+      uses: titoportas/update-project-fields@v0.1.0
+      with:
+        project-url: https://github.com/orgs/longhorn/projects/8
+        github-token: ${{ secrets.CUSTOM_GITHUB_TOKEN }}
+        item-id: ${{ steps.add-project.outputs.itemId }}
+        field-keys: Status
+        field-values: New Issues
+
```
Yu-Jack, line 76: Does this case move the issue to New Issues regardless of its current status? For example, an issue is in Implement status, does this move it to New Issues when event is triggred?

derekbit: Good catch! I separated the original workflow into two files. update-longhorn-issue.yml only works while opening and reopening a ticket.

.github/workflows/update-longhorn-issue.yml
```diff
@@ -0,0 +1,84 @@
+name: "[Issue Management] Update Longhorn Issue Status, Sprint and Milestone"
+on:
+  issues:
+    types: [opened, reopened]
```
Yu-Jack, line 4: Be careful, if you use opened to detect, it might have time difference issue.

For example, you open an issue without milestone, then you decide to add milestone in 5 seconds. The Github Action could change it back to Backlog because the GA takes some time to start up and it might get the empty milestone at that point.

derekbit: Any thought to mitigate the timing issue?

Yu-Jack: I moved the most of logic to another workflow like this https://github.com/harvester/harvester/blob/master/.github/workflows/issue-management-update-harvester-projects.yaml

derekbit: Fixed. Thank you.

Issue #7794, opened 2024-01-26
[FEATURE] V2 Volume Supports Cloning


draghuram:
Starting from Longhorn 1.3, when "type: snap" is configured in volume snapshot class, CSI snapshot requests (creation of VolumeSnapshot resource) results in a point in time Longhorn snapshot. 

However, if a PVC is created with CSI snapshot as data source, a new Longhorn volume is created and data is copied to it from the snapshot. Because of the copy, PVC is not ready immediately and when it becomes ready depends on the amount of data being copied. More over, the new PVC will by default have same number of replicas as the source volume but this can be fixed by using a different storage class that sets number of replicas to 1. While the data copy is fine in many cases, there is one particular use case where it is problematic.

Backup applications (such as CloudCasa and Velero) typically create a PVC snapshot and then backup its content. To read data from snapshot, they create a temporary PVC from the snapshot and read from it. Once backup is done, the temporary PVC is deleted. In this particular case, it is inefficient to have data copy take place when PVC is created from snapshot. Ideally, new volume should be created from snapshot with  no data copy as some storage systems do. If the new volume allows writes, modified  blocks should be saved separately while reads for other blocks can be fulfilled from the snapshot.

I don't know Longhorn internals  enough to tell whether this feature is feasible but I am opening this request after discussions with @PhanLe1010 in https://github.com/vmware-tanzu/velero/issues/6179#issuecomment-1882187472.

innobead:
This could be a topic for v2 as well if we can share the immutable snapshot by reference in different volumes. cc @derekbit @shuo-wu @DamiaSan

DamiaSan:
Instead of creating a new lvol and then copying snapshot's data into it, we could create a new clone of the snapshot.

PhanLe1010:
TODO: 

Investigate if CDI flow + Longhorn CSI snapshot/fast cloning can have achieve similar goals as the Longhorn Backing Image flow

PhanLe1010:
> Instead of creating a new lvol and then copying snapshot's data into it, we could create a new clone of the snapshot.

Thanks @DamiaSan  Looks like it is easier to be done in v2 data engine. Cool

derekbit:
In favor of https://github.com/longhorn/longhorn/issues/9106

ChanYiLin:
Hi @DamiaSan 
Suppose I have a volume with snapshot chain
“snap1->snap2->replica head”
When I clone from the volume, I will get
[...]

PR #10873, opened 2025-05-08
feat: LEP for v2 volume clone

PhanLe1010:
longhorn/longhorn#7794


enhancements/20250508-v2-volume-cloning.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Fast v2 cloning
+
+Requirements and assumptions:
+1. There should be zero data copy in this case and clone should be instant
+1. The data is depended on the source volume
+1. The newly target volume can only have 1 replica. 
+   And the replica of the target volume must be on same disk as one of the replica of the source volume
+1. Replica cannot move to new node and cannot be rebuilt 
+1. All restriction can be removed if the user decouple the volume (see [Decoupled a fast v2 cloned volume](#decoupled-a-fast-v2-cloned-volume) )
```
DamiaSan, line 110: If you use external snapshot, you can remove these restrictions as well. With external snapshot, you can create a clone of a snapshot on another logical volume store or another spdk tgt.

enhancements/20250508-v2-volume-cloning.md
```diff
@@ -0,0 +1,166 @@
...
+Flow:
+1. User create a new volume with:
+    ```yaml
+    volumeSpec:
+      dataSource: "snap://source-volume/source-snapshot-1"
+      cloneMode: "fast"
+    ```
+1. Longhorn manager create a new volume with 1 replica and schedule the replica to the node which has a healthy replica of the source volume
+1. Longhorn manager send the clone GRPC to the replica server
+1. Replica server then call SPDK API `bdev_lvol_snapshot` to create new snapshot `source-snapshot-1-clone-entry-point` from source snapshot `source-snapshot-1`
```
DamiaSan, line 123: Why do you need `source-snapshot-1-clone-entry-point`? Can't you create the clone directly from `source-snapshot-1`?

enhancements/20250508-v2-volume-cloning.md
```diff
@@ -0,0 +1,166 @@
...
+Continue from the flow in the section [Fast v2 cloning](#fast-v2-cloning) above
+1. User update the volume with:
+    ```yaml
+    volumeSpec:
+      dataSource: "snap://source-volume/source-snapshot-1"
+      cloneMode: "decoupled"
+    ```
+1. Longhorn manager send the gRPC to the replica server
+1. Replica server find the root snapshot which is the child of the snapshot `source-snapshot-1-clone-entry-point`
+1. Replica server then call SPDK API `bdev_lvol_decouple_parent` repeatedly to decouple it from all of the parents
```
DamiaSan, line 149: If you don't need to keep it thinly provisioned, you can call `bdev_lvol_inflate` once.

DamiaSan (commented):
Just some comments.

enhancements/20250508-v2-volume-cloning.md
```diff
@@ -0,0 +1,166 @@
...
+#### Longhorn API
+Add a new field into `volume.Spec` to indicate whether user wants a fast cloning or regular cloning for a new volume provisioned from a snapshot of other volume.
+For example:
+```go
+type VolumeSpec struct {
+  // +optional 
+  DataSource VolumeDataSource `json:"dataSource"`
+  // +kubebuilder:validation:Enum=fast;full;decoupled
+  // +optional 
+  CloneMode bool `json:"cloneMode"`
```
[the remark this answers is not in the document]
PhanLe1010: You are right. It should be string type. Thank you

enhancements/20250508-v2-volume-cloning.md
PhanLe1010: Yes. I thought about and tried this approach. The issue is that when user delete the source PVC, the cloned PVC is also broken after the clone already finished. This mean the cloned PVC is not really a persistent volume.

enhancements/20250508-v2-volume-cloning.md
PhanLe1010: Suppose that user want to create 2 PVC from the `source-snapshot-1`.
If we don't create the `source-snapshot-1-clone-entry-point` but clone the `source-snapshot-1` directly, when user delete the source volume, it will not be able to delete the `source-snapshot-1` because `source-snapshot-1` is having more than 1 child (the 2 cloned PVCs)

Please see the slide which demonstrate this issue https://github.com/longhorn/longhorn/blob/f93283f1a7be72474f4617f6de69c635754a02bd/enhancements/assets/v2-volume-cloning/longhorn-v2-fast-cloning.pdf

enhancements/20250508-v2-volume-cloning.md
PhanLe1010: Agree. But I think we do want to keep the volume thinly provisioned

enhancements/20250508-v2-volume-cloning.md
DamiaSan: One assumption is _The data is depended on the source volume_, so I thought you could not delete the external snapshot. If it can happen, you should inflate the clone and so it won't be anymore thinly provisioned.
Instead, if you move the source volume, you should recreate the external snapshot from the new location.

enhancements/20250508-v2-volume-cloning.md
DamiaSan: Ok, understood thanks.

Issue #10875, opened 2025-05-09
[TEST] [v1.9.0-rc2] longhorn_setting_orphan_auto_deletion missing from upgrade responder


chriscchien:
### Describe the Bug

From manual test case [Test upgrade responder extended collection](https://longhorn.github.io/longhorn-tests/manual/release-specific/v1.9.0/test-upgrade-responder-extended-collection/)
`longhorn_setting_orphan_auto_deletion` not found in upgrade responder in v1.9.0-rc2 (v1.9.0-rc1 have it)

v1.9.0-rc2
```bash
> kl get orphan
NAME                                                                      TYPE      NODE
orphan-707fa7489bcc5821e250cd13cbf4bd86c9c898f08f7a54a98910c7aa8d3594cd   replica   ip-172-31-35-152
orphan-da64054d1219c15d1e8aabd53da771c09108f25755bb870f7e05c4869dee8050   replica   ip-172-31-35-152
> 
> kubectl exec -it ${influxdb_pod} -- influx -execute 'SHOW TAG KEYS FROM upgrade_request' -database="${app_name}_upgrade_responder" | wc -l
42
>
> kubectl exec -it ${influxdb_pod} -- influx -execute 'SHOW TAG KEYS FROM upgrade_request' -database="${app_name}_upgrade_responder" | grep orphan | wc -l
0
root@ip-172-31-35-152:/home/ubuntu/longhorn/dev/upgrade-responder# kl get engineimage
NAME          INCOMPATIBLE   STATE      IMAGE                                   REFCOUNT   BUILDDATE   AGE
ei-db72f895   false          deployed   longhornio/longhorn-engine:v1.9.0-rc2   21         41h         95m
```

v1.9.0-rc1 have it
```bash
> kl get orphan
NAME                                                                      TYPE      NODE
orphan-0f0babbf3a25a28bc186149e1c3914ecee8f2a20ce4b8df1204c1a814051522f   replica   ip-172-31-34-38
orphan-26157852295d46a4a8ede4e6862e3963abdb3bec2e62d7a761fb8065a35e127e   replica   ip-172-31-34-38
> 
> kubectl exec -it ${influxdb_pod} -- influx -execute 'SHOW TAG KEYS FROM upgrade_request' -database="${app_name}_upgrade_responder" | wc -l
43
> 
[...]

PR #10876, opened 2025-05-09
chore: update install.sh in upgrade-responder

derekbit:

Issue Longhorn/longhorn#10875

Replace longhornSettingOrphanAutoDeletion with longhornSettingOrphanResourceAutoDeletion.

### DIFF
--- a/deploy/upgrade_responder_server/chart-values.yaml
+++ b/deploy/upgrade_responder_server/chart-values.yaml
@@ -172,6 +172,10 @@ configMap:
           "dataType": "string",
           "maxLen": 200
         },
+        "longhornSettingOrphanResourceAutoDeletion": {
+          "dataType": "string",
+          "maxLen": 200
+        },
         "longhornSettingPriorityClass": {
           "dataType": "string",
           "maxLen": 200
--- a/dev/upgrade-responder/install.sh
+++ b/dev/upgrade-responder/install.sh
@@ -164,7 +164,7 @@ configMap:
           "dataType": "string",
           "maxLen": 200
         },
-        "longhornSettingOrphanAutoDeletion": {
+        "longhornSettingOrphanResourceAutoDeletion": {
           "dataType": "string",
           "maxLen": 200
         },


c3y1huang (commented):
And also https://github.com/longhorn/longhorn/blob/f08185c37fe83ab625e94d717f6f57a0c70ff48c/deploy/upgrade_responder_server/chart-values.yaml#L171

derekbit:
> And also
> 
> https://github.com/longhorn/longhorn/blob/f08185c37fe83ab625e94d717f6f57a0c70ff48c/deploy/upgrade_responder_server/chart-values.yaml#L171

Do we still need `longhornSettingOrphanAutoDeletion`? It is replaced by longhornSettingOrphanResourceAutoDeletion since 1.9.0.

deploy/upgrade_responder_server/chart-values.yaml
```diff
@@ -172,6 +172,10 @@ configMap:
           "dataType": "string",
           "maxLen": 200
         },
+        "longhornSettingOrphanResourceAutoDeletion": {
```
c3y1huang, line 175: Should we add a comment to `longhornSettingOrphanAutoDeletion` explaining why we're keeping this deprecated setting in the upgrade-responder?

dev/upgrade-responder/install.sh
```diff
@@ -168,6 +168,10 @@ configMap:
           "dataType": "string",
           "maxLen": 200
         },
+        "longhornSettingOrphanResourceAutoDeletion": {
```
c3y1huang, line 171: We probably don't need the `longhornSettingOrphanAutoDeletion` in this file, since it's intended only for creating a local upgrade-responder stack for testing purposes. That said, I am fine keeping it for consistency with `upgrade_responder_server/chart-values.yaml`.

derekbit: Removed. Thanks.

deploy/upgrade_responder_server/chart-values.yaml
derekbit: JSON doesn't support comment, so I added the reason in the commit message. Thanks.

dev/upgrade-responder/install.sh
c3y1huang: We still need `longhornSettingOrphanResourceAutoDeletion` but not `longhornSettingOrphanAutoDeletion`.

c3y1huang (approved):

dev/upgrade-responder/install.sh
derekbit: Added. Thank you.

Issue #10249, opened 2025-01-19
[FEATURE] Remove v1beta1 API CRD in Longhorn v1.10


derekbit:
## What's the task? Please describe

Remove v1beta1 API CRD in Longhorn v1.10

## Describe the sub-tasks

- Remove v1beta1 API CRD
- Stop the conversion webhook service

roger-ryao:
**Verified on master-head 20250620**

- longhorn `master-head` https://github.com/longhorn/longhorn/commit/3bd56d3b80c6dc8e35480c134872d64e496997cb
- longhorn-manager `master-head` https://github.com/longhorn/longhorn-manager/commit/ede5b8285c26df2680d329027a040d19efcdae7d

**The test steps**

1. Verify fresh installation
2. Verify upgrade from v1.9 to master branch


**Result Passed**

1. The regression test for a fresh installation passed without any outstanding issues.
- https://ci.longhorn.io/job/public/job/master/job/sles/job/amd64/job/longhorn-tests-sles-amd64/1208/
2. The regression test for upgrading from `v1.9.0` to the `master` passed without any outstanding issues.
- https://ci.longhorn.io/job/public/job/master/job/sles/job/amd64/job/longhorn-upgrade-tests-sles-amd64/1017/

PR #10993, opened 2025-05-27
feat(crd): remove v1beta1 api

derekbit:

Issue Longhorn/longhorn#10249

### DIFF
--- a/chart/templates/daemonset-sa.yaml
+++ b/chart/templates/daemonset-sa.yaml
@@ -51,8 +51,6 @@ spec:
         ports:
         - containerPort: 9500
           name: manager
-        - containerPort: 9501
-          name: conversion-wh
         - containerPort: 9502
           name: admission-wh
         - containerPort: 9503
@@ -60,7 +58,7 @@ spec:
         readinessProbe:
           httpGet:
             path: /v1/healthz
-            port: 9501
+            port: 9502
             scheme: HTTPS
         volumeMounts:
         - name: boot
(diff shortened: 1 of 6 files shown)


chart/templates/daemonset-sa.yaml
```diff
@@ -51,17 +51,10 @@ spec:
         ports:
         - containerPort: 9500
           name: manager
-        - containerPort: 9501
-          name: conversion-wh
         - containerPort: 9502
           name: admission-wh
         - containerPort: 9503
           name: recov-backend
-        readinessProbe:
```
mantissahz, line 60: Don't we need `readinessProbe`?

derekbit: Conversion webhook is removed. Do we need the readinessProbe?

c3y1huang (dismissed):

chart/templates/daemonset-sa.yaml
mantissahz: It should make the container ready quickly and ensure it is ready to accept traffic. https://kubernetes.io/docs/concepts/configuration/liveness-readiness-startup-probes/#readiness-probe
Could we modify the `httpGet.port` to admission-wh port 9502?
But I'm fine if we remove it.

mantissahz (approved):

chart/templates/daemonset-sa.yaml
derekbit: Add it back. Thank you.

Issue #11010, opened 2025-05-29
[TASK] Create Pull Request Review Reminder Workflow


derekbit:
## What's the task? Please describe

Create a pull request review reminder workflow

derekbit:
cc @Yu-Jack @innobead

derekbit:
Completed.

PR #11014, opened 2025-05-29
feat(workflow): add Pull Request Review Reminder Workflow

derekbit:

Issue Longhorn/longhorn#11010

### DIFF
--- a/.github/workflows/pr-review-reminder.py
+++ b/.github/workflows/pr-review-reminder.py
@@ -1,0 +1,186 @@
+import subprocess
+import json
+import os
+import requests
+
+REPOS = ["longhorn/longhorn-manager",
+         "longhorn/longhorn-engine",
+         "longhorn/longhorn-instance-manager",
+         "longhorn/longhorn-share-manager",
+         "longhorn/backing-image-manager",
+         "longhorn/longhorn-ui",
+         "longhorn/longhorn-spdk-engine",
+         "longhorn/go-iscsi-helper",
+         "longhorn/go-spdk-helper",
+         "longhorn/backupstore",
+         "longhorn/go-common-libs",
+         "longhorn/types"]
+
+
+def flatten_issues(title, blocks, issues, user_mapping):
+    # Append the title and divider only if there are issues to display
+    print(issues)
+    if issues:
+        blocks.append(
+            {
+                "type": "section",
+                "text": {
+                    "type": "mrkdwn",
+                    "text": f"*{title}* - {len(issues)} prs"
+                }
+            }
+        )
+        blocks.append({"type": "divider"})
+
+        # Combine issues into chunks of 5
+        issue_texts = []
+        for i, issue in enumerate(issues):
+            number = issue["number"]
+            title = issue["title"]
+            issue_url = f"https://github.com/longhorn/longhorn/pull/{number}"
+            reviewers = []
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-review-reminder.py
```diff
@@ -0,0 +1,187 @@
...
+
+    payload = {
+        "blocks": blocks
+    }
+
+    headers = {
+        'Content-Type': 'application/json'
+    }
+
+    print(f"Sending notification to Slack: {webhook_url}")
```
Yu-Jack: I think we shouldn't print webhook_url since it's confidential

.github/workflows/pr-review-reminder.py
```diff
@@ -0,0 +1,187 @@
...
+            }
+        )
+        blocks.append({"type": "divider"})
+
+        # Combine issues into chunks of 5
+        issue_texts = []
+        for i, issue in enumerate(issues):
+            number = issue["number"]
+            title = issue["title"]
+            issue_url = f"https://github.com/longhorn/{title}/pull/{number}"
```
Yu-Jack: Should it be `f"https://github.com/longhorn/longhorn/pull/{number}"`?

.github/workflows/pr-review-reminder.py
```diff
@@ -0,0 +1,187 @@
...
+        }
+    })
+
+    for repo in REPOS:
+        print(f"Processing prs in {repo}...")
+        blocks = flatten_issues(repo, blocks, issues[repo], user_mapping)
+
+    payload = {
+        "blocks": blocks
+    }
```
Yu-Jack, line 93: We might take care of the limitation in the future if PRs are too much.

> https://api.slack.com/reference/block-kit/blocks
> Read our guide to [building block layouts](https://api.slack.com/block-kit/building) to learn where and how to use these blocks. You can include up to 50 blocks in each message, and 100 blocks in modals or Home tabs.

.github/workflows/pr-review-reminder.yml
```diff
@@ -0,0 +1,40 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.x'
+
+    - name: Install dependencies
+      run: |
+        python -m pip install requests
+
+    - uses: actions/checkout@v4
```
Yu-Jack, line 33: Should we checkout the code base firstly?

derekbit: what's the advantage?

.github/workflows/pr-review-reminder.py
derekbit: Removed.

.github/workflows/pr-review-reminder.py
derekbit: Fixed. Thank you.

.github/workflows/pr-review-reminder.py
derekbit: Yes, we can improve it over time.

.github/workflows/pr-review-reminder.yml
Yu-Jack: I usually checkout the code base first, then do other things. If there is no order issue, it's okay.

Issue #11031, opened 2025-06-02
[IMPROVEMENT] add support for custom annotations in the UI service on Longhorn Helm Chart


josimar-silva:
### Is your improvement request related to a feature? Please describe (👍 if you like this request)

Add support for custom annotations to the UI service on Longhorn Helm chart

### Describe the solution you'd like

E.g: values.yaml

```yaml
service:
  ui:
    type: "LoadBalancer"
    loadBalancerIP: 192.168.10.101
    annotations:
      external-dns.alpha.kubernetes.io/hostname: longhorn.homelab.lan
      metallb.io/ip-allocated-from-pool: homelab-local-pool
      metallb.universe.tf/allow-shared-ip: longhorn-svc
```

Renderer UI Service definition:
```yaml
kind: Service
apiVersion: v1
metadata:
  labels:
    app.kubernetes.io/name: longhorn
    helm.sh/chart: longhorn-1.9.0
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/instance: longhorn-test
    app.kubernetes.io/version: v1.9.0
    app: longhorn-ui
  name: longhorn-frontend
  namespace: longhorn-system
  annotations:
    external-dns.alpha.kubernetes.io/hostname: longhorn.homelab.lan
    metallb.io/ip-allocated-from-pool: homelab-local-pool
    metallb.universe.tf/allow-shared-ip: longhorn-svc
spec:
  type: LoadBalancer
  loadBalancerIP: 192.168.10.101
  selector:
    app: longhorn-ui
  ports:
  - name: http
    port: 80
    targetPort: http
    nodePort: null
```

### Additional context

https://github.com/longhorn/charts/pull/211

roger-ryao:
**Verified on master-head 20250623**

- longhorn `master-head` https://github.com/longhorn/longhorn/commit/3bd56d3b80c6dc8e35480c134872d64e496997cb

**The test steps**

https://github.com/longhorn/longhorn/issues/11031#issue-3111260160

1. Modified `values.yaml` to include custom annotations.
```diff
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -146,7 +146,10 @@ service:
     # -- NodePort port number for Longhorn UI. When unspecified, Longhorn selects a free port between 30000 and 32767.
     nodePort: null
     # -- Annotation for the Longhorn UI service.
-    annotations: {}
+    annotations:
[...]

PR #11032, opened 2025-06-02
feat: add support for custom annotations on UI service

josimar-silva:

Issue #11031 

I'm using the Longhorn Helm chart in my homelab setup.

During configuration, I found that the chart doesn't support adding custom annotations to the UI service.
As a workaround, I had to use [FluxCD postRenderers](https://github.com/josimar-silva/homelab/pull/8/files), which added unnecessary complexity.

This PR adds native support for custom annotations to the UI service, making the chart more flexible for other use cases.


derekbit (approved):
LGTM

derekbit:
Merged. Thank you @josimar-silva

PR #11082, opened 2025-06-06
chore(crd): update crds.yaml and manifests (PR longhorn/longhorn-manager#3837)

derekbit:
It was triggered by longhorn/longhorn-manager#3837.

### DIFF
--- a/chart/README.md
+++ b/chart/README.md
@@ -144,6 +144,7 @@ The `values.yaml` contains items used to tweak a deployment of this chart.
 |-----|-------------|
 | service.manager.nodePort | NodePort port number for Longhorn Manager. When unspecified, Longhorn selects a free port between 30000 and 32767. |
 | service.manager.type | Service type for Longhorn Manager. |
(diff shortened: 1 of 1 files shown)


mantissahz (approved):

c3y1huang (commented):
1. So weird, this was triggered by https://github.com/longhorn/longhorn-manager/pull/3837.
1. And perhaps we should consider fixing the DCO failure?

mantissahz:
>So weird, this was triggered by https://github.com/longhorn/longhorn-manager/pull/3837.

It seems that it will not be triggered by modifying the `longhorn` repo https://github.com/longhorn/longhorn/pull/11032 (only by changing the `longhorn-manager` repo)

derekbit:
> 1. So weird, this was triggered by [fix(recurring-job): race condition longhorn-manager#3837](https://github.com/longhorn/longhorn-manager/pull/3837).
> 2. And perhaps we should consider fixing the DCO failure?

We can refine it better. 
Currently, each merged longhorn-manager PR can trigger longhorn/longhorn update.

Issue #11106, opened 2025-06-11
[TASK] Ensure support-bundle-kit builds use vendored dependencies


c3y1huang:
## What's the task? Please describe

The current support-bundle build may produce inconsistent artifacts because Go can fetch dependencies from external sources. This can result in non-reproducible builds.

Update the build script to enforce the use of vendor dependencies to ensure consistent builds.

## Describe the sub-tasks

- Add `-mod=vendor` to all `go build` commands
- Resolve CVE issues
- Update Longhorn support-bundle-kit image version

## Additional context

`None`

yangchiu:
@c3y1huang Could you help clarify how to verify this task?

I tried building the support-bundle-kit image locally and compare its `sha256sum` with [the `support-bundle-kit:v0.0.57` image](https://github.com/rancher/support-bundle-kit/releases/tag/v0.0.57), but it seems the builds are still not consistent because the `sha256sum` are different.

```
~/support-bundle-kit$ git log --oneline
e3ea5df8 (HEAD -> master, tag: v0.0.57, origin/master, origin/HEAD) test: use constant for TestGenerateNodeZipObjects expectations

~/support-bundle-kit$ make
./.dapper default
...
...
Built rancher/support-bundle-kit:v0.0.57
INFO[0168] docker cp /go/src/github.com/rancher/support-bundle-kit/bin . 
                              Successfully copied 205MB to /home/yang/Code/support-bundle-kit/.
INFO[0169] docker cp /go/src/github.com/rancher/support-bundle-kit/dist . 
                              Successfully copied 205MB to /home/yang/Code/support-bundle-kit/.

$ sha256sum ./bin/support-bundle-kit-amd64 
5183042581ce68a4b44801b0fba42765ba9d4e7135d2d25799601d4556672179  ./bin/support-bundle-kit-amd64
```

```
~/Downloads$ tar -xzf support-bundle-kit-amd64.tar.gz
~/Downloads$ sha256sum support-bundle-kit-amd64
6de55b2f32d553fba95214c308696f18a430e34f1a4acede8c9cc3b397e50a8c  support-bundle-kit-amd64
```

c3y1huang:
> [@c3y1huang](https://github.com/c3y1huang) Could you help clarify how to verify this task?
> 
[...]

PR #11116, opened 2025-06-13
chore(chart): update dependency versions (PR longhorn/dep-versions#60)

derekbit:
This PR updates dependency versions.
It was triggered by longhorn/dep-versions#60.


c3y1huang:
Ref https://github.com/longhorn/longhorn/issues/11106

Issue #11064, opened 2025-06-04
[FEATURE] Standardized way to override container image registry


marcosbc:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

A big part of the Helm charts have standardized on using `global.imageRegistry` as a way to override the image registry, in addition to per-image modifications via `*.registry` (in the same level as `*.repository`). 

### Describe the solution you'd like

We would like this Helm chart to support overriding the container image registry both globally, using the standardized `global.imageRegistry` value, as well as per-image via `$imagePath.registry`.

yangchiu:
Verification failed on master-head (longhorn 637d9e6).

Set `privateRegistry.registrySecret` in `values.yaml` 

```
privateRegistry:
  registrySecret: docker-registry-secret
```

Then install Longhorn:

```
$ helm install longhorn ./longhorn/chart/ --namespace longhorn-system --create-namespace
```

Get an error message:

```
Error: INSTALLATION FAILED: template: longhorn/templates/uninstall-job.yaml:36:34: executing "longhorn/templates/uninstall-job.yaml" at <append>: wrong number of args for append: want 2 got 1
```

cc @marcosbc @derekbit

derekbit:
> Verification failed on master-head (longhorn [637d9e6](https://github.com/longhorn/longhorn/commit/637d9e62fd496fbdcdbfbcd007caaee4ff5116eb)).
> 
> Set `privateRegistry.registrySecret` in `values.yaml`
> 
> ```
> privateRegistry:
>   registrySecret: docker-registry-secret
> ```
> 
> Then install Longhorn:
> 
> ```
> $ helm install longhorn ./longhorn/chart/ --namespace longhorn-system --create-namespace
> ```
> 
> Get an error message:
> 
> ```
> Error: INSTALLATION FAILED: template: longhorn/templates/uninstall-job.yaml:36:34: executing "longhorn/templates/uninstall-job.yaml" at <append>: wrong number of args for append: want 2 got 1
> ```
> 
> cc [@marcosbc](https://github.com/marcosbc) [@derekbit](https://github.com/derekbit)

@yangchiu You should use `registrySecret: [docker-registry-secret]` instead.

yangchiu:
> [@yangchiu](https://github.com/yangchiu) You should use `registrySecret: [docker-registry-secret]` instead.

[...]

PR #11063, opened 2025-06-04
feat(chart): implement standardized image registry override values

marcosbc:
#### Which issue(s) this PR fixes:

Issue #11064

#### What this PR does / why we need it:

This PR adds the standardized `global.imageRegistry` value, as well as per-image via `$imagePath.registry`, to allow overriding the image.

#### Special notes for your reviewer:

N/A

#### Additional documentation or context

See issue for more details.

PR #11145, opened 2025-06-20
fix(chart): fix usage of privateRegistry.registrySecret

marcosbc:

Issue #11064

In #11063 we originally added some logic to support `global.imagePullSecrets`, but this broke the usage of `privateRegistry.registrySecret`.

We are fixing it in the following way:

* Fix usage of `privateRegistry.registrySecret` when it is a string. This was caused by the `$imagePullSecrets = append ...` line, which was implemented incorrectly.
* If `privateRegistry.registrySecret` is set but not `privateRegistry.createSecret`, no image pull secret will be used for the pods, as this is what is mentioned in the README: "This setting applies only when creation of private registry secrets is enabled."
* If `privateRegistry.registrySecret` is defined but not a string, an error is thrown. This is because the template currently expects it to be a string.

See the original issue.

### DIFF
--- a/chart/templates/daemonset-sa.yaml
+++ b/chart/templates/daemonset-sa.yaml
@@ -132,11 +132,11 @@ spec:
         secret:
           secretName: longhorn-grpc-tls
           optional: true
-      {{- with (coalesce .Values.global.imagePullSecrets .Values.privateRegistry.registrySecret) }}
+      {{- with (coalesce .Values.global.imagePullSecrets (ternary "" .Values.privateRegistry.registrySecret (empty .Values.privateRegistry.createSecret))) }}
       imagePullSecrets:
         {{- $imagePullSecrets := list }}
         {{- if kindIs "string" . }}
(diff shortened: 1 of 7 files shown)


derekbit:
Hello @yangchiu 
Please help review the PR when you are available. Thanks.

derekbit:
Hello @marcosbc 
Running into
```
Error: template: longhorn/templates/uninstall-job.yaml:32:115: executing "longhorn/templates/uninstall-job.yaml" at <.Values.privateRegistry.createSecret>: wrong type for value; expected bool; got interface {}`
```
while executing `scripts/generate-longhorn-yaml.sh`

marcosbc:
@derekbit Good catch, it should be fixed now. The `ternary` condition added an issue when the condition was not a string (an empty object by default), which should be fixed by forcing it to be a boolean via `empty .Values.privateRegistry.createSecret`.

derekbit (approved):

derekbit:
@mergify backport v1.9.x

PR #11197, opened 2025-06-30
chore(chart): update dependency versions (PR longhorn/dep-versions#69)

derekbit:
This PR updates dependency versions.
It was triggered by longhorn/dep-versions#69.


c3y1huang:
Ref https://github.com/longhorn/longhorn/issues/2259

c3y1huang:
@mergify backport v1.8.x

Issue #11186, opened 2025-06-26
[IMPROVEMENT] Helm Chart: make the `/var/lib/longhorn` hostPath configurable


pstark-code:
### Is your improvement request related to a feature? Please describe (👍 if you like this request)

I am trying to get Longhorn running on my Talos cluster. I used v1.9.0, and I followed the flux deployment suggestion utilizing the Helm Chart.
Unfortunately, I am currently unable to get the disk mounted under `/var/lib/longhorn` in the kubelet's scope. Instead it can only be mounted at `/var/mnt/my-user-volume-name`. (To be fair, this could be due to user error. I am still investigating.)

The setting `defaultSettings.defaultDataPath` can be changed in the Helm values, but that changes neither the mount point nor the source, i.e. the hostPath.

### Describe the solution you'd like

It would be useful to be able to adjust the `spec.template.spec.volumes[name=="longhorn"].hostPath` via Helm values. It can be found right [here](https://github.com/longhorn/longhorn/blob/master/chart/templates/daemonset-sa.yaml#L122-L124)

Even if my issue turns out to be my own fault, I think it would still be nice to be able to adjust this value. 

### Describe alternatives you've considered

I tried the Talos mechanism of bind mounting the `UserVolume` to the correct spot, but that stubbornly refuses to work as expected. I am still tracking down that issue.

mantissahz:
Longhorn stores some binary images in the `/var/lib/longhorn` now (hard-coded)

Here are the `Requirements` for the Talos support.
https://longhorn.io/docs/1.9.0/advanced-resources/os-distro-specific/talos-linux-support/#data-path-mounts

derekbit:
cc @longhorn/dev @innobead

pstark-code:
## Re: Hardcoded Path

> Longhorn stores some binary images in the `/var/lib/longhorn` now (hard-coded)

That's not a problem and I wasn't proposing to change that path inside the container. 

I am talking of the volume's ***hostPath*** ([source link](https://github.com/longhorn/longhorn/blob/master/chart/templates/daemonset-sa.yaml#L124)), not the volumeMount's `mountPath` ([source link](https://github.com/longhorn/longhorn/blob/master/chart/templates/daemonset-sa.yaml#L76)). It's the directory path on the ***host*** that will be mounted ***into the container*** at the defined path of `/var/lib/longhorn`

## Re: Talos specific installation instruction

> Here are the Requirements for the Talos support.

[...]

PR #11199, opened 2025-06-30
feat(chart): improve configurability of mounts for the longhorn-manager DaemonSet

pstark-code:
Issue longhorn/longhorn#11186

Additionally, adds the option for extra Volumes and extra VolumeMounts. Longhorn supports multiple disks, but can not mount multiple volumes into the container.

- add support for a custom hostPath for the data volume attached to the longhorn-manager pod
- add support for extra `volumes` entries and extra `volumeMounts` entries.

This change does not require any adjustments for existing users that don't need this feature. It should be

### DIFF
--- a/chart/templates/daemonset-sa.yaml
+++ b/chart/templates/daemonset-sa.yaml
@@ -81,6 +81,7 @@ spec:
         - name: go-cover-dir
           mountPath: /go-cover-dir/
         {{- end }}
+        {{- toYaml .Values.longhornManager.longhornManagerContainerVolumeMounts | nindent 8 }}
         env:
         - name: POD_NAME
           valueFrom:
@@ -132,6 +133,7 @@ spec:
         secret:
           secretName: longhorn-grpc-tls
           optional: true
+      {{- toYaml .Values.longhornManager.extraVolumes | nindent 6 }}
       {{- with (coalesce .Values.global.imagePullSecrets (ternary "" .Values.privateRegistry.registrySecret (empty .Values.privateRegistry.createSecret))) }}
       imagePullSecrets:
         {{- $imagePullSecrets := list }}
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -437,6 +437,22 @@ longhornManager:
   #  annotation-key2: "annotation-value2"
   # -- Host path on the nodes that will be mounted to `/var/lib/longhorn` inside the longhorn-manager pods. The default value is "/var/lib/longhorn/".
   dataVolumeHostPath: "/var/lib/longhorn/"
+  # -- Extra volume specs to add to the `longhorn-manager` pod. The values are directly inserted into the
+  # `volumes` key in the pod spec.
+  extraVolumes: []
+  ## To add extra volumes, delete the `[]` in the line above
+  ## and uncomment this example block:
+  # - name: longhorn
+  #   hostPath:
+  #     path: /var/mnt/my-bonus-data
+  # -- Extra volumeMount specs to add to the `longhorn-manager` container in the `longhorn-manager` pod.
+  # The values are directly inserted into the `volumes` key in the pod spec.
+  longhornManagerContainerVolumeMounts: []
+  ## To mount extra volumes defined above, delete the `[]` in the line above
+  ## and uncomment this example block:
+  # - name: my-bonus-data-volume
+  #   mountPath: /var/lib/longhorn-extra/my-bonus-data
+  #   mountPropagation: Bidirectional
 longhornDriver:
   log:
     # -- Format of longhorn-driver logs. (Options: "plain", "json")


mantissahz:
Hi @pstark-code, please sign off the commit.

pstark-code:
signed-off and rebased on top of the upstream default branch.

derekbit:
The PR might not be enough since some paths in the longhorn-manager or other component images are hardcoded with the `/var/lib/longhorn` prefix.

pstark-code:
> The PR might not be enough since some paths in the longhorn-manager or other component images are hardcoded with the `/var/lib/longhorn` prefix.

This was brought up in  multiple places. Please read [my response](https://github.com/longhorn/longhorn/issues/11186#issuecomment-3018474803). This change does **_not change the path_** it gets mounted at **_in the container_**. So for the Longhorn Manager software, nothing changes.

pstark-code:
I also added the ability to configure extra volumes to be mounted into the longhorn-manager container. This mirrors the longhorn option to have multiple "disks".

derekbit:
> > The PR might not be enough since some paths in the longhorn-manager or other component images are hardcoded with the `/var/lib/longhorn` prefix.
> 
> This was brought up in multiple places. Please read [my response](https://github.com/longhorn/longhorn/issues/11186#issuecomment-3018474803). This change does **_not change the path_** it gets mounted at **_in the container_**. So for the Longhorn Manager software, nothing changes.

The binaries stored in the host’s `/var/lib/longhorn/engine-binaries` are managed by the engineImage controller ([code](https://github.com/longhorn/longhorn-manager/blob/master/controller/engine_image_controller.go#L875)) and the EngineImage DaemonSet.

The change introduced in the PR makes the `spec.volumes['longhorn'].hostPath.path` of the longhorn-manager DaemonSet configurable. However, the EngineImage DaemonSet is not aware of this change, which results in the controllers being unable to locate the binaries at the expected path inside the container (`/var/lib/longhorn/engine-binaries`). Thus, updating the longhorn-manager codes is still required.

pstark-code:
> The binaries stored in the host’s `/var/lib/longhorn/engine-binaries` are managed by the engineImage controller ([code](https://github.com/longhorn/longhorn-manager/blob/master/controller/engine_image_controller.go#L875)) and the EngineImage DaemonSet.

Ah! That explains why I couldn't figure out how they get there!

Okay, then I will remove the configurable hostPath and leave the extraMounts and extraVolumeMounts options, because that, together with the `default settings.defaultPath` resolves all the issues I was looking to address.

Just as a curious question: why are the engine binary fetching containers not part of the longhorn-manager pods? InitContainers maybe? It seems unintuitive to have this tight coupling between two seemingly disparate DaemonSets.

derekbit:
> Okay, then I will remove the configurable hostPath and leave the extraMounts and extraVolumeMounts options, because that, together with the default settings.defaultPath resolves all the issues I was looking to address.

Is narrowing down the scope to `extraMounts` and `extraVolumeMounts` enough for addressing the issue?

> Just as a curious question: why are the engine binary fetching containers not part of the longhorn-manager pods? InitContainers maybe? It seems unintuitive to have this tight coupling between two seemingly disparate DaemonSets.

I think this is due to Longhorn can only one longhorn-manager daemonset but can support multiple engineimage daemonsets. Longhorn can support multiple egnineimages simutaneously.

pstark-code:
> Is narrowing down the scope to `extraMounts` and `extraVolumeMounts` enough for addressing the issue?

I believe so. I wanted to resolve two things:
1. I thought it was a silly limitation to require a very specific path on the host to contain longhorn things. 
  - but as it turns out there is a good reason to not change this. Changing that would require far-reaching refactoring within the longhorn-manager, as far as I can tell
2. I was trying to work around a talos-specific issue I was having. 
  - Context: Talos runs the kubelet inside a container and so any paths that need to be visible to the kubelet need to be bind-mounted into that namespace. This gave me some issues with the new UserVolume mechanism for exposing extra disks to talos nodes. And the only reason I was trying to do that is because the longhorn chart did not have any other way to access extra storage.
  - The extra disks are already automatically available inside the Talos kubelet namespace at `/var/mnt/${user-volume-name}`. So if I can just specify another Volume with that hostPath, and then add a longhorn disk with that path for storage, everything should work. (or just set it via the `defaultSettings.defaultDataPath`)

To summarize: Yes I think limiting the scope will be sufficient. In fact, I think making the hostPath of  `/var/lib/longhorn` configurable is useless and will fundamentally (although temporarily) break longhorn deployments.

shuo-wu:
> Talos runs the kubelet inside a container and so any paths that need to be visible to the kubelet need to be bind-mounted into that namespace.

- If all paths used by longhorn-manager should be bind mounted to the kubelet container, this change is probably not enough since longhorn-manager hard codes some paths when operating files, e.g., [the socket files](https://github.com/longhorn/longhorn-manager/blob/eb0bc95a65c953ae49db7dfede5f64b6ea2e39bf/types/types.go#L108-L109)

> if I can just specify another Volume with that hostPath, and then add a longhorn disk with that path for storage, everything should work. (or just set it via the defaultSettings.defaultDataPath)

- If you are trying to use a customized data path/disk path as Longhorn Disk, the main ways are modifying the customized setting before launching Longhorn or editing the Longhorn Node after launching Longhorn. IIRC, changing the volume host path mounted by longhorn-manager won't help automatically add it as Longhorn Disk, because it's InstanceManager Pod being responsible for managing those data paths.

PR #11175, opened 2025-06-25
Update OpenShift documentation

mlacko64:
Fix for OpenShift helm chart values files example (oauthProxy image)

#### Which issue(s) this PR fixes:

Issue # https://github.com/longhorn/longhorn/issues/11174

PR #11210, opened 2025-07-02
chore(crd): update crds.yaml and manifests (PR longhorn/longhorn-manager#3886)

derekbit:
It was triggered by longhorn/longhorn-manager#3886.


mantissahz (approved):
In general, LGTM, but this PR should be triggered by https://github.com/longhorn/longhorn/pull/11175.

Issue #5215, opened 2023-01-05
[FEATURE] Configurable Backup Block Size


innobead:
Make sure to introduce a backup format version, this way it's easy to implement the backwards compatibility.
We also want to use much larger blocks in the new version, consider going from current 2MB -> 16MB -> 64MB this will improve the compression but further also reduce the amount of block files, which will lower the cost for the S3 lookups. I am thinking 16MB could be a good middle ground. 

This is optional but for larger volumes it would be nice if we can dynamically switch to 64MB, perhaps a block size field in the new backup v2 version?

_Originally posted by @joshimoo in https://github.com/longhorn/longhorn/issues/5189#issuecomment-1372100703_

Ref:
- https://github.com/longhorn/longhorn/issues/9817

innobead:
cc @longhorn/dev

shuo-wu:
Do we need to make it configurable?

derekbit:
I'm thinking if _backup versing_ is really necessary. For the existing volumes, the `blockSize` field in backup metadata is empty, and it means the `blockSize` is 2MiB. For a new backup, the `blockSize` is configurable and is stored in to metafile. Avoiding the versioning and achieving the backward compatibility will be much simpler before backup really needs a version control.

COLDTURNIP:
The similar design exists in some popular backup systems:

- Rclone: 5MB is the default size of a S3 chunk ([doc](https://rclone.org/s3/#s3-chunk-size))
- Kopia: CAOS are packed into a 20MB pack file to store in the repository ([doc](https://kopia.io/docs/advanced/architecture), [source](https://github.com/kopia/kopia/blob/v0.20.1/repo/format/format_provider.go#L101-L104))
- Restic: blobs are packed into a pack file before uploading to the repository ([doc](https://restic.readthedocs.io/en/latest/047_tuning_backup_parameters.html#pack-size), [doc](https://github.com/restic/restic/blob/master/doc/design.rst))

dominickmarino:
Thank you for your efforts on this. We have clients getting rate-limited due to block size and operations. For Storj we would love to see this at 64mb (our block size). Defaulting to 64mb and allowing users to configure smaller may be optimal.

derekbit:
> Thank you for your efforts on this. We have clients getting rate-limited due to block size and operations. For Storj we would love to see this at 64mb (our block size). Defaulting to 64mb and allowing users to configure smaller may be optimal.

Thanks @dominickmarino for your input.

[...]

PR #11245, opened 2025-07-09
docs(lep): configurable backup block size

COLDTURNIP:

Issue #5215

Propose a design, to make the backup block size configurable. A larger backup block size may boost the transmission performance, and reduce the potential request cost for backup targets like S3. This design doc includes some comparison between the different backup block sizes.

### DIFF
--- a/enhancements/20250701-configurable-backup-block-size.md
+++ b/enhancements/20250701-configurable-backup-block-size.md
@@ -1,0 +1,215 @@
+# Configurable Backup Block Size
+
+## Summary
+
+This enhancement adds support for user configured (storage class, volumes) backup block size.
+
+### Related Issues
+
+- https://github.com/longhorn/longhorn/issues/5215
+
+## Motivation
+
+### Goals
+
+User is able to configure the backup block size other than the default size 2MiB.
+
+### Non-goals
+
+- There are pre-defined options of the valid backup block size. User cannot set the size other than the defined size.
+- The backup block size is configurable only when creating new volumes. This is immutable after volume is created.
+- The backup block size is not configurable for backup restore and disaster recovery (DR) volumes.
+- To the existing backups, the block size is immutable.
+- If a backup is created using a non-default block size, this backup cannot be restored by the engine that supports only the default block size.
+
+## Proposal
+
+Provides two available backup block size to be set in the volume CR:
+
+- 2MiB, the default size, and backward compatible with the older version
+- 16MiB
+
+Benchmark:
+
+- 3-node K3s cluster locally hosted by Vagrant-libvirt
+- 5GB v1 volume with 4GB random content single file, v2 engine is disabled in this cluster
+- Backup target minio is also locally hosted inside the same cluster.
+
+| Block Size | Num. of Blk files | Accumulate size on bkup target | Avg Transmission Time |
+|:----------:|:-----------------:|:------------------------------:|:---------------------:|
+|    2MiB    |       2063        |            4211852             |        29 sec         |
+|   16MiB    |        266        |            4197748             |        19 sec         |
+
+### User Stories
+
+Users can select an appropriate backup block size when creating a volume for their application, based on both the application requirements and the evaluation results regarding the characteristics of the backup target. The larger block size will improve the compression but further also reduce the number of block files, which will potentially lower the cost for the lookups of backup targets.
+
+#### Configure The Default Backup Block Size
+
+Users can configure global default backup block size setting `default-backup-block-size`. This configuration applies to newly created volumes.
+
+#### Configure The Block Size In Storage Class
+
+Users can specify the desired backup block size as a parameter in the storage class. This parameter is used when creating volumes. If no backup block size is specified, the system uses the globally configured default value.
+
+#### Configure The Block Size In Volume CR
+
+A field is available to specify the backup block size when creating a volume CR. This field becomes immutable once the CR is created. If no backup block size is specified, the system uses the globally configured default value.
+
+#### Configure The Block Size In Volume Creation UI
+
+Users specify the backup block size during the volume creation. It provides the available options to the user, and sets to 2MiB by default. The backup block size is listed in the volume details.
+
+While creating the volume form a backup, it also provides the available backup block size options to the user. This option affects the new backup creation for this restored volume.
+
+While creating the disaster recovery volume, the backup block size is fixed to the existing backups.
+
+#### View The Block Size In Backup List UI
+
+Users can check the size from backup list UI, so that they can confirm the backup block size.
+
+## Design
+
+### Setting
+
+A new setting `default-backup-block-size` is introduced as the default backup block size for volume creation. The setting is an integer to specify the block size in MB, , one of `2` (2 MiB) or `16` (16 MiB). 
+
+### CSI
+
+A new parameter `backupBlockSize` is introduced to the storage class.
+
+```yaml
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: pvc-larger-backup-block-size
+provisioner: driver.longhorn.io
+parameters:
+  ...
+  backupBlockSize: 16Mi
+```
+
+- A string of quantity.
+- Longhorn accepts only the value of `2Mi` and `16Mi`
+
+### Longhorn Manager
+
+- The new string field `BackupBlockSize` is configured in the volume CR and accepts a size integer in bytes, one of `2097152` (2 MiB) or `16777216` (16 MiB).
+    ```go
+    type VolumeSpec struct {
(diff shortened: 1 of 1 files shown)


enhancements/20250701-configurable-backup-block-size.md
```diff
@@ -0,0 +1,148 @@
...
+Benchmark:
+
+- 3-node K3s cluster locally hosted by Vagrant-libvirt
+- 5GB v1 volume with 4GB random content single file, v2 engine is disabled in this cluster
+- Backup target minio is also locally hosted inside the same cluster.
+
+| Block Size | Num. of Blk files | Accumulate size on bkup target | Avg Transmission Time |
+|:----------:|:-----------------:|:------------------------------:|:---------------------:|
+|    2MiB    |       2063        |            4211852             |        29 sec         |
+|   16MiB    |        266        |            4197748             |        19 sec         |
```
COLDTURNIP, line 41: Need more detail for benchmark:

- Add the case for sparse disk
- Describe the steps to reproduce
- Need the statistic for resource consumption, including CPU and memory

enhancements/20250701-configurable-backup-block-size.md
```diff
@@ -0,0 +1,148 @@
...
+    }
+    ```
+- Backup controller creates monitor for the backup CR, and passes the block size to the sync agent by set a new parameter `backup-block-size` in the volume descriptor.
+
+#### Webhooks
+
+- The volume mutator assigns the default block size (`2Mi`) when the specified value is `0`.
+- The volume validator ensures the validity of the new string field `backupBlockSize`, and prohibits modifications.
+  - The block size must be `2Mi` or `16Mi`.
+  - The size of the volume must be an integer multiple of the backup block size.
```
COLDTURNIP: The volume validator should also check the size of expansion.

enhancements/20250701-configurable-backup-block-size.md
```diff
@@ -0,0 +1,148 @@
...
+- The backup block size is configurable only when creating new volumes. This is immutable after volume is created.
+- The backup block size is not configurable for backup restore and disaster recovery (DR) volumes.
+- To the existing backups, the block size is immutable.
+
+## Proposal
+
+Provides two available backup block size to be set in the volume CR:
+
+- 2MiB, the default size, and backward compatible with the older version
+- 16MiB
```
COLDTURNIP, line 30: Need more survey for the available block size. Provide some strategy for the choice if possible.

enhancements/20250701-configurable-backup-block-size.md
derekbit: Also mention the tradeoff of block size, performance, and resource consumption

enhancements/20250701-configurable-backup-block-size.md
COLDTURNIP: Here are the default upload chunk size used by some popular backup systems for reference:

- Rclone: 5MB is the default size of a S3 chunk ([doc](https://rclone.org/s3/#s3-chunk-size))
- Kopia: CAOS are packed into a 20MB pack file to store in the repository ([doc](https://kopia.io/docs/advanced/architecture), [source](https://github.com/kopia/kopia/blob/master/repo/format/format_provider.go#L101-L104))
- Restic: blobs are packed into a pack file before uploading to the repository ([doc](https://restic.readthedocs.io/en/latest/047_tuning_backup_parameters.html#pack-size), [doc](https://github.com/restic/restic/blob/master/doc/design.rst))

enhancements/20250701-configurable-backup-block-size.md
```diff
@@ -0,0 +1,148 @@
...
+
+- During the initialization, load the block size from the backup metadata file, and validate the size. The volume size must be an integer multiple of the block size.
+- To incremental restoring, accept the block size of the new backup that is identical with the last one.
+- During the restoring, load the blocks in the configured block size.
+
+### Longhorn Manager RESTful API
+
+- Add a new field, `BackupBlockSize`, to the volume spec.
+- While Longhorn manager handles volume creation request, it validates the backup block size in the spec, and sets to the new volume CR.
+- Don't provide any method to edit the backup block size on existing volume.
```
derekbit, line 168: Handle it in the validating webhook.

derekbit (commented):
Do we need to update the `backupBlockSize` of existing volumes while upgrade?

COLDTURNIP:
> Do we need to update the `backupBlockSize` of existing volumes while upgrade?

Updating the `backupBlockSize` would make it more clear even when we're going to use default 2MB size by default.

derekbit:
> > Do we need to update the `backupBlockSize` of existing volumes while upgrade?
> 
> Updating the `backupBlockSize` would make it more clear even when we're going to use default 2MB size by default.

Yes, please add it into the LEP.

enhancements/20250701-configurable-backup-block-size.md
```diff
@@ -0,0 +1,158 @@
...
+
+User is able to configure the backup block size other than the default size 2MiB.
+
+### Non-goals
+
+- There are pre-defined options of the valid backup block size. User cannot set the size other than the defined size.
+- The backup block size is configurable only when creating new volumes. This is immutable after volume is created.
+- The backup block size is not configurable for backup restore and disaster recovery (DR) volumes.
+- To the existing backups, the block size is immutable.
+- If a backup is created using a non-default block size, this backup cannot be restored by the engine that supports only the default block size.
```
COLDTURNIP, line 23: Customized backup block size is incompatible with the earlier versions of engines. The restored volumes will be corrupted because of the segment calculation. Will warn the user for this incompatibility.

derekbit: I don’t think we need to care about old engines. It is fine that the old engines cannot recognize non-2MB blocks. The broken filesystem won't lead to any disaster.
Adding this limitation in the important notes and the backup/restore page is good enough.

derekbit (commented):
Introduce a global setting for the backup block size as other setting like compression setting.

derekbit:
@COLDTURNIP Is the LEP ready to merge?

COLDTURNIP:
> @COLDTURNIP Is the LEP ready to merge?

Yes, we've merged all the implementation.

enhancements/20250701-configurable-backup-block-size.md
```diff
@@ -0,0 +1,175 @@
...
+        ...
+    }
+    ```
+- While creating a new backup CR, the new field `BackupBlockSize` is copied from the volume CR.
+    ```go
+    type BackupSpec struct {
+        ...
+        // The backup block size
+        // +optional
+        BackupBlockSize BackupBlockSize `json:"backupBlockSize"`
```
derekbit: data type is int64.

enhancements/20250701-configurable-backup-block-size.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Longhorn Manager
+
+- The new string field `BackupBlockSize` is configured in the volume CR and accepts a quantity string, one of `2Mi` or `16Mi`.
+    ```go
+    type VolumeSpec struct {
+        ...
+        // +kubebuilder:validation:Enum=2Mi;16Mi
+        // +optional
+        BackupBlockSize BackupBlockSize `json:"backupBlockSize"`
```
derekbit: data type is int64

enhancements/20250701-configurable-backup-block-size.md
```diff
@@ -0,0 +1,175 @@
...
+|    2MiB    |       2063        |            4211852             |        29 sec         |
+|   16MiB    |        266        |            4197748             |        19 sec         |
+
+### User Stories
+
+Users can select an appropriate backup block size when creating a volume for their application, based on both the application requirements and the evaluation results regarding the characteristics of the backup target. The larger block size will improve the compression but further also reduce the number of block files, which will potentially lower the cost for the lookups of backup targets.
+
+#### Configure The Default Backup Block Size
+
+Users can configure global default backup block size settings. This configuration applies to newly created volumes.
```
derekbit: Add the setting name

enhancements/20250701-configurable-backup-block-size.md
```diff
@@ -0,0 +1,175 @@
...
+        ...
+        // The backup block size
+        // +optional
+        BackupBlockSize BackupBlockSize `json:"backupBlockSize"`
+        ...
+    }
+    ```
+- Backup controller creates monitor for the backup CR, and passes the block size to the sync agent by set a new parameter `backup-block-size` in the volume descriptor.
+
+#### Webhooks
```
derekbit, line 127: We have some handling for `-1` backup block size. Please briefly explain it here.

Issue #11264, opened 2025-07-10
[BUG] Incorrect value of `remove-snapshots-during-filesystem-trim` in longhorn chart/values.yaml


derekbit:
### Describe the Bug

Incorrect value of `remove-snapshots-during-filesystem-trim` in longhorn chart/values.yaml

- https://github.com/longhorn/longhorn/blob/master/chart/values.yaml#L210
- https://github.com/longhorn/longhorn-manager/blob/master/types/setting.go#L1270-L1280

### Expected Behavior

Should set a correct value

### Support Bundle for Troubleshooting

N/A

### Environment

- Longhorn version: master-head
- Impacted volume (PV): 
- Installation method (e.g. Rancher Catalog App/Helm/Kubectl):
- Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version:
  - Number of control plane nodes in the cluster:
  - Number of worker nodes in the cluster:
- Node config
  - OS type and version:
  - Kernel version:
  - CPU per node:
  - Memory per node:
  - Disk type (e.g. SSD/NVMe/HDD):
  - Network bandwidth between the nodes (Gbps):
- Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal):
- Number of Longhorn volumes in the cluster:

chriscchien:
Verified **pass** on longhorn master(https://github.com/longhorn/longhorn/commit/9cda8e6a64ab91002fbb2a4aef2ff93134f48b36)

The value(`ignore`, `enabled`, `disabled`) set in `persistence.unmapMarkSnapChainRemoved` in chart/values.yaml is applied to storageclass `longhorn` and carried through to dynamically provisioned volumes.

PR #11272, opened 2025-07-11
fix(chart): fix wrong setting value and storageClass parameter

derekbit:

Issue Longhorn/longhorn#11264

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -207,7 +207,7 @@ persistence:
     # -- Node selector for the default Longhorn StorageClass. Longhorn uses only nodes with the specified tags for storing volume data. (Examples: "storage,fast")
     selector: ""
   # -- Setting that allows you to enable automatic snapshot removal during filesystem trim for a Longhorn StorageClass. (Options: "ignored", "enabled", "disabled")
-  removeSnapshotsDuringFilesystemTrim: ignored
+  unmapMarkSnapChainRemoved: ignored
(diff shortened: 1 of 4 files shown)


derekbit:
@mergify backport v1.9.x v1.8.x

chart/values.yaml
```diff
@@ -207,7 +207,7 @@ persistence:
     # -- Node selector for the default Longhorn StorageClass. Longhorn uses only nodes with the specified tags for storing volume data. (Examples: "storage,fast")
     selector: ""
   # -- Setting that allows you to enable automatic snapshot removal during filesystem trim for a Longhorn StorageClass. (Options: "ignored", "enabled", "disabled")
-  removeSnapshotsDuringFilesystemTrim: ignored
+  removeSnapshotsDuringFilesystemTrim: "false"
```
mantissahz:
```suggestion
  unmapMarkSnapChainRemoved: ignored
```
Here are the `persistence` settings for the StorageClass as you modified in the `templates/storagclass.yaml`.
It is still an enum for `ignore`, `enabled` and `disabled.` [Volume.Spec.UnmapMarkSnapChainRemoved](https://github.com/longhorn/longhorn-manager/blob/12563caa8a8eeb28ec808b793b9b0ecff8fe48a1/k8s/pkg/apis/longhorn/v1beta2/volume.go#L76-L82)

We also need to modify the README.md and questions.yaml.

derekbit: Good catch! Fixed. Thank you.

mantissahz (approved):
LGTM

Issue #11179, opened 2025-06-26
[IMPROVEMENT] Add the possiblity of setting floating point values for `guaranteed-instance-manager-cpu` and `node.spec.instanceManagerCPURequest`


gigabyte132:
### Is your improvement request related to a feature? Please describe (👍 if you like this request)

Following the discussion on slack https://cloud-native.slack.com/archives/CNVPEL9U3/p1750701390272929 , I was suggested to open an improvement ticket to add additional functionality to the `guaranteed-instance-manager-cpu` and `node.spec.instanceManagerCPURequest` variables.

From slack, to describe our current usage of longhorn and setup:  

We are currently running longhorn 1.8.1 on a mostly GPU Kubernetes cluster. We are using longhorn to facilitate an in-cluster shared filesystem across the GPU nodes, that multiple workloads can use at the same time. We are currently trying to setup adequate CPU allocations for workloads that request GPUs. At the moment we are running into the fact that the longhorn instance manager, by default requests 12% of the cpu capacity available on the node, which in high core (384) environments this is quite a lot. We understand that this parameter is configurable via instanceManagerCPURequest , although from our understanding this can only take integer values. In our case, even 1% ( 3.84) cores would be more than we would like. I went through the longhorn manager code where the instanceManagerCPURequest  is calculated/parsed, and it indeed seems that we cannot set it to floating point values.

### Describe the solution you'd like

Have the ability to set floating point values for `guaranteed-instance-manager-cpu` and `node.spec.instanceManagerCPURequest`

derekbit:
cc @DamiaSan (community coordinator)

DamiaSan:
Since Kubernetes let we to express CPU requests in millicores, so it is a fine grain, I think that we could let the user to express  cpu limits with a floating point percentage.
What do you think @derekbit ?

derekbit:
This is feasible. Let's add it to Longhorn Sprint. Thanks.

derekbit:
@gigabyte132 Would you like to contribute the improvement?

gigabyte132:
> [@gigabyte132](https://github.com/gigabyte132) Would you like to contribute the improvement?

Sure @derekbit , sounds good!

DamiaSan:
A matter can be the backward compatibility of the change from integer to floating point, could you make some test @gigabyte132 ?

derekbit:
Hello @gigabyte132 
[...]

PR #11278, opened 2025-07-11
chore(chart): update guaranteedInstanceManagerCPU type to float

gigabyte132:

Issue #11179 

This PR updates the data type for `guaranteedInstanceManagerCPU` from int to float, following the update done in https://github.com/longhorn/longhorn-manager/pull/3892


derekbit (approved):
LGTM. Thank you @gigabyte132

Issue #11283, opened 2025-07-11
[CI] Replace PAT with longhorn-github-bot APP Token for Workflows


derekbit:
## What's the CI task to develop? Please describe

Replacing the PAT with the longhorn-github-bot App Token improves security through short-lived, limited-scope tokens. It avoids dependency on personal accounts, ensures more stable automation, supports cross-repo actions, and provides clear audit trails.

## Describe the items of the task development (DoD, definition of done) you'd like

- [x] https://github.com/longhorn/longhorn-manager/blob/master/.github/workflows/create-crd-update-pr-in-longhorn-repo.yml
    - https://github.com/longhorn/longhorn-manager/pull/3909
    - https://github.com/longhorn/longhorn-manager/pull/3913
- [x] https://github.com/longhorn/dep-versions/blob/master/.github/workflows/create-dep-versions-update-pr-in-longhorn-repo.yml
    - https://github.com/longhorn/dep-versions/pull/75
- [x] https://github.com/longhorn/dep-versions/blob/master/.github/workflows/trigger-builds.yaml
    - https://github.com/longhorn/dep-versions/pull/75
    - https://github.com/longhorn/dep-versions/pull/76 
- [x] https://github.com/longhorn/longhorn/blob/master/.github/workflows/update-community-issue.yml
    - https://github.com/longhorn/longhorn/pull/11294
    - https://github.com/longhorn/longhorn/pull/11295 
- [x] https://github.com/longhorn/longhorn/blob/master/.github/workflows/update-longhorn-issue.yml
    - https://github.com/longhorn/longhorn/pull/11294
    - https://github.com/longhorn/longhorn/pull/11295 
- [x] https://github.com/longhorn/longhorn/blob/master/.github/workflows/create-issue.yml
    - https://github.com/longhorn/longhorn/pull/11296 
- [x] https://github.com/longhorn/longhorn/blob/master/.github/workflows/add-issue-to-projects.yml
    - https://github.com/longhorn/longhorn/pull/11296

derekbit:
cc @Yu-Jack @innobead

derekbit:
Completed.

innobead:
Thanks @derekbit

PR #11294, opened 2025-07-14
chore(workflow): replace PAT with longhorn-github-bot APP Token

derekbit:

Issue Longhorn/longhorn#11283

### DIFF
--- a/.github/workflows/update-community-issue.yml
+++ b/.github/workflows/update-community-issue.yml
@@ -13,6 +13,15 @@ jobs:
   community:
     runs-on: ubuntu-latest
     steps:
+    - id: app-token
+      uses: actions/create-github-app-token@v2
+      with:
+        app-id: ${{ secrets.LONGHORN_GITHUB_BOT_APP_ID }}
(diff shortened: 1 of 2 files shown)


.github/workflows/update-community-issue.yml
```diff
@@ -13,6 +13,15 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
+    - id: app-token
+      uses: actions/create-github-app-token@v2
+      with:
+        app-id: ${{ secrets.LONGHORN_GITHUB_BOT_APP_ID }}
+        private-key: ${{ secrets.LONGHORN_GITHUB_BOT_PRIVATE_KEY }}
+        owner: ${{ github.repository_owner }}
+        permission-contents: write
+        permission-pull-requests: write
```
c3y1huang, line 23: Does this also need `permission-actions: write`?
https://github.com/longhorn/dep-versions/pull/76/files

derekbit: I think no, because the workflow doesn't need to call other actions.

c3y1huang (approved):

PR #11296, opened 2025-07-14
chore(workflow): replace PAT with longhorn-github-bot APP Token

derekbit:

Issue Longhorn/longhorn#11283


innobead (approved):

innobead:
@derekbit need to follow up on checking if this works as intended .

derekbit:
> @derekbit need to follow up on checking if this works as intended .

Sure. The permissions are very granular and need to be adjusted, unlike a traditional PAT which typically has broad and powerful access.

Issue #11270, opened 2025-07-10
[IMPROVEMENT] Enhance Offline Rebuilding with Resource Awareness and Retry Backoff


derekbit:
### Is your improvement request related to a feature? Please describe (👍 if you like this request)

When the offline rebuilding feature is enabled, the system will automatically attach a detached but degraded volume and attempt to rebuild its replicas. However, if there are not enough available nodes or disks, the volume remains in an attached state without the awareness of the insufficient resources.

Apply the manifest in the one-node cluster
```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: longhorn-test
provisioner: driver.longhorn.io
allowVolumeExpansion: true
reclaimPolicy: Delete
volumeBindingMode: Immediate
parameters:
  numberOfReplicas: "1"
  staleReplicaTimeout: "2880"
  fromBackup: ""
  fsType: "ext4"
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: longhorn-volv-pvc
  namespace: default
spec:
  accessModes:
    - ReadWriteOnce
  storageClassName: longhorn-test
  resources:
    requests:
      storage: 2Gi
```

You can observe 
```
NAME                                       DATA ENGINE   STATE   ROBUSTNESS   SCHEDULED   SIZE         NODE   AGE
pvc-30ecd3be-2522-4ef9-bdc9-8b71e85aa534   v1                                             2147483648          0s
pvc-30ecd3be-2522-4ef9-bdc9-8b71e85aa534   v1                                             2147483648          0s
pvc-30ecd3be-2522-4ef9-bdc9-8b71e85aa534   v1            creating                            2147483648          0s
pvc-30ecd3be-2522-4ef9-bdc9-8b71e85aa534   v1            detaching   unknown                  2147483648          0s
pvc-30ecd3be-2522-4ef9-bdc9-8b71e85aa534   v1            detached    unknown                  2147483648          1s
pvc-30ecd3be-2522-4ef9-bdc9-8b71e85aa534   v1            detached    unknown                  2147483648          1s
pvc-30ecd3be-2522-4ef9-bdc9-8b71e85aa534   v1            attaching   unknown                  2147483648          1s
pvc-30ecd3be-2522-4ef9-bdc9-8b71e85aa534   v1            attaching   unknown                  2147483648          2s
[...]

PR #11356, opened 2025-07-21
feat(lep): resource awareness and backoff retry for offline rebuilding

mantissahz:

Issue # longhorn/longhorn#11270


enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
```diff
@@ -0,0 +1,151 @@
...
+  ...
+  return len(replicaDiskCandidates) > 0, nil
+}
+```
+
+### Backoff Retry
+
+Follow the volume condition `Scheduled` that means a new replica can be scheduled to a disk candidate.
+
+When there is a volume rebuild VA ticket and the volume is attached, the volume rebuild controller will check if the rebuilding is starting. It will have a backoff retry if the rebuilding is not starting.
```
shuo-wu: I cannot understand how Longhorn achieves the rebuilding backoff relying on condition `Scheduled`?

BTW, will we use a single backoff for both online rebuilding and offline rebuilding?

enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
```diff
@@ -0,0 +1,151 @@
...
+      // There is no reusable replica or disk candidate to schedule a new replica 
+      return va, nil
+    }
+    createOrUpdateAttachmentTicket(va, attachmentID, vol.Status.OwnerID, longhorn.AnyValue, longhorn.AttacherTypeVolumeRebuildingController)
+  }
+  ...
+}
+...
+func (vbc *VolumeRebuildingController) isVolumeReplicasReusableOrSchedulable(vol *longhorn.Volume, rs map[string]*longhorn.Replica) (bool, error) {
+  reusableFailedReplica, err := vbc.scheduler.CheckAndReuseFailedReplica(rs, vol, "")
```
shuo-wu, line 92: Shouldn't we consider `hardNodeAffinity` ?

enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
mantissahz: The volume condition `VolumeConditionTypeScheduled` is used to check if the replica can be scheduled.

https://github.com/longhorn/longhorn-manager/blob/5bb438e64f5399c9a690afabca689d68797777c3/controller/volume_controller.go#L1744-L1835

https://github.com/longhorn/longhorn-manager/blob/5bb438e64f5399c9a690afabca689d68797777c3/controller/volume_controller.go#L2329-L2348

> BTW, will we use a single backoff for both online rebuilding and offline rebuilding?

Modified the description that the replica scheduled check (resource check) offline rebuilding follows the volume replica schedule mechanism, including the replenishment backoff.

enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
mantissahz: Because the volume is detached, the `volume.Spec.NodeID` and `engine.Spec.NodeID` will be empty. And the volume state is degraded, and the healthy replica count is less than `volume.Spec.numberOfReplicas`, so I believe we can follow this way like the method `replenishReplicas` in https://github.com/longhorn/longhorn-manager/blob/5bb438e64f5399c9a690afabca689d68797777c3/controller/volume_controller.go#L829-L852

enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Proposal
+
+We will add a resource awareness mechanism into the offline rebuilding flow, so it will only mention the modified parts from the [previous proposal](./20250407-volume-offline-rebuilding#proposal).
+
+### User Stories
+
+- Users want to automatically rebuild replicas while a degraded volume is detached to ensure maintain data redundancy.
+- A worker node is up with the offline rebuilding enabled.
+- A worker node is down during the offline rebuilding process.
```
derekbit: What do they mean? Can you elaborate more on the two points?

enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
```diff
@@ -0,0 +1,151 @@
...
+
+#### Enable Global Volumes Or An Individual Volume Offline Rebuilding
+
+When offline rebuilding is enabled and volumes are degraded, the degraded volumes will be attached until the volume becomes healthy from the [previous design](./20250407-volume-offline-rebuilding#manually-enable-an-individual-volume-offline-rebuilding). It is not reasonable to keep the degraded volumes attached if the resources is insufficient, for example, only 2 schedulable nodes or disks for 3 replicas.
+
+Therefore, before the volume rebuild controller starts to attach the volume, it will check if there are the disk candidates. If there is no disk candidate, the replica rebuilding will not start until the resources are sufficient.
+
+#### A Worker Node Down During Volume Offline Rebuilding
+
+If a worker node goes down and resources become insufficient, users must wait for an interval to confirm that the resources are indeed unavailable for rebuilding the replica.
```
derekbit: When and how to check resources are insufficient when a worker node is down?

> Users must wait for an interval to confirm

How to determine the interval?

enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
```diff
@@ -0,0 +1,151 @@
...
+Then the volume will be detached automatically.
+
+#### A Worker Node Up When Volume Offline Rebuilding Enabled
+
+When the offline rebuilding is enabled and a worker node get ready in the cluster, degraded volumes will be attached to rebuild the replica, provided that the node and its disk are schedulable. After the replica on this node is rebuilt, the volume will be detached if it is healthy or wait for an interval to confirm that the resources are insufficient for rebuilding another replica and then the volume will be detached as well.
```
derekbit: Same question
When and how to check resources are sufficient? Need to reveal more details.

enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
```diff
@@ -0,0 +1,151 @@
...
+  ...
+  replicaDiskCandidates, multiError, err := vbc.scheduler.FindDiskCandidates(replica, rs, vol)
+  ...
+  return len(replicaDiskCandidates) > 0, nil
+}
+```
+
+### Backoff Retry
+
+Follow the volume condition `Scheduled` that means a new replica can be scheduled to a disk candidate.
```
derekbit: The statement is not easy to understand.
Volume `Scheduled` false means it needs one or more replicas.

enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
```diff
@@ -0,0 +1,151 @@
...
+
+When there is a volume rebuild VA ticket and the volume is attached, the volume rebuild controller will check if the rebuilding is starting by the volume condition `Scheduled`. Volume will be detached and if the volume condition `Scheduled` is `False`.
```
derekbit, line 159: Why does the volume need to be detached if `Schedule` is `False`?

enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
```diff
@@ -0,0 +1,151 @@
...
+    return true, nil
+  }
+  ...
+  replicaDiskCandidates, multiError, err := vbc.scheduler.FindDiskCandidates(replica, rs, vol)
+  ...
+  return len(replicaDiskCandidates) > 0, nil
+}
+```
+
+### Backoff Retry
```
derekbit, line 105: 1. Backoff flow is not elaborated in the section.
2. Should we introduce backoff settings?

enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
```diff
@@ -0,0 +1,151 @@
...
+| `ignored` | ✓ | X |
+| `enabled` | ✓ | ✓ |
+| `disabled` | X | X |
+
+- ✓: The offline rebuilding is enabled.
+- X: The offline rebuilding is disabled.
+
+#### Enable Global Volumes Or An Individual Volume Offline Rebuilding
+
+When offline rebuilding is enabled and volumes are degraded, the degraded volumes will be attached until the volume becomes healthy from the [previous design](./20250407-volume-offline-rebuilding#manually-enable-an-individual-volume-offline-rebuilding). It is not reasonable to keep the degraded volumes attached if the resources is insufficient, for example, only 2 schedulable nodes or disks for 3 replicas.
```
innobead:
```suggestion
When offline rebuilding is enabled and volumes are degraded, the degraded volumes will be attached until the volume becomes healthy from the [previous design](./20250407-volume-offline-rebuilding#manually-enable-an-individual-volume-offline-rebuilding). It is not reasonable to keep repeating attaching degraded volumes if the resources are insufficient, for example, only 2 schedulable nodes or disks for 3 replicas.
```

enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
mantissahz: We attach the volume to the node if the volume is degraded and the `offline-replica-rebuilding` setting is enabled. The `hardNodeAffinity` will be considered as the general replica rebuilding flow, and we depend on the volume condition `Scheduled` to check if the replica can be scheduled to wait for rebuilding or detach the volume.

enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
mantissahz: Modified. Thanks.

enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
mantissahz: Added the descriptions and partial code flow of the volume controller. Thanks.

enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
mantissahz: Modified the descriptions. Thanks

enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
mantissahz: Modified the descriptions. Thanks

enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
mantissahz: Modified the descriptions. Thanks

enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
mantissahz: Modified the descriptions. Thanks

enhancements/20250721-volume-offline-rebuilding-with-resource-awareness.md
mantissahz: Modified the descriptions. Thanks

PR #11373, opened 2025-07-24
docs(lep): v2 volume expansion #8022

davidcheng0922:

Issue [#8022](https://github.com/longhorn/longhorn/issues/8022)

Propose a design to support live expanding for v2 volume

### DIFF
--- a/enhancements/20250721-v2-volume-expansion.md
+++ b/enhancements/20250721-v2-volume-expansion.md
@@ -1,0 +1,265 @@
+# V2 Volume Expansion (Nvmf frontend)
+
+## Summary
+
+Support for user to expand v2 volumes with nvmf frontend.
+
+### Related Issues
+
+- https://github.com/longhorn/longhorn/issues/8022
+
+## Motivation
+
+- Promote Longhorn Engine v2 from experimental to production-ready by implementing missing features, improving stability, and achieving functional parity with Engine v1.
+
+### Goals
+
+- Allow users to perform online expansion of Engine v2 volumes with the NVMe-oF frontend.
+- Ensure the expansion workflow and logic are aligned with Engine v1.
+
+### Non-goals
+
+- Support for online expansion of v2 volumes with the `ublk` frontend.
+
+## Proposal
+
+### User Stories
+
+- V2 volume expansion is needed for  general users
+
+### API changes
+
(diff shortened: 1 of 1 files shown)


enhancements/20250721-v2-volume-expansion.md
```diff
@@ -0,0 +1,190 @@
...
+- Promote Longhorn Engine v2 from experimental to production-ready by implementing missing features, improving stability, and achieving functional parity with Engine v1.
+
+### Goals
+
+- Allow users to perform online expansion of Engine v2 volumes with the NVMe-oF frontend.
+- Ensure the expansion workflow and logic are aligned with Engine v1.
+
+### Non-goals
+
+- Support for online expansion of v2 volumes with the `ublk` frontend.
```
PhanLe1010, line 22: Is there a blocker for this?

davidcheng0922: In my previous experience, when deleting the underlying RAID bdev, the associated ublk device immediately disappears or becomes invalid—even if `dmsetup suspend` has been applied on top of it. As a result, errors occur when attempting to recreate the frontend with the same ublk device.

I suspect this is because the current `ublk kernel module` does not support dynamic backend replacement. Therefore, the ublk device must be explicitly removed and re-added in order to bind it to the newly recreated RAID bdev.

There might be a better solution to handle this scenario, but I'm not sure if it's currently supported.

davidcheng0922: I may need more research on it

PhanLe1010 (commented):
The overall design LGTM. Only 1 question left is about the dm device suspension approach to support both nvmf and ublk frontend as we discussed in the LEP meeting

e -i HEAD~2
```diff
@@ -0,0 +1,15781 @@
+[33mcommit 24357b56a060c4bfb2b3f6534fc1438df99bb57b[m[33m ([m[1;36mHEAD[m[33m -> [m[1;32missue-11345-rbac-permission[m[33m, [m[1;31morigin/issue-11345-rbac-permission[m[33m)[m
```
c3y1huang: What is this file for?

davidcheng0922: sorry, it's a mistake

Issue #11452, opened 2025-08-11
[BUG] Helm chart does not apply customized values


johnrichter:
### Describe the Bug

The default-settings configmap is not populated when running Helm install. This is causing problems when installing longhorn on a cluster. I need to use `createDefaultDiskLabeledNodes: true` and node annotations/labels to set up the disks on my hosts, but the chart validates each `defaultSettings` option with `KindIs "invalid"` which, as far as I understand, will never set the default settings. Strangly, when I immediately run a helm upgrade after the install, the settings do appear in the config map, but its too late for my disk settings to apply and I have to do everything manually

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: longhorn-default-setting
  namespace: {{ include "release_namespace" . }}
  labels: {{- include "longhorn.labels" . | nindent 4 }}
data:
  default-setting.yaml: |-
    {{- if not (kindIs "invalid" .Values.defaultSettings.allowRecurringJobWhileVolumeDetached) }}
    allow-recurring-job-while-volume-detached: {{ .Values.defaultSettings.allowRecurringJobWhileVolumeDetached }}
    {{- end }}
    {{- if not (kindIs "invalid" .Values.defaultSettings.createDefaultDiskLabeledNodes) }}
    create-default-disk-labeled-nodes: {{ .Values.defaultSettings.createDefaultDiskLabeledNodes }}
    {{- end }}
...
```

As far as the data type for these fields, its confusing. The docs suggest the bool primitive type, but its defined as a string in other places in this repo. https://github.com/longhorn/longhorn/blob/master/deploy/upgrade_responder_server/chart-values.yaml#L142-L144
```
        "longhornSettingCreateDefaultDiskLabeledNodes": {
          "dataType": "string",
          "maxLen": 200
        },
```

### To Reproduce

Just try to install Longhorn via Helm with any default value modified. Here's mine

```yaml
defaultSettings:
  createDefaultDiskLabeledNodes: "true"
  defaultDataLocality: best-effort
  replicaAutoBalance: least-effort # longhorn best practices recommendation
  storageMinimalAvailablePercentage: 10 # percentage
  storageOverProvisioningPercentage: 100 # percentage
  v1DataEngine: "true"
  v2DataEngine: "true"
  allowRecurringJobWhileVolumeDetached: "true"
  snapshotMaxCount: 12
  logLevel: Debug
```

### Expected Behavior

Longhorn deploys with helm chart values correctly.

[...]

PR #11474, opened 2025-08-13
fix(chart): Update Helm templates to type check correctly

johnrichter:

Issue #11452 

Updates template logic to properly validate that chart values are the correct primitive type (string) for `defaultSettings` values.

Just a simple update to move from accepting any value that isn't the Go type `invalid` to moving to checking if values are the Go type `string`. Only inserts the setting if the value is a `string`.

https://helm.sh/docs/chart_template_guide/function_list/#kind-functions

I didn't add any validation errors. If values are not strings, the settings are silently dropped. This would be a future improvement to the Helm chart.

I'm based off of the v1.9.1 tag. Maybe I missed a contribution guideline. If so let me know and I'll rebase off the correct branch. Many of the listed changes are not mine and are being pulled in from the git diff between `v1.9.1` and `master`


derekbit:
@johnrichter Could you fix the conflicts ?Thank you.

derekbit:
@johnrichter Feel free to reopen and update the PR if you want to continue the improvement and fix. Thank you.

Issue #5948, opened 2023-05-17
[IMPROVEMENT] Launching a new mechanism to collect instance manager logs


shuo-wu:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

The reason why we need to launch a new kind of instance manager log handling mechanism:
1. In SPDK engine server, we cannot fetch the log of a specific SPDK engine or replica. `spdk_tgt` is not managed by this server, and `spdk_tgt` does not provide such an option to fetch logs for a specific object. Hence, there is no way to have a API like `ProcessLog`, which helps collects logs for crashed engines/replicas.
2. In the current implementation, each time when there is an engine/replica crash, the longhorn manager log will be flooded. It’s inconvenient for debugging. If we can directly check the instance manager log, there is no need to fetch the instance manager logs to longhorn managers.

## Describe the solution you'd like
The reason why Longhorn fetches the instance manager logs to longhorn managers is that, the instance manager pods may crash and the log may be lost. Once we can **persist** instance manager logs, we can deprecate the current log collection mechanism:
1. Mount and use a host path to store instance manager logs.
2. Split the log by day
3. Retain 7-day logs only. And if the log size exceeds the limit (1Gi for example), we can remove the oldest logs in advance.
4. Include the instance manager log files during support bundle generation

## Describe alternatives you've considered

N/A

## Additional context

cc @derekbit @innobead

innobead:
This should be for all `managed` pods by longhorn, and the log collector pod should be a pod managed by K8s controller (deployment) w/ ephemeral logging.

hookak:
I have a related requirement regarding this topic.

Currently, I simply want to persist the logs of the instance-manager pod that runs spdk_tgt to a hostPath volume. The reason is that when Longhorn Manager detects an issue with an instance manager, it deletes and recreates the pod. In this process, any logs from spdk_tgt are lost unless we use an external logging system.

This becomes especially problematic during debugging. For example, in v1.9.0, I've observed that the instance-manager pod may be deleted during rebuilds, making it difficult to investigate what actually went wrong with SPDK.

I’d like to ask if there is any plan to support persisting instance-manager logs to a hostPath or similar mechanism in a future release. This would be very helpful for debugging issues involving SPDK. @shuo-wu @derekbit

shuo-wu:
How about simply introducing a setting that indicates the host path for something like instance manager pod logs? The default value can be the same as (v1) default disk path.

derekbit:
[...]

PR #11523, opened 2025-08-20
feat: add log-path setting

derekbit:

Issue Longhorn/longhorn#5948

### DIFF
--- a/chart/templates/default-setting.yaml
+++ b/chart/templates/default-setting.yaml
@@ -191,6 +191,9 @@ data:
     {{- if not (kindIs "invalid" .Values.defaultSettings.logLevel) }}
     log-level: {{ .Values.defaultSettings.logLevel }}
     {{- end }}
+    {{- if not (kindIs "invalid" .Values.defaultSettings.logPath) }}
+    log-path: {{ .Values.defaultSettings.logPath }}
+    {{- end }}
     {{- if not (kindIs "invalid" .Values.defaultSettings.backupCompressionMethod) }}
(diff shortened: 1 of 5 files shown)


chart/values.yaml
```diff
@@ -354,6 +354,8 @@ defaultSettings:
   backupConcurrentLimit: ~
   # -- Specifies the default backup block size, in MiB, used when creating a new volume. Supported values are 2 or 16.
   defaultBackupBlockSize: ~
+  # -- Specifies the directory on the host where Longhorn stores log files for the instance manager pod. Currently, it is only used for instance manager pods in the v2 data engine.
+  logPath: ~
```
c3y1huang: Should we move this next to `logLevel`?

derekbit: Good idea. Updated. Thank you.

chart/templates/default-setting.yaml
```diff
@@ -203,6 +203,9 @@ data:
     {{- if not (kindIs "invalid" .Values.defaultSettings.defaultBackupBlockSize) }}
     default-backup-block-size: {{ .Values.defaultSettings.defaultBackupBlockSize }}
     {{- end }}
+    {{- if not (kindIs "invalid" .Values.defaultSettings.logPath) }}
+    log-path: {{ .Values.defaultSettings.logPath }}
+    {{- end }}
```
c3y1huang: Move to next to `log-level`?

derekbit: Updated. Thank you.

COLDTURNIP (approved):

PR #11592, opened 2025-08-29
fix: update image for data migration yaml

shuo-wu:

Some customers cannot pull this outdated image from the external. I think we can use the base image of longhorn components instead. (Or we can use `registry.suse.com/bci/bci-base:15.6`)


PhanLe1010 (approved):
LGTM

derekbit (approved):

Issue #7066, opened 2023-11-08
[FEATURE] Support v2 volume without hugepage


derekbit:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

hugepage is evitable for supporting v2 volume feature. Although huge page can improve the IO performance, but there are some drawbacks

- not easy to modify the size dynamically 
- dedicated memory size (waste)
- memory pressure for low-spec machines

There is a discussion in https://github.com/spdk/spdk/issues/2973 for the no hugepage support.  Although it is not implemented yet, but we can continue investigating the design, implementing it and integrating it into v2 volume.

derekbit:
Assigned to @c3y1huang. She will dive into the spdk internals.
cc @innobead

derekbit:
Applied the PRs https://github.com/longhorn/longhorn-manager/pull/4045 and https://github.com/longhorn/longhorn-instance-manager/pull/1069 and enabling v2 data engine, 

The two values, VmPeak and VmSize, are 2188696 kB if we set the data-engine-memory-size to 2048 MiB. 

Note: VmPeak Peak is virtual memory size, and VmSize is virtual memory size.

```
instance-manager-be923d729f85982832a698b93e0c0491:/ # ps aux
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root         1  0.0  0.0   2680  1132 ?        Ss   04:35   0:00 /tini -- instance-manager --spdk-log all --spdk-cpumask 0x1 --spdk-memory-size 2048 --enable-spdk --debug daemon --spdk-enabled --listen 0.0.0.0:8500 --spdk-no-h
root         7  0.0  0.0   4272  3620 ?        S    04:35   0:00 /bin/bash /usr/local/bin/instance-manager --spdk-log all --spdk-cpumask 0x1 --spdk-memory-size 2048 --enable-spdk --debug daemon --spdk-enabled --listen 0.0.0.0:
root        17 99.3  3.6 2188696 282732 ?      RLl  04:35   1:40 spdk_tgt -L all -m 0x1 --mem-size 2048 --no-huge
root        18  0.0  0.0   2696  1288 ?        S    04:35   0:00 tee -a /log/spdk_tgt.log
root        23  0.0  0.0   4272  1704 ?        S    04:35   0:00 /bin/bash /usr/local/bin/exec-logrotate
[...]

PR #11685, opened 2025-09-05
feat(setting): support v2 data engine without hugepage

derekbit:

Issue Longhorn/longhorn#7066


mantissahz (approved):

derekbit:
@mergify backport v1.10.x

innobead:
cc @Vicente-Cheng @tserong see if need to fix on Harvester side, as there is an integration there with v2.

Issue #11810, opened 2025-09-18
[BUG] failed to load DataEngineSpecific boolean setting from configmap


COLDTURNIP:
### Describe the Bug

If a setting is engine specific boolean, and overridden in the configmap `longhorn-default-setting`, Longhorn manager will reject to load this customized setting:

```
longhorn-manager-98lgf longhorn-manager time="2025-09-18T08:49:11.7956391Z" level=error msg="Invalid value {\"v2\" => \"true\"} for the boolean setting data-engine-interrupt-mode-enabled" func=types.GetCustomizedD
efaultSettings file="setting.go:1870" error="strconv.ParseBool: parsing \"{\\\"v2\\\" => \\\"true\\\"}\": invalid syntax"
longhorn-manager-98lgf longhorn-manager time="2025-09-18T08:49:11.797109055Z" level=error msg="Invalid value {\"v2\" => \"false\"} for the boolean setting data-engine-hugepage-enabled" func=types.GetCustomizedDefa
ultSettings file="setting.go:1870" error="strconv.ParseBool: parsing \"{\\\"v2\\\" => \\\"false\\\"}\": invalid syntax"
longhorn-manager-s66h7 longhorn-manager time="2025-09-18T08:49:12.382715747Z" level=error msg="Invalid value {\"v2\" => \"false\"} for the boolean setting data-engine-hugepage-enabled" func=types.GetCustomizedDefa
ultSettings file="setting.go:1870" error="strconv.ParseBool: parsing \"{\\\"v2\\\" => \\\"false\\\"}\": invalid syntax"
longhorn-manager-8fqb9 longhorn-manager time="2025-09-18T08:47:40.62243316Z" level=error msg="Invalid value {\"v2\" => \"false\"} for the boolean setting data-engine-hugepage-enabled" func=types.GetCustomizedDefau
ltSettings file="setting.go:1870" error="strconv.ParseBool: parsing \"{\\\"v2\\\" => \\\"false\\\"}\": invalid syntax"
longhorn-manager-s66h7 longhorn-manager time="2025-09-18T08:49:12.384976638Z" level=error msg="Invalid value {\"v2\" => \"false\"} for the boolean setting data-engine-hugepage-enabled" func=types.GetCustomizedDefa
ultSettings file="setting.go:1870" error="strconv.ParseBool: parsing \"{\\\"v2\\\" => \\\"false\\\"}\": invalid syntax"
[...]

PR #11826, opened 2025-09-21
fix(settings): quote values of settings

derekbit:

Issue Longhorn/longhorn#11810

### DIFF
--- a/chart/templates/default-setting.yaml
+++ b/chart/templates/default-setting.yaml
@@ -7,64 +7,144 @@ metadata:
 data:
   default-setting.yaml: |-
     {{- if not (kindIs "invalid" .Values.defaultSettings.allowRecurringJobWhileVolumeDetached) }}
-    allow-recurring-job-while-volume-detached: {{ .Values.defaultSettings.allowRecurringJobWhileVolumeDetached }}
+    {{- if kindIs "string" .Values.defaultSettings.allowRecurringJobWhileVolumeDetached }}
+    allow-recurring-job-while-volume-detached: {{ .Values.defaultSettings.allowRecurringJobWhileVolumeDetached | quote }}
+    {{- else }}
+    {{- fail "defaultSettings.allowRecurringJobWhileVolumeDetached must be a string" }}
+    {{- end }}
     {{- end }}
     {{- if not (kindIs "invalid" .Values.defaultSettings.createDefaultDiskLabeledNodes) }}
-    create-default-disk-labeled-nodes: {{ .Values.defaultSettings.createDefaultDiskLabeledNodes }}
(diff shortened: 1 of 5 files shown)


derekbit:
@mergify backport v1.10.x

COLDTURNIP (approved):

deploy/longhorn.yaml
```diff
@@ -77,8 +77,8 @@ metadata:
     app.kubernetes.io/version: v1.9.0-dev
 data:
   default-setting.yaml: |-
-    priority-class: longhorn-critical
-    disable-revision-counter: true
+    priority-class: "longhorn-critical"
+    disable-revision-counter: "{\"v1\":\"true\"}"
```
COLDTURNIP, line 81: NIT: maybe single quote is more readable
```suggestion
    disable-revision-counter: '{"v1":"true"}'
```

deploy/longhorn-okd.yaml
```diff
@@ -79,8 +79,8 @@ metadata:
     app.kubernetes.io/version: v1.9.0-dev
 data:
   default-setting.yaml: |-
-    priority-class: longhorn-critical
-    disable-revision-counter: true
+    priority-class: "longhorn-critical"
+    disable-revision-counter: "{\"v1\":\"true\"}"
```
COLDTURNIP, line 83: NIT: maybe single quote is more readable
```suggestion
    disable-revision-counter: '{"v1":"true"}'
```

chart/templates/default-setting.yaml
```diff
@@ -87,179 +167,415 @@ data:
...
+    {{- else }}
+    {{- fail "defaultSettings.replicaRebuildingBandwidthLimit must be a string" }}
+    {{- end }}
     {{- end }}
     {{- if not (kindIs "invalid" .Values.defaultSettings.instanceManagerPodLivenessProbeTimeout) }}
-    instance-manager-pod-liveness-probe-timeout: {{ .Values.defaultSettings.instanceManagerPodLivenessProbeTimeout }}
+    {{- if kindIs "string" .Values.defaultSettings.instanceManagerPodLivenessProbeTimeout }}
+    instance-manager-pod-liveness-probe-timeout: {{ .Values.defaultSettings.instanceManagerPodLivenessProbeTimeout | quote }}
+    {{- else }}
+    {{- fail "defaultSettings.instanceManagerPodLivenessProbeTimeout must be a string" }}
```
innobead, line 579: As many setting value by default is ~, which means null. Will the fail statement cause the template rendering failed?

derekbit: No, don't see any failure.

innobead: Yes, overlooked. The invalid check is still kept.

Issue #11615, opened 2025-09-01
[IMPROVEMENT] volume `spec.backingImage` and `spec.encrypted` shouldn't allow to update for both v1 and v2 data engines


derekbit:
### Is your improvement request related to a feature? Please describe (👍 if you like this request)

The volume fields `spec.backingImage` and `spec.encrypted` should not be updatable:
- Modifying `spec.backingImage` may result in a corrupted filesystem.
- Longhorn does not support converting a non-encrypted volume to an encrypted one, or vice versa.

Ref: https://github.com/longhorn/longhorn-manager/pull/4024#issuecomment-3242656718

derekbit:
Hello @yulken 
The PRs are merged. Could you update https://github.com/longhorn/longhorn/issues/11615#issuecomment-3293066646 and move the status to `Ready For Testing`? Thank you.

<img width="381" height="144" alt="Image" src="https://github.com/user-attachments/assets/1931af98-8d66-457f-84a1-4192eed53b87" />

yulken:
@derekbit 
I updated the comment but I'm unable to move the card. It seems I'm lacking permissions for that.

roger-ryao:
**Verified on master-head 20251017**
- longhorn master-head https://github.com/longhorn/longhorn/commit/fecc3bb87c3e6cc59d9e4ce95fa2dd8fa861f6df
- longhorn-manager master-head https://github.com/longhorn/longhorn-manager/commit/e808602f1e03c505a161a3bb8c70022ee53d08e9


**The test steps**
1. https://github.com/longhorn/longhorn/issues/11615#issuecomment-3293066646

**Result Passed**

Below are the commands and results: 

```bash
# Attempt to update spec.backingImage & spec.encrypted (expected failure)
ryao@ryao-laptop:~/Downloads/test-record> kubectl -n longhorn-system patch volume vol-3   --type=merge   -p '{"spec":{"encrypted":false}}'
The Volume "vol-3" is invalid: spec.encrypted: Invalid value: "boolean": Encrypted is immutable
ryao@ryao-laptop:~/Downloads/test-record> kubectl -n longhorn-system patch volume vol-0   --type=merge   -p '{"spec":{"encrypted":true}}'
The Volume "vol-0" is invalid: spec.encrypted: Invalid value: "boolean": Encrypted is immutable
[...]

PR #11838, opened 2025-09-22
feat: set encrypted and backingImage as immutable

yulken:

Issue https://github.com/longhorn/longhorn/issues/11615


COLDTURNIP (approved):
LGTM

Issue #11854, opened 2025-09-24
[BUG] [v1.10.0-rc4] charts - checks `if kindIs "string" ...`


mcerveny:
### Describe the Bug

Upgrade with helm charts fails with error `Error: UPGRADE FAILED: execution error at (longhorn/templates/default-setting.yaml:229:8): defaultSettings.detachManuallyAttachedVolumesWhenCordoned must be a string
`. 

New commit https://github.com/longhorn/charts/commit/5c0d8a1c26f298b9a197cdf054f9242de971e7c1 introduce check values in default-settings, **string** is checked ` {{- if kindIs "string" .Values.defaultSettings..... }}`.

For example [detachManuallyAttachedVolumesWhenCordoned string check](https://github.com/longhorn/charts/blob/5c0d8a1c26f298b9a197cdf054f9242de971e7c1/charts/longhorn/templates/default-setting.yaml#L225-L231) is not compatible with previous helm stored values (from the Harvester) because detachManuallyAttachedVolumesWhenCordoned is **boolean**.
```
node3:/home/rancher # helm get values -n longhorn-system longhorn  
USER-SUPPLIED VALUES:
defaultSettings:
  defaultDataPath: /var/lib/harvester/defaultdisk
  detachManuallyAttachedVolumesWhenCordoned: true
  nodeDrainPolicy: allow-if-replica-is-stopped
  taintToleration: kubevirt.io/drain:NoSchedule
```
Solutions:
- convert `true` to string `"true"` 
- or `{{- if kindIs "bool" .Values.defaultSettings.detachManuallyAttachedVolumesWhenCordoned }}`

### To Reproduce

Set default bool helm values helm before upgrade.

### Expected Behavior

boolean should be accepted

### Support Bundle for Troubleshooting

none

### Environment

- Longhorn version:  v1.10.0-rc4
- Installation method (e.g. Rancher Catalog App/Helm/Kubectl): helm
- Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: harvester v1.6.0 -> rke2

derekbit:
> `convert true to string "true"`

I don't think this is a bug. This is the purpose of the `if kindIs "string" ...` check.

roger-ryao:
**Verified on master-head 20250925**
- longhorn master-head https://github.com/longhorn/longhorn/commit/b7da0f8b98d4ec9a7e714a4d6b4f03e8592e0feb

**Verified on v1.10.x-head 20250925**
[...]

PR #11857, opened 2025-09-25
fix(setting): no need to check boolean settings

derekbit:

Issue Longhorn/longhorn#11854


c3y1huang (approved):

derekbit:
@mergify backport v1.10.x

Issue #11617, opened 2025-09-01
[FEATURE] "hard" podAntiAffinity for csi-attacher/csi-provisioner/csi-resizer/csi-snapshotter


danielskowronski:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

This is follow-up to https://github.com/longhorn/longhorn/issues/2894 - still considering Helm deployment on small clusters.
 
`preferredDuringSchedulingIgnoredDuringExecution` implemented for that issue work fine upon deployment. However on small cluster where replica count = workers count we may get outage of one worker and then multiple instances of CSI components will get deployed to single worker and nothing in pure k8s will rectify that.

### Describe the solution you'd like

Ability to pass "Bitnami-style" `podAntiAffinityPreset` (as `soft` or `hard`, defaulting to `soft`) per CSI component.

This could turn `preferredDuringSchedulingIgnoredDuringExecution` to `requiredDuringSchedulingIgnoredDuringExecution`.

### Describe alternatives you've considered

1. Letting CSI components run on control-plane node, but ReplicaSet managed on runtime doesn't accept tolerations to be passed.
2. Using Kyverno to patch things live 🫤 
3. Wait for https://github.com/longhorn/longhorn/issues/6959 to get implemented

### Additional context

One of my clusters is extremely small - 3 nodes, but 1 is control-plane only (tainted worker). This effectively means that (without manual intervention) single worker handles CSI components.

Please do not consider this a duplicate of https://github.com/longhorn/longhorn/issues/6959 as I'd like to tackle single fix delegating responsibility solely to k8s. If this one makes sense, I could contribute (I believe it should be just a matter of conditionally construction object [here](https://github.com/longhorn/longhorn-manager/blob/6a4d4a9586869a57981bd4ab739a53e6039e0f48/csi/deployment_util.go#L66))

derekbit:
~~@andyjeffries Would you like to submit a PR to improve the pod affinity?~~
Sorry, ping wrong person.

andyjeffries:
Hi @derekbit I haven't worked full time with Kubernetes in about 2 years, or in any capacity for about 10 months, so I'm definitely not the right fit to submit any PR for this project.

derekbit:
> Bitnami-style" podAntiAffinityPreset (as soft or hard, defaulting to soft)

Good to know.

[...]

PR #11872, opened 2025-09-26
feat: expose csi.podAntiAffinityPreset setting

yulken:

Issue https://github.com/longhorn/longhorn/issues/11617

### DIFF
--- a/chart/README.md
+++ b/chart/README.md
@@ -184,6 +184,7 @@ The `values.yaml` contains items used to tweak a deployment of this chart.
 |-----|-------------|
 | csi.attacherReplicaCount | Replica count of the CSI Attacher. When unspecified, Longhorn uses the default value ("3"). |
 | csi.kubeletRootDir | kubelet root directory. When unspecified, Longhorn uses the default value. |
+| csi.podAntiAffinityPreset | Configures Pod anti-affinity to prevent multiple instances on the same node. Use soft (tries to separate) or hard (must separate). When unspecified, Longhorn uses the default value ("soft"). |
 | csi.provisionerReplicaCount | Replica count of the CSI Provisioner. When unspecified, Longhorn uses the default value ("3"). |
 | csi.resizerReplicaCount | Replica count of the CSI Resizer. When unspecified, Longhorn uses the default value ("3"). |
 | csi.snapshotterReplicaCount | Replica count of the CSI Snapshotter. When unspecified, Longhorn uses the default value ("3"). |
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -230,6 +230,13 @@ questions:
         type: string
         label: Kubelet Root Directory
         group: Longhorn CSI Driver Settings
+      - variable: csi.podAntiAffinityPreset
+        type: string
+        label: Longhorn CSI Pod AntiAffinity Preset
+        group: Longhorn CSI Driver Settings
+        default: soft
+        description: >-
+          Configures Pod anti-affinity to prevent multiple instances on the same node. Use soft (tries to separate) or hard (must separate). When unspecified, Longhorn uses the default value ("soft").
       - variable: csi.attacherReplicaCount
         type: int
         default: 3
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -220,6 +220,8 @@ preUpgradeChecker:
 csi:
   # -- kubelet root directory. When unspecified, Longhorn uses the default value.
   kubeletRootDir: ~
+  # -- Configures Pod anti-affinity to prevent multiple instances on the same node. Use soft (tries to separate) or hard (must separate). When unspecified, Longhorn uses the default value ("soft").
+  podAntiAffinityPreset: ~
   # -- Replica count of the CSI Attacher. When unspecified, Longhorn uses the default value ("3").
   attacherReplicaCount: ~
   # -- Replica count of the CSI Provisioner. When unspecified, Longhorn uses the default value ("3").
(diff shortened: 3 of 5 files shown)


chart/questions.yaml
```diff
@@ -230,6 +230,13 @@ questions:
         type: string
         label: Kubelet Root Directory
         group: Longhorn CSI Driver Settings
+      - variable: csi.podAntiAffinityPreset
+        type: string
+        label: Longhorn CSI Pod AntiAffinity Preset
+        group: Longhorn CSI Driver Settings
+        default: soft
+        description: >-
+          Pod antiAffinityPreset. When unspecified, Longhorn uses the default value ("soft").
```
derekbit: Need to mention the setting supports either `soft` or `hard`. Also elaborate what do them mean in the description.

chart/README.md
```diff
@@ -184,6 +184,7 @@ The `values.yaml` contains items used to tweak a deployment of this chart.
 |-----|-------------|
 | csi.attacherReplicaCount | Replica count of the CSI Attacher. When unspecified, Longhorn uses the default value ("3"). |
 | csi.kubeletRootDir | kubelet root directory. When unspecified, Longhorn uses the default value. |
+| csi.podAntiAffinityPreset | Pod antiAffinityPreset. When unspecified, Longhorn uses the default value ("soft"). |
```
derekbit: Same as above

chart/values.yaml
```diff
@@ -220,6 +220,8 @@ preUpgradeChecker:
 csi:
   # -- kubelet root directory. When unspecified, Longhorn uses the default value.
   kubeletRootDir: ~
+  # -- Pod antiAffinityPreset. When unspecified, Longhorn uses the default value ("soft").
```
derekbit: Same as above

derekbit:
Hello @sushant-suse 
Could you review the description? Thank you.

c3y1huang (commented):
Should we also apply these changes to [longhorn/deploy/*.yaml](https://github.com/longhorn/longhorn/tree/master/deploy) ?

yulken:
> Should we also apply these changes to [longhorn/deploy/*.yaml](https://github.com/longhorn/longhorn/tree/master/deploy) ?

@c3y1huang I don't think it is needed. The variable defaults to "soft", so even when its not provided, longhorn should work the same.

c3y1huang:
> @c3y1huang I don't think it is needed. The variable defaults to "soft", so even when its not provided, longhorn should work the same.

Got it. Should we make this environment variable visible on the manifests so users know where to modify it if they want to use the "hard" podAntiAffinityPreset?

derekbit:
> > @c3y1huang I don't think it is needed. The variable defaults to "soft", so even when its not provided, longhorn should work the same.
> 
> Got it. Should we make this environment variable visible on the manifests so users know where to modify it if they want to use the "hard" podAntiAffinityPreset?

Hello @yulken 
Could you address the comment? Thank you.

yulken:
@c3y1huang Do you mean like this?
```yaml
          env:
          - name: POD_NAMESPACE
            valueFrom:
              fieldRef:
                fieldPath: metadata.namespace
          - name: NODE_NAME
            valueFrom:
              fieldRef:
                fieldPath: spec.nodeName
          - name: SERVICE_ACCOUNT
            valueFrom:
              fieldRef:
                fieldPath: spec.serviceAccountName
          - name: CSI_ATTACHER_IMAGE
            value: "longhornio/csi-attacher:v4.9.0-20250826"
          - name: CSI_PROVISIONER_IMAGE
            value: "longhornio/csi-provisioner:v5.3.0-20250826"
          - name: CSI_NODE_DRIVER_REGISTRAR_IMAGE
            value: "longhornio/csi-node-driver-registrar:v2.14.0-20250826"
          - name: CSI_RESIZER_IMAGE
            value: "longhornio/csi-resizer:v1.14.0-20250826"
          - name: CSI_SNAPSHOTTER_IMAGE
            value: "longhornio/csi-snapshotter:v8.3.0-20250826"
          - name: CSI_LIVENESS_PROBE_IMAGE
            value: "longhornio/livenessprobe:v2.16.0-20250826"
            # new env \/
          - name: CSI_POD_ANTI_AFFINITY_PRESET
            value: "soft"
```
I think it would pollute the manifest. I'm not sure the flag is relevant enough for that. If you mean something else, please let me know.
cc @derekbit

derekbit:
> I think it would pollute the manifest. I'm not sure the flag is relevant enough for that. If you mean something else, please let me know. cc @derekbit


Adding the flag to the manifests should be fine.

c3y1huang:
> I think it would pollute the manifest. I'm not sure the flag is relevant enough for that. If you mean something else, please let me know.

Could you clarify how adding the environment variable would pollute the manifest? From what I see, we also provide instructions to guide users to update it [via kubectl](https://github.com/longhorn/website/pull/1222/files#diff-f0b840f17f9b76f8fbb60d9ea6d5beaf614e74d8fd4f4d117ed85d4306999182R41-R48). IMO, making the env visible would help users to find where to make the change.

yulken:
@derekbit @c3y1huang 
I think I was looking at this the wrong way.
My main question was: Why do we explicitly define only the CSI image versions as variables in the manifest, but not the replica count variables?
When I used the word "pollution," I meant: Is this new variable as important as the CSI component image versions?

Now that I've thought about it more, I don't fully agree with my initial concern, so I fixed it.

derekbit:
> ```yaml
> value: "longhornio/csi-provisioner:v5.3.0
> ```

Good catch.
We have defined `CSI_{COMPONENT}_REPLICA_COUNT` in the helm chart, but the `CSI_{COMPONENT}_REPLICA_COUNT` are not shown because 
```
          {{- if .Values.csi.attacherReplicaCount }}
          - name: CSI_ATTACHER_REPLICA_COUNT
            value: {{ .Values.csi.attacherReplicaCount | quote }}
          {{- end }}
```

The default values of the replica counts are harded in longhorn-manager code base.
 I think we can still put the `CSI_POD_ANTI_AFFINITY_PRESET` in the helm chart.

Issue #11869, opened 2025-09-25
[BUG] UI fails to deploy when only IPv4 is enabled on nodes with v1.10.0 version


ADN182:
### Describe the Bug

After upgrading from v1.9.2 to v1.10.0  the longhorn-ui fail to deploy with the 1.10.0 version so pod 1.9.2 stay active until 1.10.0 is ready

❯ kubectl logs longhorn-ui-94f998669-jwjcf -n longhorn-system
**nginx: [emerg] socket() [::]:8000 failed (97: Address family not supported by protocol)**

k get pod : 
longhorn-ui-677b5b64c7-4bq8f                        1/1     Running            0              35m (1.9.2)
longhorn-ui-94f998669-jwjcf                         0/1     CrashLoopBackOff   8 (3m1s ago)   19m (1.10.0)
longhorn-ui-94f998669-z28k8                         0/1     CrashLoopBackOff   8 (3m4s ago)   19m (1.10.0)

### To Reproduce

Have a node with ipv4 enable only and upgrade chart from 1.9.2 version to 1.10.0

### Expected Behavior

Longhorn-ui deployed even if node as only ipv4 enabled.

### Support Bundle for Troubleshooting

N/A

### Environment

- Longhorn version: 1.10.0
- Impacted volume (PV): 
- Installation method (e.g. Rancher Catalog App/Helm/Kubectl): Helm
- Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: K3s
  - Number of control plane nodes in the cluster: 3
  - Number of worker nodes in the cluster: 3
- Node config
  - OS type and version: Ubuntu 22.02
  - Kernel version:
  - CPU per node:
  - Memory per node:
  - Disk type (e.g. SSD/NVMe/HDD):
  - Network bandwidth between the nodes (Gbps):
- Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal): Baremetal
- Number of Longhorn volumes in the cluster: 15

### Workaround and Mitigation

https://github.com/longhorn/longhorn/issues/11869#issuecomment-3336638753

derekbit:
Where is `1.10.1` from?

c3y1huang:
**Workaround** 

**Option 1:**
Patch and recreate the Longhorn UI deployment to remove the IPv6 listen line from the nginx configuration.

[...]

PR #11874, opened 2025-09-26
fix(ui): pod in crashloop on ipv4 only nodes

c3y1huang:

Issue longhorn/longhorn#11869

This is a temporary workaround. Do not merge.

`None`

`None`


derekbit:
Fixed in v1.10.1.

Issue #10269, opened 2025-01-21
[FEATURE] Support extra network interface (not only storage network) on the share manager pod


Vicente-Cheng:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

After the feature: [[FEATURE] Support storage network for RWX volumes](https://github.com/longhorn/longhorn/issues/8184).
We can now make all data paths on the storage network but for the Harvester use case (which means RWX volume support for guest cluster).
We would not want to expose the storage network on the guest cluster. Could we consider supporting the other extra network interface instead of the storage network?

For example, the Harvester cluster creates vlan100 for the guest VM. Can we have the endpoint with vlan100 on the share-manager pod?
(In this case, we could not set the storage network for an endpoint on the share-manager pod)

Related issue: [[ENHANCEMENT] Support various vlan for RWX volume on the guest cluster](https://github.com/harvester/harvester/issues/7218)

cc @innobead, @derekbit, @c3y1huang, @bk201 

### Describe the solution you'd like

Could we add the network related info to the Longhorn attachment ticket?
Then, the share-manager pod would know what interface should be used for the endpoint.

innobead:
This requires creating and passing a specific NAD for the VM network (VLAN) when creating an RWX volume from the guest cluster via the Harvester CSI driver, which interacts with the underlay Longhorn on the Harvester cluster.

innobead:
Discussed with @derekbit , evaluate first in 1.10.

c3y1huang:
@Vicente-Cheng 

From my understanding of the issue description, the expectation is that a workload pod in the guest cluster (downstream K8s) will be assigned a VLAN network interface to access RWX volumes on the Harvester cluster (upstream K8s). The VLAN traffic will be only to the connection between the guest cluster workload and the share-manager pod in the upstream Harvester cluster, while all other traffic remains internal to Longhorn within the Harvester cluster (upstream K8s) with or without the storage network.

Could you confirm if this expectation is correct and clarify a few points:
1. How is the VLAN created and managed for guest workloads?
1. Could you provide a sample or expected YAML of a workload pod using this VLAN to mount an RWX volume?
1. If possible, could you share an architecture or flow diagram illustrating this setup? Or, we could discuss this offline to help clarify my understanding.

cc @derekbit @innobead

Vicente-Cheng:
Hi @c3y1huang, sorry for the late reply

[...]

PR #11978, opened 2025-10-17
docs(lep): rwx volume endpoint network

c3y1huang:

Issue longhorn/longhorn#10269

LEP for https://github.com/longhorn/longhorn/issues/10269.

### DIFF
--- a/enhancements/20251017-rwx-volume-endpoint-network.md
+++ b/enhancements/20251017-rwx-volume-endpoint-network.md
@@ -1,0 +1,105 @@
+# RWX Volume Endpoint Network
+
+## Summary
+
+This proposal introduces the `endpoint-network-for-rwx-volume` setting, allowing users to assign a dedicated Network Attachment Definition (NAD) for RWX (ReadWriteMany) volume endpoints. This improves network isolation for RWX NFS traffic, keeping it separate from the primary Longhorn networks (cluster and storage networks).
+
+### Related Issues
+
+- https://github.com/longhorn/longhorn/issues/10269
+
+## Motivation
+
+In current versions (<=v1.10), RWX volume endpoints are limited to using either the cluster network or the storage network. Introducing a dedicated endpoint network enables:
+
+- Isolation of RWX NFS server and client traffic from the primary Longhorn networks and other application workloads.
+- Users have more control over network paths for RWX volumes.
+
+### Goals
+
+- Allow RWX volume NFS server endpoints to use a dedicated NAD.
+- Preserve backward compatibility for existing Longhorn systems that use the storage network for RWX volume.
+
+### Non-goals
+
+- **Remote mount support:** Using the host network interface directly is outside Kubernetes network namespace. This means the relevant component pods loses cluster DNS resolution.
+- **StorageClass/Volume-based networking:** Per-SC/volume network configuration is not supported, as the CSI plugin requires full network access to additional networks, and implementing per-SC/volume networking would be complicated.
+
+## Proposal
+
+### User Stories
+
+#### Story 1: Improved Network Control And Isolation For RWX Volumes
+
+- **Before:** RWX volume endpoints can only use the cluster or storage network.
+- **After:** RWX volume endpoints can use a dedicated NAD, isolating traffic from the cluster and storage networks.
+
+### User Experience
+
+#### Fresh Install
+
+- The user creates a dedicated NAD for RWX volume endpoints.
+- The user sets the `endpoint-network-for-rwx-volume` global setting to the NAD (`<namespace>/<name>`), following the same convention as `storage-network`.
+- When a workload pod is created, its RWX volume is mounted via the NFS endpoint using the NAD-assigned IP on the share-manager pod's `lhnet2` interface.
+- If the `endpoint-network-for-rwx-volume` global setting is not set, Longhorn will default to using the Kubernetes cluster network, keeping the same behavior when `storage-network-for-rwx-volume-enabled=true`.
+
+#### Upgrade
+
+- If `storage-network-for-rwx-volume-enabled=true`, the manager migrates the setting to `endpoint-network-for-rwx-volume=storage-network`, preserving existing behavior.
+- If `storage-network-for-rwx-volume-enabled=false`, the manager sets `endpoint-network-for-rwx-volume` to the default (empty) value, keeping behavior unchanged.
+
+### API changes
+
+`None`
+
+## Design
+
+### Implementation Overview
+
+#### New Global Setting
+
+`endpoint-network-for-rwx-volume`:
+- **Type:** `String`
+- **Default:** `""`
+- **DataEngineSpecific**: `false`,
+
+Specifies a dedicated network for mounting RWX (ReadWriteMany) volumes. Leave this field blank to use the default Kubernetes cluster network, which behaves the same as when the deprecated `storage-network-for-rwx-volume-enabled` setting is set to `false`.
+ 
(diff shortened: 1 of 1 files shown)


c3y1huang:
@Vicente-Cheng can you check if this is close to what Harvester is expecting? Thank you.

enhancements/20251017-rwx-volume-endpoint-network.md
```diff
@@ -0,0 +1,103 @@
...
+
+- **Before:** RWX volume endpoints can only use the cluster or storage network.
+- **After:** RWX volume endpoints can use a dedicated NAD, isolating traffic from the cluster and storage networks.
+
+### User Experience
+
+#### Fresh Install
+
+- The user creates a dedicated NAD for RWX volume endpoints.
+- The user sets the `endpoint-network-for-rwx-volume` global setting to the NAD (`<namespace>/<name>`), following the same convention as `storage-network`.
```
Vicente-Cheng, line 42: One slightly off-topic question here:
Do we have any limitations on the namespace of NAD?

I mean, in the Harvester scenario, the VM might have a different namespace, so the namespace of NAD might be limited to the specific VM.

enhancements/20251017-rwx-volume-endpoint-network.md
```diff
@@ -0,0 +1,103 @@
...
+#### Deprecate Old Global Setting
+
+- `SettingNameStorageNetworkForRWXVolumeEnabled` will be deprecated.
+- During upgrade, if the storage network for RWX volumes was previously enabled, the new `endpoint-network-for-rwx-volume` setting will inherit the value of the `storage-network` setting to preserve existing behavior.
+
+#### Component CNI
+
+- Before this change, CSI plugin and share-manager pods were annotated with `k8s.v1.cni.cncf.io/networks` using the `lhnet1` interface (storage network).
+- With this change, RWX NFS endpoints use a new interface, `lhnet2`, for network isolation.
+- Other components unrelated to the NFS endpoint continue to use `lhnet1` for storage network.
```
Vicente-Cheng, line 79: Could I understand the above part like:
The sharemanager pod will have two extra interfaces, one for the storage network and another for the endpoint?

Vicente-Cheng:
BTW, I couldn't find out how to add the dedicated NAD to the sharemanager pod. Did it leverage the longhorn volumeattachments ticket?

enhancements/20251017-rwx-volume-endpoint-network.md
c3y1huang: The Share Manager will have a single additional interface, `lhnet2`, dedicated to the NFS server endpoint. And, Longhorn’s primary data-plane traffic (handled by the instance manager) will still use the storage network on `lhnet1` if configured.

enhancements/20251017-rwx-volume-endpoint-network.md
c3y1huang: I am not aware of any namespace restrictions for a NAD. We have used a NAD in a different namespace (`kube-system`) when testing the storage network feature.

c3y1huang:
> BTW, I couldn't find out how to add the dedicated NAD to the sharemanager pod. Did it leverage the longhorn volumeattachments ticket?

This feature doesn’t require a Longhorn volume attachment, as it isn't per SC/volume-based. The dedicated NAD will be added via the global  `endpoint-network-for-rwx-volume` setting. Then the endpoint controller updates the endpoint subset with the CNI IP.

derekbit (commented):
In general, LGTM. 
Only one question. If the global setting is set to empty, will the RWX volumes be remounted?

enhancements/20251017-rwx-volume-endpoint-network.md
```diff
@@ -0,0 +1,103 @@
...
+
+#### Endpoint Slice
+
+The endpoint controller updates the endpoint subset with the share-manager pod's `lhnet2` interface IP from the `k8s.v1.cni.cncf.io/network-status` annotation.
```
derekbit, line 83: cc @davidcheng0922

c3y1huang:
> In general, LGTM. Only one question. If the global setting is set to empty, will the RWX volumes be remounted?

The RWX volume needs to be detached for the setting to apply. This behavior is the same as the deprecating setting [Storage Network For RWX Volume Enabled](https://longhorn.io/docs/1.10.0/references/settings/#storage-network-for-rwx-volume-enabled).

derekbit:
> > In general, LGTM. Only one question. If the global setting is set to empty, will the RWX volumes be remounted?
> 
> The RWX volume needs to be detached for the setting to apply. This behavior is the same as the deprecating setting [Storage Network For RWX Volume Enabled](https://longhorn.io/docs/1.10.0/references/settings/#storage-network-for-rwx-volume-enabled).

Thanks. Can you add the behavior in the LEP? Thanks.

derekbit (approved):
LGTM

enhancements/20251017-rwx-volume-endpoint-network.md
```diff
@@ -0,0 +1,105 @@
...
+- **After:** RWX volume endpoints can use a dedicated NAD, isolating traffic from the cluster and storage networks.
+
+### User Experience
+
+#### Fresh Install
+
+- The user creates a dedicated NAD for RWX volume endpoints.
+- The user sets the `endpoint-network-for-rwx-volume` global setting to the NAD (`<namespace>/<name>`), following the same convention as `storage-network`.
+- When a workload pod is created, its RWX volume is mounted via the NFS endpoint using the NAD-assigned IP on the share-manager pod's `lhnet2` interface.
+- If the `endpoint-network-for-rwx-volume` global setting is not set, Longhorn will default to using the Kubernetes cluster network, keeping the same behavior when `storage-network-for-rwx-volume-enabled=true`.
```
innobead, line 44: It's better to mention that if users want to use the same storage network in this case, they need to explicitly configure the setitng.

enhancements/20251017-rwx-volume-endpoint-network.md
innobead: After checking with @c3y1huang , the traffic should be this:

VM + endpoint-nad (should be just VM (vlan or overlay) network) → Share manager pod + endpoint-nad → IM pod + storage nad.

The traffic between the share manager pod and the corresponding IM pod still goes through the default cluster network. 

cc @Vicente-Cheng @WebberHuang1118

WebberHuang1118: @c3y1huang Does this mean Longhorn can have two NADs — one for the storage network and one for the share manager?

For example:

* **nad1** → storage network
* **nad2** → share manager endpoint

So:

* The IM pod’s **lhnet1** uses **nad1** for Longhorn data traffic
* The share manager pod’s **lhnet2** uses **nad2** for RWX NFS traffic

Do I understand this correctly?

WebberHuang1118 (commented):
LGTM, just a question https://github.com/longhorn/longhorn/pull/11978#discussion_r2526511593, thanks.

enhancements/20251017-rwx-volume-endpoint-network.md
c3y1huang: Yes, that’s correct. Longhorn can have two separate NADs: one for the Longhorn storage network and another for the share manager endpoint. This keeps storage traffic and NFS endpoint traffic isolated.

Vicente-Cheng: It's better to have two multus on the sharemanager pod. Like below:
- cluster network
- nad1 -> for LH storage network
- nad2 -> for nfs endpoint

The current design was already good enough. The above suggestion could be the improvement on later version (and we can get more feedback if the bottleneck is cluster network)

Issue #12019, opened 2025-10-23
[IMPROVEMENT] Improve Helm Chart defaultSettings handling with automatic quoting and multi-type support


derekbit:
### Is your improvement request related to a feature? Please describe (👍 if you like this request)

Currently, many defaultSettings values in the Longhorn Helm Chart are rendered with a simple | quote and rely on kindIs "string" checks. This causes issues when some settings can be either a string, a number, or a map (e.g., defaultReplicaCount may sometimes be a number, sometimes a map).

- Helm templates fail or render incorrectly when a value is a number or map instead of a string.
- Users may see "map[v1:2 v2:1]" instead of a proper JSON-quoted string.

chriscchien:
Verified **pass** on longhorn master(https://github.com/longhorn/longhorn-manager/commit/00771819815794bc2688d50252134cb1cd5c4678) with test [steps](https://github.com/longhorn/longhorn/issues/12019#issuecomment-3434898209)

PR #12022, opened 2025-10-23
feat(setting): handle with automatic quoting and multi-type support

derekbit:

Issue Longhorn/longhorn#12019

### DIFF
--- a/chart/templates/default-setting.yaml
+++ b/chart/templates/default-setting.yaml
@@ -13,118 +13,58 @@ data:
     create-default-disk-labeled-nodes: {{ .Values.defaultSettings.createDefaultDiskLabeledNodes }}
     {{- end }}
     {{- if not (kindIs "invalid" .Values.defaultSettings.defaultDataPath) }}
-    {{- if kindIs "string" .Values.defaultSettings.defaultDataPath }}
     default-data-path: {{ .Values.defaultSettings.defaultDataPath | quote }}
-    {{- else }}
-    {{- fail "defaultSettings.defaultDataPath must be a string" }}
-    {{- end }}
     {{- end }}
     {{- if not (kindIs "invalid" .Values.defaultSettings.replicaSoftAntiAffinity) }}
     replica-soft-anti-affinity: {{ .Values.defaultSettings.replicaSoftAntiAffinity }}
     {{- end }}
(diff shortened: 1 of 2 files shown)


derekbit:
@mergify backport v1.10.x

chart/templates/default-setting.yaml
```diff
@@ -13,118 +13,58 @@ data:
     create-default-disk-labeled-nodes: {{ .Values.defaultSettings.createDefaultDiskLabeledNodes }}
     {{- end }}
     {{- if not (kindIs "invalid" .Values.defaultSettings.defaultDataPath) }}
-    {{- if kindIs "string" .Values.defaultSettings.defaultDataPath }}
     default-data-path: {{ .Values.defaultSettings.defaultDataPath | quote }}
-    {{- else }}
-    {{- fail "defaultSettings.defaultDataPath must be a string" }}
-    {{- end }}
```
COLDTURNIP, line 20: Is it safe to remove such type check in the chart? Do we intend to defer this check until the configuration is loaded?

chart/templates/default-setting.yaml
```diff
@@ -13,118 +13,58 @@ data:
...
-    {{- fail "defaultSettings.upgradeResponderURL must be a string" }}
-    {{- end }}
     {{- end }}
     {{- if not (kindIs "invalid" .Values.defaultSettings.defaultReplicaCount) }}
-    {{- if kindIs "string" .Values.defaultSettings.defaultReplicaCount }}
-    default-replica-count: {{ .Values.defaultSettings.defaultReplicaCount | quote }}
-    {{- else }}
-    {{- fail "defaultSettings.defaultReplicaCount must be a string" }}
-    {{- end }}
+    default-replica-count: {{ include "longhorn.multiTypeSetting" .Values.defaultSettings.defaultReplicaCount }}
```
c3y1huang, line 40: Why not use a similar approach for the other settings?

chart/templates/default-setting.yaml
derekbit: The setting is automatically quoted, so everything is string.

chart/templates/default-setting.yaml
derekbit: For other settings, they are not dataEngine specific settings and don't support json-formatted string, so we don't need to use the approach.

c3y1huang (approved):

Issue #10583, opened 2025-03-13
[FEATURE] Provide a helm chart setting to define the managerUrl


fmunteanu:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

I have a Cilium Gateway API `HTTPRoute` for Longhorn frontend `longhorn.noty.cc`. Running:

```
$ curl -s https://longhorn.noty.cc/v1/backupvolumes/pvc-1ee55f51-839a-4dbc-bb6e-484cefa49905-c0fad521 | jq
{
  "actions": {
    "backupDelete": "http://10.42.6.107:9500/v1/backupvolumes/pvc-1ee55f51-839a-4dbc-bb6e-484cefa49905-c0fad521?action=backupDelete",
    "backupGet": "http://10.42.6.107:9500/v1/backupvolumes/pvc-1ee55f51-839a-4dbc-bb6e-484cefa49905-c0fad521?action=backupGet",
    "backupList": "http://10.42.6.107:9500/v1/backupvolumes/pvc-1ee55f51-839a-4dbc-bb6e-484cefa49905-c0fad521?action=backupList",
    "backupListByVolume": "http://10.42.6.107:9500/v1/backupvolumes/pvc-1ee55f51-839a-4dbc-bb6e-484cefa49905-c0fad521?action=backupListByVolume",
    "backupVolumeSync": "http://10.42.6.107:9500/v1/backupvolumes/pvc-1ee55f51-839a-4dbc-bb6e-484cefa49905-c0fad521?action=backupVolumeSync"
  },
  "backingImageChecksum": "",
  "backingImageName": "",
  "backupTargetName": "default",
  "created": "2025-03-13T07:22:17Z",
  "dataStored": "29360128",
  "id": "pvc-1ee55f51-839a-4dbc-bb6e-484cefa49905-c0fad521",
  "labels": {
    "KubernetesStatus": "{\"pvName\":\"pvc-1ee55f51-839a-4dbc-bb6e-484cefa49905\",\"pvStatus\":\"Bound\",\"namespace\":\"media\",\"pvcName\":\"application-config\",\"lastPVCRefAt\":\"\",\"workloadsStatus\":[{\"podName\":\"application-7b74cd7ffc-dtt62\",\"podStatus\":\"Running\",\"workloadName\":\"application-7b74cd7ffc\",\"workloadType\":\"ReplicaSet\"}],\"lastPodRefAt\":\"\"}",
    "VolumeRecurringJobInfo": "{}",
[...]

PR #12050, opened 2025-10-27
feat(chart): add manager-url setting for external API access

lexfrei:

Issue longhorn/longhorn#10583

This PR adds Helm chart support for the new `manager-url` setting that fixes external API access when using Ingress or Gateway API HTTPRoute.

**Problem**: When accessing Longhorn Manager API through Ingress or Gateway API, the API responses contain internal pod IPs in `actions` and `links` fields, making the API unusable from external clients.

**Solution**: The `manager-url` setting (implemented in longhorn-manager) allows users to explicitly configure the external URL. This PR exposes that setting through the Helm chart so it can be configured via:
- `values.yaml`
- Helm CLI (`--set defaultSettings.managerUrl="..."`)
- Rancher UI (via `questions.yaml`)

**Changes**:
1. **`chart/values.yaml`**: Added `managerUrl` field to `defaultSettings` with documentation and examples
2. **`chart/templates/default-setting.yaml`**: Added conditional template block for `manager-url` setting
3. **`chart/questions.yaml`**: Added Rancher UI question for easy configuration

**Configuration examples**:
```bash
# Via Helm install
helm install longhorn longhorn/longhorn \
  --namespace longhorn-system \
  --set defaultSettings.managerUrl="https://longhorn.example.com"

# Via values.yaml
defaultSettings:
  managerUrl: "https://longhorn.example.com:8443"
```

1. **Non-breaking change**: Purely additive with safe defaults (null = not set)
2. **Depends on manager PR**: This chart update requires the corresponding longhorn-manager PR to be merged first. The setting has no effect until users upgrade to a longhorn-manager version that includes the middleware implementation.
3. **Naming convention**: Uses `camelCase` (`managerUrl`) in Helm values, `kebab-case` (`manager-url`) in Kubernetes setting name (consistent with existing patterns like `upgradeResponderURL`)
4. **Template validation**: Tested with `helm template` to verify correct rendering and null value handling

**Testing**:
```bash
# Test rendering with value
helm template longhorn ./chart \
  --set defaultSettings.managerUrl="https://longhorn.example.com" \
  | grep -A 2 "manager-url"

# Expected output: manager-url: "https://longhorn.example.com"

# Test null value handling
helm template longhorn ./chart | grep "manager-url"
# Expected: No output (setting not rendered when null)
```

**Companion PR**: longhorn-manager implementation at [longhorn/longhorn-manager#4245](https://github.com/longhorn/longhorn-manager/pull/4245)

**Related discussion**: This implements the Helm chart portion of the proposal I posted in [issue comment](https://github.com/longhorn/longhorn/issues/10583#issuecomment-3437500849).

### DIFF
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -374,7 +374,7 @@ questions:
       - variable: defaultSettings.managerUrl
         label: Manager URL
         description: >-
-          External URL to access Longhorn Manager API. When set, this URL will be used in API responses (actions and links fields) instead of the internal pod IP. This is useful when accessing the API through Ingress or Gateway API HTTPRoute. Format: scheme://host[:port] (e.g., https://longhorn.example.com or https://longhorn.example.com:8443). Leave empty to use default behavior.
+          The external URL used to access the Longhorn Manager API. When set, this URL is returned in API responses (the actions and links fields) instead of the internal pod IP. This is useful when accessing the API through Ingress or Gateway API HTTPRoute. Format: scheme://host[:port] (for example, https://longhorn.example.com or https://longhorn.example.com:8443). Leave it empty to use the default behavior.
         group: Longhorn Default Settings
         type: string
         default: ''
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -260,7 +260,7 @@ defaultSettings:
   upgradeChecker: ~
   # -- The Upgrade Responder sends a notification whenever a new Longhorn version that you can upgrade to becomes available. The default value is https://longhorn-upgrade-responder.rancher.io/v1/checkupgrade.
   upgradeResponderURL: ~
-  # -- External URL to access Longhorn Manager API. When set, this URL will be used in API responses (actions and links fields) instead of the internal pod IP. This is useful when accessing the API through Ingress or Gateway API HTTPRoute. Format: scheme://host[:port] (e.g., https://longhorn.example.com or https://longhorn.example.com:8443). Leave empty to use default behavior.
+  # -- The external URL used to access the Longhorn Manager API. When set, this URL is returned in API responses (the actions and links fields) instead of the internal pod IP. This is useful when accessing the API through Ingress or Gateway API HTTPRoute. Format: scheme://host[:port] (for example, https://longhorn.example.com or https://longhorn.example.com:8443). Leave it empty to use the default behavior.
   managerUrl: ~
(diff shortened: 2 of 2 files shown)


chart/values.yaml
```diff
@@ -253,6 +251,8 @@ defaultSettings:
   upgradeChecker: ~
   # -- The Upgrade Responder sends a notification whenever a new Longhorn version that you can upgrade to becomes available. The default value is https://longhorn-upgrade-responder.rancher.io/v1/checkupgrade.
   upgradeResponderURL: ~
+  # -- External URL to access Longhorn Manager API. When set, this URL will be used in API responses (actions and links fields) instead of the internal pod IP. This is useful when accessing the API through Ingress or Gateway API HTTPRoute. Format: scheme://host[:port] (e.g., https://longhorn.example.com or https://longhorn.example.com:8443). Leave empty to use default behavior.
```
derekbit:
```suggestion
  # -- The external URL used to access the Longhorn Manager API. When set, this URL is returned in API responses (the actions and links fields) instead of the internal pod IP. This is useful when accessing the API through Ingress or Gateway API HTTPRoute. Format: scheme://host[:port] (for example, https://longhorn.example.com or https://longhorn.example.com:8443). Leave it empty to use the default behavior.
```

chart/questions.yaml
```diff
@@ -371,6 +364,13 @@ questions:
         group: Longhorn Default Settings
         type: string
         default: 'https://longhorn-upgrade-responder.rancher.io/v1/checkupgrade'
+      - variable: defaultSettings.managerUrl
+        label: Manager URL
+        description: >-
+          External URL to access Longhorn Manager API. When set, this URL will be used in API responses (actions and links fields) instead of the internal pod IP. This is useful when accessing the API through Ingress or Gateway API HTTPRoute. Format: scheme://host[:port] (e.g., https://longhorn.example.com or https://longhorn.example.com:8443). Leave empty to use default behavior.
```
derekbit:
```suggestion
          The external URL used to access the Longhorn Manager API. When set, this URL is returned in API responses (the actions and links fields) instead of the internal pod IP. This is useful when accessing the API through Ingress or Gateway API HTTPRoute. Format: scheme://host[:port] (for example, https://longhorn.example.com or https://longhorn.example.com:8443). Leave it empty to use the default behavior.
```

chart/values.yaml
lexfrei: Done

chart/questions.yaml
lexfrei: Done

PR #12169, opened 2025-11-14
feat(setting): update chart and upgrade responder

c3y1huang:

Issue longhorn/longhorn#10269

Add `endpoint-network-for-rwx-volume` setting to chart and upgrade responder.

### DIFF
--- a/deploy/upgrade_responder_server/chart-values.yaml
+++ b/deploy/upgrade_responder_server/chart-values.yaml
@@ -248,6 +248,10 @@ configMap:
           "dataType": "string",
           "maxLen": 200
         },
+        "longhornSettingEndpointNetworkForRWXVolume": {
+          "dataType": "string",
+          "maxLen": 200
+        },
         "longhornSettingSystemManagedComponentsNodeSelector": {
           "dataType": "string",
           "maxLen": 200
--- a/dev/upgrade-responder/install.sh
+++ b/dev/upgrade-responder/install.sh
@@ -222,6 +222,10 @@ configMap:
           "dataType": "string",
           "maxLen": 200
         },
+        "longhornSettingEndpointNetworkForRWXVolume": {
+          "dataType": "string",
+          "maxLen": 200
+        },
         "longhornSettingSystemManagedComponentsNodeSelector": {
           "dataType": "string",
           "maxLen": 200


chart/README.md
```diff
@@ -370,6 +370,7 @@ During installation, you can either allow Longhorn to use the default system set
 | defaultSettings.snapshotMaxCount | Maximum snapshot count for a volume. The value should be between 2 to 250 |
 | defaultSettings.storageMinimalAvailablePercentage | Percentage of minimum available disk capacity. When the minimum available capacity exceeds the total available capacity, the disk becomes unschedulable until more space is made available for use. The default value is "25". |
 | defaultSettings.storageNetwork | Storage network for in-cluster traffic. When unspecified, Longhorn uses the Kubernetes cluster network. |
+| defaultSettings.endpointNetworkForRWXVolume | A dedicated network for mounting RWX (ReadWriteMany) volumes. When unspecified, Longhorn uses the Kubernetes cluster network. |
```
derekbit: Can we make all descriptions aligned to simplify the maintainance?

```suggestion
| defaultSettings.endpointNetworkForRWXVolume | Specifies a dedicated network for mounting RWX (ReadWriteMany) volumes. Leave this blank to use the default Kubernetes cluster network. **Caution**: This setting should change after all RWX volumes are detached because some Longhorn component pods must be recreated to apply the setting. You cannot modify this setting while RWX volumes are still attached. |
```

chart/values.yaml
```diff
@@ -332,6 +332,8 @@ defaultSettings:
   orphanResourceAutoDeletionGracePeriod: ~
   # -- Storage network for in-cluster traffic. When unspecified, Longhorn uses the Kubernetes cluster network.
   storageNetwork: ~
+  # -- Dedicated network for mounting RWX (ReadWriteMany) volumes. When unspecified, Longhorn uses the Kubernetes cluster network.
```
derekbit: Same as above.
```suggestion
  # -- Specifies a dedicated network for mounting RWX (ReadWriteMany) volumes. Leave this blank to use the default Kubernetes cluster network. **Caution**: This setting should change after all RWX volumes are detached because some Longhorn component pods must be recreated to apply the setting. You cannot modify this setting while RWX volumes are still attached.
```

Issue #12219, opened 2025-11-21
[BUG] Some default settings in questions.yaml are placed incorrectly.


derekbit:
### Describe the Bug

Some default settings in questions.yaml are placed incorrectly.

https://github.com/longhorn/longhorn/blob/b013e79794f6eb4be1ad998126aefaa03445ae54/chart/questions.yaml#L1171-L1209

### Expected Behavior

N/A

### Support Bundle for Troubleshooting

N/A

### Environment

- Longhorn version: 
- Impacted volume (PV): 
- Installation method (e.g. Rancher Catalog App/Helm/Kubectl):
- Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version:
  - Number of control plane nodes in the cluster:
  - Number of worker nodes in the cluster:
- Node config
  - OS type and version:
  - Kernel version:
  - CPU per node:
  - Memory per node:
  - Disk type (e.g. SSD/NVMe/HDD):
  - Network bandwidth between the nodes (Gbps):
- Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal):
- Number of Longhorn volumes in the cluster:

derekbit:
cc @mantissahz

roger-ryao:
**Verified on master-head 20251125**
- longhorn master-head https://github.com/longhorn/longhorn/commit/ab10bd1a7359ea0b599edf8f323e1b084a32f17c

**Result Passed**

- Longhorn installation and settings were displayed correctly.
- Referring to https://github.com/rancher/rancher/issues/52626, we verified the `chart/questions.yaml` in the Rancher App Catalog (see screenshot)

<img width="1795" height="825" alt="Image" src="https://github.com/user-attachments/assets/da44796b-1168-4d2b-82a6-aa09a195d80f" />

. The related commit contains no errors.

PR #12220, opened 2025-11-21
fix(chart): some default settings in questions.yaml are placed incorrectly

derekbit:

Issue Longhorn/longhorn#12219


derekbit:
@mergify backport v1.10.x

Issue #12225, opened 2025-11-21
[IMPROVEMENT] Add Resource Limits Configuration for Longhorn manager/instance-manager


hookak:
### Is your improvement request related to a feature? Please describe (👍 if you like this request)

As I mentioned in [https://github.com/longhorn/longhorn/issues/12224](https://github.com/longhorn/longhorn/issues/12224) regarding the CSI components, this request focuses on adding resource requests/limits for **longhorn-manager** and **instance-manager**.

The `longhorn-manager` DaemonSet is currently deployed without CPU or memory requests/limits. To ensure predictable scheduling and production-grade resource control, the Helm chart should allow users to configure resource requirements for the longhorn manager and instance-manager container

related: https://github.com/longhorn/longhorn/issues/8332

derekbit:
cc @PhanLe1010

chriscchien:
https://github.com/longhorn/longhorn/pull/12241 is in review, moved the status back

chriscchien:
Verified **pass** on longhorn master(https://github.com/longhorn/longhorn-manager/commit/0c5eb1d2b77b9a50d39d25b7d7387e2637ed8a4b) with test [steps](https://github.com/longhorn/longhorn/issues/12225#issuecomment-3578008863)

PR #12241, opened 2025-11-25
feat(chart): allow configuring manager resources via values

hookak:

Issue # https://github.com/longhorn/longhorn/issues/12225

### DIFF
--- a/chart/templates/daemonset-sa.yaml
+++ b/chart/templates/daemonset-sa.yaml
@@ -102,6 +102,10 @@ spec:
         - name: GOCOVERDIR
           value: /go-cover-dir/
         {{- end }}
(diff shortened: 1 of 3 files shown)


chart/templates/daemonset-sa.yaml
```diff
@@ -102,6 +102,10 @@ spec:
         - name: GOCOVERDIR
           value: /go-cover-dir/
         {{- end }}
+        {{- with .Values.longhornManager.resources }}
```
c3y1huang, line 105: Missing in the values.yaml?

hookak: Thanks for checking it. I've added it now

derekbit (approved):
LGTM

Issue #12240, opened 2025-11-25
[IMPROVEMENT] Configure rolling update strategy for longhorn-manager and CSI deployments


hookak:
### Is your improvement request related to a feature? Please describe (👍 if you like this request)

While upgrading Longhorn images, I found that several Longhorn components are redeployed without defining `maxUnavailable` in their update strategies:

* `longhorn-manager` (DaemonSet)
* `longhorn-csi-plugin` (DaemonSet)
* CSI components such as `csi-attacher`, `csi-provisioner`, `csi-resizer`, `csi-snapshotter` (Deployments)

Because no `maxUnavailable` value is specified, Kubernetes treats it as allowing **all Pods to be unavailable at once** during a rollout.
As a result, **all Pods terminate simultaneously**, leading to a temporary shutdown of the Longhorn control plane and CSI operations during upgrades.

### Describe the solution you'd like

For **longhorn-manager (DaemonSet)**, since it is deployed through the Helm chart, the `maxUnavailable` value should be made **configurable by users** through `values.yaml`. This allows each environment—especially large clusters—to choose an appropriate level of parallelism during rolling updates (e.g., `10%`, `20%`, or a fixed number).

For **CSI components (Deployments)** such as `csi-attacher`, `csi-provisioner`, `csi-resizer`, and `csi-snapshotter`, they are created directly by Longhorn and typically run with **3 replicas**. Because these components are not user-managed through Helm and their availability is critical for core CSI operations, their `maxUnavailable` value should be **set to `1` by default in the code**. This ensures that at least one replica remains available during upgrades, providing safer and more predictable rolling updates.

chriscchien:
Verified **pass** on longhorn master(https://github.com/longhorn/longhorn-manager/commit/d5dd9e77b4d5fc7475a3d767e40278a3660e0464) v1.11.x(https://github.com/longhorn/longhorn-manager/commit/57847a468b998d628155de580761a9c5937ae8d1)

Helm upgrade from below path with `longhornManager.updateStrategy.rollingUpdate.maxUnavailable` to "20%", the upgrade completed without error and did not see all pods terminate simultaneously happen during the upgrade.
- Fresh install v1.10.1 then upgrade to v1.11.x-head
- Fresh install v1.11.x-head then upgrade to master-head

PR #12253, opened 2025-11-26
feat(chart): allow overriding manager update strategy

hookak:

Issue # https://github.com/longhorn/longhorn/issues/12240

### DIFF
--- a/chart/templates/crds.yaml
+++ b/chart/templates/crds.yaml
@@ -1323,6 +1323,14 @@ spec:
                 type: boolean
               nodeID:
                 type: string
+              rebuildConcurrentSyncLimit:
+                description: |-
+                  RebuildConcurrentSyncLimit controls the maximum number of file synchronization operations that can run
+                  concurrently during a single replica rebuild.
+                  It is determined by the global setting or the volume spec field with the same name.
+                maximum: 5
+                minimum: 0
+                type: integer
               replicaAddressMap:
                 additionalProperties:
                   type: string
@@ -1474,6 +1482,13 @@ spec:
                   type: object
                 nullable: true
                 type: object
+              rebuildConcurrentSyncLimit:
+                description: |-
+                  RebuildConcurrentSyncLimit controls the maximum number of file synchronization operations that can run
+                  concurrently during a single replica rebuild.
+                  It is determined by the global setting or the volume spec field with the same name.
+                minimum: 0
+                type: integer
               rebuildStatus:
                 additionalProperties:
                   properties:
@@ -1483,7 +1498,13 @@ spec:
                     error:
                       type: string
                     fromReplicaAddress:
+                      description: Deprecated. We are now using FromReplicaAddressList
+                        to list all source replicas.
                       type: string
+                    fromReplicaAddressList:
+                      items:
+                        type: string
+                      type: array
                     isRebuilding:
                       type: boolean
                     progress:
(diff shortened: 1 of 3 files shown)


chart/templates/daemonset-sa.yaml
```diff
@@ -170,9 +170,10 @@ spec:
         {{- end }}
       {{- end }}
       serviceAccountName: longhorn-service-account
+  {{- $defaultUpdateStrategy := dict "rollingUpdate" (dict "maxUnavailable" "100%") }}
+  {{- $updateStrategy := default $defaultUpdateStrategy .Values.longhornManager.updateStrategy }}
   updateStrategy:
-    rollingUpdate:
-      maxUnavailable: "100%"
+{{ toYaml $updateStrategy | indent 4 }}
```
derekbit: I'm thinking if the longhorn-manager rolling upgrade can impact the upgrade path.
WDYT? @c3y1huang @shuo-wu @PhanLe1010 @innobead

c3y1huang: If both new and old manager pods reconcile the same CR same time, the sequence could be:
1. New manager pod updates the CR -> succeeds
1. Old manager tries to update the same CR ->
	1. Get conflict error, triggers a retry
  	1. Retried reconciliation -> read the CR and applied the changes
1. Another new manager updates the CR -> succeeds

I think by the end of the upgrade, after the last manager reconciles, the CR should be correct.

innobead: In the current longhorn-manager design, the upgrade won't begin on the chosen leader node until the old manager pods are removed, so having this setting won't impact the upgrade process.

However, this setting might not improve control-plane (longhorn-manager) availability because the leader election mechanism is used during the upgrade. As a result, controllers in each manager pod won't start until the upgrade is complete. Simply said, the value might not matter for longhorn-manager in the current design.

[/upgrade/upgrade.go#L292-L314](https://github.com/longhorn/longhorn-manager/blob/b6485bbbc3eb071fda4b74c6de9811a430c37b91/upgrade/upgrade.go#L292-L314)
```go
func waitForOldLonghornManagersToBeFullyRemoved(namespace, managerImage string, kubeClient *clientset.Clientset) error {
	logrus.Info("Waiting for old Longhorn manager pods to be fully removed")
	for i := 0; i < 600; i++ {
		managerPods, err := upgradeutil.ListManagerPods(namespace, kubeClient)
		if err != nil {
			return err
		}
		foundOldManager := false
		for _, pod := range managerPods {
			if isOldPod, oldImage := isOldManagerPod(pod, managerImage); isOldPod {
				logrus.Infof("Found old longhorn manager: %v with image %v", pod.Name, oldImage)
				foundOldManager = true
				break
			}
		}
		if !foundOldManager {
			return nil
		}
		time.Sleep(1 * time.Second)
	}

	return fmt.Errorf("timed out while waiting for old Longhorn manager pods to be fully removed")
}
```

innobead: @derekbit @hookak do you think we still need this PR?

hookak: Thanks for the review and the detailed explanation. @innobead 

Honestly, I wasn’t aware that manager controllers effectively wait until the upgrade is complete due to the current leader-election-based upgrade flow (in our case we were “upgrading” within the same version but with a different commit, so I missed this behavior).

That said, this PR still keeps the default behavior unchanged, and only makes it configurable for users who want a different policy. In our environment, we have a concrete requirement to roll the longhorn-manager DaemonSet in a rolling manner to avoid overloading the Kubernetes API server during upgrades. So we’d like to keep the default as-is, but allow users to opt into a rolling upgrade strategy based on their operational needs.

Given the variety of environments and constraints, I believe this configurability is still valuable.

innobead: Sounds good to me. We might also give a brief explanation of the setting in doc to ensure the right understanding from users, given the current design.

chart/templates/daemonset-sa.yaml
```diff
@@ -174,9 +174,10 @@ spec:
         {{- end }}
       {{- end }}
       serviceAccountName: longhorn-service-account
+  {{- $defaultUpdateStrategy := dict "rollingUpdate" (dict "maxUnavailable" "100%") }}
```
innobead: Instead of dealing with the Helm template lang here, suggest moving the default setting of the upgrade strategy to values.yaml so we can keep all default values there. WDYT? @hookak

hookak: Agree

Issue #11635, opened 2025-09-03
[IMPROVEMENT] Introduce Concurrent Job Limit for Snapshot Operations


derekbit:
### Is your improvement request related to a feature? Please describe (👍 if you like this request)

Some snapshot operations may trigger a merge process that temporarily consumes additional storage space, increasing the risk of disk exhaustion. To mitigate this, we can introduce a global setting to limit the number of concurrent jobs for such operations.

derekbit:
cc @shuo-wu @innobead

liyimeng:
Can [this](https://github.com/longhorn/longhorn/issues/12031) be covered?

By the way, I am really curious if this is an issue in data engine v2. will v1 eventually replaced with v2 at some point in the future?

davidcheng0922:
I think so. Snapshot purge in v1 is quite expensive, especially when many snapshots are purged at the same time. Limiting the concurrency can help reduce resource usage, and the design is still under discussion.

IIRC, snapshot purge in the v2 is a much lighter operation.

Btw, v2 is planned to become GA in a future release.

derekbit:
Hello @davidcheng0922 
longhorn-manager PR is merged. Please create a longhorn PR and website PR to update settings. Thank you.

davidcheng0922:
> Hello [@davidcheng0922](https://github.com/davidcheng0922) longhorn-manager PR is merged. Please create a longhorn PR and website PR to update settings. Thank you.

Got it

derekbit:
Moved back to `Review`. The longhorn/longhorn PR is not ready yet.

yangchiu:
@davidcheng0922 Could you help confirm whether this feature and [the test steps](https://github.com/longhorn/longhorn/issues/11635#issuecomment-3552019369) also supports `v2` volumes? 

I followed [the test steps](https://github.com/longhorn/longhorn/issues/11635#issuecomment-3552019369) on a v2 volumes, but no snapshot purge was triggered in `step 5. Remove the snapshot snapshot 2 -> trigger snapshot purge`. I also tried triggering snapshot purge manually multiple times right after deleting the snapshot, and they all returned `200 OK`.

davidcheng0922:
@yangchiu 
Sorry — this improvement applies to the v1 engine, where snapshot purging can take lots of  time. In the v2 engine, snapshot purging is much faster, so manual testing may not clearly reproduce the blocking behavior.

yangchiu:
[...]

PR #12254, opened 2025-11-26
feat(chart,upgrade-responder): add snapshot-heavy-task-concurrent-limit setting

davidcheng0922:

https://github.com/longhorn/longhorn/issues/11635

### DIFF
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -893,6 +893,13 @@ questions:
         min: 1
         max: 60
         default: 10
+      - variable: defaultSettings.snapshotHeavyTaskConcurrentLimit
+        label: Snapshot Heavy Task Concurrent Limit
+        description: >-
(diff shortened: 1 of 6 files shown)


chart/questions.yaml
```diff
@@ -893,6 +893,12 @@ questions:
         min: 1
         max: 60
         default: 10
+      - variable: defaultSettings.snapshotHeavyTaskConcurrentLimit
+        label: Snapshot Heavy Task Concurrent Limit
+        description: >-
+          Setting that controls how many snapshot heavy task operations (such as purge and clone) can run concurrently per node. This is a best-effort mechanism: due to the distributed nature of the system, temporary oversubscription may occur. The limiter reduces worst-case overload but does not guarantee perfect enforcement.        type: int
+        min: 0
+        default: 5
```
derekbit, line 902:
```suggestion
        description: >-
          Setting that controls how many snapshot heavy task operations (such as purge and clone) can run concurrently per node. This is a best-effort mechanism: due to the distributed nature of the system, temporary oversubscription may occur. The limiter reduces worst-case overload but does not guarantee perfect enforcement.
        type: int
        min: 0
        default: 5
```

davidcheng0922: thank you, my mistake. Fix.

Issue #11754, opened 2025-09-11
[FEATURE] custom annotation/label of UI's k8s service on value.yaml of helm chart


mo124121:
#### Situation

Now I am using longhorn with istio in ambient mesh. In order to enable CUSTOM authorization policy that requires L7 security policy, I needed to add label on `longhorn-frontend` k8s service. Related issue is below.
https://github.com/istio/istio/issues/51214#issuecomment-2809020176

As a workaround, I labeled the service by kubectl and succeeded to enable the authorization policy.
```
k label -n longhorn-system service/longhorn-frontend istio.io/ingress-use-waypoint="true"
```

However, I would like to avoid manual operation to keep IaC and GitOps by Argo CD.

#### Environment
longhorn: 1.9.1
istio: 1.26.3

### Describe the solution you'd like

Adding sections like `service.ui.labels` on values.yaml of helm chart. An example is as below.

```
service:
  ui:
    # -- Service type for Longhorn UI. (Options: "ClusterIP", "NodePort", "LoadBalancer", "Rancher-Proxy")
    type: ClusterIP
    # -- NodePort port number for Longhorn UI. When unspecified, Longhorn selects a free port between 30000 and 32767.
    nodePort: null
    # -- service labels for UI
    labels: {}
    # -- service annotations for UI
    annotations: {}
  manager:
    # -- Service type for Longhorn Manager.
    type: ClusterIP
    # -- NodePort port number for Longhorn Manager. When unspecified, Longhorn selects a free port between 30000 and 32767.
    nodePort: ""
    # -- service labels for manager
    labels: {}
    # -- service annotations for manager
    annotations: {}
```
https://github.com/longhorn/charts/blob/43dea24e79455e141179355d212c828ecbf1244a/charts/longhorn/values.yaml#L142-L152

We also need to update templates of helm chart to apply those parameters on the services.

### Describe alternatives you've considered

It seems that there is already `longhornManager.serviceAnnotations` section on the values.yaml, but no section for others.
https://github.com/longhorn/charts/blob/43dea24e79455e141179355d212c828ecbf1244a/charts/longhorn/values.yaml#L427

In order to keep backward compatibility, it might be better to add sections like `longhornUI.serviceLabels`. But personally it seems to be less organized compared with the first proposal.

### Additional context

I am writing this issue here because I can not see issues tab in GitHub repository. But please guide me if there is better place.
https://github.com/longhorn/charts

[...]

PR #12276, opened 2025-11-30
feat(chart): add labels setting for longhorn ui and manager services

lucasl0st:
implements Issue #11754

Adds parameters to the helm chart to optionally provide additional labels to the longhorn ui service and for the manager service.
Useful when using teleport which can expose the service to specific users/groups using labels.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -465,6 +465,11 @@ longhornManager:
   ## and uncomment this example block
   #  annotation-key1: "annotation-value1"
   #  annotation-key2: "annotation-value2"
+  serviceLabels: {}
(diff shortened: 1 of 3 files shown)


chart/values.yaml
```diff
@@ -153,6 +153,11 @@ service:
     ## and uncomment this example block
     #  annotation-key1: "annotation-value1"
     #  annotation-key2: "annotation-value2"
+    labels: { }
```
davidcheng0922: redundant space

lucasl0st: fixed

derekbit:
Thanks @lucasl0st's contribution!

Issue #8737, opened 2024-06-12
[IMPROVEMENT] Speed up replica rebuilding by reuse the last transferred data 


PhanLe1010:
## Is your improvement request related to a feature? Please describe (👍 if you like this request)

Longhorn already has the fast replica rebuilding feature which is great. It relies on the checksums of snapshot disk files which is calculated sometime after the snapshot is created and the rebuilding has finished (ref [link](https://github.com/longhorn/longhorn-manager/blob/b5f679491ad34955c0574c645bcd8c3b6f1bcc70/controller/monitor/snapshot_monitor.go#L363-L365))

However, the current feature fast replica rebuilding feature cannot solve the following use case:
1. Users create a volume of 1 replica
2. Users increase number of replicas to 2
3. Rebuilding progress reach 99% and failed
4. Longhorn doesn't try to skip transferring the last transferred data because it doesn't have a checksum yet. Longhorn would have to fallback to the mechanism:
    1. Loop over each data chunk
    2. Compare the checksum of data chunk on the source and on the rebuilding replica (the checksum calculation cost CPU)
    3. Transfer the data if mismatch 

Some ideas to help with the above use case:

Idea 1: (simple implementation - limited benefit)
1. Do the continuous checksum while transferring the data to the rebuilding replica on the rebuilding replica side 
2. Once the snapshot file is transferred record the checksum immediately.
3. If the rebuilding replica failed, the next time we rebuild by reusing this replica, we can skip this snapshot 
4. This only helps to speed up the process at the snapshot completion milestones

Idea 2: (complex implementation - more benefit)
@shuo-wu has an idea of:
1. Do the continuous checksum while transferring the data to the rebuilding replica on the rebuilding replica side 
2. Record the continuous checksum in a separate file 
3. If the rebuilding replica fails, next time we rebuild by reusing this replica, we can skip the transferred data region if it has the same checksum as the one recorded in the file above
4. This helps to speed up the process at the data chunk completion milestones

## Additional context:
We need to investigate how trim feature can affect these ideas 
 

cc @shuo-wu @ejweber @innobead @derekbit

derekbit:
Should we issue a sync request for each rebuilt snapshot file?

PhanLe1010:
> Should we issue a sync request for each rebuilt snapshot file or each data region?

[...]

PR #12298, opened 2025-12-08
docs(lep): reuse transferred data when rebuilding

mantissahz:

Issue # longhorn/longhorn#8737


enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,238 @@
...
+### Related Issues
+
+- https://github.com/longhorn/longhorn/issues/8737
+
+## Motivation
+
+### Goals
+
+- Integrate a lightweight SQLite database into the sparse-tools client.
+- Calculate and store checksums (e.g., xxHash or SHA-256) for every data interval sent during the delta rebuilding process.
```
shuo-wu: Do we need to use CRC64 to make consistency with the v2 implementation? CRC64 generates a 64-bit checksum, which is space-efficient in our use case.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,238 @@
...
+    offset INTEGER NOT NULL,
+    size INTEGER NOT NULL,
+    checksum TEXT NOT NULL,
+    timestamp DATETIME DEFAULT CURRENT_TIMESTAMP
+);
+```
+
+3. Skip transferring data or calculate and store data checksum:
+
+- Check if the snapshot image files on the client and server sides are modified
```
shuo-wu: For each snapshot file, we need to record the replica directory name in case of multiple replicas of one volume in the same disk

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,238 @@
...
+- If no, calculate the checksum of the data buffer, and sending data buffer with the checksum to the server.
+  - Write the data to the image file and store the checksum into the database.
+
+Inside the data transmission loop, Longhorn will:
+
+- Check if the data checksum exists, and whether it matches.
+- If yes, skip transferring data and process the next data chunk.
+- If not, calculate the checksum of the data buffer.
+- Send the checksum with the data buffer to the server (the rebuilding node).
+- Execute an SQL INSERT statement to record the data interval details on both the client and server sides:
```
shuo-wu: Will we use 2 threads handling them simultaneously here? Will the SQL INSERT or even the above SQL GET degrade the rebuilding speed?

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,238 @@
...
+```
+
+3. Skip transferring data or calculate and store data checksum:
+
+- Check if the snapshot image files on the client and server sides are modified
+- If yes, go to get checksums in the database and check if they are identical on both sides.
+- If no, calculate the checksum of the data buffer, and sending data buffer with the checksum to the server.
+  - Write the data to the image file and store the checksum into the database.
+
+Inside the data transmission loop, Longhorn will:
```
shuo-wu: After the rebuilding, when a snapshot is modified (due to snapshot deletion), will Longhorn do any active cleanup?

mantissahz: I design to do cleanup after every rebuilding now (step 4), but it might be better to do cleanup when a snapshot is modified (snapshot deletion or purge).
WDYT, @shuo-wu?

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Modified to use go routines to execute SQL INSERT on both sides.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Added the description to point out for each replica in different directory.
Modified `snapshots` table to save the replica directory.
Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: If we don't have any security concerns, I believe we can use `xxxhash` to get the checksum quickly.
Yes, we might make our behaviors consistent for Longhorn volumes.
I'm fine either way.
WDYT, @derekbit?

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
shuo-wu: We should not do any cleanup immediately after the rebuilding. Otherwise, there is another rebuilding happening later on, Longhorn should read and calculate each chunk for the disk, which does not make sense.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
shuo-wu: I mean, "Send the checksum with the data buffer to the server (the rebuilding node)" and `SQL INSERT` can be handled simultaneously with multiple goroutines, since the src replica may have no DB record for the 1st rebuilding.

shuo-wu:
Had a discussion with @mantissahz and @derekbit, here is the summary:
1. The chunk (checksum) table can be a general solution:
    1. It can be generated not only during rebuilding, but also during regular snapshot checksum calculation (like v2). Then, during a rebuilding, there is a higher chance for replicas to skip the chunk checksum table generation stage then speed up the progress.
    2. If Longhorn has the table before rebuilding, then there is no need to retrieve the fiemap as the chunk checksum table will tell which are data vs holes.
2. Need to mention how to operate or update the tables in all 3 cases: full rebuilding, delta rebuilding, fast rebuilding
3. Need to explain in detail how to rely on those timestamps to verify and reuse a sync-in-progress snapshot file in the rebuilding-failed replica. e.g., In the previous failed rebuilding, Longhorn synced up 1TB of data for a huge snapshot file then got failed by timeout. If Longhorn retries, how can it reuse the failed rebuilding replica and skip the first 1TB data?

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
+# Keep Checksums Of Delta Rebuilding
```
derekbit: The title is incorrect. The goal of the LEP is to speed up a failed replica rebuilding rahter than keeping checksums. Need to update the tittle and summary to reflect the purpose.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Summary
+
+This enhancement outlines a method to persist checksums of transmitted data chunks during the Longhorn volume delta rebuilding process. By integrating SQLite into the sparse-tools client, we can store checksums immediately after data is successfully sent. This feature is currently supported only for the V1 data engine.
```
derekbit: Update the summary.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
...
+
+### Related Issues
+
+- https://github.com/longhorn/longhorn/issues/8737
+
+## Motivation
+
+### Goals
+
+- A global setting to enable or disable this feature.
```
derekbit: As the Longhorn convention, besides of the global setting, need to add a per-volume field to enable or disable the feature.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
...
+
+This enhancement outlines a method to persist checksums of transmitted data chunks during the Longhorn volume delta rebuilding process. By integrating SQLite into the sparse-tools client, we can store checksums immediately after data is successfully sent. This feature is currently supported only for the V1 data engine.
+
+### Related Issues
+
+- https://github.com/longhorn/longhorn/issues/8737
+
+## Motivation
+
+### Goals
```
derekbit: Refine the goals, as mentioned above, the goals should be
- Speed up failed delta replica rebuilding
- Persistent checksums (file checksum or ?) for what...

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
...
+
+- A global setting to enable or disable this feature.
+- Integrate a lightweight SQLite database into the sparse-tools client.
+- Calculate and store checksums (e.g., CRC64 or SHA-256) for every data interval sent during the delta rebuilding process.
+- Provide persistent records of the replica's data intervals for future verification.
```
derekbit: Probaby no need to mention the implementation methods. They are really not the goals.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
...
+- Skip transferred data when retrying the replica rebuilding process.
```
derekbit: Mention why we need to skip. For what?

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
...
+
+### Non-goals [optional]
+
+- This enhancement does not cover the logic for verifying the data against these checksums on the server side (receiver).
```
derekbit: This is important for data integrity. Why don't we support it?

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
...
+- Calculate the data interval checksums and store checksums into the database when the snapshot is created.
+
+## Proposal
+
+### User Stories
```
derekbit: The user story is incomplete. The follow items are more for implementation. Instead, please elaborate
- The current pain points without the implement.
- Why do we need to introduce the feature/improvement?

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
...
+
+- Integrity: It ensures that the data sent during a rebuild matches the source data by cross-referencing the stored checksums.
+- Resume-like functionality: Longhorn will skip re-sending chunks that have already been confirmed as sent during a previous rebuild attempt, using the local checksum database as a state tracker.
+
+### User Experience In Detail
+
+The user experience remains largely unchanged for the end-user operating Longhorn via the UI or kubectl. However, for developers and advanced users debugging the system:
+
+1. A new SQLite database file (e.g., rebuild_tracking.db) will be created in the Longhorn working directory (`/var/lib/longhorn/`).
```
derekbit:
```suggestion
1. A new SQLite database file (e.g., rebuild_tracking.db) will be created in the Longhorn replica directory (`/var/lib/longhorn/replicas/<replica directory>`).
```

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
...
+2. During sync operations, a rebuilding starts, a slight performance degradation may be noticeable due to checksum calculation.
```
derekbit: What does `sync operations` mean? 

> slight performance degradation

This is important. Please provide the number (percentage) of the performance degradation

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
...
+3. If a rebuild fails, the database file can be retrieved by Longhorn to analyze the progress and data integrity state.
```
derekbit: How does Longhorn analyze the progress and data integrity state? Please elaborate on how to do it.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
...
+## Proposal
+
+### User Stories
+
+- Integrity: It ensures that the data sent during a rebuild matches the source data by cross-referencing the stored checksums.
+- Resume-like functionality: Longhorn will skip re-sending chunks that have already been confirmed as sent during a previous rebuild attempt, using the local checksum database as a state tracker.
+
+### User Experience In Detail
+
+The user experience remains largely unchanged for the end-user operating Longhorn via the UI or kubectl. However, for developers and advanced users debugging the system:
```
derekbit: 1. The newly introduced gobal setting and per-volume setting are missing.
2. UI needs to be updated as well.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Design
+
+### Implementation Overview
+
+The core logic changes will occur in `longhorn-engine/pkg/sync/rpc/server.go`, `sparse-tools/sparse/client.go` and `sparse-tools/sparse/rest/handlers.go`.
+
+#### A Global Setting `skip-transferred-data-when-rebuilding`
+
+This enhancement will increase the rebuilding time and disk usage for calculating the data interval checksums and inserting the records into the database.
```
derekbit:
```suggestion
This enhancement will increase the rebuilding time and disk usage for calculating the data interval checksums and inserting the records into the database.
```

This is an important information. Please provide the numbers (percentages) of the increase of the rebuilding time.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
...
+```
+
+#### Skip Transferred Data During Rebuilding
+
+Calculate and store data checksum if the data interval is not transferred.
+
+##### Full Rebuilding
+
+- Whether snapshot image file is modified
+  - Check if the `ctime` and `mtime` of snapshot image file in its directory (multiple replicas of a volume on the same disk) on the client and server sides are different from the record in the database.
```
derekbit: Correct?

```suggestion
  - Check if the `ctime` and `mtime` of snapshot image file in its directory (multiple replicas of a volume on the same disk) on the client (healthy replica) and server (rebuilt replica) sides are different from the record in the database.
```

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
...
+- If yes, back to the original implementation, it will calculate the checksum of the data interval on both sides and check if two checksums are the same.
+  - Save these checksums into the database on both sides.
+- If no, compare the checksums in both databases (client and server sides) for each data interval.
+  - Check if checksums for the data interval in both databases exist and are the same
+    - If yes, skip transferring this data interval, and handle next data interval.
+    - If no, back to the original implementation, it will calculate the checksum of the data interval on both sides and check if two checksums are the same.
+      - Save these checksums into the database on both sides.
+
+Pseudocode modification for `sparse-tools/sparse/client.go` (client side):
+
```
derekbit: Please add a paragraph or section describing how to avoid the race contention while update the database.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
derekbit: What are the ctime and mtime? Where are they from? The information is missing.

Please elaborate on
- checksum database open
- while rebuilding
- checksum database close

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
...
+  // New: Update the snapshot list (delete obsolete snapshots, such as those that have been deleted or merged) 
+  err := fromClient.UpdateSnapshotList(s.volumeName, req.SyncFileInfoList)
+  err := updateSnapshotList(s.volumeName, req.SyncFileInfoList)
+  ...
+  return nil
+}
+```
+
+### Test plan
+
```
derekbit: Also need to add the test cases

1. Performance: compare the performance with and without the feature
2. Corrupt the database and make sure the database can be cleaned up and fall back to the original replica rebuilding

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
...
+    mtime DATETIME NOT NULL
+);
+CREATE TABLE IF NOT EXISTS sent_chunks (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    begin INTEGER NOT NULL,
+    end INTEGER NOT NULL,
+    checksum INTEGER NOT NULL
+);
+```
+
```
derekbit: If the database is somehow corrupted, how to handle it? Please address it.

derekbit:
> Had a discussion with @mantissahz and @derekbit, here is the summary:
> 
> 1. The chunk (checksum) table can be a general solution:
>    
>    1. It can be generated not only during rebuilding, but also during regular snapshot checksum calculation (like v2). Then, during a rebuilding, there is a higher chance for replicas to skip the chunk checksum table generation stage then speed up the progress.
>    2. If Longhorn has the table before rebuilding, then there is no need to retrieve the fiemap as the chunk checksum table will tell which are data vs holes.
> 2. Need to mention how to operate or update the tables in all 3 cases: full rebuilding, delta rebuilding, fast rebuilding
> 3. Need to explain in detail how to rely on those timestamps to verify and reuse a sync-in-progress snapshot file in the rebuilding-failed replica. e.g., In the previous failed rebuilding, Longhorn synced up 1TB of data for a huge snapshot file then got failed by timeout. If Longhorn retries, how can it reuse the failed rebuilding replica and skip the first 1TB data?

@mantissahz The point 3 is not addressed in the latest LEP.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
derekbit: What will happen if a snapshot having a checksum database is merged? How to handle the checksum database? Please address it in the LEP.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
...
+    Required:           true,
+    ReadOnly:           false,
+    DataEngineSpecific: true,
+    Default:            fmt.Sprintf("{%q:\"true\"}", longhorn.DataEngineTypeV1),
+  }
+```
+
+2. Send the value of the setting `skip-transferred-data-when-rebuilding` from `longhorn-manager/engine_controller` through `longhorn-instance-manager` to `longhorn-engine`
+
+#### Database Initialization
```
derekbit: The management and life cycle of a database is not explained in the LEP.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
derekbit: > transmitted data chunks during the 


What is the size of a data chunk, and how is it determined? Is it related to extents? If so, when two image files have identical content, are their database records the same? Is it possible for the records to differ even if the content is identical?

For example

Different records in the databased
- database 1 of disk file A
  - 0-1M: checksumxx
  - 1M-2M: checksumyy

- database 2 of disk file B
  - 0-2M: checksumzz

, but the file contents are same.

The infromation is missing in the LEP.

derekbit: > transmitted data chunks 

What determines the size of a data chunk? Is it tied to extents? If two image files have the same content, would they result in the same database records, or could the records differ even though the content is identical?

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
```diff
@@ -0,0 +1,286 @@
...
+
+N/A
+
+## Design
+
+### Implementation Overview
+
+The core logic changes will occur in `longhorn-engine/pkg/sync/rpc/server.go`, `sparse-tools/sparse/client.go` and `sparse-tools/sparse/rest/handlers.go`.
+
+#### A Global Setting `skip-transferred-data-when-rebuilding`
```
derekbit:
```suggestion
#### A Global Setting `reuse-transferred-data-when-rebuilding`
```

mantissahz: Modified. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Renamed the file name and modified the title. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Updated. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Added. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Refined. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Updated and removed some. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Updated. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Removed this non-goal.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Updated the user story. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Removed this part.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Removed this part. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Updated. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Leave it until the implementation.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Updated the descriptions. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Added. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Added. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Added. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Added descriptions. Thanks

mantissahz: Added it in the databse life cycle. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Added. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Now the source replica node will store the checksum in the database, and target node will not.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Yes, the database file will be deleted when snapshot is deleted or merged, or the database file is corrupted when the rebuild starts.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: leave it until the implementation. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Use SHA256 for low collision.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: > If so, when two image files have identical content, are their database records the same?

I believe they will be the same, otherwise this function [CompareLocalAndRemoteInterval](https://github.com/longhorn/sparse-tools/blob/ae3aca526fd7fabbe0f8a23a3d7f49ca397a232f/sparse/client.go#L499-L543) will fail.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,300 @@
...
+    begin INTEGER NOT NULL,
+    end INTEGER NOT NULL,
+    checksum INTEGER NOT NULL
+);
+```
+
+##### Life Cycle
+
+- Databases will be created only on client and server sides (health and rebuilding replica nodes) when a rebuild starts.
+- System will always keep the database on client side, and it will delete the database on server side if there is no record in the database after rebuilding.
```
derekbit: > if there is no record in the database after rebuilding.

What does `no record` mean?
Why do we need to remove it?

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,300 @@
...
+- If yes, back to the original implementation, it will calculate the checksum of the data interval on both sides and check if two checksums are the same.
+  - Save these checksums into the database on both sides.
+- If no, get the checksum in the client database, and reading data and calculate the checksum on the server side for each data interval.
+  - Check if two checksums are the same
+    - If yes, skip transferring this data interval, and handle next data interval.
+    - If no, reading data and calculate the checksum of this data interval on client sides and check if two checksums are the same.
+      - If yes, skip transferring this data interval.
+      - If no, send data buffer to the server.
+      - Save these checksums into the database on both sides.
+- Update snapshot `ctime` and `mtime` in the database on client side.
```
derekbit: Should we get the timestamps after or before closing the snapshot image file? Please add the order of the snapshot file and database close operations.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,300 @@
...
+  The `client.syncBatchSize` will be `defaultSyncBatchSize   = 512 * Blocks` as 2 MiB by default.
+
+#### Skip Transferred Data During Rebuilding
+
+Get snapshot image file information, `ctime` and `mtime`:
+
+- **ctime**:
+  Change time (inode change time), the last time the file’s metadata (inode) changed.
+- **mtime**:
+  Modification time, the last time the file’s content was modified.
```
derekbit: Should we get the timestamps after or before closing the snapshot image file?

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,300 @@
...
+- If a snapshot database can not be accessed or corrupted (opening or reading error), the database can be deleted and recreate a new one when a rebuild starts.
+
+#### Data Interval And Chunk Size
+
+- Data interval size is depending on the Filesystem:
+
+  | Filesystem | Minimal Data Interval Size | Maximal Data Interval Size |
+  | :---: | :---: | :---: |
+  | Ext4 | 4 KiB | 128 MiB |
+  | XFS | 4 KiB| <= Allocation Group size |
```
derekbit: Can you provide the reference doc for the values?

> Allocation Group size

What does it mean?

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,300 @@
...
+```
+
+##### Life Cycle
+
+- Databases will be created only on client and server sides (health and rebuilding replica nodes) when a rebuild starts.
+- System will always keep the database on client side, and it will delete the database on server side if there is no record in the database after rebuilding.
+- When a snapshot is deleted or merged, Longhorn needs to delete its database for rebuilding if exists.
+- If a snapshot database can not be accessed or corrupted (opening or reading error), the database can be deleted and recreate a new one when a rebuild starts.
+
+#### Data Interval And Chunk Size
```
derekbit: Need to answer 
> If two image files have the same content, would they result in the same database records, or could the records differ even though the content is identical?

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
derekbit: > I believe they will be the same

Can you elaborate more on the statement?

derekbit: > otherwise this function [CompareLocalAndRemoteInterval](https://github.com/longhorn/sparse-tools/blob/ae3aca526fd7fabbe0f8a23a3d7f49ca397a232f/sparse/client.go#L499-L543) will fail.

Well, the logic is different than the scenario in my question...
`CompareLocalAndRemoteInterval()` is calculating the same data interval, but my question is if the records in the database can match to the data interval of the file. If yes, we can reuse the database.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,300 @@
...
+
+- ✓: The skipping transferred data is enabled.
+- X: The skipping transferred data is disabled.
+
+**Debugging:**
+
+For developers and advanced users:
+
+1. A new SQLite database file (e.g., `volume-snap-*.img.db`) will be created in the replica directory.
+2. During the first rebuilding, a slight performance degradation may be noticeable due to checksum calculation and database operations.
```
derekbit: Need to provide the number of the performance degradation as I commented before.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,300 @@
...
+  type UpdateReuseTransferredDataWhenRebuildingInput struct {
+    ReuseTransferredDataWhenRebuilding bool `json:"reuseTransferredDataWhenRebuilding"`
+  }
+  ```
+
+## Design
+
+### Implementation Overview
+
+This enhancement will increase the rebuilding time and disk usage for calculating the data interval checksums and inserting the records into the database.
```
derekbit: > increase the rebuilding time and disk usage

Please provide the increase percentages for the statement.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,300 @@
...
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    begin INTEGER NOT NULL,
+    end INTEGER NOT NULL,
+    checksum INTEGER NOT NULL
+);
+```
+
+##### Life Cycle
+
+- Databases will be created only on client and server sides (health and rebuilding replica nodes) when a rebuild starts.
```
derekbit: Should we use cache mode or sync mode to open the database?

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,300 @@
...
+1. Check if the snapshot information is stored in the database before the transferring data, if not, create a record with the snapshot `ctime` and `mtime`, and send a flag as the snapshot is modified to the transferring data loop.
+
+2. Check if the `ctime` and `mtime` of snapshot image file in its replica directory (multiple replicas of a volume on the same disk) are different from the record in the database.
+
+##### Full Rebuilding
+
+- If yes, update the snapshot information in database, and calculate and save the checksum into the database (on client side), and then send data buffer to the server for each data interval.
+- If no, send data buffer to the server for each data interval.
+- Remove the database on the server side.
+- Update snapshot `ctime` and `mtime` in the database on client side.
```
derekbit: Same question as above.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
derekbit: Same question as above.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,300 @@
...
+    - If yes, skip transferring this data interval, and handle next data interval.
+    - If no, reading data and calculate the checksum of this data interval on client sides and check if two checksums are the same.
+      - If yes, skip transferring this data interval.
+      - If no, send data buffer to the server.
+      - Save these checksums into the database on both sides.
+- Update snapshot `ctime` and `mtime` in the database on client side.
+
+##### Database Concurrency Control
+
+Longhorn uses 4 go routines to calculate the checksum and consume the data interval concurrently. To avoid race conditions when updating the database, a lock is required to ensure that database write operations are serialized.
```
derekbit: > a lock 


mutex or rwlock?

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
derekbit: If the client side (healthy replica) already has a database, do we use, recreate or delete it? 
If the server side (failed replica) already has a database, do we use, recreate or delete it?

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Removed the deleting database part. We can keep it even it is empty.
We will rebuild it if it becomes a client side when the rebuild starts next time.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Updated the descriptions. We should get the new info. after closing the snapshot file here. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Updated the descriptions. We will get the info before we open the snapshot file here.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: >Can you provide the reference doc for the values?

Added.

>Allocation Group size

Updated the new table and removed this. It needs to be assign when formating the file system. But it is not the real limitation. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Use `sync` mode. The default synchronous mode is `FULL`.
Added descriptions. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Update the descriptions. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: RWMutex for multipe reads and single write. Updated the decription. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: We will use it if already has a database for retrying rebuilding.
And we will check if the records of database are consistent with another side (reading the checksum from the database or calculating the data buffer checksum)
Updated the description. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Updated the description. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: A record `25` of the table `sent_chunks` in database "snapshot-A.img.db" stores, for example:
```
begin: 2
end: 5
checksum: ABC
```
The checksum of the data interval{begin: 2, end: 5} of the snapshot `snapshot-A.img` (real data) should be `ABC` if the snapshot `snapshot-A.img` is not modified. (They, `ctime` and `mtime`, will be compared before the rebuild starts, and we stored this record `25` by previous rebuilding)

That's why they will be the same.

derekbit: No, I don't mean this.

If the record [2, 5] is from the client side (healthy replica), the server side needs to read data from [2, 5] this is not determined by the server side itself. However, can you make sure the [2, 5] is in the same extent? If the server side read the data interval from the file, can you make sure it must be [2, 5] rather than [0, 5]?

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
derekbit: The question is not clarified.
https://github.com/longhorn/longhorn/pull/12298#discussion_r2621262973

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+### Related Issues
+
+- https://github.com/longhorn/longhorn/issues/8737
+
+## Motivation
+
+### Goals
+
+- A global setting and per volume setting to enable or disable this feature.
+- Speed up failed delta replica rebuilding via calculating the data interval checksums during rebuilding to skip transferred data when retrying the replica rebuilding process and provide persistent records for future verification.
```
derekbit: > provide persistent records for future verification.

How to use the records for verification?

derekbit:
@mantissahz Remember to update the misleading titles of the PR and commit.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+
+## Proposal
+
+### User Stories
+
+Users will want the rebuilding process to skip transferring data chunks that have already been successfully synchronized in a previous attempt, rather than starting from scratch, to minimize the time the volume is in a degraded state and reduce network bandwidth usage and speeding up recovery.
+
+### User Experience In Detail
+
+In some situations, for instance, environments with unstable network connections, replica rebuilding may frequently fail due to interruptions. Without this feature, each retry starts the rebuilding process from the beginning. If the network instability persists, the rebuilding process might never complete, leaving the volume in a degraded state. By enabling this feature, Longhorn calculates the checksums of transferred data of the snapshots and persists the records in the local database. This allows the rebuilding process to skip transferred data from the beginning to the point of failure, ensuring that the replica eventually becomes healthy even in poor environment conditions.
```
derekbit: >  This allows the rebuilding process to skip transferred data from the beginning to the point of failure

The LEP doesn't really skip transferred data. It only saves the calculation of the checksums of data intervals.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+| `false`                                       | `disabled`                                    | No                         |
+
+- ✓: The skipping transferred data is enabled.
+- X: The skipping transferred data is disabled.
+
+**Debugging:**
+
+For developers and advanced users:
+
+1. A new SQLite database file (e.g., `volume-snap-*.img.db`) will be created in the replica directory.
```
derekbit: Can you separate the cases for 
- Client side
	1. healthy replica that doesn't have the database file
	2. healthy replica that doesn't have the database file. Encounter a failure and then restart.
	3. healthy replica that has the database
	4. ....
- Server side
  ...

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+    ReuseTransferredDataWhenRebuilding bool `json:"reuseTransferredDataWhenRebuilding"`
+  }
+  ```
+
+## Design
+
+### Implementation Overview
+
+This enhancement will increase the rebuilding time and disk usage for calculating the data interval checksums and inserting the records into the database.
+Therefore, users can disable the feature via settings if they don't want to waste time building the checksum database for rebuilding.
```
derekbit: Suggest removing it because it is not the implementation detail. We can provide the numbers and add the notice in the official doc instead.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+
+2. Send the value of the setting `reuse-transferred-data-when-rebuilding` or `Volume.Spec.ReuseTransferredDataWhenRebuilding` from `longhorn-manager/engine_controller` through `longhorn-instance-manager` to `longhorn-engine`
+
+#### Database Initialization And Life Cycle
+
+##### Initialization
+
+Longhorn will need to add a Go SQLite driver (e.g., pure Go libraries `modernc.org/sqlite`) to the `sparse-tools` go.mod file.
+
+Before the synchronizing snapshot content loop begins, the client will initialize database connections on client and server sides and create tables if they don't exist. The database file name will be the snapshot image file name with a suffix `.db`, for example `volume-snap-de52a2e1-5267-4a50-a031-2ba273548c47.img.db` and will be stored in the replica directory of the worker node. The database will be saved after rebuilding.
```
derekbit: > The database will be saved after rebuilding.

Sould we save the databases on both server and client sides?

derekbit: > the client will initialize database connections on client and server sides and create tables if they don't exist.

Do we need to reuse the database in the server side (failed replica)?
Is it possible that a record is in the database but the data is not written into the snapshot image file?

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+
+Use the default synchronous mode `FULL`: it will wait for all changes to be flushed to disk before continuing.
```
derekbit: `FULL` is `sync` and might significally slow down the rebuilding process. Do we really need to use it?

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+ctime := time.Unix(int64(stat.Ctim.Sec), int64(stat.Ctim.Nsec)).String()
+```
+
+1. Check if the snapshot information is stored in the database before the transferring data, if not, create a record with the snapshot `ctime` and `mtime`, and send a flag as the snapshot is modified to the transferring data loop.
+
+2. Check if the `ctime` and `mtime` of snapshot image file in its replica directory (multiple replicas of a volume on the same disk) are different from the record in the database.
+
+##### Full Rebuilding
+
+- If yes, update the snapshot information in database, and calculate and save the checksum into the database (on client side), and then send data buffer to the server for each data interval.
```
derekbit: What's the snapshot information?

derekbit: Need the peformance benchmark for enabling the checksum calculation.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+- If no, send data buffer to the server for each data interval.
+- Remove the database on the server side.
```
derekbit: Why do we need to remove the database? Can you add the reason here?

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+
+```go
+fileInfo, err := os.Stat(snapshotName)
+mtime := fileInfo.ModTime().String()
+
+stat := fileInfo.Sys().(*syscall.Stat_t)
+ctime := time.Unix(int64(stat.Ctim.Sec), int64(stat.Ctim.Nsec)).String()
+```
+
+1. Check if the snapshot information is stored in the database before the transferring data, if not, create a record with the snapshot `ctime` and `mtime`, and send a flag as the snapshot is modified to the transferring data loop.
```
derekbit: > Check if the snapshot information is stored in the database  ...

What's the information?

derekbit: > send a flag as the snapshot is modified to the transferring data loop.

What's the purpose for sending a flag?

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+- If yes, update the snapshot information in database, and calculate and save the checksum into the database (on client side), and then send data buffer to the server for each data interval.
+- If no, send data buffer to the server for each data interval.
+- Remove the database on the server side.
+- Get the snapshot `ctime` and `mtime` after closing the snapshot file and update snapshot `ctime` and `mtime` in the database on client side.
+
+##### Delta Rebuilding
+
+- If yes, back to the original implementation, it will calculate the checksum of the data interval on both sides and check if two checksums are the same.
+  - Save these checksums into the database on both sides.
+- If no, get the checksum in the client database, and reading data and calculate the checksum on the server side for each data interval.
```
derekbit: The formatting looks a bit off. Which decision do these yes and no refer to?

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
derekbit: The formatting looks a bit off. Which decision do these yes and no refer to?

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+  - Check if two checksums are the same
+    - If yes, skip transferring this data interval, and handle next data interval.
+    - If no, reading data and calculate the checksum of this data interval on client sides and check if two checksums are the same.
+      - If yes, skip transferring this data interval.
+      - If no, send data buffer to the server.
+      - Save these checksums into the database on both sides.
```
derekbit: Do we need to save the checksums in the server side?

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+- Get the snapshot `ctime` and `mtime` after closing the snapshot file and update snapshot `ctime` and `mtime` in the database on client side.
```
derekbit: For what?

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+1. Check if the snapshot information is stored in the database before the transferring data, if not, create a record with the snapshot `ctime` and `mtime`, and send a flag as the snapshot is modified to the transferring data loop.
+
+2. Check if the `ctime` and `mtime` of snapshot image file in its replica directory (multiple replicas of a volume on the same disk) are different from the record in the database.
+
+##### Full Rebuilding
+
+- If yes, update the snapshot information in database, and calculate and save the checksum into the database (on client side), and then send data buffer to the server for each data interval.
+- If no, send data buffer to the server for each data interval.
+- Remove the database on the server side.
+- Get the snapshot `ctime` and `mtime` after closing the snapshot file and update snapshot `ctime` and `mtime` in the database on client side.
```
derekbit: For what?

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+    }
+    ...
+  }
+  return nil
+}
+...
+```
+
+### Test plan
+
```
derekbit, line 300: Add benchmark for the full rebuilding and compare the performance enabling and disabling the feature.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+Use the default synchronous mode `FULL`: it will wait for all changes to be flushed to disk before continuing.
+
+```sql
+CREATE TABLE IF NOT EXISTS snapshot (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    ctime TEXT NOT NULL,
+    mtime TEXT NOT NULL
+);
+CREATE TABLE IF NOT EXISTS sent_chunks (
```
derekbit:
```suggestion
CREATE TABLE IF NOT EXISTS chunks (
```

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    begin INTEGER NOT NULL,
+    end INTEGER NOT NULL,
```
derekbit:
```suggestion
    offsize INTEGER NOT NULL,
    size INTEGER NOT NULL,
```

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+    checksum INTEGER NOT NULL
```
derekbit: Why is a checksum integer? SHA256 is a string.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+```
+
+##### Life Cycle
+
+- Databases will be created if they do not exist on client and server sides (health and rebuilding replica nodes) when a rebuild starts.
+- System will always keep the databases after rebuilding.
+- When a snapshot is deleted or merged, Longhorn needs to delete its database for rebuilding if exists.
+- If a snapshot database can not be accessed or corrupted (opening or reading error), the database can be deleted and recreate a new one when a rebuild starts.
+
+#### Data Interval And Chunk Size
```
derekbit: What does the `Chunk` mean here?
Can we use either `Data Interval` or `Chunk` in the LEP to make it consistent?

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+
+- Data interval size is depending on the Filesystem:
+
+  Minimal Data Interval Size: 1 Block, 4 KiB.
+
+  | Feature | EXT4 | XFS |
+  | :---: | :---: | :---: |
+  | Length Field | 16-bit (effectively 15-bit*) | 21-bit |
+  | Max Blocks | 32,768 (2^15) | 2,097,151 (2^21−1) |
+  | Max Extent Size | 128 MiB | 8 GiB |
```
derekbit, line 195: Does it mean the max data interval while rebuilding is 8GiB and consume 8GiB memory for one data interval?

mantissahz: The data interval size will up to 8 GiB.
And We will read and send the data chunk in 2MiB.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Updated the LEP and added a graph to explain the difference between the data interval and the data size. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Yes, modified to `TEXT`. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Modified. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Modified. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+```
+
+2. Send the value of the setting `reuse-transferred-data-when-rebuilding` or `Volume.Spec.ReuseTransferredDataWhenRebuilding` from `longhorn-manager/engine_controller` through `longhorn-instance-manager` to `longhorn-engine`
+
+#### Database Initialization And Life Cycle
+
+##### Initialization
+
+Longhorn will need to add a Go SQLite driver (e.g., pure Go libraries `modernc.org/sqlite`) to the `sparse-tools` go.mod file.
+
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+- Get the snapshot `ctime` and `mtime` after closing the snapshot file and update snapshot `ctime` and `mtime` in the database on client side.
+
+##### Delta Rebuilding
+
+- If yes, back to the original implementation, it will calculate the checksum of the data interval on both sides and check if two checksums are the same.
+  - Save these checksums into the database on both sides.
+- If no, get the checksum in the client database, and reading data and calculate the checksum on the server side for each data interval.
+  - Check if two checksums are the same
+    - If yes, skip transferring this data interval, and handle next data interval.
+    - If no, reading data and calculate the checksum of this data interval on client sides and check if two checksums are the same.
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+    end INTEGER NOT NULL,
+    checksum INTEGER NOT NULL
+);
+```
+
+##### Life Cycle
+
+- Databases will be created if they do not exist on client and server sides (health and rebuilding replica nodes) when a rebuild starts.
+- System will always keep the databases after rebuilding.
+- When a snapshot is deleted or merged, Longhorn needs to delete its database for rebuilding if exists.
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+    - If yes, skip transferring this data interval, and handle next data interval.
+    - If no, reading data and calculate the checksum of this data interval on client sides and check if two checksums are the same.
+      - If yes, skip transferring this data interval.
+      - If no, send data buffer to the server.
+      - Save these checksums into the database on both sides.
+- Get the snapshot `ctime` and `mtime` after closing the snapshot file and update snapshot `ctime` and `mtime` in the database on client side.
+
+##### Database Concurrency Control
+
+Longhorn uses 4 go routines to calculate the checksum and consume the data interval concurrently. To avoid race conditions when updating the database, a lock is required to ensure that database write operations are serialized.
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+  ```go
+  func (client *syncClient) syncDataInterval(dataInterval Interval) error {
+    // Process data in chunks
+    for offset := dataInterval.Begin; offset < dataInterval.End; {
+      size := getSize(offset, client.syncBatchSize, dataInterval.End)
+    ...
+  }
+  ```
+
+  The `client.syncBatchSize` will be `defaultSyncBatchSize   = 512 * Blocks` as 2 MiB by default.
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+  - Save these checksums into the database on both sides.
+- If no, get the checksum in the client database, and reading data and calculate the checksum on the server side for each data interval.
+  - Check if two checksums are the same
+    - If yes, skip transferring this data interval, and handle next data interval.
+    - If no, reading data and calculate the checksum of this data interval on client sides and check if two checksums are the same.
+      - If yes, skip transferring this data interval.
+      - If no, send data buffer to the server.
+      - Save these checksums into the database on both sides.
+- Get the snapshot `ctime` and `mtime` after closing the snapshot file and update snapshot `ctime` and `mtime` in the database on client side.
+
```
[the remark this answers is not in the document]
mantissahz: Not modify. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+  7. Trigger another replica rebuild.
+  8. The subsequent rebuild should create a new database. The rebuilding must complete successfully.
+
+- Rebuilding should succeed even if the checksum database is corrupted:
+  1. Create a volume with 3 replicas and write some data.
+  2. Trigger a replica rebuild (e.g., by deleting a replica).
+  3. Wait for the rebuild to make some progress and the database file to be created.
+  4. Corrupt the SQLite database file (e.g., `echo "garbage" > <path-to-db>`).
+  5. Interrupt the rebuild process.
+  6. The subsequent rebuild should detect the database file corruption (or fail to open the DB), log a warning, and fall back to a delta rebuilding (and recreate the DB and proceed). The rebuilding must complete successfully.
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+    ctime TEXT NOT NULL,
+    mtime TEXT NOT NULL
+);
+CREATE TABLE IF NOT EXISTS sent_chunks (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    begin INTEGER NOT NULL,
+    end INTEGER NOT NULL,
+    checksum INTEGER NOT NULL
+);
+```
```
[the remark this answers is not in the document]
mantissahz: Will not Modify. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+
+Introduce a new volume Action API `reuseTransferredDataWhenRebuilding`:
+
+  | API | Input | Output | Comments | HTTP Endpoint |
+  | --- | --- | --- | --- | --- |
+  | Update | N/A | err error | Enable/Disable volume reuses transferred data when rebuilding | **POST** `/v1/volumes/{VolumeName}?action=reuseTransferredDataWhenRebuilding` |
+
+  ```golang
+  type UpdateReuseTransferredDataWhenRebuildingInput struct {
+    ReuseTransferredDataWhenRebuilding bool `json:"reuseTransferredDataWhenRebuilding"`
```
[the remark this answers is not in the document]
mantissahz: Will not modify.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+##### Life Cycle
+
+- Databases will be created if they do not exist on client and server sides (health and rebuilding replica nodes) when a rebuild starts.
+- System will always keep the databases after rebuilding.
+- When a snapshot is deleted or merged, Longhorn needs to delete its database for rebuilding if exists.
+- If a snapshot database can not be accessed or corrupted (opening or reading error), the database can be deleted and recreate a new one when a rebuild starts.
+
+#### Data Interval And Chunk Size
+
+- Data interval size is depending on the Filesystem:
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+        log.Warnf("Failed to get the checksum from database for data interval %+v: %v", batchInterval, err)
+      }
+      dataBuffer, err = client.CompareLocalAndRemoteInterval(batchInterval, checksumFromDB)
+    }
+    ...
+    if dataBuffer != nil {
+      // NEW: Checksum and Store, (e.g., sha256.Sum256(dataBuffer))
+      checksum := calculateChecksum(dataBuffer)
+      go func() {
+        // Here needs a RWMutex Lock (multiple reads, one write) or a channel to ensure the records are not written simultaneously
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+    checksum INTEGER NOT NULL
+);
+```
+
+##### Life Cycle
+
+- Databases will be created if they do not exist on client and server sides (health and rebuilding replica nodes) when a rebuild starts.
+- System will always keep the databases after rebuilding.
+- When a snapshot is deleted or merged, Longhorn needs to delete its database for rebuilding if exists.
+- If a snapshot database can not be accessed or corrupted (opening or reading error), the database can be deleted and recreate a new one when a rebuild starts.
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+
+#### Data Interval And Chunk Size
+
+- Data interval size is depending on the Filesystem:
+
+  Minimal Data Interval Size: 1 Block, 4 KiB.
+
+  | Feature | EXT4 | XFS |
+  | :---: | :---: | :---: |
+  | Length Field | 16-bit (effectively 15-bit*) | 21-bit |
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
```diff
@@ -0,0 +1,307 @@
...
+...
+  SettingDefinitionReuseTransferredDataWhenRebuilding = SettingDefinition{
+    DisplayName:        "Reuse Transferred Data When Rebuilding",
+    Description:        "When enabled, Longhorn will reuse transferred data chunks on the rebuilding replica node during a previous rebuild attempt. This improves rebuild efficiency and reduces unnecessary data transfer, but may increase disk usage and slightly impact performance due to checksum calculation and database storage.",
+    Category:           SettingCategoryGeneral,
+    Type:               SettingTypeString,
+    Required:           true,
+    ReadOnly:           false,
+    DataEngineSpecific: true,
+    Default:            fmt.Sprintf("{%q:\"true\"}", longhorn.DataEngineTypeV1),
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Updated. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Updated. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: No, we don't need. Updated. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Updated. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Updated. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Updated. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Updated. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Updated. Thanks.
We will not have the database on server side.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: We can test 'NORMAL'.
Updated. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Updated. Thanks.

mantissahz: Only on client side.
Updated. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Updated. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Added and Updated. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Updated. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Updated. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: As we discussed, we only have the database on client side.
Updated. Thanks.

enhancements/20251126-reuse-transferred-data-when-rebuilding.md
mantissahz: Updated. Thanks.

enhancements/20251126-keep-checksums-of-delta-rebuilding.md
mantissahz: Updated. Thanks.

PR #12631, opened 2026-02-07
fix(scripts): clarify helm version in generate-longhorn-yaml error message

### LINKED ISSUE #12630: [BUG] generate-longhorn-yaml.sh OKD flag not reset in loop
## Summary

The `scripts/generate-longhorn-yaml.sh` script has two issues:

### 1. OKD flag not reset in loop iteration

The `OKD_ENABLED_FLAG` variable is not reset at the start of each loop iteration. While the current array order (`longhorn.yaml` first, then `longhorn-okd.yaml`) works correctly, if the order changes or more files are added, the flag could persist incorrectly.

### 2. Incorrect error message

The error message says "Please install helm v4 first" but the grep check is `v3\|v4`, so it should say "helm v3 or v4".

## Steps to reproduce

1. Review `scripts/generate-longhorn-yaml.sh`
2. Note that `OKD_ENABLED_FLAG` is only set conditionally but never reset
3. Note the error message mismatch

## Expected behavior

- `OKD_ENABLED_FLAG` should be reset to empty at the start of each loop iteration
- Error message should match the actual version check

luojiyin1987:
## Summary

Clarify the Helm version error message in `scripts/generate-longhorn-yaml.sh` so it matches the existing version check behavior.

Fixes #12630

## Changes

1. Update the error message to: `Please install helm v4 (recommended) or v3 (backward compatible) before generating ...`
2. No logic change in version validation (it still checks `v3|v4`).

## Test plan

- [x] Confirm script behavior is unchanged when Helm v3 or v4 is installed.
- [x] Confirm the error message is now accurate when Helm is missing or unsupported.

### DIFF
--- a/scripts/generate-longhorn-yaml.sh
+++ b/scripts/generate-longhorn-yaml.sh
@@ -10,7 +10,7 @@ DEPLOY_YAML_TMP="$PRJ_DIR/deploy/longhorn.yaml.tmp"
 NAMESPACE=${NAMESPACE:-longhorn-system}
 
 if ! command -v helm &> /dev/null || ! helm version --short | grep -q "v3\|v4"; then
-  echo "Please install helm v4 first before generating $DEPLOY_YAML!"
+  echo "Please install helm v4 (recommended) or v3 (backward compatible) before generating $DEPLOY_YAML!"
   exit 1
 fi
 
@@ -33,4 +33,3 @@ EOD
   mv "$DEPLOY_YAML_TMP" "$DEPLOY_YAML"
 
 done
-


mantissahz:
Hi @luojiyin1987, 

Thanks for contributing the correction.
But for `reset OKD flag in loop`, can you explain how to reset it in this commit?
And please sign off on the commit.

derekbit:
ping @luojiyin1987

luojiyin1987:
@derekbit   I had sign off on the commit

derekbit:
> Hi @luojiyin1987,
> 
> Thanks for contributing the correction. But for `reset OKD flag in loop`, can you explain how to reset it in this commit? And please sign off on the commit.

@luojiyin1987 Could you clarify the question? Thank you.

luojiyin1987:
@derekbit Sorry for the confusion. After reviewing my commit again, I realize the "reset OKD flag" change is not included in the diff - the PR title is incorrect.

  The current code actually works fine because longhorn.yaml is processed first (when OKD_ENABLED_FLAG is empty), then longhorn-okd.yaml second.

  I can either:
  1. Update the PR title to reflect only the error message fix, or
  2. Add the defensive fix (using else OKD_ENABLED_FLAG="") if you think it's worth having

  Which would you prefer?

luojiyin1987:
@derekbit I have updated the PR title and description to match the actual diff. This PR now only includes the Helm error message clarification (no OKD flag reset logic change). Could you please take another look? Thank you.

derekbit:
cc @COLDTURNIP

derekbit:
@luojiyin1987 Could you rebase the PR? Thanks.

luojiyin1987:
@derekbit   Rebased on the latest `master` and force-pushed. Thanks.

derekbit (approved):

Issue #9205, opened 2024-08-07
[BUG] Encrypt volume provided size is 16MB shorter than the claimed size


WebberHuang1118:
## Describe the bug

LH encrypt volume's actual size is 16MB shorter than the claimed size, this could cause integrity issues for space-sensitive applications.

## To Reproduce (follow test plan in https://github.com/longhorn/longhorn/issues/4883#issuecomment-1619514908)
- Creating secret
  ```
  apiVersion: v1
  kind: Secret
  metadata:
    name: longhorn-crypto
    namespace: longhorn-system
  stringData:
    CRYPTO_KEY_VALUE: "simple"
    CRYPTO_KEY_PROVIDER: "secret"
  ```
- Creating storage class
  ```
  kind: StorageClass
  apiVersion: storage.k8s.io/v1
  metadata:
    name: longhorn-migratable
  provisioner: driver.longhorn.io
  allowVolumeExpansion: true
  parameters:
    numberOfReplicas: "1"
    staleReplicaTimeout: "2880" # 48 hours in minutes
    fromBackup: ""
    migratable: "true"
    encrypted: "true"
    csi.storage.k8s.io/provisioner-secret-name: "longhorn-crypto"
    csi.storage.k8s.io/provisioner-secret-namespace: "longhorn-system"
    csi.storage.k8s.io/node-publish-secret-name: "longhorn-crypto"
    csi.storage.k8s.io/node-publish-secret-namespace: "longhorn-system"
    csi.storage.k8s.io/node-stage-secret-name: "longhorn-crypto"
    csi.storage.k8s.io/node-stage-secret-namespace: "longhorn-system"
  ```
- Creating encrypted volume
  ```
  apiVersion: v1
  kind: PersistentVolumeClaim
  metadata:
    name: longhorn-block-pvc
    namespace: default
  spec:
    accessModes:
      - ReadWriteMany
    volumeMode: Block
    storageClassName: longhorn-migratable
    resources:
      requests:
        storage: 5Gi
  ```
- Attaching encrypted volume to a Linux pod
  ```
  apiVersion: v1
  kind: Pod
  metadata:
    name: block-volume-test
    namespace: default
  spec:
    containers:
    - name: volume-test
      image: ubuntu:latest
      imagePullPolicy: IfNotPresent
      securityContext:
        privileged: true
      command: ["/bin/sleep"]
      args: ["3600"]
      volumeDevices:
      - devicePath: /dev/longhorn-block-pvc
[...]

PR #12765, opened 2026-03-09
docs(lep): luks2 header space preservation for encrypted volumes

mantissahz:

Issue # longhorn/longhorn#9205


enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,191 @@
...
+
+```go
+type InstanceSpec struct {
+    ...
+    // +optional
+    VolumeSize          int64 `json:"volumeSize,string"`
+    // +optional
+    // VolumeEncrypted     bool `json:"volumeEncrypted"`
+    // +optional
+    VolumeEncryptedSize int64 `json:"volumeEncrypted,string"`,
```
derekbit: As we discussed in the LEP, do we really need to field? If we have `VolumeEncrypted` field, we can directly get the encrypted volume size from `VolumeSize`.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,191 @@
...
+1. The volume expansion request is received, and it updates `Volume.Spec.Size`.
+2. The volume controller calculates the new `VolumeEncryptedSize` = `NewSize + 16MiB`.
+3. It updates `Engine.Spec.VolumeEncryptedSize` and `Replica.Spec.VolumeSize` with the new correct expansion size.
+
+## Upgrade Plan
+
+**Existing Encrypted Volumes:**
+
+Update the fields `Spec.VolumeSize` `Spec.VolumeEncryptedSize` of the Engine and Replica of existing encrypted volumes if necessary.  
+When upgrading the engine, the volume controller will create new replicas with the correct replica size.
```
derekbit: The following questions are not addressed
- Do we need to handle the existing encrypted volumes?
- Can the existing encrypted volumes support the LUKS2 header space preservation?
  - If yes, how to handle it
  - If no, how does Longhorn distinguish existing volumes and new volumes?

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,191 @@
...
+
+### Goals
+
+- Abstract the 16 MiB LUKS2 header overhead from the user.
+- Correctly propagate the calculated backend size (Requested Size + 16 MiB) through all core data paths: Creation, Rebuilding, Restoring, and Expanding.
+- Provide backward compatibility for existing encrypted volumes.
+
+### Non-goals
+
+N/A
```
derekbit: Please check the questions in the upgrade section.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,191 @@
+# LUKS2 Header Space Reservation for Encrypted Volumes
+
+## Summary
+
+This proposal addresses the issue where the actual block device size presented to the user is 16 MiB smaller than the requested volume size when the volume is created with encryption enabled. This 16 MiB is currently consumed by the LUKS2 encryption header. This enhancement will transparently handle this overhead at the backend level, ensuring the user gets the exact usable capacity they requested, and properly managing this offset during volume rebuilding, restoring, and expansion.
```
derekbit: Please provide the reference link for the 16-MiB header size. Is it possible the size differ?

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Motivation
+
+When users create a volume of a specific size (e.g., `10 GiB`) with encryption enabled, Longhorn currently carves out 16 MiB of that space for the LUKS2 metadata header. As a result, the `lsblk` output and the usable filesystem size inside the pod appear as `10 GiB - 16MiB`. This discrepancy confuses users and can trigger strict size-check validations in certain workloads or database applications.
+
+### Goals
+
+- Abstract the 16 MiB LUKS2 header overhead from the user.
+- Correctly propagate the calculated backend size (Requested Size + 16 MiB) through all core data paths: Creation, Rebuilding, Restoring, and Expanding.
+- Provide backward compatibility for existing encrypted volumes.
```
derekbit: The LEP only address the v1 data engine. Please also address the v2 data engine as well.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,191 @@
...
+N/A
+
+## Design
+
+### Implementation Overview
+
+Introduce helper utilities to centralize the calculation of the backend size.
+
+```go
+const LUKS2HeaderSize = 16 * 1024 * 1024 // 16 MiB
```
derekbit: Should move the constant and size calculation helper function to go-common-libs as @shuo-wu mentioned.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,191 @@
...
+
+```go
+func (vc *VolumeController) createEngine(v *longhorn.Volume, ...) *longhorn.Engine {
+    backendSize := utils.GetBackendSize(v.Spec.Size, v.Spec.Encrypted)
+    
+    engine := &longhorn.Engine{
+        Spec: longhorn.EngineSpec{
+            VolumeSize:          v.Spec.Size,       // The frontend size for the user
+            // VolumeEncrypted:     v.Spec.Encrypted,
+            VolumeEncryptedSize: backendSize,
```
derekbit: I prefer `VolumeEncrypted` instead.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,178 @@
...
+- Allow Longhorn to automatically rebuild the degraded replica.
+- **Expected:** Rebuild completes successfully. The new replica file size matches the existing replicas' `2 GiB + 16 MiB` size, and data integrity, for example the `md5sum` of a stored file, remains intact.
+
+### Backup & Restore
+
+- Write a 500 MiB payload to the `2 GiB` encrypted volume and calculate payload checksum.
+- Create a backup to the remote backup server.
+- Restore the backup to a new volume.
+- Attach the restored volume to a workload.
+- **Expected:** The restored volume presents exactly `2.0G` to the workload. The restored payload checksum matches.
```
derekbit: cc @WebberHuang1118 See if there is any suggestion for the test plan

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,178 @@
...
+3. It updates `Engine.Spec.VolumeSize` and `Replica.Spec.VolumeSize` to the correct expanded size.
+
+## Upgrade Plan
+
+**Existing Encrypted Volumes:**
+
+Update the `Spec.VolumeSize` and `Spec.VolumeEncryptedSize` fields of the engine and replica for existing encrypted volumes if necessary.  
+When upgrading the engine, the volume controller creates new replicas with the correct replica size.
+
+## Test Plans
```
derekbit, line 406: Please add test plans for existing encrypted volumes

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,178 @@
+# LUKS2 Header Space Reservation for Encrypted Volumes
+
+## Summary
+
+This enhancement addresses an issue in which the actual block device size presented to the user is 16 MiB smaller than the requested volume size when encryption is enabled. The missing 16 MiB is currently consumed by the LUKS2 encryption header. This proposal handles that overhead transparently at the backend layer, ensuring that users receive the exact usable capacity they requested while correctly managing the offset during volume rebuild, restore, and expansion operations.
```
derekbit: Is the 16 MiB a fixed value for different cryptsetup/LUKS2 configuration? We can make the design and implementation more flexible and keep the value configurable.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,178 @@
...
+
+### User Experience In Detail
+
+From the end user's perspective, this functionality is fully transparent and requires no configuration changes.
+
+1. A user creates a standard PVC with a storage class name that points to an encrypted Longhorn `StorageClass` and specifies `size: 1GiB`.
+2. Longhorn automatically creates a volume with `Size: 1073741824`, while internally adjusting the backend storage allocation to `1 GiB + 16 MiB`.
+3. When the workload mounts the volume and the user runs `lsblk` or `df -h`, the reported partition size exactly matches `1G`, rather than the previous `1008M` equivalent.
+4. Restoring a backup, taking a snapshot, and expanding a volume all respect this calculation implicitly. If the user expands the volume to `2 GiB`, the Longhorn UI and the Kubernetes PVC reflect `2 GiB`, while Longhorn automatically expands the encrypted backend replica to `2 GiB + 16 MiB`. Existing legacy volumes continue to behave as they did previously without disruption.
```
derekbit: - Do we store the `Encrypted` in the backup.cfg or volume.cfg?
  - If no, we should persist it.
- How do we handle the existing backups for the existing encrypted volumes?
- How do we distinguish between the the backups for existing encrypted volumes and newly created encrypted volumes while restoring them?

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,178 @@
...
+
+1. The volume expansion request is received and updates `Volume.Spec.Size`.
+2. The volume controller calculates the new `NewBackendSize = NewVolumeSize + 16 MiB`.
+3. It updates `Engine.Spec.VolumeSize` and `Replica.Spec.VolumeSize` to the correct expanded size.
+
+## Upgrade Plan
+
+**Existing Encrypted Volumes:**
+
+Update the `Spec.VolumeSize` and `Spec.VolumeEncryptedSize` fields of the engine and replica for existing encrypted volumes if necessary.
```
derekbit: The sentence is not clear. Should we handle the `existing encrypted volumes`? What does `if necessary` mean here?

derekbit (commented):
The LEP does not clearly define how to distinguish between existing encrypted volumes and newly created encrypted volumes. These two cases must be handled carefully to avoid compatibility issues. Please enhance this section and update the document accordingly.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Removed two fields. Add the extra size to `VolumeSize` directly.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: > Do we need to handle the existing encrypted volumes?

No, not handle them, and use `EI.Status.CLIAPIVersion` to distinguish existing volumes and new volumes

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Updated. Thanks.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Added. https://gitlab.com/cryptsetup/cryptsetup/-/wikis/FrequentlyAskedQuestions#10-luks2-questions
And `yes`, it is configurable.
Thanks.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Still checking the flow and testing.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Moved it and updated the LEP. Thanks.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Removed two fields. Add the extra space directly to `VolumeSize`. Thanks.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,191 @@
...
+
+### User Experience In Detail
+
+From the end user's perspective, this functionality is completely transparent and requires zero configuration changes.
+
+1. A user creates a standard PVC with a storage class name pointing to an encrypted Longhorn `StorageClass` and `size: 1GiB`.
+2. Longhorn automatically creates a Volume with `Size: 1073741824`, but internally adjusts the backend storage allocation (`VolumeEncryptedSize` or similar parameter) to `1 GiB + 16 MiB`.
+3. When the workload mounts the volume and the user executes `lsblk` or `df -h`, they see the partition size exactly matches `1G`, as opposed to the previous `1008M` equivalent.
+4. Restoring backups, taking snapshots, and expanding volumes respect this calculation implicitly. If a user expands the volume to `2 GiB`, the Longhorn UI and Kubernetes PVC reflect `2 GiB`, while Longhorn automatically handles expanding the encrypted backend replica to `2 GiB + 16 MiB`. Existing legacy volumes will continue to behave as they previously did without disruption.
```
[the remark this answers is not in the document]
mantissahz: Updated. Thanks.

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Added. Thanks.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
derekbit: How can `EI.Status.CLIAPIVersion` be used for distinguishing existing volumes and new volumes?
If the volume is detached and attached, the volume will use new engine image. Isn't it?

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: It's configurable.
But if we make it configurable, it might be more complex than the current LEP.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: > Do we store the Encrypted in the backup.cfg or volume.cfg?
> - If no, we should persist it.

1. No. And `yes`, we should persist in it.
2. We will treat existing and newly created backups as the same.
3. During restoring a backup to a volume, we will add the extra 16 MiB space to the engine and replicas only when `Volume.Spec.Encrypted` is enabled.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Updated this section.
We will not handle the existing encrypted volumes until their volume engines are upgraded.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,197 @@
...
+- On the node to which the volume is attached, verify that the backend replica file size is exactly `1 GiB + 16 MiB`.
+
+### Volume Expansion
+
+1. Create a new `1 GiB` encrypted volume.
+   - Expand the `1 GiB` encrypted volume from Test 1 to `2 GiB`.
+   - Wait for expansion to complete.
+   - **Expected:** `lsblk` or `fdisk -l` inside the pod shows `2G`. The replica image file on the worker node shows `2 GiB + 16 MiB`.
+2. Use an existing `1 GiB` encrypted volume that uses the old engine image.
+   - Expand the `1 GiB` encrypted volume from Test 1 to `2 GiB`.
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,197 @@
...
+    // CLIAPIVersion used to communicate with user e.g. longhorn-manager
+    CLIAPIVersion    = 12
+    CLIAPIMinVersion = 8
+    ...
+)
+```
+
+1. Bump the `CLIAPIVersion` to distinguish existing encrypted volumes from newly created encrypted volumes.
+2. When creating/rebuilding/expanding an encrypted volume with `CLIAPIVersion >= 12`, the extra 16 MiB space will be added to the engine and replicas.
+3. Otherwise, for `CLIAPIVersion < 12`, the operations will follow the current implementation and will not add the extra 16 MiB.
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,197 @@
...
+When users create a volume of a specific size, for example `10 GiB`, with encryption enabled, Longhorn currently reserves 16 MiB of that space for the LUKS2 metadata header. As a result, the `lsblk` output and the usable filesystem size inside the pod appear as `10 GiB - 16 MiB`. This discrepancy is confusing and can trigger strict size validation checks in some workloads or database applications.
+
+### Goals
+
+- Abstract the 16 MiB LUKS2 header overhead from the user.
+- Correctly propagate the calculated backend size (`Requested Size + 16 MiB`) through all core data paths: creation, rebuilding, restoring, and expansion.
+
+### Non-goals
+
+- Handle existing encrypted volumes during and after upgrading Longhorn.
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,197 @@
+# LUKS2 Header Space Reservation for Encrypted Volumes
+
+## Summary
+
+This enhancement addresses an issue in which the actual block device size presented to the user is 16 MiB smaller than the requested volume size when encryption is enabled. The missing 16 MiB is currently consumed by the [LUKS2 encryption header](https://gitlab.com/cryptsetup/cryptsetup/-/wikis/FrequentlyAskedQuestions#10-luks2-questions). Although the LUKS2 header size is configurable, it is set to a fixed value in Longhorn. This proposal handles that overhead transparently at the backend layer, ensuring that users receive the exact usable capacity they requested while correctly managing the offset during volume rebuild, restore, and expansion operations.
```
derekbit:
```suggestion
This enhancement addresses an issue in which the actual block device size presented to the user is 16 MiB smaller than the requested volume size when encryption is enabled. The missing 16 MiB is currently consumed by the [LUKS2 encryption header](https://gitlab.com/cryptsetup/cryptsetup/-/wikis/FrequentlyAskedQuestions#10-luks2-questions). Although the LUKS2 header size is configurable, it is set to a fixed value in Longhorn. This proposal handles that overhead transparently at the backend layer, ensuring that users receive the exact usable capacity they requested while correctly managing the offset during volume rebuild, backup, restore, clone and expansion operations.
```

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,197 @@
...
+
+Before this enhancement, if a user requests a `10 GiB` PersistentVolumeClaim (PVC) backed by an encrypted Longhorn volume, the actual block device presented to the pod in Kubernetes is `10 GiB - 16 MiB`. This occurs because Longhorn uses 16 MiB internally to store the LUKS2 metadata header. The discrepancy can cause application deployment workflows, such as those used by strict database systems, to fail if they perform hard validation on block device sizes.
+
+After this enhancement, when a user requests a `10 GiB` encrypted volume, they receive exactly `10 GiB` of usable decrypted storage space in the pod. The 16 MiB overhead is automatically added as hidden backend replica capacity and is managed entirely by Longhorn.
+
+### User Experience In Detail
+
+From the end user's perspective, this functionality is fully transparent and requires no configuration changes.
+
+1. A user creates a standard PVC with a storage class name that points to an encrypted Longhorn `StorageClass` and specifies `size: 1GiB`.
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,197 @@
...
+- https://github.com/longhorn/longhorn/issues/9205
+
+## Motivation
+
+When users create a volume of a specific size, for example `10 GiB`, with encryption enabled, Longhorn currently reserves 16 MiB of that space for the LUKS2 metadata header. As a result, the `lsblk` output and the usable filesystem size inside the pod appear as `10 GiB - 16 MiB`. This discrepancy is confusing and can trigger strict size validation checks in some workloads or database applications.
+
+### Goals
+
+- Abstract the 16 MiB LUKS2 header overhead from the user.
+- Correctly propagate the calculated backend size (`Requested Size + 16 MiB`) through all core data paths: creation, rebuilding, restoring, and expansion.
```
derekbit: nit: mentioned before.

```suggestion
- Correctly propagate the calculated backend size (`Requested Size + 16 MiB`) through all operations
```

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Non-goals
+
+- Provide automatic, transparent migration for all existing encrypted volumes during a Longhorn upgrade.
```
derekbit:
```suggestion
- Provide automatic, transparent migration for all existing encrypted volumes without the extra overhead during a Longhorn upgrade.
```

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Proposal
+
+Add 16 MiB to `Engine.Spec.VolumeSize` and `Replica.Spec.VolumeSize` when `Volume.Spec.Encrypted` is `true`.
```
derekbit: If the Volume.Spec.Size is 10G, will the `Engine.Spec.VolumeSize` and `Replica.Spec.VolumeSize` becomes 10G+16M? Is it possible to keep `Engine.Spec.VolumeSize` and `Replica.Spec.VolumeSize`  as 10G, but the real file size is 16G+16M to avoid the confusion and mismatch?

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
derekbit: Is it possible that the overhead is bigger than 16MB?

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: >If the volume is detached and attached, the volume will use new engine image

If the volume starts using the new engine image, it will be like an engine upgrade. Therefore, it should use a new engine and new replicas, and the new size with the extra space will be assigned to them.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,197 @@
...
+```go
+func getBackendSize(requestedVolumeSize int64, encrypted bool) int64 {
+    // The default LUKS2 header size is 16 MiB, so it must be added to the replica size when the volume is encrypted. Otherwise, the device
+    // presented to the user will be 16 MiB smaller than the requested size.
+    // https://gitlab.com/cryptsetup/cryptsetup/-/wikis/FrequentlyAskedQuestions
+    if encrypted {
+        return requestedVolumeSize + LUKS2HeaderSize
+    }
+    return requestedVolumeSize
+}
```
derekbit: How about moving the helper function to go-common-libs?

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
derekbit: When upgrading replicas and its engine of an existing volume, how to assign new size with the extra space to the existing volume? After an upgrade, a 10G disk file is still 10G.  Do you mean creating a new disk file and copying data from the old disk file to the new one? Can you elaborate on it more?

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,197 @@
...
+Introduce helper utilities to centralize backend size calculation.
+
+#### `go-common-libs/types/crypto.go`
+
+```go
+const (
+    CryptoKeyProvider = "CRYPTO_KEY_PROVIDER"
+    CryptoKeyValue    = "CRYPTO_KEY_VALUE"
+    ...
+    LUKS2HeaderSize = 16 * 1024 * 1024 // 16 MiB
```
derekbit, line 107: What's the max value of the header?

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,197 @@
...
+    // CLIAPIVersion used to communicate with user e.g. longhorn-manager
+    CLIAPIVersion    = 12
+    CLIAPIMinVersion = 8
+    ...
+)
+```
+
+1. Increment the `CLIAPIVersion` by 1 to introduce a new API level that distinguishes existing encrypted volumes from newly created encrypted volumes.
+2. When creating/rebuilding/expanding an encrypted volume with `CLIAPIVersion >= 12`, add the extra 16 MiB space to the engine and replicas.
+3. Otherwise, for `CLIAPIVersion < 12`, follow the current implementation and do not add the extra 16 MiB.
```
derekbit: This is not clear. The steps doesn't address the existing volumes.
During and after upgrading the engine image of the existing volumes, what will happen?

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,197 @@
...
+2. Use an existing `1 GiB` encrypted volume that uses the old engine image.
+   - Delete a replica of the attached `1 GiB` encrypted volume.
+   - Allow Longhorn to automatically rebuild the degraded replica.
+   - **Expected:** Rebuild completes successfully. The new replica file size matches the existing replicas' `1 GiB` size, and data integrity, for example the `md5sum` of a stored file, remains intact.
+
+### Backup Restore
+
+- Create a `1 GiB` encrypted volume.
+- Write a 100 MiB payload to the `1 GiB` encrypted volume and calculate payload checksum.
+- Create a backup to the remote backup server.
```
derekbit, line 441: Add a step to check volume.cfg or(?) backup.cfg's `Encrypted` field.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Affected Operations
+
+#### Rebuilding
+
+During rebuilding, the volume controller instructs the new replica to provision the correct size. Data synchronization reconstructs the target blocks, and because the underlying block device file accommodates `VolumeSize + 16 MiB`, the snapshot tree and raw sync can safely contain both the LUKS metadata and the encrypted payload.
+
+#### Backup and Restore
+
+Send the `LonghornBackupParameterEncrypted` parameter to `backupstore` so that the volume encryption state is preserved when a backup is created.
```
derekbit: Need to elaborate on how to back up existing encrypted volumes and new encrypted volumes.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,197 @@
...
+
+When a user restores from a backup, the backup target reads the metadata and the controller provisions a new volume. The volume controller determines whether the backup originated from an encrypted volume, calculates the user-requested `Volume.Spec.Size`, and creates the engine and replicas with the correct `BackendSize`.
```
derekbit: Need to mention how restore a volume without/with 16 MB. How do you find the correct backend size?

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,197 @@
...
+
+#### Expanding
+
+When a user expands a volume:
+
+1. The volume expansion request is received and updates `Volume.Spec.Size`.
+2. The volume controller calculates the new `NewBackendSize = NewVolumeSize + 16 MiB`.
+3. It updates `Engine.Spec.VolumeSize` and `Replica.Spec.VolumeSize` to the correct expanded size.
```
derekbit: Describe how to handle existing and new encrypted volumes' expansion.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,197 @@
...
+        }
+    }
+}
+```
+
+### Affected Operations
+
+#### Rebuilding
+
+During rebuilding, the volume controller instructs the new replica to provision the correct size. Data synchronization reconstructs the target blocks, and because the underlying block device file accommodates `VolumeSize + 16 MiB`, the snapshot tree and raw sync can safely contain both the LUKS metadata and the encrypted payload.
```
derekbit: Discuss the existing and new volumes' rebuilding.

derekbit (commented):
The design for newly created volume might be straightforward. However, the LEP should discuss more about the compatibility and upgrade.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
derekbit: Can we make VolumeSize same as volume.spec.size but only handle it in the data path, because we already know the volume is encryted?

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Modified. Thanks.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Modified. Thanks

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Modified to keep the values of fields as `Volume.Spec.VolumeSize` and use the corrected backend size when starting instance processes. Thanks

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Yes, the maximum should be 136 MiB.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Moved. Thanks.

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: 136 MiB when I tested the command `cryptsetup --type luks2 luksFormat --luks2-metadata-size 4096k --luks2-keyslots-size 128M`

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Added descriptions during and after upgrading the engine image of the existing volumes. Thanks.

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Removed adding the `Encrypted` field to the volume.cfg file, so we won't check it.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
derekbit: Then, isn't the constant value in go-common-libs suitable? The design needs to consider the different values of the overhead.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Refined this section, and removed `Send the "LonghornBackupParameterEncrypted" parameter to "backupstore"`

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Updated the descriptions. Thanks.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Added the descriptions for expanding the existing encrypted volumes. Thanks.

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,257 @@
+# LUKS2 Header Space Reservation for Encrypted Volumes
+
+## Summary
+
+This enhancement addresses an issue in which the actual block device size presented to the user is 16 MiB smaller than the requested volume size when encryption is enabled. The missing 16 MiB is currently consumed by the [LUKS2 encryption header](https://gitlab.com/cryptsetup/cryptsetup/-/wikis/FrequentlyAskedQuestions#10-luks2-questions). Although the LUKS2 header size is configurable, it is set to a fixed value in Longhorn. This proposal handles that overhead transparently at the backend layer, ensuring that users receive the exact usable capacity they requested while correctly managing the offset during volume rebuild, backup, restore, clone and expansion operations.
```
derekbit, line 5: The overhead is adjustable and not a constant for different encryption configuration based on your input. When creating an encrypted volume, can we know what the size of the overhead before executing cryptsetup?

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Added the description for existing encrypted volumes. Thanks.

derekbit:
@mantissahz As we discussed today, since the overhead is not a fixed for different configuration, can you elaborate on the 16MB overhead and add non-16MB value in the no-goals and emphasize we only support 16MB size?

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Yes, it's updated as you mentioned.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: >how to assign new size with the extra space to the existing volume

There will be an implicit expansion before starting the engine upgrade.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Updated. Thanks.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: As we discussed yesterday, we should have it fixed in 16 MiB.

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Added a section to introduce the LUKS header, and we will use the constant value 16 MiB first. But yes, we can make the header size fixed with a specific parameter.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
WebberHuang1118: So if we stick to a 16MB header, is it possible that the 16MB could be exhausted? For example, if the volume grows to a massive size?

mantissahz: >is it possible that the 16MB could be exhausted?

I don't quite understand what you mean.
Do you mean will the header be affected by writing data to a volume or expanding a volume?
I think the answer is `NO`.
The header stores the "keys" and some configurations (metadata), and they will be immutable unless using the cryptsetup commands.
Only adding too many passwords (only 32 slots) or re-encrypting the volume will affect the header (it needs extra metadata space, but it seems it will use the alignment padding to store temporary metadata).

WebberHuang1118: > Only adding too many passwords (only 32 slots) 
Is there any safeguard to prevent this from happening? 

Also, under normal circumstances, would only changing the cryptsetup/LUKS2 configuration affect the header size?

mantissahz: > Is there any safeguard to prevent this from happening?

We don't provide a method to add passwords now(`cryptsetup luksAddKey`).

> Also, under normal circumstances, would only changing the cryptsetup/LUKS2 configuration affect the header size?

Only the parameters `--luks2-metadata-size`, `--luks2-keyslots-size`, and `--offset` will affect the header size when formatting the device.
And we don't provide a way to tune the parameters.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
WebberHuang1118: What if an encrypted volume that was previously sent to a remote target is 16MB shorter? What will happen when restoring it to the new engine version?

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,302 @@
...
+2. When creating/rebuilding/expanding an encrypted volume with `CLIAPIVersion >= 12`, add the extra 16 MiB space to the engine and replicas.
+
+Existing encrypted volumes:
+
+1. For `CLIAPIVersion < 12`, follow the current implementation and do not add the extra 16 MiB until upgrading the engine.
+2. During engine upgrade,
+   1. Check the `cryptsetup` version on the host before expanding the encrypted volume.
+   2. An implicit volume expansion will be triggered to make the backend size correct for encrypted volumes.
+   3. Any error that occurs during the expansion process should return an error or create an event to record the error, and should not affect other volume operations.
+3. After the upgrade, the replica image size on the host will be expanded to the volume size plus an extra 16 MiB, and the device size inside the workload will match the requested volume size.
```
WebberHuang1118, line 402: Harvester supports live migration for block-mode, migratable volumes, so for a given volume, there will be a time window where the old engine and new engine exist simultaneously.

IIUC, for encrypted volumes, the old engine is 16MB shorter, while the new engine is not. How will Longhorn handle this situation?

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: For example, you backed up a 1 GiB encrypted volume (maybe it wrote 200 MiB data).
1. The backup(snapshot) sent to the remote backup target will be 200 MiB in size.
2. When restoring this backup to a 1 GiB encrypted volume (`Volume.Spec.Encrypted` should be true), this restore process will attach the volume to a node, and it will provision a sparse file that is 1 GiB plus 16 MiB in size as a replica image file (volume-head image file).
3. And then restoring the backup data to this new volume.

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: >Harvester supports live migration for block-mode, migratable volumes, so for a given volume, there will be a time window where the old engine and new engine exist simultaneously.

Do you mean that when the Harvester is doing a live migration, it will also start the engine upgrade?
Before upgrading the engine for encrypted volumes with the old image, Longhorn will expand the volume by 16 MiB.
Then the engine upgrade process can start.

WebberHuang1118: >  it will also start the engine upgrade?

Harvester doesn't explicitly trigger an engine upgrade. Instead, Harvester leverages live migration so that the Longhorn engines eventually start using the new version.

For a running VM in Harvester, prior to the upgrade, it uses an older version of the engine. During the upgrade, the VM will live-migrate to another node, and Longhorn will start a new version of the engine on the target node. IIUC, the new engine and the old engine's replica set share the same on-disk files.

mantissahz: > IIUC, the new engine and the old engine's replica set share the same on-disk files.

Yes, that's why the expansion is needed.
But I believe it should not have two engines running at the same time.

WebberHuang1118: For a Longhorn volume, there will be two running engines at the same time: one with the old version and one with the new version. Once the CSI detachment is issued from the original node, the engine with the old version will be deleted along with its replica set. 

This is a common scenario during a Harvester upgrade.

mantissahz: When the new version engine is assigned to the `Engine.Spec.Image`, the engine controller will handle the expansion first if the volume is encrypted with the old engine image.
The Longhorn engine controller will start the upgrade process only after the expansion.

WebberHuang1118: I don't think there is an explicit `engine upgrade` for Harvester's use case. Instead, live migration starts the new engine on the target node while the old engine is still running on the source node.

WebberHuang1118 (commented):
Please help answer [this question](https://github.com/longhorn/longhorn/pull/12765#discussion_r2964327126), thanks.

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: I will review the migration flow and will have an update on the LEP for the migration.
And yes, it will be an issue that old and new engines have different backend sizes, but they point to the same replica directory.
There is always an expansion for the encrypted volume when moving the engine to a new version image.
And I believe it should be at the beginning of the migration with the old engine.

mantissahz: Hi @WebberHuang1118,
Added a section to the migration.
But I still have some questions about the Harvester scenarios for live migration.
We can discuss it tomorrow.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,338 @@
...
+    ...
+    log = log.WithField("migrationEngine", migrationEngine.Name)
+
+    // expand the backend size of the migration engine for an encrypted volume with the old engine image (CLIAPIVersion <= 11).
+    requiredBackendSize, err := c.getRequiredBackendSizeForEncryptedVolume(v, migrationEngine)
+    if requiredBackendSize != migrationEngine.Status.CurrentSize {
+        // expand backend size. engineClientProxy.VolumeExpandd()
+        // wait for backend size expanded
+        return nil
+    }
```
WebberHuang1118: A question:

As mentioned,
```
func getBackendSize(requestedVolumeSize int64, encrypted bool, cliAPIVersion int) int64 {
    // The default LUKS2 header size is 16 MiB, so it must be added to the replica size when the volume is encrypted. Otherwise, the device
    // presented to the user will be 16 MiB smaller than the requested size.
    // https://gitlab.com/cryptsetup/cryptsetup/-/wikis/FrequentlyAskedQuestions
    if encrypted && cliAPIVersion >= 12 {
        return requestedVolumeSize + LUKS2HeaderSize
    }
    return requestedVolumeSize
}
```

Is the header size added only when `CLIAPIVersion >= 12`? If so, does the expansion during this phase exclude the header size?

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,338 @@
...
+
+![encrypted volume engine upgrade process flow](./image/encrypted-volumes-engine-upgrade-flow.png)
+
+## Test Plans
```
WebberHuang1118, line 406: We may consider adding a Harvester upgrade to the test plan. For Harvester VM volumes during a system upgrade, the scenario is: live migration + engine upgrade.

cc @derekbit @Vicente-Cheng

WebberHuang1118 (commented):
A question and a suggestion. Thanks!

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
derekbit: @mantissahz @WebberHuang1118 
Do we have a conclusion for the live migration case?

WebberHuang1118: IMO, the test case for live migration + engine upgrade should be considered, as it is similar to the Harvester system upgrade scenario.

derekbit: @mantissahz Please help clarify and finish the following items

1. Should we make sure all the encrypted volumes are expanded before the live migration to make the flow simple?

2. Update the live migration flow and test plan accordingly.

mantissahz: > IMO, the test case for live migration + engine upgrade should be considered, as it is similar to the Harvester system upgrade scenario.

Added the test plan for the live migration.

> Should we make sure all the encrypted volumes are expanded before the live migration to make the flow simple?

Yes, we can do it, and it will be simpler.
Updated the migration flow.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: > Is the header size added only when CLIAPIVersion >= 12? If so, does the expansion during this phase exclude the header size?

Yes, the header size will not be added if the volume is an existing encrypted volume (using the old engine image, such as v1.11.x) for the expansion, rebuild, or other operations.

Only the live migration is an exception; we will expand the old encrypted volumes before the live migration.

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: Added the test case for live migration with Harvester.

enhancements/20260309-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,348 @@
...
+
+#### Live Migration
+
+When a user try to migrate an encrypted volume with an old engine image:
+
+1. The live migration request is received and update the volume attachment ticket.
+2. The volume attachment controller will start handling the migration and update the field `Volume.Spec.MigrationNodeID`.
+3. The volume controller start handling the migration process.
+4. Before the migration process starts, the volume controller ensure the backend size is expanded to the correct size for the LUK2 overhead.
+5. After the backend size is expanded, the migration process starts as the current implementation.
```
WebberHuang1118: So, does live migration require the engine to be upgraded to the new version first? Will the engine upgrade cause the pod to terminate?

For the Harvester use case, the VM and volume need to remain running throughout the live migration.

mantissahz: > So, does live migration require the engine to be upgraded to the new version first? Will the engine upgrade cause the pod to terminate?

Yes, it requires upgrading the engine to the new version, and the pod will not be terminated.

And I have a proposal for this part. Please check Slack.

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,364 @@
...
+        }
+        // Check if the engine image CLIAPIVersion is 11
+        // If yes, reject the request.
+        return nil
+    ...
+    }
+}
+}
+```
+
```
WebberHuang1118, line 384: Can we also mention the engine and instance-manager pod combination in this case? As I understand it, before migration, the old engine was running in the new instance-manager pod. Then, after the old engine was upgraded and the LUKS header was expanded, live migration could start, right?

WebberHuang1118 (commented):
Just one question. Please take a look. Overall, LGTM. Thanks.

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
mantissahz: > Can we also mention the engine and instance-manager pod combination in this case? 

Sure, I will elaborate on this part. 

> As I understand it, before migration, the old engine was running in the new instance-manager pod. 

No. The new replicas are in the new instance manager pod after the engine image upgrade. The engine is still in the old one because of the attachment. 

> Then, after the old engine was upgraded and the LUKS header was expanded, live migration could start, right?

Yes, it is, and as you mentioned, the VM needs to restart to resize the file system.

WebberHuang1118: > > As I understand it, before migration, the old engine was running in the new instance-manager pod.
> 
> No. The new replicas are in the new instance manager pod after the engine image upgrade. The engine is still in the old one because of the attachment.

I mean, at first, the old-version engine runs in the old-version instance-manager pod. Then, a live engine update upgrades the engine to the new version while it is still running in the old-version instance-manager pod. After this live engine upgrade, live migration can start.

> Yes, it is, and as you mentioned, the VM needs to restart to resize the file system.

Just to make sure we’re on the same page: throughout the process from engine upgrade to volume live migration, the VM can remain running, right? The underlying block device has already been expanded, while the filesystem size remains unchanged. That part is purely an in-guest situation, correct?

mantissahz: > I mean, at first, the old-version engine runs in the old-version instance-manager pod. Then, a live engine update upgrades the engine to the new version while it is still running in the old-version instance-manager pod. After this live engine upgrade, live migration can start.

Yes, it should work as you described.

> Just to make sure we’re on the same page: throughout the process from engine upgrade to volume live migration, the VM can remain running, right?

It should work as the current process when upgrading Harvester, except that you need to upgrade the volume engine. 

>The underlying block device has already been expanded, while the filesystem size remains unchanged. That part is purely an in-guest situation, correct?

Yes, the filesystem size remains unchanged if the VM is not restarted.

WebberHuang1118: > > Just to make sure we’re on the same page: throughout the process from engine upgrade to volume live migration, the VM can remain running, right?
> 
> It should work as the current process when upgrading Harvester, except that you need to upgrade the volume engine.

Need to double-check: what does `except that you need to upgrade the volume engine` mean?

Harvester sets `concurrentAutomaticEngineUpgradePerNodeLimit` to `3`, as defined in the [values.yaml](https://github.com/harvester/harvester/blob/c73fdf790c0f3d4f553b948cfa07504d6d387292/deploy/charts/harvester/values.yaml#L502). 

IIUC, based on the LEP, the live engine upgrade starts first. After the engine upgrade succeeds, the next step is live volume migration to other nodes.

mantissahz: > Need to double-check: what does except that you need to upgrade the volume engine mean?

Upgrading the engine is the only additional step needed before the migration for the current Harvester upgrade.

> IIUC, based on the LEP, the live engine upgrade starts first. After the engine upgrade succeeds, the next step is live volume migration to other nodes.

Yes, it's the correct procedure.

WebberHuang1118 (dismissed):
LGTM, and thanks for the detailed explanation.

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,443 @@
...
+            msg := fmt.Sprintf("cannot have second CSI ticket for migratable volume %v while it is in state %v", vol.Name, vol.Status.State)
+            return werror.NewInvalidError(msg, "spec.attachmentTickets")
+        }
+        // Check if the engine image CLIAPIVersion is 11
+        // If yes, reject the request.
+        return nil
+    ...
+    }
+}
+}
```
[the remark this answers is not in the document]
mantissahz: Removed. Thanks

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,443 @@
...
+
+- Install a k3s cluster with Longhorn v1.11.x and Kubevirt.
+- Create a VM with a `1 GiB` Longhorn encrypted volume.
+- The volume presents `1.0G - 16 MiB` inside the VM.
+- Write a 100 MiB payload to the encrypted volume and calculate payload checksum in the VM.
+- Upgrade Longhorn to v1.12.
+- Live migration before upgrading the volume engine.
+- **Expected:**: Get an error from the Longhorn validator.
+- After upgrade the volume, start live migration.
+- **Expected:**: The migration is completed. The migrated volume presents `1.0G` inside the VM. The payload checksum matches.
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,443 @@
...
+- The version 2.1.0 is the exact minimal version of `cryptsetup` where the default LUKS2 header size became 16 MiB. (See [Cryptsetup 2.1.0 Release Notes](https://www.kernel.org/pub/linux/utils/cryptsetup/v2.1/v2.1.0-ReleaseNotes))
+
+#### Parameters Affecting Header Size
+
+During the `cryptsetup luksFormat` process, the following parameters dictate the final header size:
+
+- `--type`  
+  The Longhorn default value is `luks2`. If using the type `luks`, it will result in a 2 MiB header.
+- `--luks2-metadata-size` and `--luks2-keyslots-size`  
+    These allow you to explicitly define how much space to reserve for the JSON text and the binary keyslot data, overriding the defaults. Longhorn uses the default values  when the format command is executed.
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,443 @@
...
+        ...
+    }
+    func IsValidForExpansion(engine *longhorn.Engine, cliAPIVersion, imAPIVersion int, encrypted bool) (bool, int64, error) {
+        if engine.Status.IsExpanding {
+        return false, engine.Status.CurrentSize, nil
+        }
+
+        engineVolumeSize, err := util.GetActualBackendSize(engine.Spec.VolumeSize, encrypted, cliAPIVersion)
+        if err != nil {
+        return false, engine.Status.CurrentSize, err
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,443 @@
...
+1. The volume expansion request is received and updates `Volume.Spec.Size`.
+2. The volume controller updates `Engine.Spec.VolumeSize` and `Replica.Spec.VolumeSize` as `Volume.Spec.Size`.
+3. The engine controller starts expanding process with `Engine.Spec.VolumeSize`.
+
+`engineapi/proxy_volume.go`
+
+```go
+func (p *Proxy) VolumeExpand(e *longhorn.Engine) (err error) {
+    v, err := p.ds.GetVolumeRO(e.Spec.VolumeName)
+    cliAPIVersion, err := ec.ds.GetDataEngineImageCLIAPIVersion(e.Spec.Image, e.Spec.DataEngine)
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

enhancements/20260413-luks2-header-space-preservation-for-encrypted-volumes.md
```diff
@@ -0,0 +1,443 @@
...
+    return p.grpcClient.VolumeExpand(string(e.Spec.DataEngine), e.Name, e.Spec.VolumeName, p.DirectToURL(e),
+        lhcrypto.getBackendSize(e.Spec.VolumeSize, v.Spec.Encrypted, cliAPIVersion, false))
+}
+```
+
+#### Live Migration
+
+When a user try to migrate an encrypted volume with an old engine image:
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

WebberHuang1118 (approved):

derekbit (approved):

Issue #4014, opened 2022-05-21
[BUG] Annotations not being applied to ingress


johnwc:
## Describe the bug

We set a annotation on the ingress, but it never was applied to the ingress.

## To Reproduce

1. Deploy chart using annotation on ingress.
2. View ingress that was created, no annotations were created as you added.

## Expected behavior

Ingress gets deployed with applied annotations.

## Environment

 - Longhorn version: 1.2.4
 - Installation method (e.g. Rancher Catalog App/Helm/Kubectl): Rancher Catalog App
 - Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: RKE2
   - Number of management node in the cluster: 3
   - Number of worker node in the cluster: 2

## Additional context

```YAML
ingress:
  annotations:
    cert-manager.io/cluster-issuer: letsencrypt
  enabled: true
  host: longhorn.ourdomain.tld
  ingressClassName: nginx-lb
  secrets: null
  tls: true
  tlsSecret: longhorn.local-tls
```

PhanLe1010:
Thanks @johnwc for reporting
I am able to reproduce this issue using Longhorn chart on the Rancher market place.
The problem doesn't happen when install using pure Helm. 
 Need to do more investigation

PhanLe1010:
I am scratching my head not understanding why setting the helm value [here](https://github.com/rancher/charts/blob/4b64d2984fb6f7e0959d7a3e84e3ee44665486b3/charts/longhorn/100.1.2%2Bup1.2.4/values.yaml#L202-L204)  doesn’t take effect.

I am asking for help from Rancher team

PhanLe1010:
Rancher UI ticket is filled at https://github.com/rancher/dashboard/issues/6038

riptidewave93:
For what it's worth I see the same issue with kustomize here, using:

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

namespace: longhorn-system

helmCharts:
- name: longhorn
  repo: https://charts.longhorn.io
  version: 1.6.1 # https://artifacthub.io/packages/helm/longhorn/longhorn
  releaseName: longhorn
  namespace: longhorn-system
  includeCRDs: true
  valuesInline:
    ingress:
      enabled: true
      annotations:
        cert-manager.io/cluster-issuer: XXXX
        external-dns.alpha.kubernetes.io/hostname: XXXX
      ingressClassName: nginx
      host: XXXXX
      tls: true
[...]

PR #12793, opened 2026-03-13
fix: add a non-null default value for ingress annotations

DodoLeDev:

Issue #4014

The default values for ingress annotations is `null`. And [Rancher does not like it](https://github.com/rancher/dashboard/issues/6038#issuecomment-3598681834).

I have added an empty dict instead, so we can finally deploy ingress annotations through the Rancher UI

_This PR should not be too long to review_ :stuck_out_tongue_winking_eye:


shuo-wu (approved):

DodoLeDev:
I have edited the title of this PR to fix Github lint errors

derekbit:
Thanks @DodoLeDev for the contribution!

Issue #12778, opened 2026-03-12
[TASK] Add distro information to upgrade responder requests


davidcheng0922:
## What's the task? Please describe

Provide distro information to the upgrade responder if the distro environment variable is set in the Longhorn manager.

## Additional context

- https://github.com/rancher/suse-storage-mgmt/issues/116

yangchiu:
Tested on master-head (longhorn/longhorn 6bb49f6) by following [the test steps](https://github.com/longhorn/longhorn/issues/12778#issuecomment-4044826959).

1. Set `Enable Upgrade Checker` to `false`:
```
$ kubectl get settings -n longhorn-system | grep upgrade

upgrade-checker    false    true    3m23s
```
2. Deploy `longhorn-manager`, make sure having the env `LONGHORN_DISTRO`:
```
$ kubectl get pods -n longhorn-system -l app=longhorn-manager
NAME                     READY   STATUS    RESTARTS   AGE
longhorn-manager-zjrz7   2/2     Running   0          4m16s

$ kubectl exec -it longhorn-manager-zjrz7 -n longhorn-system -- /bin/sh
Defaulted container "longhorn-manager" out of: longhorn-manager, pre-pull-share-manager-image
sh-4.4# echo $LONGHORN_DISTRO
oss
```
3. Install `upgrade-responder` (It is easier to trace the log by setting the replicas of deployment to `1`); ref: `longhorn/dev/upgrade-responder/install.sh`:
```
$ ./install.sh 
secret/influxdb-creds created
persistentvolumeclaim/influxdb created
deployment.apps/influxdb created
service/influxdb created
Deployment influxdb is running.
Cloning into 'upgrade-responder'...
remote: Enumerating objects: 1319, done.
remote: Counting objects: 100% (208/208), done.
remote: Compressing objects: 100% (82/82), done.
remote: Total 1319 (delta 139), reused 128 (delta 126), pack-reused 1111 (from 2)
Receiving objects: 100% (1319/1319), 55.06 MiB | 3.92 MiB/s, done.
Resolving deltas: 100% (713/713), done.
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /home/yang/Code/automation/terraform2/aws/ubuntu/k3s.yaml
WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /home/yang/Code/automation/terraform2/aws/ubuntu/k3s.yaml
Release "longhorn-upgrade-responder" does not exist. Installing it now.
[...]

PR #13043, opened 2026-05-06
fix: correct upgrade responder image tag

davidcheng0922:

https://github.com/longhorn/longhorn/issues/12778

Currently, it seems that we build the image with the tag `master-head` instead `longhorn-head`

### DIFF
--- a/dev/upgrade-responder/install.sh
+++ b/dev/upgrade-responder/install.sh
@@ -4,7 +4,7 @@ UPGRADE_RESPONDER_REPO="https://github.com/longhorn/upgrade-responder.git"
 UPGRADE_RESPONDER_REPO_BRANCH="master"
 UPGRADE_RESPONDER_VALUE_YAML="upgrade-responder-value.yaml"
 UPGRADE_RESPONDER_IMAGE_REPO="longhornio/upgrade-responder"
-UPGRADE_RESPONDER_IMAGE_TAG="longhorn-head"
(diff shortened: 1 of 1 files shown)


derekbit:
checked with @mantissahz. We cannot use master-head and should use longhorn-head here. Am I right? @mantissahz 
.

mantissahz:
>checked with @mantissahz. We cannot use master-head and should use longhorn-head here. Am I right? @mantissahz

Yes, almost all services use the `longhorn-head` image.

The `longhorn` branch contains some patches specific to the Longhorn project, which can handle legacy Longhorn manager versions

derekbit:
Then, we don't need the change.
@mantissahz when will the EIO build and deploy the new image?

mantissahz:
>Then, we don't need the change.
@mantissahz when will the EIO build and deploy the new image?

We should rebuild the image first, and then ask the EIO team to upgrade the server.
https://github.com/longhorn/upgrade-responder/pull/76

derekbit:
> > Then, we don't need the change.
> > @mantissahz when will the EIO build and deploy the new image?
> 
> We should rebuild the image first, and then ask the EIO team to upgrade the server. [longhorn/upgrade-responder#76](https://github.com/longhorn/upgrade-responder/pull/76)

Got it. Could you help it? BTW, can you update the flow in the wiki page?

Issue #13059, opened 2026-05-07
[IMPROVEMENT] Introduce longhorn-global-manager Deployment for cluster-wide controllers


hookak:
### Is your improvement request related to a feature? Please describe (👍 if you like this request)

Today every longhorn-manager pod runs a cluster-wide Pod informer (KubernetesPodController, KubernetesPVController). On large clusters this scales poorly:

- kube-apiserver watch fan-out grows as O(nodes) for the same Pod stream — each node opens its own cluster-wide Pod watch.
- Per-pod memory of longhorn-manager grows with cluster Pod count, because every replica caches every Pod in the cluster.

Operators on clusters with hundreds of nodes / tens of thousands of Pods report apiserver pressure and longhorn-manager memory growth that are both attributable to this fan-out, even though only two controllers actually need cluster-wide Pod visibility.

The other ~6 Pod-informer consumers in longhorn-manager already only need a namespace-scoped view (longhorn-system) and could drop the cluster-wide watch entirely if the two outliers were hosted elsewhere.

Full design: https://github.com/hookak/longhorn/blob/feat/global-manager-enhancement/enhancements/20260506-global-longhorn-manager.md

### Describe the solution you'd like

Introduce a new longhorn-global-manager Deployment (leader-elected singleton, replicas configurable for standby) that hosts the controllers requiring cluster-wide Pod visibility — initially KubernetesPodController and KubernetesPVController. The DaemonSet keeps all node-local controllers and switches its Pod informer to a namespace-scoped watch.

Key properties:
- Opt-in via Helm value (`globalManager.enabled`, default false) — existing clusters are unchanged on upgrade until the operator flips the toggle.
- Ownership election via the existing `isControllerResponsibleFor` / `Status.OwnerID` machinery; no new CRDs, no data-plane changes.
- Wave 1 scope is intentionally narrow (the two cluster-wide Pod informer consumers). Further controller migration is not committed by this change.

hookak:
Hi, could you take a look at this issue when you have a moment? I'd appreciate your opinion on a couple of architectural points before going further. @derekbit @innobead 

The enhancement proposes introducing a leader-elected Deployment alongside the existing DaemonSet, so that controllers needing a cluster-wide view no longer have to run on every node. As currently drafted, the new topology ships as opt-in, with both the existing single-workload model and the new split model supported in parallel.

Two things I'd love to hear your thoughts on:

[...]

PR #13060, opened 2026-05-07
docs(enhancement): add Global Longhorn Manager design

hookak:
Design for hosting Longhorn's cluster-wide Pod informer in a leader-elected longhorn-global-manager Deployment instead of every DaemonSet replica, addressing the kube-apiserver watch fan-out and per-pod memory growth that scale linearly with cluster size today.

This release migrates only the cluster-wide Pod informer consumers (KubernetesPVController and KubernetesPodController). Other controllers remain in the DaemonSet; whether any of them is moved in the future is a separate decision and not committed by this enhancement. The longhorn-global-manager Deployment is always deployed — there is no enable toggle; the DaemonSet simply stops hosting these two controllers and narrows its Pod informer to the longhorn-system namespace. There is no CRD, on-disk format, or REST API change (the only new objects are a Lease and the Deployment), so a downgrade is a chart/image rollback and the brief rollout overlap is convergence-safe.

Longhorn 13059

Issue https://github.com/longhorn/longhorn/issues/13059

Adds the enhancement proposal document only (no code in this PR). It introduces longhorn-global-manager and scopes this release to the two controllers that genuinely need cluster-wide Pod visibility.

This revision changes the design from the earlier default-off / opt-in toggle to a single always-on topology (no `globalManager.enabled`). Rationale: the split is rollback-safe via chart/image rollback with no CRD or data change; standby replicas keep warm informer caches for near-instant leader failover; and a permanent toggle would add a third behavior to maintain. Re-requesting review since this supersedes the earlier opt-in framing.

### DIFF
--- a/enhancements/20260506-global-longhorn-manager.md
+++ b/enhancements/20260506-global-longhorn-manager.md
@@ -1,0 +1,330 @@
+# Global Longhorn Manager
+
+## Summary
+
+`longhorn-manager` runs as a DaemonSet — one pod per node, each hosting
+every Longhorn control-plane reconciler alongside the node-local work
+(replica/engine/instance-manager lifecycle, disk monitoring, environment
+checks). As clusters have grown, two scaling issues stand out:
+
+1. **`kube-apiserver` watch fan-out from the cluster-wide Pod informer.**
+   Every Pod change in the cluster is re-delivered by `kube-apiserver` to
+   every manager pod. The watch traffic — and the work `kube-apiserver`
+   does to serve it — scales with both cluster Pod count and node count.
+
+2. **Per-pod memory from informer caches.** Each manager pod caches every
+   object its informers watch. The Pod cache size is determined by the
+   cluster's Pod count, not the node's, so adding nodes multiplies the
+   same large Pod cache N times across DaemonSet pods.
+
+Both costs grow with the cluster even though the controllers driving them
+are owner-elected — only one daemon pod does useful work at a time; the
+other N − 1 carry the watch and cache cost just to be ready for failover.
+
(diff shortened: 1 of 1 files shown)


derekbit:
cc @longhorn/dev for review

derekbit:
cc @shuo-wu @c3y1huang @innobead

derekbit:
cc @PhanLe1010

derekbit:
In general, the design LGTM.
@shuo-wu @c3y1huang @PhanLe1010 Could you review it when you are available? Thanks.

hookak:
Really appreciate you reviewing this. @derekbit 

At the moment, the split topology is opt-in, but that means the controllers need to support both execution models. While coexistence is technically possible, I’m concerned that carrying both paths will make the controller logic harder to follow and maintain over time.

Because of that, I think making the split topology the default would be a better direction.

shuo-wu (dismissed):
LGTM.

You are right. I just double checked. Only ` KubernetesPVController` and `KubernetesPodController ` truly require global resources (mainly for update `volume.Status.KubernetesStatus` IIRC). And other controllers do not need to use cluster-wide informers at all!

hookak:
Really appreciate your review, @shuo-wu!

I wanted to flag a change since your approval: longhorn-global-manager is now always deployed, with no `globalManager.enabled` opt-in toggle.

We've also validated the implementation end-to-end on an internal multi-node cluster, and it behaved as expected.

I'll update the related chart and code accordingly.

hookak:
The full set is ready for review and in sync with this document.
@shuo-wu @derekbit When you have a chance, a re-review would be much appreciated, happy to clarify any part of the design.

derekbit (approved):

Issue #2035, opened 2020-12-03
[BUG] Uninstall Job should remove all Longhorn resources (CRDs, PodSecurityPolicies, ClusterRoles, and ClusterRolebindings)


meldafrawi:
**Describe the bug**
see title

**To Reproduce**
1. install Longhorn using Helm Chart
2. uninstall Longhorn

**Expected behavior**
All Longhorn related resources should be deleted

**Log**
N/A

**Environment:**
 - Longhorn version: `master` & `1.0.2`
 - Kubernetes version: 1.18
 - Node config
   - OS type and version
   - CPU per node:
   - Memory per node:
   - Disk type
   - Network bandwidth and latency between the nodes:
 - Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal):

**Additional context**
- Longhorn `v1.0.2`: CRDs, PodSecurityPolicies, ClusterRoles, and ClusterRolebindings are not deleted
- Longhorn `master`: CRDs are not deleted

yasker:
I am not sure if CRDs should be cleaned up though. Rancher chart (2.5) seems deals with it differently. CC @shuo-wu

shuo-wu:
This CRD not-clean-up issue is caused by the annotation `helm.sh/resource-policy: keep` introduced in #1818.

After adding this annotation, the CRD will be retained not only during the upgrade but also during the deletion. And I found there is a similar case in the Helm github community: https://github.com/helm/helm/issues/7839#issuecomment-606044441

I will investigate if we can fix this issue via a hook. And maybe we need to put this issue in v1.1.0 first so that users won't be affected. 

---

As for Longhorn installation/uninstallation in Rancher Cluster Explorer, the CRD creation and deletion is handled by a separate sub-project hence we don't need to worry about it.

yasker:
If we cannot delete it automatically (since we don't really know who is installing/uninstalling), we can call it out in the uninstallation doc.

joshimoo:
related for installations outside of the longhorn-system namespace:
https://github.com/longhorn/longhorn/issues/2034

shuo-wu:
> related for installations outside of the longhorn-system namespace:
> #2034

As I mentioned in that issue, Helm 3 won't clean up the namespace during the uninstallation. I am not sure if the cleanup failure of the resource except for CRD is related to this namespace setting issue.

shuo-wu:
[...]

PR #7644, opened 2024-01-12
fix: add backupbackingimage to clusterrole

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7643

Add backupbackingimage to the uninstall clusterrole

PR #5135, opened 2022-12-23
fix(uninstall): missing resource in ClusterRole

c3y1huang:
- https://github.com/longhorn/longhorn/issues/5132
- https://github.com/longhorn/longhorn/issues/5133

PR #13068, opened 2026-05-10
fix: include enginefrontends in uninstall ClusterRole

immanuwell:
Issue #2035 (related)

`uninstall/uninstall.yaml` was one CRD behind `deploy/longhorn.yaml`.

`enginefrontends.longhorn.io` is shipped now, but the uninstall ClusterRole did not include `enginefrontends`. Tiny RBAC drift but yeah, it can bite during cleanup.

Repro:
```bash
python3 - <<'PY'
import yaml
from pathlib import Path

crds=[]
for doc in yaml.safe_load_all(Path("deploy/longhorn.yaml").read_text()):
    if doc and doc.get("kind") == "CustomResourceDefinition":
        name = doc.get("metadata", {}).get("name", "")
        if name.endswith(".longhorn.io"):
            crds.append(name.split(".")[0])

role=[]
for doc in yaml.safe_load_all(Path("uninstall/uninstall.yaml").read_text()):
    if doc and doc.get("kind") == "ClusterRole":
        for rule in doc.get("rules", []):
            if rule.get("apiGroups") == ["longhorn.io"]:
                role = rule.get("resources", [])

print("missing from uninstall role:", sorted(set(crds) - set(role)))
PY
```

Before this PR:
```text
missing from uninstall role: ['enginefrontends']
```

After this PR:
```text
missing from uninstall role: []
```

Generated with `python3 scripts/update-uninstall-manifest.py`.

No cloud/provider quota weirdness here, Kubernetes RBAC is API-server enforced, so any cluster with this CRD can hit it.

Similar prior cleanup: #5135 and #7644.

### DIFF
--- a/uninstall/uninstall.yaml
+++ b/uninstall/uninstall.yaml
@@ -74,6 +74,7 @@ rules:
   - backups
   - backuptargets
   - backupvolumes
+  - enginefrontends
   - engineimages
   - engines
   - instancemanagers


derekbit:
Good catch @immanuwell! 
Could you create a new BUG ticket in https://github.com/longhorn/longhorn/issues? Thank you.

derekbit (approved):

derekbit:
@mergify backport v1.12.x

PR #13090, opened 2026-05-14
chore: update CONTRIBUTING.md

derekbit:

Issue #

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -198,7 +198,7 @@ For documentation pull requests, please make sure:
 - The content is accurate and matches the current Longhorn behavior.
 - The change is linked to a related issue when it affects user-facing behavior, troubleshooting, installation, upgrade, settings, or feature documentation.
 - The wording is clear and concise.
-- Examples, commands, and YAML snippets are tested when possible.
+- Examples, commands, and YAML snippets are tested.
 
 ## Review Process
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,73 +1,239 @@
-# Contributing Guideline
+# Contributing to Longhorn
 
-Welcome contributing to Longhorn!
+Welcome, and thank you for your interest in contributing to Longhorn!
 
-This guideline applies to all the repositories under Longhorn.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,73 +1,239 @@
...
+- Installation method:
+
+Test steps:
+1.
+2.
+3.
+
+Result:
+- PASS / FAIL
+- Relevant logs, screenshots, or command output if applicable.
```
innobead, line 77: Is this consistent with our PR template across repos?

CONTRIBUTING.md
```diff
@@ -1,73 +1,239 @@
...
+- https://github.com/longhorn/longhorn/wiki/coding-convention
+
+In particular, Go imports must follow the project import grouping convention. Import groups should be organized consistently with the existing codebase and separated by blank lines.
+
+The expected grouping is generally:
+
+1. Go standard library packages.
+2. Third-party packages.
+3. Kubernetes-related packages.
+4. Longhorn component packages outside the current repository.
```
innobead, line 173: Is this coming from the output formatted by go fmt, or is it specifically designed?

CONTRIBUTING.md
```diff
@@ -1,73 +1,239 @@
...
+## Documentation Changes
+
+Documentation improvements are welcome.
+
+For documentation pull requests, please make sure:
+
+- The content is accurate and matches the current Longhorn behavior.
+- The change is linked to a related issue when it affects user-facing behavior, troubleshooting, installation, upgrade, settings, or feature documentation.
+- The wording is clear and concise.
+- Examples, commands, and YAML snippets are tested when possible.
```
innobead:
```suggestion
- Examples, commands, and YAML snippets are tested.
```

innobead (dismissed):
LGTM, a few questions.

CONTRIBUTING.md
derekbit: IIUC, it is just a Longhorn convention.

CONTRIBUTING.md
derekbit: After rethinking this section, contributor should update the pre-check list in the source issue ticket instead.

CONTRIBUTING.md
innobead: I see. It's okay, but I think it may be easier for users to use a standard way. This can also mean we don't need to customize anything specific. 

cc @c3y1huang @shuo-wu

innobead (approved):

PR #13100, opened 2026-05-14
feat(chart): add always-on longhorn-global-manager Deployment

hookak:
Add a `globalManager` block in values.yaml and a templated
longhorn-global-manager Deployment that owns the cluster-wide Pod informer.
The cluster-wide Pod controllers (KubernetesPVController,
KubernetesPodController) run in this leader-elected Deployment as a singleton
instead of in every longhorn-manager DaemonSet pod, collapsing the cluster-wide
Pod watch fan-out from O(node count) to one active leader.

The split is the only topology (single-mode): the Deployment is always
rendered and the DaemonSet no longer hosts those controllers. There is no
opt-in toggle; the longhorn-manager binary excludes the migrated controllers
from `daemon` unconditionally. Multiple replicas with hostname anti-affinity
provide hot standby; a livenessProbe hits /v1/healthz (port 9505), which
reports process liveness regardless of leader status so standby replicas stay
Ready.

See enhancements/20260506-global-longhorn-manager.md.

longhorn-13059


derekbit (approved):

derekbit:
@mergify backport v1.13.x

Issue #9104, opened 2024-07-29
[FEATURE] V2 Volume Supports Live Upgrade


derekbit:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

Support V2 volume to remain lively and uninterrupted during the Longhorn engine upgrade, starting from 1.12 to 1.13. 

## Additional context

https://github.com/longhorn/longhorn/issues/9456

derekbit:
Because of the introduction of UBLK, we will redesign the live upgrade.

derekbit:
Because of the introduction of UBLK, @PhanLe1010 and I will redesign the live upgrade.

cc @PhanLe1010

PhanLe1010:
Please see this report for testing data and pros and cons of different approaches: 
https://www.notion.so/V2-Live-Upgrade-2a3758c45123800ba179c8093686746e?source=copy_link

Thank you!

cc @derekbit @shuo-wu @innobead

shuo-wu:
After discussions with @derekbit and @PhanLe1010 based on the above test result, we decide to pick Derek’s design as our solution. 
The main concerns of not choosing the suspension solution are:
1. In the above test, there is no continuous IOs when suspending the frontend. If there is, "Volume Subspension Time" may be not a minor number.
2. The above test does not count the instance manager restart time, which may be 10~20s in general.
3. Even though the above 2 parts are not in count, the test already takes **over 2 minutes** to finish the upgrade when there are 20 volumes in an instance manager pod.
4. It seems that "Volume Scaling Up Time" will increase linearly when there are more volumes. In other words, the total suspension time is unpredictable in real world.

roger-ryao:
Hi @davidcheng0922 
     I have updated the test cases to support `v2 upgrade`. You can refer to the following branch https://github.com/roger-ryao/longhorn-tests/tree/issue9104.  
    Or use the following settings to run the tests on jenkins: 
     - `LONGHORN_TESTS_CUSTOM_IMAGE` : `ryaoio/longhorn-e2e-test:9104`
     - `CUSTOM_TEST_OPTIONS` : `-t \"Test Replicas Not Accumulate During Engine Upgrade\" -t \"Pre-release Checks\" -t \"Test System Upgrade with New Instance Manager\" -v LOOP_COUNT:1 -v RETRY_COUNT:600 -v DATA_ENGINE:v2 -v test_v2_only:true`

davidcheng0922:
[...]

PR #13214, opened 2026-05-27
chore(crd): v2 live upgrade

davidcheng0922:

https://github.com/longhorn/longhorn/issues/9104


enhancements/20260415-v2-live-upgrade.md
```diff
@@ -0,0 +1,575 @@
...
+
+Three new settings are introduced to control the v2 live upgrade behavior:
+
+#### (1) Take Snapshot Before V2 Data Engine Upgrade
+
+- **Name**: `take-snapshot-before-v2-data-engine-upgrade`
+- **Type**: Boolean
+- **Default**: `false`
+- **Category**: General
+- **Description**: If enabled, Longhorn takes a snapshot and waits for its checksum calculation to complete before relocating a v2 data engine to another node during the Instance Manager Upgrade. This helps minimize replica rebuild time if a failure occurs during the upgrade. This setting must be enabled to utilize the Fast Replica Rebuild feature (`fast-replica-rebuild-enabled`) during the upgrade process.
```
shuo-wu: - We cannot guarantee that all users will enable the snapshot integrity setting. 
- The snapshot integrity setting cannot guarantee that the snapshot checksum will be calculated immediately after the snapshot creation. IIRC, the trigger of the calculation takes time, and the calculation may be slow if the snapshot is huge
- Besides, IIRC, there are bugs for the v2 data integrity. Somehow it does not work, so the fast rebuild cannot be applied.

davidcheng0922: yeah, after re-thinking, I may remove this feature first, focus on basic ability. Maybe we can improve it in the future.

shuo-wu: Actually, I discussed this with @derekbit yesterday, and the conclusion is that rebuilding during a live upgrade is unavoidable as long as we don't want to block I/O for a long time. And the current fast rebuilding is not a good solution in this live upgrade case. 
Instead, we may need to rely on this new rebuilding mechanism https://github.com/longhorn/longhorn/pull/7031 in the future. The brief summary for the new feature is: Recording the modified sector locations within a few minutes after a replica failure. Once the failed replica is back within a certain time (let's say 5 min), we can directly copy the data of the recorded sector locations from any healthy replicas to the failed replica head without any snapshot creation and data verification.

derekbit (approved):

derekbit:
@mergify backport v1.13.x

Issue #11521, opened 2025-08-19
[IMPROVEMENT] Add usage metrics for volume size


innobead:
### Is your improvement request related to a feature? Please describe (👍 if you like this request)

To understand volume size usage (for reference in sharding development), add corresponding metrics to https://metrics.longhorn.io/.

cc @longhorn/maintainer

kunalworldwide:
For volume size usage metrics, the most operationally useful metrics would be:

- `longhorn_volume_actual_size_bytes` — actual disk space consumed (after compression/dedup if applicable)
- `longhorn_volume_capacity_bytes` — provisioned capacity
- `longhorn_volume_usage_ratio` — actual/capacity as a ratio for alerting thresholds

The distinction between provisioned capacity and actual usage matters a lot for thin-provisioned volumes where the underlying space consumption can differ significantly from what's declared. If Longhorn already tracks these values internally for the UI, exposing them as Prometheus metrics should be a matter of registering new collectors against the existing data.

Is there a design doc for the metrics.longhorn.io analytics platform, or is this specifically about Prometheus-style metrics on the Longhorn manager?

chriscchien:
Verified **pass** on longhorn master(https://github.com/longhorn/longhorn-manager/commit/8c4ca75648e2638247f0d1760a059871fcc24ec2)

1. Set `Enable Upgrade Checker` to `false`
2. Create volumes with size 2Gi, 20Gi, 60Gi, 100Gi, 200Gi
3. Install `upgrade-responder` (It is easier to trace the log by setting the replicas of deployment to `1`); ref: `longhorn/dev/upgrade-responder/install.sh`
4. Set Upgrade Responder URL to `http://longhorn-upgrade-responder.default.svc.cluster.local:8314/v1/checkupgrade` (Use the internal cluster service url)
5. Set `Enable Upgrade Checker` to `true`
6. Check the InfluxDB upgrade_request field keys from upgrade-responder.

```
$ kubectl -n default exec influxdb-586bd78744-8hmv7 -- \
  influx -database longhorn_upgrade_responder \
  -execute 'SELECT longhorn_volume_size2_to5_gi_b_count, longhorn_volume_size20_to50_gi_b_count, longhorn_volume_size50_to100_gi_b_count, longhorn_volume_size100_to200_gi_b_count, longhorn_volume_size200_to500_gi_b_count FROM upgrade_request ORDER BY time DESC LIMIT 1'
name: upgrade_request
[...]

PR #13370, opened 2026-06-22
chore(upgrade-responder): update chart-values.yaml and install.sh

derekbit:

Issue Longhorn/longhorn#11521


mantissahz (approved):
LGTM

Issue #12889, opened 2026-04-02
[IMPROVEMENT] Improve Longhorn Helm chart defaults and security/availability configurability


alita1991:
### Is your improvement request related to a feature? Please describe (👍 if you like this request)

We are reviewing the rendered Longhorn Helm manifests and found several areas where the chart could be improved to provide stronger default hardening, clearer workload behavior, and better configurability for production environments.

This is not about Longhorn’s inherent storage architecture constraints such as privileged host access or hostPath usage. It is about chart-level gaps that appear addressable in templates, defaults, or exposed Helm values.

**Areas for improvement**

  - Add resource requests and limits consistently across rendered workloads
  - Add ephemeral-storage requests and limits where applicable
  - Add PodDisruptionBudgets for rendered Deployments
  - Strengthen anti-affinity so workloads that should not co-schedule can enforce that behavior
  - Explicitly define deployment update strategy instead of relying on Kubernetes defaults
  - Add or expose support for:
      - allowPrivilegeEscalation: false where possible
      - automountServiceAccountToken: false
      - SELinux labeling
      - AppArmor
      - seccomp configuration
      - explicit seccompProfile: RuntimeDefault or Localhost
      - configurable pod nodeAffinity
  - Make replica counts adjustable for all rendered Deployments where appropriate
  - Improve default high-availability posture for the chart
  - Set terminationMessagePolicy: FallbackToLogsOnError where appropriate
  - Document or support graceful shutdown expectations for workloads
  - Expose explicit dual-stack service configuration where supported
  - Provide a clear story for image signing support

**Current observations from rendered manifests**

  - Some workloads do not have full resource requests/limits
  - ephemeral-storage is only configured for part of the rendered workload set
  - Rendered Deployments do not have PodDisruptionBudgets
  - Anti-affinity is currently soft rather than enforced
  - Some deployment behavior relies on defaults rather than explicit manifest fields
  - Multiple hardening-related fields are absent or not consistently configurable
  - Some placement and HA controls are only partially exposed
  - Dual-stack networking is not explicitly configurable from the chart

**Why this matters**

These improvements would make the chart easier to use in environments with stronger operational, security, and availability requirements, while reducing the amount of downstream patching needed by adopters.

**Review the chart and:**

  - add missing manifest hardening where feasible
  - expose missing values in values.yaml
  - improve default workload availability behavior
  - document any remaining unsupported areas clearly

**Acceptance criteria**

  - Missing workload hardening and availability settings are added where supported
[...]

PR #13378, opened 2026-06-22
feat(chart): add opt-in PodDisruptionBudget for Longhorn UI

somaz94:
Add an opt-in PodDisruptionBudget for the Longhorn UI Deployment, gated on
`longhornUI.podDisruptionBudget.enabled` (default false). The UI already defaults
to `replicas: 2` with pod anti-affinity, so this lets adopters keep a UI pod
available during voluntary disruptions such as node drains, with no change to
default behavior.

This is the Longhorn UI slice of #12889 (chart PDB gap). Other rendered
Deployments can follow in separate PRs.

Validation (helm lint + helm template, chart 1.12.0-dev):
- helm lint: 0 chart(s) failed
- default values: no PodDisruptionBudget rendered
- podDisruptionBudget.enabled=true: PDB renders with minAvailable: 1
- enabled=true + maxUnavailable set: PDB renders with maxUnavailable, minAvailable omitted (mutually exclusive)

related: #12889

related: #13466

### DIFF
--- a/chart/templates/poddisruptionbudget-ui.yaml
+++ b/chart/templates/poddisruptionbudget-ui.yaml
@@ -1,0 +1,21 @@
+{{- if and .Values.longhornUI.podDisruptionBudget.enabled (gt (int .Values.longhornUI.replicas) 1) }}
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  labels: {{- include "longhorn.labels" . | nindent 4 }}
+    app: longhorn-ui
+  name: longhorn-ui
+  namespace: {{ include "release_namespace" . }}
+spec:
+  {{- $pdb := .Values.longhornUI.podDisruptionBudget }}
+  {{- if ne (toString $pdb.maxUnavailable) "" }}
+  maxUnavailable: {{ $pdb.maxUnavailable }}
+  {{- else if ne (toString $pdb.minAvailable) "" }}
+  minAvailable: {{ $pdb.minAvailable }}
+  {{- else }}
+  minAvailable: 1
+  {{- end }}
+  selector:
+    matchLabels:
+      app: longhorn-ui
+{{- end }}
--- a/chart/README.md
+++ b/chart/README.md
@@ -234,6 +234,10 @@ Longhorn consists of user-deployed components (for example, Longhorn Manager, Lo
 |-----|------|---------|-------------|
 | longhornUI.affinity | object | `{"podAntiAffinity":{"preferredDuringSchedulingIgnoredDuringExecution":[{"podAffinityTerm":{"labelSelector":{"matchExpressions":[{"key":"app","operator":"In","values":["longhorn-ui"]}]},"topologyKey":"kubernetes.io/hostname"},"weight":1}]}}` | Affinity for Longhorn UI pods. Specify the affinity you want to use for Longhorn UI. |
 | longhornUI.nodeSelector | object | `{}` | Node selector for Longhorn UI. Specify the nodes allowed to run Longhorn UI. |
+| longhornUI.podDisruptionBudget | object | `{"enabled":false,"maxUnavailable":"","minAvailable":1}` | Pod Disruption Budget for Longhorn UI. Keeps a minimum number of UI pods available during voluntary disruptions such as node drains. Effective only when `longhornUI.replicas` is greater than 1. |
+| longhornUI.podDisruptionBudget.enabled | bool | `false` | Setting that allows you to enable the Pod Disruption Budget for Longhorn UI. |
+| longhornUI.podDisruptionBudget.maxUnavailable | string | `""` | Maximum number or percentage of Longhorn UI pods that can be unavailable during a disruption. When set, it takes precedence over `longhornUI.podDisruptionBudget.minAvailable`. |
+| longhornUI.podDisruptionBudget.minAvailable | string | `1` | Minimum number or percentage of Longhorn UI pods that must remain available during a disruption. Mutually exclusive with `longhornUI.podDisruptionBudget.maxUnavailable`. |
 | longhornUI.priorityClass | string | `"longhorn-critical"` | PriorityClass for Longhorn UI. |
 | longhornUI.replicas | int | `2` | Replica count for Longhorn UI. |
 | longhornUI.tolerations | list | `[]` | Toleration for Longhorn UI on nodes allowed to run Longhorn components. |
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -580,6 +580,14 @@ longhornUI:
   ## and uncomment this example block
   #  label-key1: "label-value1"
   #  label-key2: "label-value2"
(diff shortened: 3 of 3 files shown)


COLDTURNIP:
Please remember to update [the chart README](https://github.com/longhorn/longhorn/blob/master/chart/README.md#longhorn-ui-settings) for the new configurable value `podDisruptionBudget`

chart/templates/poddisruptionbudget-ui.yaml
```diff
@@ -0,0 +1,18 @@
...
+  labels: {{- include "longhorn.labels" . | nindent 4 }}
+    app: longhorn-ui
+  name: longhorn-ui
+  namespace: {{ include "release_namespace" . }}
+spec:
+  {{- if .Values.longhornUI.podDisruptionBudget.maxUnavailable }}
+  maxUnavailable: {{ .Values.longhornUI.podDisruptionBudget.maxUnavailable }}
+  {{- else }}
+  minAvailable: {{ .Values.longhornUI.podDisruptionBudget.minAvailable | default 1 }}
+  {{- end }}
```
COLDTURNIP: Please preserve valid zero values for the PDB fields. Helm treats numeric `0` as false/empty, so `if .Values...maxUnavailable` and `default 1` convert `0` into `minAvailable: 1`.

One option is to check for an empty string explicitly:

```yaml
     {{- $pdb := .Values.longhornUI.podDisruptionBudget }}
     {{- if ne (toString $pdb.maxUnavailable) "" }}
       maxUnavailable: {{ $pdb.maxUnavailable }}
     {{- else }}
       minAvailable: {{ $pdb.minAvailable }}
     {{- end }}
```

This keeps `maxUnavailable: 0` and `minAvailable: 0` renderable while preserving the current empty-string default behavior.

COLDTURNIP (changes requested):

somaz94:
Thanks, good catch. Pushed a fix using `ne (toString ...) ""` so `maxUnavailable: 0` and `minAvailable: 0` are now preserved.

I kept a three-branch form rather than dropping `default 1` entirely: if both `maxUnavailable` and `minAvailable` are empty, it still falls back to `minAvailable: 1` so the rendered PDB stays valid. Verified with `helm template` for `maxUnavailable=0`, `minAvailable=0`, both-empty (→ `minAvailable: 1`), and `50%`; `helm lint` passes.

chart/templates/poddisruptionbudget-ui.yaml
somaz94: Done in 65be701 — switched to the explicit `ne (toString $pdb.maxUnavailable) ""` guard you suggested. `maxUnavailable: 0` and `minAvailable: 0` now render as-is, while the empty-string default still falls back to `minAvailable: 1`. Thanks for catching this!

somaz94:
Rebased onto the latest `master` to clear the BEHIND state. The `0`-value fix @COLDTURNIP requested is in — the PDB template uses `ne (toString $pdb.maxUnavailable) ""`, so `maxUnavailable: 0` / `minAvailable: 0` render as-is while the empty-string default still falls back to `minAvailable: 1` (verified via `helm template`). PTAL / could you resolve the thread when you get a chance? Thanks!

derekbit:
Hello @somaz94 
Thanks for your contribution!
Since https://github.com/longhorn/longhorn/issues/12889 includes several areas, could you create an imporvement ticket in https://github.com/longhorn/longhorn/issues and link it to https://github.com/longhorn/longhorn/issues/12889? It will help manage and validate the improvement. Thanks.

somaz94:
Thanks @derekbit! I've opened a dedicated improvement ticket for this UI PDB slice and linked it to #12889:

- #13466 — [IMPROVEMENT] Add opt-in PodDisruptionBudget for the Longhorn UI Deployment

This PR (#13378) implements that ticket. Happy to adjust the scope or split further if that helps validation/backport. PTAL 🙏

derekbit:
@somaz94 Could you check if the comments from copilot need to be addressed? Thanks.

somaz94:
@derekbit Thanks for the ping — both Copilot comments are addressed in 23518fe:

1. **Single-replica PDB gating** — the template now renders only when `podDisruptionBudget.enabled` **and** `longhornUI.replicas > 1`, matching the documented "effective only when replicas > 1" behavior. This avoids creating an always-blocking PDB (`minAvailable: 1`) that could stall node drains on single-replica setups.
2. **`minAvailable` type** — documented as `string` (IntOrString) via a helm-docs `(string)` annotation, so percentage values like `"50%"` are correctly reflected (consistent with `maxUnavailable`).

Verified with `helm template`: replicas=2 renders the PDB, replicas=1 renders nothing, and `maxUnavailable: 0` is still preserved. `helm lint` is clean.

c3y1huang (approved):
LGTM.

Nit: I suggest cleaning up the commit history by folding fixups into the appropriate commits or squashing them before the merge.

somaz94:
@COLDTURNIP @derekbit — I believe everything raised is now addressed; summarizing so the stale review can be re-checked:

- **Preserve valid zero values** (@COLDTURNIP): fixed in `2c6085c`. The template uses `{{ if ne (toString $pdb.maxUnavailable) "" }} … {{ else if ne (toString $pdb.minAvailable) "" }} … {{ else }} minAvailable: 1 {{ end }}`, so `0` is preserved. Verified: `minAvailable: 0` renders `minAvailable: 0` and `maxUnavailable: 0` renders `maxUnavailable: 0` (neither collapses to the `minAvailable: 1` default).
- **Copilot — gate on replicas > 1**: done in `8499f5a`; the PDB template only renders when `(gt (int .Values.longhornUI.replicas) 1)`, matching the documented "effective only when replicas > 1" behavior.
- **Copilot — minAvailable type**: the README/values now document it as `string` (IntOrString), so percentages like `"50%"` are clearly supported.
- **Improvement ticket** (@derekbit): opened #13466 and linked it to #12889.

No functional changes outstanding on my side — @COLDTURNIP could you take another look when you have a moment? Thanks!

derekbit:
> LGTM.
> 
> Nit: I suggest cleaning up the commit history by folding fixups into the appropriate commits or squashing them before the merge.

Hello @somaz94 Could you address the comment? Thank you.

somaz94:
Thanks @c3y1huang @derekbit — done. The four commits are now folded into a single commit (`7466f58`), rebased onto the latest `master` (this also clears the out-of-date branch state). The net change is unchanged: 3 files, 33 insertions.

Re-verified after the squash with `helm lint` (0 failed) and `helm template`:
- default values → no PDB rendered
- `enabled=true` → `minAvailable: 1`
- `replicas=1` → no PDB rendered (the `replicas > 1` gate)
- `maxUnavailable` set → `minAvailable` omitted (mutually exclusive)
- `minAvailable=0` → preserved as `0`

@COLDTURNIP the points from your review are addressed in this commit — could you take another look when you get a chance? The `CHANGES_REQUESTED` from 06-23 predates the fixes and is still blocking the merge.

COLDTURNIP (approved):

Issue #13417, opened 2026-06-29
[BUG] Kernel Workqueue Lockup and Unstable RKE2 Service After Enabling LH V2 in Harvester


WebberHuang1118:
### Describe the Bug

Kernel Soft Lockup and Unstable RKE2 Service After Enabling LH V2

Related kernel call trace:
```
Jun 25 02:30:37 hp-114-tink-system kernel: BUG: workqueue lockup - pool cpus=2 node=0 flags=0x0 nice=0 stuck for 338s!
Jun 25 02:30:37 hp-114-tink-system kernel: Showing busy workqueues and worker pools:
Jun 25 02:30:37 hp-114-tink-system kernel: workqueue events: flags=0x0
Jun 25 02:30:37 hp-114-tink-system kernel:   pwq 2: cpus=0 node=0 flags=0x0 nice=0 active=1 refcnt=2
Jun 25 02:30:37 hp-114-tink-system kernel:     pending: drm_fb_helper_damage_work
Jun 25 02:30:37 hp-114-tink-system kernel:   pwq 10: cpus=2 node=0 flags=0x0 nice=0 active=2 refcnt=3
Jun 25 02:30:37 hp-114-tink-system kernel:     pending: output_poll_execute, fqdir_free_fn
Jun 25 02:30:37 hp-114-tink-system kernel: workqueue events_long: flags=0x0
Jun 25 02:30:37 hp-114-tink-system kernel:   pwq 10: cpus=2 node=0 flags=0x0 nice=0 active=1 refcnt=2
Jun 25 02:30:37 hp-114-tink-system kernel:     pending: br_fdb_cleanup [bridge]
Jun 25 02:30:37 hp-114-tink-system kernel: workqueue events_freezable: flags=0x4
Jun 25 02:30:37 hp-114-tink-system kernel:   pwq 10: cpus=2 node=0 flags=0x0 nice=0 active=1 refcnt=2
Jun 25 02:30:37 hp-114-tink-system kernel:     pending: pci_pme_list_scan
Jun 25 02:30:37 hp-114-tink-system kernel: workqueue events_power_efficient: flags=0x80
Jun 25 02:30:37 hp-114-tink-system kernel:   pwq 10: cpus=2 node=0 flags=0x0 nice=0 active=3 refcnt=4
Jun 25 02:30:37 hp-114-tink-system kernel:     in-flight: 3192461:gc_worker [nf_conntrack]
[...]

PR #13426, opened 2026-06-30
feat(chart): support adjusting irqaffinity

derekbit:

Issue Longhorn/longhorn#13417


derekbit:
Addressed in https://github.com/longhorn/longhorn/pull/13496/changes

Issue #12837, opened 2026-03-24
[BUG] V2 Data Engine: SPDK block disk initialization takes exactly 12 minutes per disk to become Ready after node startup


jiangbinzhe:
### Describe the Bug

After starting Longhorn, the initialization process for block-type disks (used by the V2 SPDK Data Engine) under a node experiences a severe delay. Each individual physical disk takes exactly 12 minutes to transition into the Ready/Schedulable state.

During this 12-minute disk initialization phase, the longhorn-instance-manager logs reveal a strictly 1-minute interval between debug logs originating from the SPDKBackingImageList function. This indicates that the instance manager is caught in a synchronous polling loop or a recurrent timeout mechanism while enumerating or initializing the SPDK block devices on the node.

### To Reproduce

Configure a Longhorn cluster with the V2 Data Engine (SPDK) enabled.

Add block devices (e.g., PCIe NVMe drives) as disks to a Longhorn Node for the V2 engine. Observe that this initial disk addition and initialization process completes quickly without issues.

Trigger a disk rescan phase by either rebooting the physical node or manually deleting/restarting the longhorn-instance-manager pod on that node.

Observe the Longhorn UI "Nodes" -> "Disks" section during this restart and rescan phase: each previously configured SPDK block disk now takes exactly 12 minutes to transition back to the Schedulable / Ready status.

Monitor the longhorn-instance-manager pod logs and observe the strict 1-minute synchronous polling interval originating from the SPDKBackingImageList function.

### Expected Behavior

When a node starts or a block disk is added to the V2 engine, the disk initialization and the status update to Schedulable should complete within a few seconds, without prolonged 1-minute periodic polling cycles per disk.

### Support Bundle for Troubleshooting

I will generate the support bundle from the Longhorn UI and send it to longhorn-support-bundle@suse.com immediately after submitting this issue. I will include the assigned GitHub Issue number in the email subject line for reference.

### Environment

- Longhorn version: v1.11.1
- Impacted volume (PV): N/A (Disk initialization issue, not PV specific)
- Installation method (e.g. Rancher Catalog App/Helm/Kubectl): [ Helm]
- Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: K3s, v1.34.5+k3s1
  - Number of control plane nodes in the cluster: 1
  - Number of worker nodes in the cluster: 0 (Single node cluster)
- Node config
  - OS type and version: Ubuntu 24.04.3 LTS
[...]

PR #13463, opened 2026-07-06
docs(lep): spdk batch dirty recovery

davidcheng0922:

https://github.com/longhorn/longhorn/issues/12837

### DIFF
--- a/enhancements/20260706-spdk-batch-dirty-recovery.md
+++ b/enhancements/20260706-spdk-batch-dirty-recovery.md
@@ -1,0 +1,783 @@
+# SPDK Batched Dirty Recovery
+
+## Summary
+
+Introduce a batched dirty recovery path for SPDK blobstore. The new path is enabled by
+default through SPDK blobstore options, with an internal rollback knob for development
+and validation. It reduces dirty recovery time by:
+
+- Scanning the blob metadata page region with larger contiguous reads.
+- Replaying discovered blob metadata chains with bounded concurrency.
+- Keeping commit serial and reusing the existing metadata parse and used-mask write path.
+
+This targets Longhorn V2/SPDK recovery cases where a large lvstore has a large blob
+metadata page region but only a small number of active lvols/blobs.
+
+Related issue: https://github.com/longhorn/longhorn/issues/12837
+
+## Problem
+
+The current dirty recovery path, `bs_load_replay_md()`, scans blob metadata one page at a
+time:
+
+```
+┌─────────────────────────────────────────┐
+│   Legacy Serial Recovery Loop           │
+├─────────────────────────────────────────┤
+│                                         │
+│  1. Read one metadata page (4 KB I/O)   │
+│     ↓                                   │
+│  2. Validate CRC and root-page fields   │
+│     ↓                                   │
+│  3. If root page:                       │
+│       → Replay its chain                │
+│       → Commit it                       │
+│     ↓                                   │
+│  4. Continue with next unclaimed page   │
+│     ↓                                   │
+│  Repeat for all md_len pages            │
+│                                         │
+└─────────────────────────────────────────┘
+```
+
+For a 512 GiB blobstore with 1 MiB clusters, SPDK's default metadata sizing reserves one
+metadata page per cluster:
+
+```
+┌─────────────────────────────────────┐
+│  Example: 512 GiB Blobstore         │
+├─────────────────────────────────────┤
+│  md_len         = 524,288 pages     │
+│  md_page_size   = 4 KiB             │
+│  scan size      = 2 GiB             │
+└─────────────────────────────────────┘
+```
+
+The legacy path therefore issues hundreds of thousands of small 4 KiB device reads. On
+some NVMe devices this low-QD/small-I/O pattern is much slower than the device's normal
+sequential read capability.
+
+## Goals
+
+- Reduce dirty recovery scan time by reducing device I/O count and improving I/O size.
+- Preserve recovery correctness by keeping metadata validation at page granularity.
+- Keep all global state mutation serial during commit.
+- Provide a safe fallback to the existing serial recovery path.
+- Avoid changing the on-disk blobstore format.
+
+## Non-Goals
+
+- Change clean recovery.
+- Change blobstore or lvol metadata layout.
+- Parallelize commit/global state updates.
+- Optimize normal SPDK read/write I/O paths.
+
+## Relevant Blobstore Layout
+
+### Key Concepts
+
+**Blobstore** = A logical storage pool backed by one block device.
+
+**Blob** = A logical volume (lvol) backed by metadata + data clusters. Each blob has a unique blob ID.
+
+**Cluster** = Data allocation unit (typically 1 MiB). Blobs allocate data in cluster granularity.
+
+**Page** = Metadata allocation unit (4 KiB). Blob metadata is stored as a chain of pages.
+
+**Relationship:**
+- One blobstore contains many blobs.
+- One blob = one metadata chain (linked pages) + N data clusters.
+- `total_clusters = blobstore_size / cluster_size` (blobstore.c:11363)
+- `md_len ≈ total_clusters` (default: one metadata page reserved per cluster)
+- Metadata region itself occupies some clusters, reducing usable data clusters.
+
+### Super Block Structure
+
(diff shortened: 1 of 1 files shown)


enhancements/20260706-spdk-batch-dirty-recovery.md
```diff
@@ -0,0 +1,600 @@
...
+
+```text
+CRC must match
+sequence_num must be 0
+page->id must equal bs_page_to_blobid(page_num)
+page must not be an extent page
+```
+
+Only valid root pages are recorded. A CRC mismatch during scan means "not a valid root
+candidate" and is skipped, matching the discovery nature of this phase.
```
shuo-wu, line 313: Does this mean that we will ignore the CRC-mismatched root pages? Will we clean up those invalid roots in the future?

enhancements/20260706-spdk-batch-dirty-recovery.md
```diff
@@ -0,0 +1,600 @@
...
+extent pages are not accepted as chain pages
+sequence_num must match the chain position
+root page id must match bs_page_to_blobid(root_page)
+continuation page id must match the root blob id
+next must be < md_len unless it is SPDK_INVALID_MD_PAGE
+```
+
+Any uncertainty during replay falls back to the existing serial recovery path. Unlike
+Phase 1, these pages are expected chain pages, so silently skipping them could produce an
+incomplete blob.
```
shuo-wu: Do you mean falling back to start the existing serial recovery flow from phase 1? Will the required time be the time the old method required (12 min, for example) plus the time consumed by this new method?

enhancements/20260706-spdk-batch-dirty-recovery.md
```diff
@@ -0,0 +1,600 @@
...
+
+Phase 2 creates one replay chain per discovered root page:
+
+```text
+replay_qd = min(recovery_qd, chain_count)
+```
+
+It keeps at most `replay_qd` chains active. For each active chain, it reads one metadata
+page at a time, validates it, appends a copy to the in-memory chain, then follows
+`page->next` until the chain ends.
```
shuo-wu: For each md chain (of each blob in this blobstore), there will be a thread that handles serial page reads, right? Will this be no improvement for some cases in this phase? e.g, when there is only one giant blob (512G) occupying all space in a blobstore (512G), this step means reading all 524,288 md pages in the single chain one by one.

~~Besides, after each read, will it cache the page? If yes, does this mean that spdk_tgt has to cache all md pages (2GB) in memory?~~ The below **Replay Buffer** explains that spdk_tgt will not cache all md pages

shuo-wu (dismissed):
In general LGTM.

This new proposal would improve the Phase 1 (scan Root Pages) handling a lot anyway.
And the explanations in section `Relevant Blobstore Layout` really helps!

enhancements/20260706-spdk-batch-dirty-recovery.md
```diff
@@ -0,0 +1,600 @@
...
+- The super block must validate before dirty recovery starts.
+- The blob metadata page region must be addressed through `bs_md_page_to_lba()`.
+- Chunk scanning must continue to validate pages individually.
+- Phase 2 must not silently skip expected chain pages.
+- Global used masks and cluster accounting must be updated serially.
+- The existing used-mask write path must write rebuilt masks after commit.
+
+## Rollback
+
+Set `recovery_qd` to `1` to force the legacy serial dirty recovery path.
```
c3y1huang: Will this be exposed to users via a setting?

enhancements/20260706-spdk-batch-dirty-recovery.md
```diff
@@ -0,0 +1,600 @@
+# SPDK Batched Dirty Recovery
+
+## Summary
+
+Introduce an opt-in batched dirty recovery path for SPDK blobstore. The new path
+reduces dirty recovery time by:
```
c3y1huang: `opt-in` might be slightly misleading, as the code seems to use the new batched path by default and users can't actually opt in or out.

c3y1huang (commented):
In general LGTM.

enhancements/20260706-spdk-batch-dirty-recovery.md
davidcheng0922: Yes. If a page has a CRC mismatch during Phase 1, we cannot trust it as a valid root page, so it is skipped. This matches the existing serial recovery behavior: a metadata page must pass CRC validation before it can be treated as a valid root.

The recovery path does not explicitly wipe that invalid page. Instead, dirty recovery rebuilds the used masks from valid metadata chains. A **CRC-mismatched page will not be marked as used** and **can be overwritten by future metadata allocation**. If the corrupted page was the only valid root of an active blob, the existing serial recovery would not be able to recover it reliably either.

enhancements/20260706-spdk-batch-dirty-recovery.md
davidcheng0922: Yes, fallback will take longer time.

So in the worst case, the elapsed time can be the time already spent in batched recovery plus the time required by the legacy serial recovery. This is intentional: fallback is a correctness safety net for unexpected or uncertain states, not the expected fast path.

enhancements/20260706-spdk-batch-dirty-recovery.md
davidcheng0922: Probably yes for that specific case. 

If there is only one blob and its metadata is one long chain, Phase 2 will have limited benefit. A metadata chain is effectively a linked list, so we need to read the current page before knowing `page->next`; parallelizing within one chain is not straightforward or obviously safe without extra indexing/layout changes.

enhancements/20260706-spdk-batch-dirty-recovery.md
davidcheng0922: As the discussion in the meeting, I will investigate this part. Thank you

enhancements/20260706-spdk-batch-dirty-recovery.md
davidcheng0922: Maybe not, it is quick fallback for us currently.

enhancements/20260706-spdk-batch-dirty-recovery.md
davidcheng0922: I will fix it, thank you

davidcheng0922:
### Improvement

- Phase 1 now retries only the failed scan batch/chunk with single-page reads instead of falling back the entire recovery path immediately.

- Phase 2 now supports summary-based range replay, so long metadata chains can be replayed with batched range/chunk reads instead of many individual 4 KiB reads. After summary, we do know the chain head and tail for the entire chain. [Design Detail](https://github.com/davidcheng0922/longhorn/blob/issue-12837-batch-dirty-recovery/enhancements/20260706-spdk-batch-dirty-recovery.md#page-summary)



### Findings

- Metadata page order is often ascending in practice, but SPDK does not guarantee it, so recovery must follow next rather than predict page order.

- Metadata pages should not overlap across valid chains; if overlap is detected, recovery treats it as invalid metadata.

- Metadata chain length is not directly related to blob size or written data size. In my test, a 500 GiB volume with 200 GiB written data still had only 2 metadata chain pages. Chain length is more likely related to the number and size of blob metadata descriptors, such as extents, xattrs, snapshots/clones, or other metadata-heavy SPDK operations.

- Recovery cannot scan only active blobs because dirty recovery rebuilds blobstore state, and active/inactive blob information is not trusted after an unclean shutdown.

### Current Test Benchmark

```
Env  : aws c5d.xlarge
Disk Size  : 512Gi
CPU Mask  :  0x8 (1 cpu) 


Volumes: 
- volume 1: 500Gi with 200Gi IO
- volume 2: 20Gi  with 1Gi IO
```

Original dirty recovery time: 3 minutes
Current method: 15 seconds

cc @shuo-wu

enhancements/20260706-spdk-batch-dirty-recovery.md
shuo-wu: The improvement suggestion I mentioned in the discussion for this retry: Only retry the serial recovery for **the pages in CRC-mismatched chunks**, which is similar to the below `a chunk scan I/O fails` case

derekbit:
cc @hookak @bachmanity1

Issue #13476, opened 2026-07-08
[IMPROVEMENT] Support per-component PriorityClass overrides for system-managed components


swasabs:
## Is your feature request related to a problem? Please describe

Longhorn currently has a single `priority-class` setting for system-managed components. This is useful as a default, but it does not allow operators to assign different PriorityClasses to node-local storage components and cluster-level CSI sidecars.

During node memory pressure/OOM scenarios, some node-local Longhorn components may need stronger scheduling/preemption protection than cluster-level components. For example, an operator may want `instance-manager`, `engine-image`, and `longhorn-csi-plugin` to use `system-node-critical`, while CSI controller sidecars continue to use `system-cluster-critical` or a custom Longhorn PriorityClass.

## Describe the solution you'd like

Add optional per-component PriorityClass overrides for Longhorn system-managed components while preserving the existing global `priority-class` setting as the default/fallback.

The proposed setting is:

`system-managed-components-priority-classes`

The setting value maps component names to PriorityClass names. Supported component keys:

- `instance-manager`
- `engine-image`
- `longhorn-csi-plugin`
- `csi-attacher`
- `csi-provisioner`
- `csi-resizer`
- `csi-snapshotter`

If a component is not present in the map, it should continue to inherit the existing `priority-class` setting.

## Compatibility

This should be fully backward compatible:

- Existing `priority-class` behavior remains unchanged.
- The new setting is optional.
- If the new setting is unset or a component is omitted, the component continues to use the global `priority-class` fallback.

Changing PriorityClass for these components should continue to be treated as a danger-zone operation requiring detached volumes, consistent with the existing `priority-class` behavior.

## Additional context

Implementation PR:

- https://github.com/longhorn/longhorn-manager/pull/4978
- https://github.com/longhorn/longhorn/pull/13485

swasabs:
## Pre Ready-For-Testing Checklist
* [ ] **Context & Reproduction**
    Feature request, no bug repro. Test steps: install/upgrade with
    `defaultSettings.systemManagedComponentsPriorityClasses` set to a
    per-component map; verify instance-manager/engine-image/CSI sidecar
    pods pick up the mapped PriorityClass, and components omitted from
    the map still fall back to `defaultSettings.priorityClass`.

* [ ] **Workaround**
    None — today all system-managed components share the single global
    `priority-class` setting.

* [x] **Fix / Feature Explanation**
    Yes, both PRs describe Why/What changed/Compatibility clearly.

[...]

PR #13485, opened 2026-07-09
feat: expose system-managed priority class overrides

swasabs:
## Why

Expose the manager-side fine-grained priority class support through Helm values while preserving the existing global default behavior.

## What changed

- Add `defaultSettings.systemManagedComponentsPriorityClasses`
- Render the new Longhorn setting only when explicitly set
- Document the new value in chart docs

## Compatibility

- Existing `defaultSettings.priorityClass` behavior is unchanged
- If the new value is omitted, all components continue to inherit the global priority class

## Testing

- `helm lint chart`
- Rendered the setting from both JSON-string and YAML-map values
- Verified the unset default does not render the setting

## Related

- Feature issue: https://github.com/longhorn/longhorn/issues/13476
- Manager companion PR: https://github.com/longhorn/longhorn-manager/pull/4978


swasabs:
.

Issue #1061, opened 2020-02-18
feature: shard support


yasker:
Support sharding across different disks (same or different nodes) for one volume.

jcambronero:
hi @yasker, what is the status of this feature please?

yasker:
@jcambronero We haven't scheduled this feature yet. What's your use case? Currently, we're recommending using LVM to aggregate disks on one node to help.

jcambronero:
thank you @yasker , we would like to have the ability to create volumes that are larger that the maximum capacity available on a single node. We're hoping sharding across different nodes would allow us to accomplish this.

yasker:
In that case, yes you will need this feature. We are in the development cycle of v1.1.0 release, targeting October. Once we're done with v1.1.0, we can take a look and see what we can do about this feature.

PhilParisot:
Any updates for this?

yasker:
@PhilParisot we've raised the priority of this issue. We will start working on it in the near future.

PhilParisot:
Awesome thanks @yasker!

yasker:
Also, as an alternative, we can consider LVM in striping mode. But need to know what we should do to deal with the disk failure.

michaelandrepearce:
An approach ive seen in other distributed storage solutions is that a block volume is broken into chunks and chunks are then managed in similar way to how the currently the whole volume is, where then the chunks are individually replicated and moved etc.

innobead:
> I dont see how lvm stripping mode will work as lvm is still local to a single node. The issue i have is that i need a pv larger than a single node. An approach ive seen in other distributed storage solutions is that a block volume is broken into chunks and chunks are then managed in similar way to how the currently the whole volume is, where then the chunks are individually replicated and moved etc.

The definite end goal is to try to leverage all disk resources across the nodes in the future, but right now, what we are looking for is based on the current architecture and design to efficiently enhance in-node disk-based volume management to in-node across-disks-based.

michaelpearce-gain:
@innobead thats pretty much useless, i can manage that in node with LVM today, the issue is that we need PV's larger than disks within a node.

mfbrj:
Sharding in the same node seems to be useless. Anyone can create a node with larger disks. What we neeed is transpararent sharding leveraging the space on several nodes, pehaps with another concept beyond simple replicas. Chunks?

An example of a new Model:
[...]

PR #13524, opened 2026-07-15
docs(examples): reorganize v2 examples and introduce sharding

c3y1huang:

Issue longhorn/longhorn#1061

Add example YAMLs for V2 sharding volumes.

`None`

`None`

### DIFF
--- a/examples/v2/replication/storageclass.yaml
+++ b/examples/v2/replication/storageclass.yaml
@@ -1,16 +1,19 @@
 kind: StorageClass
 apiVersion: storage.k8s.io/v1
 metadata:
-  name: longhorn-v2-data-engine
+  name: longhorn-v2-replicated
 provisioner: driver.longhorn.io
 allowVolumeExpansion: true
 reclaimPolicy: Delete
 volumeBindingMode: Immediate
 parameters:
-  numberOfReplicas: "3"
+  dataEngine: "v2"
+  # dataLayout is optional for replicated volumes - RAID1 is the default.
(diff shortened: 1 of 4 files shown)


examples/v2/replication/storageclass.yaml
```diff
@@ -1,16 +1,19 @@
...
 provisioner: driver.longhorn.io
 allowVolumeExpansion: true
 reclaimPolicy: Delete
 volumeBindingMode: Immediate
 parameters:
-  numberOfReplicas: "3"
+  dataEngine: "v2"
+  # dataLayout is optional for replicated volumes - RAID1 is the default.
+  dataLayout.type: "replicated"
+  dataLayout.mode: "raid1"
```
derekbit, line 13: If the two parameters are not assigned
```
   dataLayout.type: "replicated"
   dataLayout.mode: "raid1"
```
the default values should be `replicated` and `raid1` ,right?

c3y1huang: [Leaving it empty will be accepted as the replicated volume](https://github.com/longhorn/longhorn-manager/blob/bc1d7ff798c791552777f7408d2b7b0365f22686/webhook/resources/volume/validator.go#L759-L765). We don't normalize it, so they stay empty on the CR.

derekbit (approved):

derekbit:
Might need to update the path to the v2 example in the official doc.

c3y1huang:
@mergify backport v1.12.x

Issue #10745, opened 2025-04-18
[FEATURE] Configurable block disk block size


jfroy:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

In Longhorn v1.8 with the V2 engine, it is not possible to configure the block size. The hardcoded default `512` size is suboptimal, particularly for NVMe namespaces using a 4K block size format.

See https://github.com/longhorn/longhorn-manager/blob/master/controller/monitor/disk_utils.go#L25

### Describe the solution you'd like

Add a new field to `DiskSpec` to specify the block size.

PhanLe1010:
Looks like the setting 512b block size only applied to disk with [AIO driver](https://github.com/longhorn/longhorn-spdk-engine/blob/d357f88a37d692f673b350265802a6c66b9821bc/vendor/github.com/longhorn/go-spdk-helper/pkg/spdk/client/basic.go#L44-L59) . For other disk driver like [NVME](https://github.com/longhorn/longhorn-spdk-engine/blob/d357f88a37d692f673b350265802a6c66b9821bc/pkg/spdk/disk/nvme/nvme.go#L36-L66), [DiskDriverVirtioBlk](https://github.com/longhorn/longhorn-spdk-engine/blob/d357f88a37d692f673b350265802a6c66b9821bc/pkg/spdk/disk/virtio-blk/virtio-blk.go#L25-L55), [DiskDriverVirtioScsi](https://github.com/longhorn/longhorn-spdk-engine/blob/d357f88a37d692f673b350265802a6c66b9821bc/pkg/spdk/disk/virtio-scsi/virtio-scsi.go#L25-L55), the block size would be inferred from the hardware information 


Are you using AIO disk driver @jfroy ?

Since AIO driver is the only use case for this, I thinking we should wait to see if there are a lot of user who use AIO driver and how much is the performance difference. WDYT @derekbit @shuo-wu @DamiaSan ?

derekbit:
Yes, as @PhanLe1010, only AIO disk driver supports configurable block size. However, we still stick to 512 bytes for the compatibility of the backing image feature.

PhanLe1010:
> Yes, as [@PhanLe1010](https://github.com/PhanLe1010), only AIO disk driver supports configurable block size. However, we still stick to 512 bytes for the compatibility of the backing image feature.

Thanks @derekbit 
This would raise a concern is that if the NVME hardware disk has 4096 byte block size, it would break our backing image feature? 
[...]

PR #13584, opened 2026-07-21
docs(lep): define configurable v2 logical block size

christophersherman:

Issue longhorn/longhorn#10745

This proposal defines the safety contract behind configurable AIO block sizes. A per-disk field alone is not enough: SPDK RAID cannot combine replicas with different logical block sizes, so Longhorn also needs an immutable V2 volume block size and scheduler enforcement.

The proposal keeps 512 bytes as the default, initially supports 512 and 4096, records the size actually reported by every V2 block disk, and covers legacy disks and volumes without assuming that every unset resource is 512. It also specifies how clone, backup, restore, DR, EC shards, and failed-replica reuse preserve the volume contract.

The main design choice is volume-level ownership. I considered a cluster-wide setting, disk tags, and inferring from the first replica, but each either prevents heterogeneous clusters or leaves correctness dependent on scheduling order and operator convention.

V2 backing-image support remains removed as of v1.12 and is not reintroduced here. The proposal only records the 512-byte compatibility constraint that must apply if that feature returns without a separate image compatibility design.

The implementation remains split across component PRs. longhorn-manager#5021 stays in draft while this contract is reviewed.

Documentation-only change. The test plan includes Linux unit/controller coverage, SPDK integration checks, synthetic 4Kn loop-device E2E coverage, and a physical 4Kn NVMe manual test.

### DIFF
--- a/enhancements/20260721-configurable-v2-logical-block-size.md
+++ b/enhancements/20260721-configurable-v2-logical-block-size.md
@@ -1,0 +1,467 @@
+# Configurable V2 Logical Block Size
+
+## Summary
+
+Longhorn currently creates every AIO-backed V2 disk with a hard-coded 512-byte logical block size. This prevents a 4Kn block device from becoming ready because SPDK refuses to expose an AIO bdev with a logical block size smaller than the block device reports. It also prevents users from intentionally presenting a 512-byte-native device as a 4096-byte AIO bdev.
+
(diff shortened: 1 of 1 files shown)


derekbit:
@christophersherman The PR is still in draft state. Is it ready for review?

christophersherman:
Ready for review now. Thank you!

derekbit:
cc @c3y1huang @shuo-wu @hookak @PhanLe1010 @innobead

christophersherman:
Rebased onto current master. The implementation draft is current too. Cheers

Issue #12552, opened 2026-01-28
[FEATURE] V2 Data Engine Fast Cloning 


derekbit:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

The v2 data engine generally does not support fast cloning based on COW, as SPDK lvol snapshots and clones operate only on the same disk and require a strict parent/child relationship. Consequently, COW-based cloned volumes cannot be rebuilt on another disk and will fail permanently if the parent snapshot or the original disk is lost.

In the initial phase, @PhanLe1010  implemented a general solution for v2 volume cloning. However, starting a new VM from a golden image under this approach performs significantly worse than v1, and the copy process consumes substantial bandwidth and disk space.

For more details, see the enhancement proposal: [v2 Volume Cloning](https://github.com/longhorn/longhorn/blob/master/enhancements/20250508-v2-volume-cloning.md).

Recently, we explored a method to aggregate multiple disks into a single logical disk, added as a single AIO disk and lvstore (ref: [V2 Disk Size Aggregation](https://longhorn.io/blog/20251223-v2-disk-size-aggregation/). We need to investigate whether COW-based cloning can be made feasible.

derekbit:
Discussion https://suse.slack.com/archives/C02DR3N5T24/p1770347509393899

shuo-wu:
## Manual Tests

- Comprehensive linked-clone volume feature set tests. Refer to the step in https://github.com/shuo-wu/longhorn-tests/blob/2bc8fed5b7953d6ac78750c1fc0b78d17739cac8/e2e/tests/regression/test_linked_clone.robot#L436. (You can directly run this robot test. It works. But other linked-clone tests are not bug free right now. I will fix them before creating a robot test PR)

- Upgrade test with existing/legacy linked-clone volume: 
    1. Create some linked-clone volumes in v1.12.0
    2. Attach some of the volumes 
    3. Upgrade the system to the latest
    4. Make sure the legacy volumes still have basic functionalities: attach, detech, and backup creation
    5. Delete all legacy volumes.
    6. Create some new linked-clone volumes. Make sure the feature set works.

derekbit:
Hello @shuo-wu Is it ready to move to `Ready For Testing`?

roger-ryao:
Hi @shuo-wu 
    While testing `linked-clone` volume and `nested-clone` volume, I found that the snapshots created do not show up under `Snapshots and Backups`.
   Attached are the screenshot and the support bundle

```bash
[...]

PR #13593, opened 2026-07-23
feat(crd): add linkedCloneSrcReplicaName field to replica CRD schema

shuo-wu:

Issue https://github.com/longhorn/longhorn/issues/12552


derekbit:
Updated by  https://github.com/longhorn/longhorn/pull/13597

Issue #13634, opened 2026-07-30
[IMPROVEMENT] User-deployed Longhorn pods missing longhorn.io/component label (inconsistent with system-managed components)


rafaelgaspar:
Thanks for stopping by to let us know something could be better!

### Describe the Bug

Longhorn labels system-managed pods with `longhorn.io/component`, but user-deployed pods do not get that label — only `app:` is set. That inconsistency breaks a uniform labeling model and makes network policies, monitoring, and operational selectors harder than the [networking reference](https://longhorn.io/docs/1.12.0/references/networking/) implies.

### To Reproduce

1. Install Longhorn v1.12.0 (Helm or manifest).
2. Inspect pod labels:

```sh
kubectl get pod -n longhorn-system -l app=longhorn-manager -o jsonpath='{.items[0].metadata.labels}' | jq .
kubectl get pod -n longhorn-system -l longhorn.io/component=instance-manager -o jsonpath='{.items[0].metadata.labels}' | jq .
kubectl get pod -n longhorn-system -l app=longhorn-csi-plugin -o jsonpath='{.items[0].metadata.labels}' | jq .
```

**System-managed** pods (created by longhorn-manager) have `longhorn.io/component`:

```yaml
# instance-manager
longhorn.io/component: instance-manager
longhorn.io/managed-by: longhorn-manager
```

**User-deployed** pods do **not** have `longhorn.io/component`:

```yaml
# longhorn-manager
app: longhorn-manager
longhorn.io/admission-webhook: longhorn-admission-webhook
longhorn.io/recovery-backend: longhorn-recovery-backend
# no longhorn.io/component

# longhorn-csi-plugin (created by driver deployer, not Helm)
app: longhorn-csi-plugin
# no longhorn.io/component
```

Same for `longhorn-ui`, `longhorn-driver-deployer`, and CSI sidecars (`csi-attacher`, `csi-provisioner`, `csi-resizer`, `csi-snapshotter`) — all `app:` only.

`GetManagerLabels()` in longhorn-manager only sets `app: longhorn-manager`; it does not set `longhorn.io/component`. The v1.12 Helm chart has no `podLabels` (or equivalent) for manager/UI/driver-deployer, and CSI resources are created at runtime by `deploy-driver`, so they never receive the component label either.

### Expected Behavior

All Longhorn pods should use the same component labeling scheme: **`longhorn.io/component: <name>`** on every workload, in addition to existing `app:` labels (backward compatible).

[...]

PR #13635, opened 2026-07-30
feat(chart): add longhorn.io/component labels to chart workloads

rafaelgaspar:

Issue longhorn/longhorn#13634

Helm-managed pods (manager, UI, driver-deployer) lack `longhorn.io/component` while system-managed pods have it. This adds the label to chart pod templates and updates bundled chart network policies and `examples/network-policy/` to select on `longhorn.io/component` instead of `app:`.

- `longhorn-manager` → `manager`
- `longhorn-ui` → `ui`
- `longhorn-driver-deployer` → `driver-deployer`

Network policy ingress rules referencing CSI use `csi-plugin` (pods labeled by longhorn-manager).

Existing `app:` labels are unchanged.

- Chart-only; no manager/CSI runtime code in this PR.
- Bundled network policies under `chart/templates/network-policies/` and matching `examples/network-policy/` files updated together.

https://github.com/longhorn/longhorn/issues/13634

#### Test plan
```
helm lint chart
Result: PASS (0 failures)
```

### DIFF
--- a/chart/templates/daemonset-sa.yaml
+++ b/chart/templates/daemonset-sa.yaml
@@ -13,6 +13,7 @@ spec:
     metadata:
       labels: {{- include "longhorn.labels" . | nindent 8 }}
         app: longhorn-manager
+        longhorn.io/component: manager
       {{- with .Values.annotations }}
       annotations:
         {{- toYaml . | nindent 8 }}
--- a/chart/templates/deployment-driver.yaml
+++ b/chart/templates/deployment-driver.yaml
@@ -13,6 +13,7 @@ spec:
     metadata:
       labels: {{- include "longhorn.labels" . | nindent 8 }}
         app: longhorn-driver-deployer
+        longhorn.io/component: driver-deployer
     spec:
       initContainers:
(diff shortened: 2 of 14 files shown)


derekbit:
Thanks @rafaelgaspar for your contribution.

I think we should keep the original labels for backward compatibility, since other applications or projects integrating with Longhorn may still rely on them. We can add `longhorn.io/component` as a consistent label across all Longhorn workloads, but removing or changing the existing labels could break existing selectors and introduce a breaking change.

WDYT? @shuo-wu @c3y1huang @COLDTURNIP @innobead

rafaelgaspar:
Thanks @derekbit — agreed.

This PR already keeps the existing `app:` labels on manager / UI / driver-deployer and only *adds* `longhorn.io/component` for consistency with system-managed pods.

I’ve reverted the NetworkPolicy / example selector changes so we don’t alter existing `app:`-based selectors (also safer while CSI pods still lack `longhorn.io/component`). We can align those selectors in a follow-up once all workloads carry the component label.

Happy to adjust further if you’d prefer a different approach.

Issue #13636, opened 2026-07-31
[IMPROVEMENT] Help large replicas on full disks reclaim the space and squash all data into the heads


shuo-wu:
## What's the task? Please describe

When a v1 replica is large and the disk is full, there is no way to rescue the replica. Since the snapshot deletion & purge require extra free space, which is impossible on a full disk.
To help salvage such a replica, we can provide a script that reclaims space for the replica even though the replica is not running and there is no extra space on the disk.

#### The proposal of the script:
1. Scan all snap & head files to build a sector location list/map that indicates which snap file contains the latest data chunk at a specific offset. This can refer to how Longhorn replica initializes a [sector location list](https://github.com/longhorn/longhorn-engine/blob/3bae247d3a909f5b2ff4b4efa3810ec0006f651c/pkg/replica/diff_disk.go#L19-L24): https://github.com/longhorn/longhorn-engine/blob/3bae247d3a909f5b2ff4b4efa3810ec0006f651c/pkg/replica/diff_disk.go#L406-L437
2. For each sector [offset, length], figure out if there is the latest data in a snap based on the above sector location map. If YES, directly and blindly punch holes for all snaps older than the snap. (.
3. Walk through the sector location map sector-by-sector, move the data from the snap file to the head, then punch a hole to the snap file.
4. All snap files should already be empty here. Remove these snap files and update the meta files for the replica.

#### An example
- When a replica contains such a snapshot chain:
```
snap-old    _xxx
 |
 v
snap-new    __yy
 |
 v
head        ___z
```

- In the 1st step: build a sector location map like below(, which can be `[]byte`, the list index is the offset while the list value is the snaps file index). The list index means data offset:

| Offset  |  File |
|---|---|
| 0 | none |
| 1 | snap-old |
| 2 | snap-new |
| 3 | head |
 
- In the 2nd step: iterate all blocks based on the above list:
    - For Offset 0, there is no data. Do nothing.
    - For Offset 1, the latest data `x` is in `snap-old`. Try to punch a hole at offset 1 for all snapshot files older than `snap-old`. Since there is no older snap file, do nothing.
    - For Offset 2, the latest data `y` is in `snap-new`. Try to punch a hole at offset 2 for all snapshot files older than `snap-new`. Since the only older snap is `snap-old`, punch a hole at offset 2 for `snap-old`.
[...]

PR #13652, opened 2026-08-04
Allows replicas on full disks reclaim the space and squash all data into the heads

apoorvajagtap:
Issue #13636 

This PR so far:
- Scan all snap & head files to build a sector location list/map that indicates which snap file contains the latest data chunk at a specific offset. 
- For each sector [offset, length], if there is the latest data in a snap based on the above sector location map, directly and blindly punches holes for all snaps older than the snap.
- moves the data from the snap file to the head then punches hole for those specific sectors in the snap file.


derekbit:
Hello @apoorvajagtap 
Thanks for your first implementation. 
I thought it is a shell script before. Would you be able to move the implementation and merge it into https://github.com/longhorn/cli that is used for troubleshooting?

apoorvajagtap:
Closing this in favour of https://github.com/longhorn/cli/pull/572
Let me know in case the location is not appropriate for the script in longhorn/cli. Thanks!

Issue #5220, opened 2023-01-06
[FEATURE] Encrypt volume backup to remote backup store without in-cluster volume encryption


innobead:
## Is your feature request related to a problem? Please describe (👍 if you like this request)

For volume data encryption, Longhorn supports FS volume encryption (block volume encryption will be #4883 ), so it can do volume encryption in transit and at rest. When encrypting a volume and backing it up to the remote backup store, the data will be encrypted as well and this is how to achieve at-rest encryption. However, this at-rest backup encryption needs to rely on in-cluster volume encryption.

To make the encryption operation flexible and meet users' different encryption compliance requirements, supporting backup volume encryption to the remote backup store can be independent of in-cluster volume encryption. 

## Describe the solution you'd like

- Have a global setting or volume-specific setting to enable remote backup encryption. By default, it's disabled.
- Users can provide an encryption key via different ciphers like AES-256
- Users can use external KMS providers for their key source

## Describe alternatives you've considered

Just rely on the server-side encryption of backup store, but it will be vendor-lockin

## Additional context

cc @longhorn/dev 

## Related Tickets

https://github.com/longhorn/longhorn/issues/8453

PatrickHuetter:
This feature would be awesome since it would be possible to use cheap third party s3 storage providers like https://www.idrive.com/object-storage-e2/ without having the fear of getting unauthorized data access to backups. Especially if you don't want to encrypt volumes locally.

rebeccazzzz:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Longhorn-Sprint-59d8105b2fbdf4155c4589ef/issues/longhorn/longhorn/5220?planning-poker) with Zenhub @chriscchien @ejweber @ChanYiLin @mantissahz @james-munson @PhanLe1010 @jillian-maroket @FrankYang0529 @roger-ryao @shuo-wu @yangchiu

rebeccazzzz:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Longhorn-Sprint-59d8105b2fbdf4155c4589ef/issues/longhorn/longhorn/5220?planning-poker) with Zenhub @derekbit

PatrickHuetter:
Just to add some thoughts here: I would prefer client-side encryption since you can't trust any external vendor. A hybrid solution would be very nice where the user could choose between client-side and server-side (s3 based) encryption but more important is the client-side encryption.

runningman84:
[...]

PR #13693, opened 2026-08-10
feat(lep): client-side backup encryption

carterli0407-cell:
Issue #5220

### What this PR does / why we need it:

Adds a LEP for client-side backup encryption (#5220).

Longhorn can encrypt a volume in-cluster but cannot encrypt only that volume's backup. Users who trust their cluster but not their backup destination currently have to enable in-cluster volume encryption purely to protect remote backup data.

This LEP proposes encrypting backup data locally, after compression and before the backup store, so the target only ever receives ciphertext.
It is configured independently of volume encryption, applies to every target `backupstore` supports (S3, NFS, CIFS) by sitting above `BackupStoreDriver`, and uses a Kubernetes Secret as the key source.

Configuration has two levels: `Volume.Spec.BackupEncryptionSecret` overrides the global `backup-encryption-secret` setting. If neither is set, backups behave exactly as they do today.

### Special notes for your reviewer:

- Server-side encryption removed and moved to Non-goals, tracked in #14037 .
- `inherit` state removed; resolution is Volume, then global.
- `BackupTarget`-level configuration dropped for the first version.
- `Enabled` removed from backup status; a non-empty Secret name means encrypted.
- S3-only restriction removed; encryption sits above `BackupStoreDriver`.
- `S3-Compatible Providers` section removed.
- Streaming is no longer required; `BackupStoreDriver.Write` takes an
  `io.ReadSeeker`, so encrypted temp files are used for whole-file paths.

Three points worth a close look, since they are new since the last round:

### DIFF
--- a/enhancements/20260810-backup-store-s3-encryption.md
+++ b/enhancements/20260810-backup-store-s3-encryption.md
@@ -1,255 +1,421 @@
-# Client-Side Encryption for S3 Volume Backups
+# Backup Store Encryption for S3 Targets
 
 ## Summary
 
-Longhorn currently supports volume encryption inside the Kubernetes cluster. When an encrypted volume is backed up, the backup data stored in the remote backup target is also protected because the source volume data is already encrypted.
+Longhorn currently supports encryption for in-cluster volumes. When an encrypted Longhorn volume is backed up, the backup data is also protected because the source volume data is already encrypted.
 
-However, this couples backup-at-rest encryption to in-cluster volume encryption. Users who do not need or do not want the runtime overhead and operational complexity of encrypted Longhorn volumes cannot independently protect their remote backups with client-side encryption.
+However, this couples backup-at-rest encryption with in-cluster volume encryption.
 
-This enhancement introduces **client-side encryption for Longhorn S3 backups**, independent of Longhorn in-cluster volume encryption.
+Some users want to keep Longhorn volumes unencrypted inside the Kubernetes cluster while independently encrypting backup data before sending it to a remote S3 backup store.
 
-When enabled, Longhorn encrypts backup data locally before sending it to the S3 or S3-compatible backup target and decrypts it locally while restoring. The S3 provider only receives encrypted data and does not receive the encryption key.
+This enhancement introduces **client-side encryption for Longhorn backups stored in S3 or S3-compatible backup targets**, independently of Longhorn in-cluster volume encryption.
 
-The feature is disabled by default to preserve the existing behavior.
+When client-side backup encryption is enabled:
 
-The initial implementation supports:
+```text
+Volume data
+    |
+    v
+Backup processing
+    |
+    v
+Compression
+    |
+    v
+Client-side encryption
+    |
+    v
+Existing S3 backupstore
+    |
+    v
+Remote S3-compatible storage
+```
 
-* S3 and S3-compatible backup targets only.
-* Client-side authenticated encryption using AES-256.
-* Encryption keys stored in a Kubernetes Secret.
-* Transparent backup, restore, backup listing, synchronization, and deletion.
-* Existing unencrypted backups without migration.
-* AWS S3 and MinIO interoperability.
-* An extensible key-provider design so external KMS integration can be added later.
+The remote object-storage provider receives ciphertext instead of plaintext backup data.
+
+The feature supports encryption configuration at three levels:
+
+1. Global
+2. BackupTarget
+3. Volume
+
+The configuration precedence is:
+
+```text
+Volume > BackupTarget > Global
+```
+
+The feature is disabled by default for backward compatibility.
+
+The initial implementation applies only to S3 and S3-compatible backup targets.
 
 ### Related Issues
 
 * #5220 - Encrypt volume backup to remote backup store without in-cluster volume encryption
 * #8453 - Backup encryption
-* #12297 - Automated test coverage for backup encryption
+* #12297 - Backup encryption automated test
 * #4883 - Block volume encryption
 
 ---
 
-## Motivation
+# Motivation
 
-Longhorn currently provides encryption at the volume level. This is useful when data must be encrypted on Longhorn storage nodes, but volume encryption and backup encryption serve different security requirements.
+Longhorn volume encryption and backup encryption address different security requirements.
 
-For example, a user might trust the storage nodes running inside their Kubernetes cluster but use an external S3-compatible provider for disaster-recovery backups.
+A user may trust the Kubernetes cluster and the Longhorn storage nodes but store disaster-recovery backups in external object storage.
 
-In this case, enabling volume encryption solely to secure remote backups introduces unnecessary coupling.
-
-Other applications commonly encrypt their backup data locally before sending it to object storage. Longhorn should provide the same option so users do not have to depend on storage-provider-specific server-side encryption or an additional encryption proxy.
-
-Server-side encryption such as SSE-S3, SSE-KMS, or SSE-C does not completely address this use case because encryption is still implemented by the object-storage service. It may also differ between AWS S3 and other S3-compatible providers.
-
-The intended security boundary for this enhancement is:
+For example:
 
 ```text
-Longhorn
-   |
-   | plaintext
-   v
-Compression
-   |
-   | compressed plaintext
-   v
-Client-side encryption
-   |
-   | ciphertext only
-   v
-S3 client
-   |
-   | ciphertext
-   v
-Remote S3 / S3-compatible storage
+Kubernetes cluster
+    |
+    | trusted
+    v
+Longhorn volume
+    |
+    | unencrypted
+    v
+Longhorn backup
+    |
+    | encrypted locally
+    v
+External S3 provider
 ```
 
-The remote backup provider must never need access to the plaintext backup payload or the Longhorn backup encryption key.
+Without independent backup encryption, users need to enable in-cluster volume encryption even if their only requirement is protecting data stored outside the cluster.
 
-### Goals
+This introduces unnecessary coupling between two separate security policies.
 
-1. Allow users to encrypt remote S3 backups without enabling Longhorn volume encryption.
+Client-side backup encryption allows users to define:
 
-2. Encrypt backup data before it leaves the Longhorn backup process.
+```text
+Volume encryption:
+    disabled
 
-3. Decrypt backup data only after it is downloaded from the remote backup target.
+Backup encryption:
+    enabled
+```
 
-4. Use authenticated encryption so corruption or malicious modification of encrypted objects is detected.
+The S3 provider receives encrypted data and does not need to implement any Longhorn-specific encryption capability.
 
-5. Support a user-managed AES-256 key stored in a Kubernetes Secret.
+This is also useful when users do not fully trust an external storage provider or when encryption must happen before data leaves the Kubernetes cluster.
 
-6. Remain compatible with AWS S3 and S3-compatible storage such as MinIO.
+---
 
-7. Preserve Longhorn's existing incremental backup and block deduplication behavior.
+# Goals
 
-8. Preserve compatibility with existing plaintext backups.
+The goals of this enhancement are:
 
-9. Keep encryption disabled by default.
+* Support client-side encryption for Longhorn S3 backups.
+* Make backup encryption independent of Longhorn volume encryption.
+* Encrypt backup data before it leaves Longhorn.
+* Support AWS S3 and S3-compatible backup targets through the existing Longhorn S3 implementation.
+* Keep the encryption layer independent from the S3 provider.
+* Allow encryption to be configured globally.
+* Allow a BackupTarget to override the global configuration.
+* Allow a Volume to override the BackupTarget and global configuration.
+* Use the following configuration precedence:
 
-10. Design the encryption metadata and key-provider abstraction so support for external KMS providers can be added without changing the encrypted data format unnecessarily.
+```text
+Volume > BackupTarget > Global
+```
 
-11. Ensure encryption keys and plaintext key material are never written to backup-store objects, logs, events, Backup CRs, or support bundles.
+* Allow users to explicitly disable encryption for a specific BackupTarget or Volume.
+* Use authenticated encryption to provide both confidentiality and integrity protection.
+* Store encryption keys in Kubernetes Secrets for the initial implementation.
+* Design key handling so external KMS providers can be supported in the future.
+* Preserve Longhorn incremental backup behavior where encryption contexts are compatible.
+* Preserve support for existing unencrypted backups.
+* Ensure existing configurations remain unchanged after upgrade.
+* Persist enough non-sensitive encryption information with every backup so existing backups remain restorable after configuration changes.
+* Never store plaintext encryption keys in backup metadata, CR status, logs, events, or support bundles.
+* Fail closed when encryption cannot be performed.
 
-### Non-goals
+---
 
-The following are outside the scope of the initial implementation:
+# Non-Goals
 
-* Client-side encryption for NFS, CIFS, Azure Blob, or other non-S3 backup targets.
+The following are outside the scope of this enhancement:
+
+* Server-side encryption.
+* SSE-S3.
+* SSE-KMS.
+* SSE-C.
+* Combining client-side encryption with server-side encryption.
+* Backup encryption for NFS or CIFS backup targets.
+* Backup encryption for non-S3 backupstore drivers.
 * Changing Longhorn in-cluster volume encryption.
-* SSE-S3, SSE-KMS, or SSE-C support as the implementation of this feature.
-* Encrypting S3 object names, object sizes, timestamps, or other storage-provider-visible metadata.
-* Automatic migration of existing plaintext backups to encrypted backups.
-* Automatic migration of encrypted backups back to plaintext.
-* In-place encryption-key rotation or re-encryption of existing backup data.
-* Generic external KMS support in the first implementation.
-* Recovering backups when the user has lost the encryption key.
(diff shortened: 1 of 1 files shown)


enhancements/20260810-backup-store-s3-encryption.md
```diff
@@ -0,0 +1,1497 @@
...
+Algorithm:   AES-256-GCM-DARE
+KeyProvider: secret
+KeyID:       backup-key-2026-01
+WrappedDEK:  ...
+Ciphertext:  ...
+```
+
+The exact binary encoding should be defined in the implementation and documented so future Longhorn versions can maintain compatibility.
+
+The header must contain no plaintext encryption key.
```
COLDTURNIP: Rather than design our own binary format to store the metadata plus the DEK, is it possible to keep such information in a JSON and archive it with the KEK in TAR or ZIP, to make it simpler without dedicate tool to extract?

carterli0407-cell: Good suggestion. I think we can avoid defining our own binary envelope and use a standard container format instead.

For example, we could package the backup object like:

backup.tar
├── metadata.json
└── data.enc

metadata.json could contain the format version, encryption algorithm, key provider, key ID, wrapped DEK, and the metadata required for key unwrapping, while data.enc contains the DARE-encrypted data stream.

But I don't think we should archive the KEK with the backup. The KEK should remain external in the Kubernetes Secret or KMS. We should only store the wrapped DEK in metadata.json; otherwise storing the KEK together with the encrypted data would defeat the key separation.

I would prefer TAR over ZIP here since our backup path is streaming-oriented, and TAR keeps the container format relatively simple. We should also make sure the metadata is authenticated/bound to the key wrapping so fields such as keyID or algorithm cannot be modified silently.

enhancements/20260810-backup-store-s3-encryption.md
```diff
@@ -0,0 +1,1497 @@
+# Client-Side Encryption for S3 Volume Backups
```
derekbit: I think we should support both `Server-Side Encryption` and `Client-Side Encryption` for this feature, but we can both one of them in the first phase. Could you create another ticket for the server-side encryption that we can implement in the future. WDYT?

enhancements/20260810-backup-store-s3-encryption.md
```diff
@@ -0,0 +1,1497 @@
...
+
+This is the proposed implementation approach.
+
+---
+
+## User Experience In Detail
+
+### Enabling backup encryption
+
+Encryption should be configured on the `BackupTarget`, because encryption is a property of the backup destination rather than of the running Longhorn volume.
```
derekbit: > Encryption should be configured on the `BackupTarget`


There are different use cases that we should support:

1. Some users may want to encrypt all backups globally.
2. Some users may want to encrypt backups only for a specific `BackupTarget`.
3. Some users may want to encrypt backups for only specific volumes, while leaving other volumes unencrypted.

Therefore, I propose supporting encryption configuration at three levels:

1. Global setting: provides the default encryption configuration.
2. Per-BackupTarget setting:我 overrides the global configuration for a specific BackupTarget.
3. Per-volume setting: overrides the BackupTarget and global configuration for a specific volume.

The configuration precedence would be: `Volume > BackupTarget > Global`

This would allow use cases such as:

- Encrypt all backups by default, while allowing specific volumes to opt out.
- Keep encryption disabled globally, but enable it for a specific BackupTarget.
- Use different encryption keys for different BackupTargets or volumes.

I think this hierarchical configuration model provides more flexibility than configuring encryption exclusively on the BackupTarget.

derekbit (commented):
@carterli0407-cell 
The LEP is incomplete, because the changes in CRDs are not mentioned.小心 Please update the LEP and address the comments.

enhancements/20260810-backup-store-s3-encryption.md
derekbit: As mentioned, please focus on client-side encryption in this implementation and create a server-side encryption  feature ticket.

derekbit (commented):
@carterli0407-cell To reduce the reviewers' burden and keep the LEP focused on client-side encryption, could you remove the server-side encryption from the LEP? I will start reviewing the LEP after re-organizing the LEP. Thanks.

Also, please create a FEATURE ticket for the server-side encryption separately and set the milestone to `Backlog`. Thanks.

enhancements/20260810-backup-store-s3-encryption.md
```diff
@@ -0,0 +1,2756 @@
...
+    }
+
+    if backupTargetPolicy != nil &&
+        backupTargetPolicy.State != BackupEncryptionStateInherit {
+        return *backupTargetPolicy
+    }
+
+    return globalPolicy
+}
+```
```
derekbit: The current design and BackupEncryptionStateInherit are a bit confusing and not easy to understand. Can we simplify the design by directly specifying the backupEncryptionSecret at each level?

For example, we could introduce backupEncryptionSecret to volume.spec, backuptarget.spec, and the global setting.

The resolution would then be straightforward:

- If volume.spec.backupEncryptionSecret is set, use it.
- Otherwise, if backuptarget.spec.backupEncryptionSecret is set, use it.
- Otherwise, use the global backupEncryptionSecret.

The backup controller can then use the key stored in the resolved Secret to encrypt the backup.

This would avoid introducing an explicit Inherit state and make the configuration and inheritance behavior easier to understand.

One thing we would need to consider is how to explicitly disable backup encryption at a lower level. If an empty backupEncryptionSecret means "inherit", we would need another way to distinguish "inherit" from "explicitly disabled".

For example, we could introduce an optional enabled field:

```
type BackupEncryptionSpec struct {
    Enabled *bool  `json:"enabled,omitempty"`
    Secret  string `json:"secret,omitempty"`
}
```
for both Volume.Spec and BackupTarget.Spec.


The semantics would be:

- enabled: nil -> inherit from the next level.
- enabled: true -> explicitly enable encryption and use the specified Secret.
- enabled: false -> explicitly disable encryption.

In this case, the resolution would still be simple: use the first level where enabled is explicitly set; otherwise, fall back to the next level and eventually the global setting.

This keeps the inheritance behavior simple while still allowing a Volume or BackupTarget to explicitly override the global setting and disable backup encryption when needed.

enhancements/20260810-backup-store-s3-encryption.md
```diff
@@ -0,0 +1,2756 @@
...
+
+Before a backup starts, Longhorn resolves the effective encryption policy.
+
+Conceptually:
+
+```go
+func resolveBackupEncryption(
+    volumePolicy *BackupEncryptionSpec,
+    backupTargetPolicy *BackupEncryptionSpec,
+    globalPolicy BackupEncryptionSpec,
```
derekbit: What is `globalPolicy BackupEncryptionSpec`? For a global setting, it is a setting like https://github.com/longhorn/longhorn-manager/blob/master/types/setting.go#L1025-L1034

enhancements/20260810-backup-store-s3-encryption.md
```diff
@@ -0,0 +1,2756 @@
...
+
+Existing Volume resources are interpreted as:
+
+```text
+state = inherit
+```
+
+---
+
+# Backup CRD Changes
```
derekbit: The Backup.Spec change is missing

enhancements/20260810-backup-store-s3-encryption.md
```diff
@@ -0,0 +1,2756 @@
...
+
+Conceptually:
+
+```go
+type BackupEncryptionStatus struct {
+    Enabled       bool   `json:"enabled,omitempty"`
+    FormatVersion string `json:"formatVersion,omitempty"`
+    Algorithm     string `json:"algorithm,omitempty"`
+    KeyProvider   string `json:"keyProvider,omitempty"`
+    KeyID         string `json:"keyID,omitempty"`
```
derekbit:
```suggestion
    Secret   string `json:"secret,omitempty"`
```

enhancements/20260810-backup-store-s3-encryption.md
```diff
@@ -0,0 +1,2756 @@
...
+The Backup CR needs to record the effective encryption information used when that backup was created.
+
+Current Volume, BackupTarget, or Global configuration must not be treated as the source of truth for an existing backup because those settings may change later.
+
+Conceptually:
+
+```go
+type BackupEncryptionStatus struct {
+    Enabled       bool   `json:"enabled,omitempty"`
+    FormatVersion string `json:"formatVersion,omitempty"`
```
derekbit: Why do we need Enabled and FormatVersion?
If is is encrypted, the Secret is non-empty, we don't need `Enabled`.

enhancements/20260810-backup-store-s3-encryption.md
```diff
@@ -0,0 +1,2756 @@
...
+
+When global encryption is enabled:
+
+* the key provider must be supported;
+* a Secret must be configured;
+* the Secret must contain valid key information.
+
+---
+
+## BackupTarget Validation
```
derekbit: CRD changes are missing

enhancements/20260810-backup-store-s3-encryption.md
```diff
@@ -0,0 +1,2756 @@
...
+```yaml
+spec:
+  backupEncryption:
+    state: disabled
+```
+
+If the field is omitted:
+
+```text
+state = inherit
```
derekbit: Can we remove `inherit` and only use `enabled` and `disabled`?

enhancements/20260810-backup-store-s3-encryption.md
```diff
@@ -0,0 +1,2756 @@
...
+
+## BackupTarget Validation
+
+Valid states:
+
+```text
+inherit
+disabled
+enabled
+```
```
derekbit: Remove `inherit`.
https://github.com/longhorn/longhorn/pull/13693/changes#diff-5f8dcc5584f8d26d9dc1faee200d934fe210599336837bc05f7abb224e72f588R494

derekbit (commented):
The LEP is quite lengthy and messy. For example, the CRD structures are mentioned in **Configuration Model** but not shown in the **API and CRD Changes**.

Could you consolidate and organize the LEP?

enhancements/20260810-backup-store-s3-encryption.md
```diff
@@ -0,0 +1,2756 @@
...
+* retries;
+* multipart uploads;
+* provider compatibility;
+* object operations.
+
+The encryption layer only transforms plaintext into ciphertext and back.
+
+---
+
+# S3-Compatible Providers
```
derekbit: We can remove `S3-Compatible Providers` because the client-side encryption is nothing to do with the storage providers.

enhancements/20260810-backup-store-s3-encryption.md
```diff
@@ -0,0 +1,2756 @@
...
+* corruption;
+* truncation;
+* incorrect keys;
+* invalid encrypted chunks.
+
+AES-CTR without an authentication mechanism is not sufficient.
+
+---
+
+# Why Streaming Encryption
```
derekbit: Can NFS and CIFS support it? We need to handle them as well. If they don't support Streaming Encryption, we can put the encrypted files in a temp directory and then upload them.

enhancements/20260810-backup-store-s3-encryption.md
```diff
@@ -0,0 +1,2756 @@
...
+---
+
+## Existing BackupTarget CRs
+
+Existing BackupTarget CRs do not contain the new encryption field.
+
+They behave as:
+
+```text
+state = inherit
```
derekbit: Remove `inherit`

enhancements/20260810-backup-store-s3-encryption.md
```diff
@@ -0,0 +1,2756 @@
...
+```
+
+---
+
+## Existing Volume CRs
+
+Existing Volume CRs also behave as:
+
+```text
+state = inherit
```
derekbit: Remove `inherit`

enhancements/20260810-backup-store-s3-encryption.md
```diff
@@ -0,0 +1,2756 @@
...
+---
+
+## Non-S3 BackupTarget
+
+Example:
+
+```text
+client-side backup encryption is currently supported only for S3 backup targets
+```
+
```
derekbit: client-side backup encryption should support ALL backup targets as we discussed. Why do we only support S3?

derekbit (commented):
Please remove `inherit` and follow the approach discussed here: https://github.com/longhorn/longhorn/pull/13693#discussion_r3790967900. This should make the design simpler. If there are any obstacles to removing `inherit`, we can discuss them.

enhancements/20260810-backup-store-s3-encryption.md
```diff
@@ -0,0 +1,2756 @@
...
+* missing Secret fails;
+* invalid Secret data fails;
+* missing key ID fails if required;
+* old key and new key are distinguishable;
+* key material never appears in errors;
+* key material never appears in logs.
+
+---
+
+# Unit Tests - S3 Integration
```
derekbit: As mentioned, the feature should support all backup targets, not only S3.

enhancements/20260810-backup-store-s3-encryption.md
```diff
@@ -0,0 +1,2756 @@
...
+* handle BackupTarget encryption configuration;
+* handle Volume encryption configuration;
+* resolve effective encryption policy using:
+
+```text
+Volume > BackupTarget > Global
+```
+
+* load the required Secret securely;
+* validate S3 compatibility;
```
derekbit:
```suggestion
* validate NFS/CIFS/S3 compatibility;
```

enhancements/20260810-backup-store-s3-encryption.md
carterli0407-cell: Earlier we discussed the use case of enabling encryption globally while allowing specific volumes to opt out. If that use case is no longer required, I think this simplified Secret-based configuration is better and I'll update the LEP accordingly.
Otherwise, we would still need a small explicit opt-out mechanism in addition to backupEncryptionSecret.
I'll reorganize the LEP around this simpler approach after confirming this point.

derekbit: We can keep it simple in the first version.

enhancements/20260810-backup-store-s3-encryption.md
carterli0407-cell: I'll remove this abstraction.

enhancements/20260810-backup-store-s3-encryption.md
carterli0407-cell: I would remove `BackupTarget.Spec.BackupEncryptionSecret` completely, remove `inherit/enabled/disabled`, and make client-side encryption work for all backup targets.

enhancements/20260810-backup-store-s3-encryption.md
carterli0407-cell: Good catch. I'll add the Backup.Spec change explicitly. Thanks.

enhancements/20260810-backup-store-s3-encryption.md
carterli0407-cell: Agree.

enhancements/20260810-backup-store-s3-encryption.md
carterli0407-cell: Agreed. I'll remove Enabled.

enhancements/20260810-backup-store-s3-encryption.md
carterli0407-cell: With the simplified first-version design, I'm going to remove BackupTarget-level encryption configuration entirely.

enhancements/20260810-backup-store-s3-encryption.md
carterli0407-cell: Yes. I'll remove inherit based on your above suggestion.

enhancements/20260810-backup-store-s3-encryption.md
carterli0407-cell: Agreed. I'll remove the S3-Compatible Providers section.

enhancements/20260810-backup-store-s3-encryption.md
carterli0407-cell: Yes, I'll extend the design to cover NFS and CIFS as well.
I'll include S3, NFS, and CIFS in the POC/test plan and verify which paths can remain fully streaming.

enhancements/20260810-backup-store-s3-encryption.md
carterli0407-cell: The S3-only restriction was left over from the earlier design.

I'll remove that restriction and make the encryption layer generic to the backupstore abstraction so client-side encryption applies to all supported backup targets, including S3, NFS, and CIFS.

enhancements/20260810-backup-store-s3-encryption.md
carterli0407-cell: I'll include S3, NFS, and CIFS paths and verify both encrypted backup and restore behavior.

enhancements/20260810-backup-store-s3-encryption.md
carterli0407-cell: I'll update this.

enhancements/20260810-backup-store-s3-encryption.md
derekbit: streaming is not really necessary in the first version. If streaming is not supported, as mentioned, we can encrypt the file, save it locally, and upload it to remote.

PR #13694, opened 2026-08-11
docs(lep): volume group snapshot

c3y1huang:

Issue longhorn/longhorn#13349

Add the LEP for volume group snapshot support.

`None`

`None`

### DIFF
--- a/enhancements/20260811-volume-group-snapshot.md
+++ b/enhancements/20260811-volume-group-snapshot.md
@@ -1,0 +1,517 @@
+# Volume Group Snapshot
+
+## Summary
+
+Many applications store state on more than one volume (a database with data and WAL on separate PVs, a sharded database across StatefulSets). Longhorn snapshots one volume at a time: nothing ties the per-volume snapshots together, at restore time nothing says which belong to the same attempt, and restoring them together can be inconsistent.
+
+This LEP adds volume group snapshot support: snapshot a set of volumes as one group, through CSI VolumeGroupSnapshot, kubectl, and the Longhorn UI.
+
+A new `SnapshotGroup` CRD and controller turn a group request into one normal `Snapshot` per member volume through the existing per-volume path; the group reports ready when every member snapshot is taken within a bounded deadline. The change is scoped to the control plane.
+
+Consistency boundary: Longhorn snapshots each volume independently, so a group snapshot is crash-consistent per volume and only loosely aligned across volumes. True cross-volume application consistency comes from pausing the application around the group snapshot - the separate snapshot-hook work in [#2128](https://github.com/longhorn/longhorn/issues/2128) that builds on top of this.
+
+### Related Issues
+
+- longhorn/longhorn#13349 (this feature)
+- longhorn/longhorn#2128 (application-consistent snapshot/backup; builds on this)
+- kubernetes/enhancements#3476 (upstream VolumeGroupSnapshot KEP)
+
+## Motivation
+
+### Goals
+
+- Implement the CSI VolumeGroupSnapshot API (GroupController service) for in-cluster group snapshots (class `type: snap`) and backup-type group snapshots (class `type: bak`): the same group snapshot, followed by a per-member upload through the existing backup path.
+- Offer the same grouping through the Longhorn UI, backed by new REST endpoints: create, inspect, and delete snapshot groups, and make group membership visible on the volume's snapshot list.
+- Reuse the existing per-volume snapshot path unchanged.
+- Report status that cannot be misread: an explicit phase (InProgress, Ready, Failed), per-member snapshot creation times, and a completion deadline that forces an unfinished group to Failed. A group that failed or snapshotted only some members never passes for a consistent one.
+- Provide the foundation #2128 builds on: check whether group snapshot is available, request a group snapshot, learn when it finishes, and verify every member snapshot was taken inside the pause window.
+
+### Non-goals
+
+- Application-level consistency (pausing the application). Future work in #2128.
+- A dedicated group-backup object (a `BackupGroup` CRD) and scheduled group backups. Consistency is decided when the snapshots are taken, not when they are uploaded, so no group-level backup machinery is needed; CSI `type: bak` groups reuse the existing per-volume backup path.
+- Recurring group snapshots: a RecurringJob that creates a `SnapshotGroup` on a schedule. Follow-up work that needs its own design. Existing per-volume recurring jobs are unaffected, except the `snapshot-delete` task: its retain count applies to every snapshot on the volume, so it can age out a member and degrade the group like any other out-of-band member deletion. This is deliberate. `snapshot-delete` enforces a hard retention limit on the volume; skipping group members would let a forgotten group hold snapshots until the volume reaches `snapshot-max-count` and new snapshots start failing. A visibly Degraded group is the safer failure. Future work if protection is ever wanted.
+
+## Proposal
+
+One layer above the existing per-volume Snapshot CR path. Nothing below it changes:
+
+```
+CSI VolumeGroupSnapshot | UI | kubectl
+        |
+        |  every entry point creates the same CR
+        v
+SnapshotGroup CR
+        |
+        |  the group controller creates one Snapshot CR per member
+        v
+member Snapshot CRs
+  (the existing per-volume path, unchanged)
+        |
+        |  the group controller mirrors each member's status into the group
+        v
+group status
+  (phase: InProgress -> Ready | Failed)
+```
+
+### User Stories
+
+#### Story 1: snapshot a multi-volume application
+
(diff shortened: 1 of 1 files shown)


enhancements/20260811-volume-group-snapshot.md
```diff
@@ -0,0 +1,511 @@
...
+| Backup target unreachable during upload (`bak`) | Uploads stall and the group never reports ready. No deadline applies, because the deadline bounds only taking the snapshots; the behavior matches a stalled per-volume backup. |
+| Backup target unreachable during delete (`bak`) | Backup deletion blocks and retries per existing per-volume semantics, for every member; the group CR is not deleted until the labeled backups are gone, so retried Delete RPCs pick up the remaining backups. |
+| Plugin restart between the group snapshot and upload completion (`bak`) | Nothing is lost, because the plugin keeps no state of its own: the handle records the group type, a retried Create creates only the member Backup CRs that are still missing, and readiness is recomputed by listing the group-labeled backups. |
+| `GetVolumeGroupSnapshot` on a `bak` group before uploads complete | `ready_to_use` is computed live from the member backups: false while any upload is in flight, true once every member backup completes, even before a Create retry writes the backups-completed annotation. Once the annotation is written, it stays true. |
+
+### Test plan
```
derekbit, line 490: Should we also add a data consistency test to verify that all member snapshots in a VolumeGroupSnapshot represent the same point-in-time state?

c3y1huang: The fan-out design doesn't promise the same point-in-time across members. This means the snapshots are taken close together, not at the exact same time (covered in the Notes section), so strict point-in-time would fail by design. Cross-volume consistency comes from #2128 (pausing the application around the group snapshot).

What I think we can test: pause writes around the group snapshot, verify all members restore to the same application state, or just simply verify every member creation time falls inside the pause window. 

Added this to the plan.

enhancements/20260811-volume-group-snapshot.md
```diff
@@ -0,0 +1,515 @@
...
+
+## Summary
+
+Many applications store state on more than one volume (a database with data and WAL on separate PVs, a sharded database across StatefulSets). Longhorn snapshots one volume at a time: nothing ties the per-volume snapshots together, at restore time nothing says which belong to the same attempt, and restoring them together can be inconsistent.
+
+This LEP adds volume group snapshot support: snapshot a set of volumes as one group, through CSI VolumeGroupSnapshot, kubectl, and the Longhorn UI.
+
+A new `SnapshotGroup` CRD and controller turn a group request into one normal `Snapshot` per member volume through the existing per-volume path; the group reports ready when every member snapshot is taken within a bounded deadline. The change is scoped to the control plane.
+
+Consistency boundary: Longhorn snapshots each volume independently, so a group snapshot is crash-consistent per volume and only loosely aligned across volumes. True cross-volume application consistency comes from pausing the application around the group snapshot - the separate snapshot-hook work in [#2128](https://github.com/longhorn/longhorn/issues/2128) that builds on top of this.
```
derekbit, line 11: This is important because users might assume that snapshots in a snapshot group are strictly aligned. We can highlight this in the official documentation.

enhancements/20260811-volume-group-snapshot.md
```diff
@@ -0,0 +1,515 @@
...
+
+### Non-goals
+
+- Application-level consistency (pausing the application). Future work in #2128.
+- A dedicated group-backup object (a `BackupGroup` CRD) and scheduled group backups. Consistency is decided when the snapshots are taken, not when they are uploaded, so no group-level backup machinery is needed; CSI `type: bak` groups reuse the existing per-volume backup path.
+- Recurring group snapshots: a RecurringJob that creates a `SnapshotGroup` on a schedule. Follow-up work that needs its own design. Existing per-volume recurring jobs are unaffected, except the `snapshot-delete` task: its retain count applies to every snapshot on the volume, so it can age out a member and degrade the group like any other out-of-band member deletion. This is deliberate. `snapshot-delete` enforces a hard retention limit on the volume; skipping group members would let a forgotten group hold snapshots until the volume reaches `snapshot-max-count` and new snapshots start failing. A visibly Degraded group is the safer failure. Future work if protection is ever wanted.
```
derekbit, line 33: Can you create a FEATURE ticket for the recurring group snapshots?

enhancements/20260811-volume-group-snapshot.md
```diff
@@ -0,0 +1,515 @@
...
+
+An operator wants backups that a database can actually recover from. With #2128, a pre-hook pauses the application, this feature snapshots all member volumes, and a post-hook resumes it.
+
+#2128 needs from this LEP only: a way to detect group support, a way to request a group snapshot of a fixed set, a signal when it finished or failed, and per-member creation times to verify every member snapshot was taken inside the pause window.
+
+### User Experience In Detail
+
+Prerequisites (documentation only):
+
+- The group CRDs installed (`VolumeGroupSnapshot`, `VolumeGroupSnapshotContent`, `VolumeGroupSnapshotClass`) - and installed before any component's feature gate is enabled anywhere.
```
derekbit: Are there any requirements for the Kubernetes version or the versions of the CSI sidecar components?

enhancements/20260811-volume-group-snapshot.md
```diff
@@ -0,0 +1,515 @@
...
+- snapshot-controller from an [external-snapshotter](https://github.com/kubernetes-csi/external-snapshotter/releases) release with group snapshot support, with the `CSIVolumeGroupSnapshot` feature gate enabled. It verifies the CRDs at startup and crash-loops loudly when they are missing - a visible, operator-fixable failure.
+- The Longhorn deploy-time toggle for CSI group snapshot support enabled (default off).
```
derekbit: Could we enable it by default?

enhancements/20260811-volume-group-snapshot.md
derekbit: You are right!

enhancements/20260811-volume-group-snapshot.md
```diff
@@ -0,0 +1,515 @@
...
+
+  | Method | Path | Behavior |
+  |---|---|---|
+  | `GET` | `/v1/snapshotgroups` | List groups. |
+  | `GET` | `/v1/snapshotgroups/{name}` | Get one group. |
+  | `POST` | `/v1/snapshotgroups` | Create a group. Fields: `name`, exactly one of `volumes` / `volumeSelector`, optional `labels`, optional `deadlineSeconds`. |
+  | `DELETE` | `/v1/snapshotgroups/{name}` | Delete the group and its members - the same path kubectl deletion takes. |
+  | `POST` | `/v1/snapshotgroups?action=preview` | Preview a create without creating anything: runs the same resolver admission uses and returns the volumes the spec would select and, when a volume cannot be a member, the reason why. Failures that would reject the create are reported in the response body, not as HTTP errors. |
+  | `GET` | `/v1/ws/snapshotgroups`, `/v1/ws/{period}/snapshotgroups` | Watch stream. Both forms are registered like every existing resource stream; the UI uses the `{period}` form (for example `/v1/ws/1s/snapshotgroups`). |
```
derekbit, line 234: nit: How about using `dryrun` instead?

derekbit (commented):
In general, LGTM. Only few comments and questions.

enhancements/20260811-volume-group-snapshot.md
c3y1huang: Prefer to keep as `preview`.

I think `dryrun` over-promises a bit: a dry run usually means running the full create path without persisting, with the same rejection behavior as the real request.

This action is intentionally different. It takes only the membership field and returns the member sets with volume ineligible reasons, so the UI can render for live preview.

enhancements/20260811-volume-group-snapshot.md
c3y1huang: IMO we should keep it off by default. Longhorn doesn't install the VolumeGroupSnapshot CRDs, so with default-enabled, a fresh install or upgrade w/o the CRDs would fail the deploy. Off by default keeps this opt-in, same as upstream - the csi snapshotter [feature gate is still off](https://github.com/kubernetes-csi/external-snapshotter/blob/v8.6.0/pkg/features/features.go#L42). We can revisit when upstream flips the default.

enhancements/20260811-volume-group-snapshot.md
c3y1huang: The requirement is an external snapshotter >v8.6.0; this is the version group snapshot API GA's to v1, which the snapshotter sidecar consumes. No new K8s requirement because the API is CRD-based, and v8.6.0 needs only K8s 1.25. 

Updated the LEP.

enhancements/20260811-volume-group-snapshot.md
c3y1huang: https://github.com/longhorn/longhorn/issues/13821

derekbit (approved):

Issue #12976, opened 2026-04-25
[FEATURE] Allow configuring Gateway API filters


98jan:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

Hi,

with ingress annotations, it is possible to configure a basic auth or oauth2-proxy integration, to secure the Longhorn UI.
With Gateway API it isn't possible to configure, therefore I would propose to add support for configuring rules filters.

### Describe the solution you'd like

For cilium hubble UI I used following config:
  
```YAML
parentRefs:
    - name: traefik-gateway
      sectionName: websecure
      namespace: traefik
  hostnames:
    - hubble.${DOMAIN}
  rules:
    - backendRefs:
        - name: hubble-ui
          namespace: cilium
          port: 80
      filters:
        - type: ExtensionRef
          extensionRef:
            group: traefik.io
            kind: Middleware
            name: oauth-forwardauth
        - type: ExtensionRef
          extensionRef:
            group: traefik.io
            kind: Middleware
            name: oauth-forwardauth-service
```

### Describe alternatives you've considered

Using ingress configuration for now.

COLDTURNIP:
It is possible to insert a configurable block in the [httproute.yaml](https://github.com/longhorn/longhorn/blob/master/chart/templates/httproute.yaml):

```yaml
rules:
      - matches:
          - path:
              type: {{ .Values.httproute.pathType | default "PathPrefix" }}
              value: {{ .Values.httproute.path | default "/" }}

        {{- with .Values.httproute.filters }}
        filters:
          {{- toYaml . | nindent 10 }}
        {{- end }}

        backendRefs:
          - name: longhorn-frontend
            port: 80
```

So that user can list the filters in the values.yaml:

```yaml
httproute:
  enabled: true
  filters:
    - type: ExtensionRef
      extensionRef:
        group: traefik.io
        kind: Middleware
        name: oauth-forwardauth
    - type: ExtensionRef
      extensionRef:
        group: traefik.io
        kind: Middleware
        name: oauth-forwardauth-service
```

derekbit:
@apoorvajagtap Could you update https://github.com/longhorn/longhorn/issues/12976#issuecomment-5302898791 and add the steps to validation and update the status to `Ready For Testing`? QA will help update the feature. Thank you.

derekbit:
BTW, remember to update the Sprint. Thank you.

[...]

PR #13708, opened 2026-08-12
docs(lep): gateway api filters

apoorvajagtap:
Issue #12976 

Add the LEP for Gateway API Filters support.

`None`

`None`

### DIFF
--- a/enhancements/20260812-gateway-api-filters.md
+++ b/enhancements/20260812-gateway-api-filters.md
@@ -1,0 +1,71 @@
+# Gateway API Filters
+
+## Summary
+
+This enhancement adds a `httproute.filters` field to the Helm chart, rendered onto the HTTPRoute's rule, so users can secure the Longhorn UI behind Gateway-API-native auth/middleware mechanisms.
+
+### Related Issues
+
+https://github.com/longhorn/longhorn/issues/12976
+
+## Motivation
+
+### Goals
+
+- Allow users installing Longhorn via the Helm chart with `httproute.enabled=true` to attach one or more `HTTPRouteFilter` entries to the generated route rule.
+- Preserve backward compatibility: existing httproute.* values and generated manifests must be unaffected when filters is unset.
+- Document the new `httproute.filters` value in the Helm chart configuration reference, including its type, default, and description.
+
+### Non-goals [optional]
+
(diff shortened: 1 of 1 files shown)


enhancements/20260812-gateway-api-filters.md
```diff
@@ -0,0 +1,71 @@
+# Gateway API Filters
+
+## Summary
+
+This enhancemenť adds a `httproute.filters` field to the Helm chart, rendered onto the HTTPRoute's rule, so users can secure the Longhorn UI behind Gateway-API-native auth/middleware mechanisms.
```
COLDTURNIP: Non-ASCii character `ť` (U+0165) here.

```suggestion
This enhancement adds a `httproute.filters` field to the Helm chart, rendered onto the HTTPRoute's rule, so users can secure the Longhorn UI behind Gateway-API-native auth/middleware mechanisms.
```

enhancements/20260812-gateway-api-filters.md
```diff
@@ -0,0 +1,71 @@
...
+  httproute:
+     enabled: true
+     filters:
+       - type: ExtensionRef
+         extensionRef:
+           group: traefik.io
+           kind: Middleware
+           name: oauth-forwardauth
+```
+4. Run `helm upgrade ...`. The chart renders filters onto `spec.rules[0].filters` of the generated HTTPRoute.
```
derekbit, line 53: We render filters onto the `spec.rules[0].filters`. When will we render the filters to multiple rules?

derekbit (commented):
In general, LGTM. Only one question. Good job @apoorvajagtap.

enhancements/20260812-gateway-api-filters.md
apoorvajagtap: IIUC, right now the Longhorn chart generates single rule via [httproute.yaml](https://github.com/longhorn/longhorn/blob/master/chart/templates/httproute.yaml#L35), so might not need to render the filters to multiple rules. 
If we plan to extend this to multiple rules, then `filters` needs to be rendered with per rule approach.
For that, I believe we need to restructure the current [values.yaml](https://github.com/longhorn/longhorn/blob/master/chart/values.yaml#L647) for fields like `path` and `pathType` as well.
Let me know in case I am missing anything.

derekbit: No worries. It is fine that there is only one rule for now.

derekbit:
The LEP looks LGTM. @apoorvajagtap https://github.com/longhorn/longhorn/pull/13719 is still a draft. Could we start reviewing it?

derekbit (approved):

PR #13719, opened 2026-08-13
feat(chart): Gateway API Filters support

apoorvajagtap:
Issue #12976 

Adds `httproute.filters` field in chart template to allow configuring filters that mirrors the upstream Gateway API `HTTPRouteFilter` schema.

- https://github.com/longhorn/longhorn/pull/13708

### DIFF
--- a/chart/README.md
+++ b/chart/README.md
@@ -261,14 +261,15 @@ Longhorn consists of user-deployed components (for example, Longhorn Manager, Lo
 
 ### HTTPRoute Settings
 
-| Key | Type | Default | Description |
-|-----|------|---------|-------------|
-| httproute.annotations | object | `{}` | Annotations for the HTTPRoute resource in the form of key-value pairs. |
-| httproute.enabled | bool | `false` | Setting that allows Longhorn to generate HTTPRoute records for the Longhorn UI service using Gateway API. |
-| httproute.hostnames | list | `[]` | List of hostnames for the HTTPRoute. Multiple hostnames are supported. |
-| httproute.parentRefs | list | `[]` | Gateway references for HTTPRoute. Specify which Gateway(s) should handle this route. |
-| httproute.path | string | `"/"` | Default path for HTTPRoute. You can access the Longhorn UI by following the full path. |
-| httproute.pathType | string | `"PathPrefix"` | Path match type for HTTPRoute. (Options: "Exact", "PathPrefix") |
+| Key                   | Type | Default | Description                                                                                          |
+|-----------------------|------|---------|------------------------------------------------------------------------------------------------------|
+| httproute.annotations | object | `{}` | Key-value annotations to apply to the HTTPRoute resource.                                            |
+| httproute.enabled     | bool | `false` | When `true`, Longhorn generates HTTPRoute records for the Longhorn UI service using the Gateway API. |
+| httproute.hostnames   | list | `[]` | List of hostnames bound to the HTTPRoute.                                                            |
+| httproute.parentRefs  | list | `[]` | List of Gateway references specifying which Gateway(s) should handle this route.                     |
+| httproute.filters     | list | `[]` | List of Gateway API filters to attach to the HTTPRoute rule for the Longhorn UI service.             |
+| httproute.path        | string | `"/"` | The routing path used to access the Longhorn UI.                                                     |
+| httproute.pathType    | string | `"PathPrefix"` | The path match type for the HTTPRoute. Valid options are `"Exact"` or `"PathPrefix"`.                |
 
 ### Private Registry Settings
 
--- a/chart/questions.yaml
+++ b/chart/questions.yaml
@@ -1320,6 +1320,13 @@ questions:
         type: string
         required: true
         label: Hostnames (JSON array)
+      - variable: httproute.filters
+        default: '[]'
+        description: >-
+          Gateway API filters (such as authentication middleware) to attach to the HTTPRoute rule for the Longhorn UI service. It is a JSON array of HTTPRouteFilter objects. Example: '[{"type":"ExtensionRef","extensionRef":{"group":"traefik.io","kind":"Middleware","name":"oauth-forwardauth"}}]'
+        type: string
+        required: false
+        label: Filters (JSON array)
       - variable: httproute.path
         default: /
         description: >-
--- a/chart/templates/httproute.yaml
+++ b/chart/templates/httproute.yaml
@@ -1,4 +1,19 @@
 {{- if .Values.httproute.enabled -}}
+  {{- $filters := .Values.httproute.filters -}}
+  {{- if $filters }}
+    {{- if not (kindIs "slice" $filters) }}
+      {{- fail "httproute.filters must be a YAML list of HTTPRouteFilter objects" }}
+    {{- end }}
+    {{- range $index, $filter := $filters }}
+      {{- if not (kindIs "map" $filter) }}
+        {{- fail (printf "httproute.filters[%d] must be an HTTPRouteFilter object" $index) }}
+      {{- end }}
+      {{- $type := get $filter "type" }}
+      {{- if or (not (kindIs "string" $type)) (empty $type) }}
+        {{- fail (printf "httproute.filters[%d].type must be a non-empty string" $index) }}
+      {{- end }}
+    {{- end }}
+  {{- end }}
 apiVersion: gateway.networking.k8s.io/v1
 kind: HTTPRoute
 metadata:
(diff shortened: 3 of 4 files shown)


chart/values.yaml
```diff
@@ -644,9 +644,11 @@ ingress:
   #   certificate:
 httproute:
   # -- Setting that allows Longhorn to generate HTTPRoute records for the Longhorn UI service using Gateway API.
-  enabled: false
+  enabled: true
```
derekbit: How about keeping it `false` by default?

chart/values.yaml
```diff
@@ -644,9 +644,11 @@ ingress:
...
   # -- Gateway references for HTTPRoute. Specify which Gateway(s) should handle this route.
-  parentRefs: []
+  parentRefs:
+    - name: longhorn-gateway
+      namespace: longhorn-system
```
derekbit: Can we keep it empty? User can update it by itself.

chart/values.yaml
apoorvajagtap: Yes, I was testing it and missed changing the value back to false. Thanks!

chart/templates/httproute.yaml
```diff
@@ -36,6 +36,10 @@ spec:
     - path:
         type: {{ .Values.httproute.pathType | default "PathPrefix" }}
         value: {{ .Values.httproute.path | default "/" }}
+    {{- with .Values.httproute.filters }}
+    filters:
+      {{- toYaml . | nindent 10 }}
+      {{- end }}
```
COLDTURNIP: The parity `end` should be at the same indent level.
```suggestion
    {{- with .Values.httproute.filters }}
    filters:
      {{- toYaml . | nindent 10 }}
    {{- end }}
```

COLDTURNIP:
Hi @sushant-suse , could you help review README and the description in chart/questions.yaml?

chart/values.yaml
```diff
@@ -661,6 +661,19 @@ httproute:
   ## Example:
   # - longhorn.example.com
   # - longhorn.example.org
+  # -- allows configuring filters.
+  filters:
```
COLDTURNIP: This expression declares the default value as `null` instead of empty list.
```suggestion
  filters: []
```

chart/questions.yaml
```diff
@@ -1320,6 +1320,13 @@ questions:
         type: string
         required: true
         label: Hostnames (JSON array)
+      - variable: httproute.filters
+        default: '[]'
+        description: >-
+          Gateway API filters (such as authentication middleware) to attach to the HTTPRoute rule for the Longhorn UI service.
+        type: string
```
COLDTURNIP, line 1327: Since the value is a string of object list, we need to provide the example and the type information of the items.
```suggestion
        default: '[]'
        description: >-
          Gateway API filters (such as authentication middleware) to attach to the HTTPRoute rule for the Longhorn UI service. It is a JSON array of HTTPRouteFilter objects. Example: '[{"type":"ExtensionRef","extensionRef":{"group":"traefik.io","kind":"Middleware","name":"oauth-forwardauth"}}]'
        type: string
```

chart/values.yaml
```diff
@@ -661,6 +661,19 @@ httproute:
   ## Example:
   # - longhorn.example.com
   # - longhorn.example.org
+  # -- allows configuring filters.
```
COLDTURNIP:
```suggestion
  # -- List of allowed HTTPRouteFilter rules.
```

chart/templates/httproute.yaml
COLDTURNIP: In addition, is there a way to ensure the rendered items are valid for `filters`?

chart/README.md
```diff
@@ -261,14 +261,15 @@ Longhorn consists of user-deployed components (for example, Longhorn Manager, Lo
...
+| httproute.annotations | object | `{}` | Annotations for the HTTPRoute resource in the form of key-value pairs.                                   |
+| httproute.enabled     | bool | `false` | Setting that allows Longhorn to generate HTTPRoute records for the Longhorn UI service using Gateway API. |
+| httproute.hostnames   | list | `[]` | List of hostnames for the HTTPRoute. Multiple hostnames are supported.                                   |
+| httproute.parentRefs  | list | `[]` | Gateway references for HTTPRoute. Specify which Gateway(s) should handle this route.                     |
+| httproute.filters     | list | `[]` | List of Gateway API filters to attach to the HTTPRoute rule for the Longhorn UI service.                 |
+| httproute.path        | string | `"/"` | Default path for HTTPRoute. You can access the Longhorn UI by following the full path.                   |
+| httproute.pathType    | string | `"PathPrefix"` | Path match type for HTTPRoute. (Options: "Exact", "PathPrefix")                                          |
```
sushant-suse:
```suggestion
|---|---|---|---|
| httproute.annotations | object | `{}` | Key-value annotations to apply to the HTTPRoute resource. |
| httproute.enabled | bool | `false` | When `true`, Longhorn generates HTTPRoute records for the Longhorn UI service using the Gateway API. |
| httproute.hostnames | list | `[]` | List of hostnames bound to the HTTPRoute. |
| httproute.parentRefs | list | `[]` | List of Gateway references specifying which Gateway(s) should handle this route. |
| httproute.filters | list | `[]` | List of Gateway API filters to attach to the HTTPRoute rule for the Longhorn UI service. |
| httproute.path | string | `"/"` | The routing path used to access the Longhorn UI. |
| httproute.pathType | string | `"PathPrefix"` | The path match type for the HTTPRoute. Valid options are `"Exact"` or `"PathPrefix"`. |
```

sushant-suse (changes requested):
Some suggestions for your consideration.
Thanks.

chart/templates/httproute.yaml
```diff
@@ -36,6 +36,15 @@ spec:
     - path:
         type: {{ .Values.httproute.pathType | default "PathPrefix" }}
         value: {{ .Values.httproute.path | default "/" }}
+    {{- with .Values.httproute.filters }}
+    filters:
+      {{- range . }}
+        {{- if not (has .type (list "RequestHeaderModifier" "ResponseHeaderModifier" "RequestRedirect" "URLRewrite" "RequestMirror" "ExtensionRef" "CORS" "ExternalAuth")) }}
```
apoorvajagtap: @COLDTURNIP I have added this additional check to confirm the types. If we want to also add a check for if the relevant block is passed correctly or not in the values.yaml, something like:
```
{{- if and (eq .type "ExtensionRef") (not .extensionRef) }}
  {{- fail "httproute.filters: type ExtensionRef requires an 'extensionRef' block with group, kind, and name" }}
{{- end }}
```
Let me know, we might need to just maintain a map with each type's block string, and then add this check for all the available types for filters.

sushant-suse (dismissed):

chart/templates/httproute.yaml
apoorvajagtap: I have added the map for an efficient check. Let me know if we prefer this way or not. Thanks!

COLDTURNIP: Listing the available enum values is too narrow. In this way, we have to always keep an eye on the upstream K8s API change and maintain this list. Suggest to simply check the desired JSON object type, then leave the content checking to K8s API server during resource applying.

At the beginning of this YAML template, we can define some type checking logic like:

```yaml
{{- if .Values.httproute.enabled -}}
{{- $filters := .Values.httproute.filters -}}
{{- if $filters }}
  {{- if not (kindIs "slice" $filters) }}
    {{- fail "httproute.filters must be a YAML list of HTTPRouteFilter objects" }}
  {{- end }}
  {{- range $index, $filter := $filters }}
    {{- if not (kindIs "map" $filter) }}
      {{- fail (printf "httproute.filters[%d] must be an HTTPRouteFilter object" $index) }}
    {{- end }}
    {{- $type := get $filter "type" }}
    {{- if or (not (kindIs "string" $type)) (empty $type) }}
      {{- fail (printf "httproute.filters[%d].type must be a non-empty string" $index) }}
    {{- end }}
  {{- end }}
{{- end }}
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
```

Later we can safely render the list under `filters`:

```yaml
  rules:
  - ...
    {{- with $filters }}
    filters:
      {{- toYaml . | nindent 6 }}
    {{- end }}
    ...
```

- If the input value is not a valid list of objects, or any object does not contain `type` information, then the Helm will reject to render the YAML before applying change to the cluster, and user will be aware to check Helm value `httproute.filters`
- While the YAML is rendered but the content contains invalid value (e.g. `type: UnknownType`), this change will be still blocked by K8s API server.

Note: Longhorn Helm chart need to keep compatibility with Helm 3. The utility functions such as `kindIs` are available under this criteria.

apoorvajagtap: Makes sense. Thanks for pointing that out!
I have made the changes & tested the configuration via traefik middleware locally as well. Now, longhorn UI pops up for the credentials when filter is configured. PTAL

COLDTURNIP (approved):

derekbit (approved):
LGTM. Thanks @apoorvajagtap

Issue #13802, opened 2026-08-20
[BUG] v1.12.1 default NetworkPolicies miss legitimate non-pod-labeled and host-originated callers, silently breaking CSI attach/detach


steled:
### Describe the Bug

After upgrading from v1.12.0 to v1.12.1, volume attach/detach started failing across the cluster (`FailedAttachVolume`, engines stuck in `starting`), even though `longhorn-manager` pods were `Running`/`Ready` and internally healthy. We tracked this down to two separate gaps in the new default per-component `NetworkPolicy` objects introduced in v1.12.1, confirmed live via our CNI's (Cilium) policy verdict logs.

**Gap 1: external CSI sidecar Deployments aren't in `longhorn-manager`'s peer allow-list**

The `longhorn-manager` NetworkPolicy's ingress `from` list only includes:

```yaml
- podSelector: {matchLabels: {app: longhorn-manager}}
- podSelector: {matchLabels: {app: longhorn-ui}}
- podSelector: {matchLabels: {app: longhorn-csi-plugin}}
- podSelector: {matchExpressions: [{key: recurring-job.longhorn.io, operator: Exists}], matchLabels: {longhorn.io/managed-by: longhorn-manager}}
- podSelector: {matchExpressions: [{key: longhorn.io/job-task, operator: Exists}]}
- podSelector: {matchLabels: {app: longhorn-driver-deployer}}
```

This does not include the external CSI sidecar Deployments — `csi-attacher`, `csi-provisioner`, `csi-resizer`, `csi-snapshotter` — which are separate Deployments from `longhorn-csi-plugin` and call the manager's REST API (port 9500) directly during attach/detach orchestration. In any environment where a NetworkPolicy provider actually enforces this list, those sidecars can't reach the manager and volume operations they mediate silently fail or hang.

**Gap 2 (more serious): the node's own kernel iSCSI initiator isn't in `instance-manager`'s peer allow-list**

The `instance-manager` NetworkPolicy's ingress `from` list only includes other pods:

```yaml
- podSelector: {matchLabels: {app: longhorn-manager}}
- podSelector: {matchLabels: {longhorn.io/component: instance-manager}}
- podSelector: {matchLabels: {longhorn.io/component: backing-image-manager}}
- podSelector: {matchLabels: {longhorn.io/component: backing-image-data-source}}
```

[...]

PR #13807, opened 2026-08-21
fix(chart): preserve storage traffic with network policies

COLDTURNIP:

Issue longhorn/longhorn#13802

### DIFF
--- a/chart/README.md
+++ b/chart/README.md
@@ -91,8 +91,10 @@ The `values.yaml` contains items used to tweak a deployment of this chart.
 |-----|------|---------|-------------|
 | networkPolicies.enabled | bool | `false` | Setting that allows you to enable network policies that control access to Longhorn pods. |
 | networkPolicies.kubeAPIServerSourceCIDRs | list | `[]` | Kubernetes API server source CIDRs allowed to access the Longhorn admission webhook. When empty, source filtering is disabled. |
+| networkPolicies.recoveryBackendAdditionalIngressPorts | list | `[]` | Additional TCP ports used by recovery backend mesh transport. |
 | networkPolicies.restrictInternalTraffic | bool | `true` | Setting that allows you to enable network policies for internal Longhorn components. When enabled, only authorized Longhorn components are allowed to communicate with each other. |
 | networkPolicies.type | string | `"k3s"` | Distribution that determines the policy for allowing access for an ingress. (Options: "k3s", "rke2", "rke1") |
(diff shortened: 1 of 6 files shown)


derekbit (approved):
LGTM. Thanks @COLDTURNIP

derekbit:
@mergify backport v1.12.x

steled:
Validated this against our exact reported scenarios by rendering the templates at the merge commit (`494c137e`) with `helm template` (no cluster changes, pure template rendering):

```
helm template longhorn ./chart \
  --set networkPolicies.enabled=true \
  --set networkPolicies.restrictInternalTraffic=true \
  --set 'networkPolicies.v1DataEngineInitiatorSourceCIDRs[0]=169.254.0.0/16' \
  --set 'networkPolicies.recoveryBackendAdditionalIngressPorts[0]=15008' \
  --show-only templates/network-policies/instance-manager-networking.yaml \
  --show-only templates/network-policies/recovery-backend-network-policy.yaml
```

- `instance-manager-networking.yaml` renders a TCP/3260 ingress rule scoped to `169.254.0.0/16` — matches our own workaround for gap 2 exactly.
- `recovery-backend-network-policy.yaml` renders the `share-manager` peer allowed on both 9503 and 15008 — covers gap 3 (Istio Ambient HBONE), and more precisely scoped than our own workaround (which had opened all ports for that peer).

Both confirmed to address the reported symptoms once the corresponding values are set.

One suggestion on the default: with `v1DataEngineInitiatorSourceCIDRs` left empty (the default), the new port-3260 rule has no `from` at all, i.e. it allows *any* source, not just host-originated/link-local traffic. That's broader than the actual need (the node's own kernel iSCSI initiator, reachable via `169.254.0.0/16` under Cilium's host-routing) — might be worth defaulting `restrictInternalTraffic` users toward the link-local range, or at least calling it out in the docs, so the "internal traffic restriction" feature doesn't quietly gain an allow-from-anywhere hole on 3260.

Also confirmed @COLDTURNIP's correction: our original gap-1 write-up (CSI sidecars needing a direct NetworkPolicy allow into `longhorn-manager`) was based on a wrong assumption about the call path — they go through the Unix socket to `longhorn-csi-plugin`, not the network. Our `networkpolicy-manager-csi-sidecars.yaml` workaround was very likely addressing a non-issue; the real root cause for that symptom was gap 2 all along. No action needed here, just confirming for the record.

Issue #13814, opened 2026-08-22
[IMPROVEMENT][Helm] Allow setting spec.trafficDistribution


mydoomfr:
### Is your improvement request related to a feature? Please describe (👍 if you like this request)

`longhorn-backend`, `longhorn-admission-webhook` and `longhorn-recovery-backend` are backed by the
`longhorn-manager` DaemonSet, so they have one endpoint per node and callers are routed to any of them.

On my cluster, 4 nodes are local and 1 is remote behind a WAN link (pod-to-pod RTT 18.6 ms vs 1.3 ms).
So the CSI plugin on the remote node reaches a manager in the other zone 3 times out of 4, and the
apiserver validates 1 Longhorn CR write in 4 against the remote admission webhook. That second one is
the real problem: a webhook timeout fails the write, it does not just slow it down.

`spec.trafficDistribution` exists for this, but the chart hardcodes the `spec` of these Services. Only
`service.manager.{type,nodePort}`, `service.ui.*` and `longhornManager.service{Labels,Annotations}` are
exposed; the webhook and recovery backend Services have no values at all.

### Describe the solution you'd like

A `trafficDistribution` value per Service, defaulting to empty so the field is omitted and nothing
changes for existing users:

```yaml
service:
  ui: { trafficDistribution: "" }
  manager: { trafficDistribution: "" }
  admissionWebhook: { trafficDistribution: "" }
  recoveryBackend: { trafficDistribution: "" }
```

### Describe alternatives you've considered

A Helm post-renderer, which is what I run today. It works, but every affected user has to reimplement
it. `internalTrafficPolicy: Local` is not an option here since it has no fallback, whereas
`PreferSameZone` degrades to current behaviour when the local zone has no endpoint.

derekbit:
cc @COLDTURNIP @c3y1huang

derekbit:
Hello @mydoomfr The PR is merged. Could you help update https://github.com/longhorn/longhorn/issues/13814#issuecomment-5411508277? Thank you.

yangchiu:
Verification passed on master-head (longhorn/longhorn 53b612b).

For each service, setting `trafficDistribution` to `PreferSameZone` during Longhorn installation is correctly reflected in the deployed services:

```
.service.ui.trafficDistribution = "PreferSameZone"
.service.manager.trafficDistribution = "PreferSameZone"
.service.admissionWebhook.trafficDistribution = "PreferSameZone"
[...]

PR #13815, opened 2026-08-22
feat(chart): expose trafficDistribution for the Longhorn Services

mydoomfr:

Issue #13814 

Adds `service.{ui,manager,admissionWebhook,recoveryBackend}.trafficDistribution`, mapping to `longhorn-frontend`, `longhorn-backend`, `longhorn-admission-webhook` and `longhorn-recovery-backend`.

Each value defaults to `""` and renders through `{{- with }}`, so the field is omitted unless set.

`helm template` output is byte-identical to `master` with default values, and with a real production `values.yaml`. Setting one value leaves the other three Services untouched; checked for all four.

No validation in the chart, on purpose: the API server already rejects bad values and lists the supported ones, and stays correct as Kubernetes adds more.
Same as `service.ui.type` today. Happy to add a `fail` guard if you prefer.

`longhorn-frontend` is included for consistency even though many users run a single UI replica.

`chart/README.md` generated with `scripts/helm-docs.sh`

Verified against Kubernetes v1.36.
Every rendered Service passed `kubectl apply --dry-run=server`, including the combinations that touch the same templates (`service.ui.type=LoadBalancer` with `loadBalancerClass`, `NodePort` with `nodePort` for both ui and manager, `openshift.enabled=true`, and all four values at once).

`helm lint` clean

### DIFF
--- a/chart/README.md
+++ b/chart/README.md
@@ -148,12 +148,16 @@ The `values.yaml` contains items used to tweak a deployment of this chart.
 
 | Key | Description |
 |-----|-------------|
+| service.admissionWebhook.trafficDistribution | Traffic distribution preference for the Longhorn admission webhook service. (Options: "PreferSameZone", "PreferSameNode") When unspecified, Kubernetes distributes traffic across all endpoints. |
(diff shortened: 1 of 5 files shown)


derekbit (approved):
LGTM. @mydoomfr thanks for your contribution!

Issue #13947, opened 2026-09-05
[IMPROVEMENT] Chart value to allow metrics scrapers through the longhorn-manager NetworkPolicy


alekc:
### Is your improvement request related to a feature? Please describe (👍 if you like this request)

Chart 1.12.1 gates six of the seven NetworkPolicy templates on `networkPolicies.restrictInternalTraffic`, which defaults to true. The longhorn-manager policy allows ingress only from Longhorn's own components, so a Prometheus-compatible scraper in another namespace loses access to port 9500 on upgrade and every `longhorn_*` series disappears. Because those series stop existing rather than dropping to zero, threshold-based Longhorn alerts keep evaluating cleanly against nothing and stay silent, so volume robustness, disk fullness and node readiness go unmonitored with nothing firing. This is #13740, closed after the documentation was improved; the chart still has no supported way to express the allowance.

### Describe the solution you'd like

A `networkPolicies.metricsScrapeSources` value taking a list of standard NetworkPolicy peers, appending one ingress rule to the longhorn-manager policy restricted to TCP/9500. Default `[]`, so rendered output is identical for existing installations and the posture changes only when an operator opts in. This matches `kubeAPIServerSourceCIDRs`, `v1DataEngineInitiatorSourceCIDRs` and `recoveryBackendAdditionalIngressPorts`, which are already opt-in default-empty lists on these same templates.

### Describe alternatives you've considered

Making `networkPolicies.enabled` a master gate, or defaulting `restrictInternalTraffic` to false: both were ruled out in #13740 on security grounds, and neither is proposed here.

The manual policy in `examples/network-policy/allow-prometheus-to-longhorn-manager.yaml` works, but it sits outside the chart: not templated with the release namespace, not covered by the chart upgrade path, and every monitored installation maintains its own copy.

### Additional context

Port 9500 also serves the Longhorn API, so the value should take full peers rather than a bare namespace list, letting `namespaceSelector` and `podSelector` be combined in one peer. The warning already added in longhorn/website#1425 applies unchanged.

Related: #12505, closed before these policies were on by default.

I have a change ready for this and can submit a PR.

derekbit:
cc @COLDTURNIP @c3y1huang

derekbit:
Hello @alekc 
Thanks for your contribution! Could you update https://github.com/longhorn/longhorn/issues/13947#issuecomment-5628441354 for helping our QA's testing? Thank you.

alekc:
[...]

Issue #13740, opened 2026-08-15
[BUG] v1.12.1 chart: new default-on `networkPolicies.restrictInternalTraffic` silently breaks Prometheus scraping of longhorn-manager (metrics/alerting blackout on patch upgrade)


antoinemichea:
## Describe the bug

Chart `1.12.1` introduces `networkPolicies.restrictInternalTraffic` with **default `true`**, and the six internal NetworkPolicy templates are gated **only** on that value — they ignore `networkPolicies.enabled`:

```
templates/network-policies/manager-network-policy.yaml:            {{- if .Values.networkPolicies.restrictInternalTraffic }}
templates/network-policies/instance-manager-networking.yaml:       {{- if .Values.networkPolicies.restrictInternalTraffic }}
templates/network-policies/webhook-network-policy.yaml:            {{- if .Values.networkPolicies.restrictInternalTraffic }}
templates/network-policies/recovery-backend-network-policy.yaml:   {{- if .Values.networkPolicies.restrictInternalTraffic }}
templates/network-policies/backing-image-manager-network-policy.yaml:     idem
templates/network-policies/backing-image-data-source-network-policy.yaml: idem
```
(only `ui-frontend-network-policy.yaml` checks `networkPolicies.enabled`)

The manager policy restricts ingress to Longhorn components only (`longhorn-manager`, `longhorn-ui`, `longhorn-csi-plugin`, jobs, driver-deployer). **There is no allowance — and no chart knob — for a metrics scraper**, so any Prometheus/VictoriaMetrics agent scraping `longhorn-manager:9500 /metrics` from another namespace is cut off.

Result on our production cluster, upgrading `1.12.0 → 1.12.1` with `networkPolicies.enabled: false` (untouched default) and no values change (we verified 0 removed keys before upgrading):

- 6 NetworkPolicies appeared in `longhorn-system` at upgrade time;
- all 12 scrape targets went `up=0` (connect timeout);
- `longhorn_*` series went **968 → 0** and stayed at 0 for ~15 h until we diagnosed it;
- every Longhorn alert was structurally unable to fire during that window — including volume-robustness alerts (while we were post-upgrade rebuilding, precisely when we needed them) and our backup-freshness alert.

Nothing in the v1.12.1 release notes mentions this behavioral change, and it lands in a **patch** release.

## To Reproduce

1. Install chart 1.12.0 with default `networkPolicies` values, scrape `longhorn-backend:9500` from Prometheus in another namespace (works).
[...]

PR #13948, opened 2026-09-05
feat(chart): allow metrics scrapers through the manager network policy

alekc:

Issue longhorn/longhorn#13947

One thing the issue does not cover: the chart ships a ServiceMonitor under `metrics.serviceMonitor.enabled` that scrapes `app: longhorn-manager` on port `manager`, TCP/9500, and also ships a longhorn-manager NetworkPolicy that `restrictInternalTraffic` renders by default whose six peers are all bare `podSelector` entries. Render master with `--set metrics.serviceMonitor.enabled=true` and both objects appear, so the chart denies the scrape configuration it just offered unless Prometheus runs in longhorn-system. The issue has the rest, including why the resulting alert silence does not announce itself.

`networkPolicies.metricsScrapeSources` takes a list of NetworkPolicy peers and appends one ingress rule restricted to TCP/9500, defaulting to `[]` so existing installations render unchanged. longhorn/longhorn#13740 ruled out a `networkPolicies.enabled` master gate and flipping the `restrictInternalTraffic` default; this revisits neither.

The allowance is a separate `from` item, so TCP/9500 constrains only the scraper peers while the existing item stays port-unrestricted for internal Longhorn traffic. The value takes full peers rather than a namespace list so `namespaceSelector` and `podSelector` can be combined in one entry, and its description says to.

`chart/questions.yaml` is untouched, matching the other list-typed `networkPolicies` values, and `chart/README.md` is helm-docs output rather than hand-edited.

Test environment:
- Longhorn version: chart from this branch, rendered against a live 1.12.1 install
- Kubernetes version: v1.34.8
- Kubernetes distribution: kubeadm
- Node count: 3
- OS: Ubuntu 24.04 and 25.10
- Data engine: v1
- Installation method: Helm (Argo CD)
- Scraper: VictoriaMetrics vmagent in the monitoring namespace

Test steps:
1. Rendered master against this branch across `enabled` x `restrictInternalTraffic` x `type` x `ingress.enabled`, 24 combinations plus the bare default and an explicit empty list: identical every time.
2. Regenerated both `deploy/` manifests with `scripts/generate-longhorn-yaml.sh`: byte-identical to what is committed.
3. Rendered the opt-in path with a selector pair, an `ipBlock` with `except`, and a `matchExpressions` selector, parsing the output back to confirm peer grouping survives.
4. Applied the rendered policy on a cluster whose `longhorn_*` series had been absent for 31 hours behind the 1.12.1 policies, then reverted it.

Result: PASS. Before: 0 of 3 `longhorn-backend` targets up, 0 `longhorn_*` series, an `absent()` alert firing. Within 15 seconds of the apply: 3 of 3 up, 11181 series, `longhorn_node_count_total` reporting from all three manager pods, alert cleared. Reverting put it back, so the added rule is what changed the outcome.

Known risks: NetworkPolicy is L3/L4, so an allowance on TCP/9500 reaches the REST API as well as `/metrics` and cannot be narrowed further in the policy; an over-broad peer exposes that API to a whole namespace, hence the selector pair in the description. Setting the value while `restrictInternalTraffic` is off is a no-op, correctly, since the manager pods are then selected by no policy from this chart. If this lands, longhorn/website#1425 is worth a follow-up pointing Helm installs at the value.


COLDTURNIP (approved):
LGTM

derekbit (approved):

derekbit:
Thanks @alekc for your contribution!

derekbit:
@mergify backport v1.13.x v1.12.x

Issue #14020, opened 2026-09-16
[IMPROVEMENT] Restrict the Secret access of CSI components


COLDTURNIP:
### Is your improvement request related to a feature? Please describe (👍 if you like this request)

To current implementation, the CSI components are launched using the `longhorn-service-account` service account, which [allows to access any Secrets](https://github.com/longhorn/longhorn/blob/v1.13.x/chart/templates/role.yaml#L10-L11).

### Describe the solution you'd like

Rather than the service account `longhorn-service-account`, create a new one for the CSI components, with less permission bounded.

derekbit:
Hello @COLDTURNIP 
Let's uppdate the validation steps in https://github.com/longhorn/longhorn/issues/14020#issuecomment-5707911676. @roger-ryao can validate the improvement. Thanks.

COLDTURNIP:
## Verification Steps

### Secret read permission removal

1. Given the Longhorn setup with Helm value `csi.allowControllerSecretAccess` disabled
2. And a Secret for volume encryption (see the [doc](https://longhorn.io/docs/1.12.1/advanced-resources/security/volume-encryption/))
3. And a StorageClass A for encrypted volumes with all 6 `csi.storage.k8s.io/node-secret-*` encryption parameters. No `csi.storage.k8s.io/provisioner-secret-*` parameters.
4. When create the encrypted volume from a PVC with this StorageClass A
5. Then the volume is created and attached without any issue
6. When expand the volume size on the PVC
7. Then the volume is expanded without any issue
8. When destroy the PVC
9. Then the volume is destroyed without any isue
10. When create a new StorageClass B for encrypted volumes with all 8 documented `csi.storage.k8s.io/*-secret-*` encryption parameters.
11. And When create the encrypted volume from a PVC with this StorageClass B
12. Then the PV creation cannot be fulfilled because of the external-provisioner does not able to read the Secret.### Backward compatibility
13. When check the permission of the CSI sidecar components, then are using the new ServiceAccount `longhorn-csi-service-account` and is not allowed to read the Secrets
    ```bash
    $ NS=longhorn-system
    $ CSI_SELECTOR='app in (csi-attacher,csi-provisioner,csi-resizer,csi-snapshotter,longhorn-csi-plugin)'
    $ kubectl -n "$NS" get pods -l "$CSI_SELECTOR" \
    
[...]

Issue #12531, opened 2026-01-25
[BUG][v1.11.0-rc3] Encrypted v2 volume stuck in detach/attach loop after deleting instance manager pod


chriscchien:
### Describe the Bug

From test case `Crash Instance Manager While RWO Encrypted Workload Pod Is Starting`, after deleting instance manager pod of a encrypted v2 volume, the volume have a chance stuck at detach/attach loop

https://github.com/user-attachments/assets/21fa2d4a-1a71-47db-88f8-469d35f8b286

https://github.com/user-attachments/assets/08a3647f-41fc-485a-9f3f-523a35cca6ef

https://10.115.5.5/job/private/job/longhorn-e2e-test/164/consoleFull log
```
23:07:39  Waiting for workload e2e-test-deployment-0 volume pvc-59f6b2bc-bbc2-41fb-9ff1-8149ec50b767 to be healthy
23:07:39  Waiting for pvc-59f6b2bc-bbc2-41fb-9ff1-8149ec50b767 attached ... (0)
23:07:39  Waiting for pvc-59f6b2bc-bbc2-41fb-9ff1-8149ec50b767 healthy (0) ...
23:07:39  Writing 100 MB random data to pod e2e-test-deployment-0-6cf6f54cfb-pvnpm file data.txt
23:07:39  Waiting for pod e2e-test-deployment-0-6cf6f54cfb-pvnpm status Running, current status Running (0) ...
23:07:39  Writing random data to pod e2e-test-deployment-0-6cf6f54cfb-pvnpm ... (0)
23:07:41  Storing pod e2e-test-deployment-0-6cf6f54cfb-pvnpm file data.txt checksum = a9f40304acca945da5441583ca972e51
23:07:41  Try to set custom resource volumes pvc-59f6b2bc-bbc2-41fb-9ff1-8149ec50b767 annotation test.longhorn.io/data-checksum-data.txt=a9f40304acca945da5441583ca972e51 ... (0)
23:07:41  Try to set custom resource volumes pvc-59f6b2bc-bbc2-41fb-9ff1-8149ec50b767 annotation test.longhorn.io/last-recorded-checksum=a9f40304acca945da5441583ca972e51 ... (0)
23:07:41  Got v2 instance manager running on node ip-10-0-1-253: instance-manager-49ed998bc49aad4666d73eea2faf1d53
23:07:41  Deleting pod instance-manager-49ed998bc49aad4666d73eea2faf1d53 in namespace longhorn-system
[...]

Issue #12490, opened 2026-01-16
[IMPROVEMENT] Refactor SmartManager disk device scanning logic to filter partitions and virtual devices


COLDTURNIP:
### Is your improvement request related to a feature? Please describe (👍 if you like this request)

This issue tracks the improvement of the disk device scanning logic within the `SmartManager`, as discussed in [review comment](https://github.com/longhorn/longhorn-manager/pull/4438#discussion_r2689376529).

The current SmartManager implementation (in the beszel fork) may attempt to scan all block devices found, potentially including:

1. Partitions (e.g., sda1, nvme0n1p1) instead of just the parent block device.
2. Virtual/Loop devices that do not support SMART metrics.

Iterating over partitions or unsupported devices causes redundant smartctl I/O risk. We need a more robust way to identify the underlying physical block devices that actually support SMART queries.

### Describe the solution you'd like

Refactor the device scanning logic in `SmartManager` to:

- Filter out partitions: Ensure we only target the parent block device (e.g., scan sda, ignore sda1).
- Exclude irrelevant device types: Explicitly skip loop devices (loop*), RAM disks, or other virtual devices that do not report physical health status.

A possible solution is to filter the devices from the output of `lsblk -d -n -o NAME,TYPE`

### Additional context

https://github.com/longhorn/longhorn-manager/pull/4438#discussion_r2689376529

derekbit:
cc @c3y1huang

PR #14026, opened 2026-09-17
feat(rbac): restrict secret access for csi sidecar components

COLDTURNIP:

Issue longhorn/longhorn#14020

Separate CSI controller service-account permissions from the manager and node plugin while retaining Secret compatibility by default.

- `csi.allowControllerSecretAccess: true` (default) renders `longhorn-csi-secret-role` and `longhorn-csi-secret-bind`, granting only cluster-wide Secret GET to `longhorn-csi-service-account` in the release namespace.
- `false` renders neither object. The base CSI role remains Secret-free.
- Generated standard and OpenShift installation manifests include the default grant. Manifest users can explicitly delete the binding and keep it out of future applies.
- Keep node-only encryption examples. No separate compatibility RBAC example is needed because the default installation manifests already grant access.
- Explain safe-disable prerequisites in the value, Rancher question, and generated chart README. Preserve all three original encryption proposals with dated v1.13.0 update sections instead of rewriting their design history.

The existing commit was amended, not followed by an additional commit. Its title now uses `feat(rbac): separate CSI controller Secret permissions` as requested in review. The companion [manager branch](https://github.com/longhorn/longhorn-manager/pull/5237) now contains only the service-account change; there is no admission validator or dynamic RBAC reconciliation. The previous dynamic implementation remains on the manager's `14020-csi-rbac-dynamic` branch.

Default Secret GET remains cluster-wide by design. The option changes only the dedicated controller grant, not manager/node-plugin permissions or unrelated administrator-created RBAC.

#### Verification

Published before running `make`, as requested. Helm lint, duplicate-key checks, default/false rendering, custom-namespace rendering, OpenShift false rendering, and group-snapshot rendering passed using `csi.allowControllerSecretAccess`. Both generated release manifests include the exact default GET grant. Static security review found no concrete implementation defect.

Live K3s upgrade with the default enabled granted Secret GET, not list/watch, in two namespaces. Helm false removed both RBAC objects and denied all three verbs; true restored GET only. Deleting the binding manually also revoked access and survived a completed deployer restart with a legacy StorageClass present. Existing encrypted payloads and PVC identities were preserved.

A fresh encrypted PVC using historical provisioner-secret parameters successfully provisioned and mounted with the default grant; its synthetic payload checksum matched. The original proposal text is preserved byte-for-byte in two documents, and in the V2 proposal except for required ASCII quote normalization.


chart/templates/clusterrole.yaml
```diff
@@ -93,3 +93,58 @@ rules:
...
   verbs: ["use"]
 {{- end }}
+---
+# Permissions for the CSI attacher, provisioner, resizer, and snapshotter.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: longhorn-csi-role
+  labels: {{- include "longhorn.labels" . | nindent 4 }}
+rules:
```
c3y1huang, line 103: With Secret access removed, would this break SCs that sets `provisioner-secret-name`?
- https://github.com/longhorn/longhorn/blob/8e78cf85a224bba0e1532ce32fe269925277b3c0/examples/crypto/storageclass-crypto-global.yaml#L12-L16
- https://github.com/kubernetes-csi/external-provisioner/blob/34c6fc2149a354cea116724d1836262516f83225/pkg/controller/controller.go#L743-L750

Also, [build #12490](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/12490/) was run with `-m coretest`, which only covered 4 CSI tests and none for encryption. Suggest rerunning with the full CSI set.
- https://github.com/longhorn/longhorn-tests/blob/2ecf9e8c04f3d2dc767be7d3ece383a962c57035/manager/integration/tests/test_csi.py#L282-L300
- https://github.com/longhorn/longhorn-tests/blob/2ecf9e8c04f3d2dc767be7d3ece383a962c57035/manager/integration/tests/test_csi.py#L315-L335

COLDTURNIP: Though the parameters `csi.storage.k8s.io/provisioner-secret-*` are listed in the [Longhorn official doc](https://longhorn.io/docs/1.12.1/advanced-resources/security/volume-encryption/), the longhorn-manager's controller server does not consume the secret during `CreateVolume`/`DeleteVolume`. Actually to Longhorn encrypted volume, `CreateVolume` refers to only the custom parameter `encrypted`. The Secret consumed by node server's `NodeStageVolume` and `NodeExpandVolume` is fetched by the Kubelet, not the CSI sidecars. We should correct controller-side Secret parameters in the documented example later.

Will launch a job to cover full CSI set.

deploy/longhorn-okd.yaml
```diff
@@ -18,6 +18,7 @@ description: "Ensure Longhorn pods have the highest priority to prevent any unex
 globalDefault: false
 preemptionPolicy: PreemptLowerPriority
 value: 1000000000
+
```
c3y1huang: Could you regenerate these with Helm v3 to **keep the diff clean**? The blank line seems to be a v3 vs. v4 rendering difference, unrelated to this PR.

The bot currently regenerates the manifest with v3, and it keeps flipping back and forth. We can fix it in another PR.

c3y1huang: https://github.com/longhorn/longhorn-manager/pull/5241

chart/templates/clusterrole.yaml
c3y1huang: Agree on the Longhorn side. My concern is one step earlier. If the SC has `provisioner-secret-name`, the CSI provisioner [fetches that Secret](https://github.com/kubernetes-csi/external-provisioner/blob/34c6fc2149a354cea116724d1836262516f83225/pkg/controller/controller.go#L743-L750) itself before it calls [CreateVolume](https://github.com/kubernetes-csi/external-provisioner/blob/34c6fc2149a354cea116724d1836262516f83225/pkg/controller/controller.go#L880), and returns if the `Get` fails. So IIUC, the request would never reach the driver, whether or not the driver would have used the `Secret`.

COLDTURNIP: The existing test case expect to refer these `csi.storage.k8s.io/provisioner-secret-*` parameters. Will update the test case to correct the expected behavior.

deploy/longhorn-okd.yaml
COLDTURNIP: Regenerate using Helm 3 for this PR.

chart/templates/clusterrole.yaml
c3y1huang: > The existing test case expect to refer these `csi.storage.k8s.io/provisioner-secret-*` parameters. Will update the test case to correct the expected behavior.

\+ doc
\+ [example](https://github.com/longhorn/longhorn/blob/8e78cf85a224bba0e1532ce32fe269925277b3c0/examples/crypto/storageclass-crypto-global.yaml#L12-L16)

COLDTURNIP: The example is updated, and pass the [updated CSI tests](https://github.com/longhorn/longhorn-tests/pull/3758) in Jenkins regression job #12531. The full regression is still undergoing.

COLDTURNIP: Doc PR longhorn/website#1452

COLDTURNIP:
Full regression longhorn-tests-regression/12530 completed. There are two failed case, and none of the case relates to the Secret permission removal.

- `test_sm_pod_recreate_backoff`: confirmed to be flaky (longhorn-tests-regression/12579)
- `test_offline_node`: failed, but this does not relates to the CSI sidecars. The node takes around 6 min to recover, and exceeded the 5 min criteria.

c3y1huang (commented):
What happens to users with encrypted SC with `csi.storage.k8s.io/provisioner-secret-*` after upgrade? Should there be a pre-upgrade check?

COLDTURNIP:
> What happens to users with encrypted SC with `csi.storage.k8s.io/provisioner-secret-*` after upgrade? Should there be a pre-upgrade check?

We have some [upgrade notes and KB post](https://github.com/longhorn/website/pull/1452) to guide the user to remove the parameters. Will update pre-upgrade checker in longhorn-manager.

c3y1huang (commented):
The change LGTM. Two minor suggestions:
- Fix the commit message drifts.
	> Separate CSI controller permissions from the manager role and grant no
	> Secret access. Remove the configurable Secret reader role and binding
	> so both new installations and upgrades use the same restricted policy.
- The title says "csi plugin" which isn't right; `longhorn-csi-plugin` still uses `longhorn-service-account`, IIUC.

c3y1huang (dismissed):

chart/values.yaml
```diff
@@ -273,6 +273,8 @@ csi:
   resizerReplicaCount: ~
   # -- Replica count of the CSI Snapshotter. When unspecified, Longhorn uses the default value ("3").
   snapshotterReplicaCount: ~
+  # -- Setting that allows the CSI controller sidecars to read Kubernetes Secrets through a static ClusterRole. This grants cluster-wide Secret GET access to the CSI controller sidecars; set to false to omit this grant and manage compatibility access manually.
+  sidecarSecretRole: true
```
c3y1huang:
```suggestion
  allowControllerSecretAccess: true
```

chart/values.yaml
```diff
@@ -273,6 +273,8 @@ csi:
   resizerReplicaCount: ~
   # -- Replica count of the CSI Snapshotter. When unspecified, Longhorn uses the default value ("3").
   snapshotterReplicaCount: ~
+  # -- Setting that allows the CSI controller sidecars to read Kubernetes Secrets through a static ClusterRole. This grants cluster-wide Secret GET access to the CSI controller sidecars; set to false to omit this grant and manage compatibility access manually.
```
c3y1huang: Suggest improving the description. It says how to turn it off, but not when it's safe to.

enhancements/20221024-pv-encryption.md
```diff
@@ -44,13 +44,12 @@ how to create and use an encrypted volume.
...
 Creation and usage of an encrypted volume requires 2 things:
 - the storage class needs to specify `encrypted: "true"` as part of its parameters.
-- secrets need to be created and reference for the csi operations need to be setup.
+- a Secret must be referenced by the node-side CSI operations.
 - see below examples for different types of secret usage.
 
-The kubernetes sidecars are responsible for retrieval of the secret and passing it to the csi driver.
-If the secret hasn't been created the PVC will remain in the Pending State.
-And the side cars will retry secret retrieval periodically, once it's available the sidecar container will call
-`Controller::CreateVolume` and pass the secret after which longhorn will create a volume.
```
c3y1huang, line 53: Suggest not rewriting the design body, this loses the record of the original design context. Can we add a dated section instead to describe the change? For example:
```
> **Update (v1.13.0, longhorn/longhorn#14020):** controller-side Secret parameters are no longer used. Do not configure `csi.storage.k8s.io/provisioner-secret-name` or `csi.storage.k8s.io/provisioner-secret-namespace` for Longhorn encryption. Provisioning does not read the encryption key; use only node-side Secret references.

...
```

c3y1huang (commented):
> fix(rbac): separate CSI controller Secret permissions

feat over fix since this introduces new value and service account.

c3y1huang (approved):

derekbit (approved):

derekbit:
@mergify backport v1.13.x

